PR #17, opened 2023-08-21
Login page

zpoken:


apps/extension/public/popup.html
```diff
@@ -7,7 +7,7 @@
     <script src="js/vendor.js"></script>
   </head>
   <body>
-    <div id="root"></div>
+    <div id="root" style="width: 400px; height: 600px"></div>
```
grod220, line 10: This should go in globals.css no? It may require renaming id to `popup-root`.

apps/extension/src/components/password-input.tsx
```diff
@@ -0,0 +1,45 @@
+import { EyeClosedIcon, EyeOpenIcon } from '@radix-ui/react-icons';
+import { useState } from 'react';
+import { Input, InputProps } from 'ui/components';
+
+interface IPasswordInputProps {
```
grod220, line 5: Don't think we need to append interface names with `I`

```diff
- interface IPasswordInputProps {
+ interface PasswordInputProps {

File: apps/extension/src/components/password-input.tsx [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+  value: string;
+  label: string;
+  isHelper: boolean;
+  helper: string;
```
grod220, line 9: Helper is the wrong name for this. These are more like validations. Why not make this general? Something like:

```typescript
interface PasswordInputProps {
   validations: Validation[]
}

interface Validation {
   checkFn: (txt: string) => boolean)
   type: "warn" | "error" // corresponds to red or yellow
}
```

apps/extension/src/components/password-input.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <div className='absolute inset-y-0 right-4 flex cursor-pointer items-center'>
+          {reveal ? (
+            <EyeOpenIcon onClick={() => setReveal(false)} />
+          ) : (
+            <EyeClosedIcon onClick={() => setReveal(true)} className='text-gray-500' />
+          )}
+        </div>
+        <Input
+          type={reveal ? 'text' : 'password'}
+          variant={!value.length ? 'default' : value.length >= 8 ? 'success' : 'warn'}
```
grod220, line 38: Should this be hard coded? if you are entering your password, we should only give color feedback when you enter it in wrong.

apps/extension/src/routes/popup/enter-password.tsx
```diff
@@ -3,36 +3,87 @@ import { useState } from 'react';
...
+    setPasswordValue(e.target.value);
+    setMatchError(true);
+  };
+
+  const gotoRestorePage = () =>
+    void (async function () {
+      await chrome.tabs.create({
+        url: `${chrome.runtime.getURL('page.html')}`,
+      });
+    })();
```
grod220, line 44: You should move business logic to the slice. Also when you do, make sure you add more tests for this functionality.

apps/extension/src/routes/popup/enter-password.tsx
```diff
@@ -3,36 +3,87 @@ import { useState } from 'react';
...
+
+  const handleChangePassword: InputProps['onChange'] = e => {
+    setPasswordValue(e.target.value);
+    setMatchError(true);
+  };
+
+  const gotoRestorePage = () =>
+    void (async function () {
+      await chrome.tabs.create({
+        url: `${chrome.runtime.getURL('page.html')}`,
```
grod220, line 42:
```diff
- url: `${chrome.runtime.getURL('page.html')}`,
+ url:  chrome.runtime.getURL('page.html'),
```

apps/extension/src/routes/popup/enter-password.tsx
```diff
@@ -3,36 +3,87 @@ import { useState } from 'react';
...
+              Unlock
+            </Button>
+          </form>
+          <Button className='text-white' variant='link' onClick={gotoRestorePage}>
+            Forgot Password?
+          </Button>
+          <p className='text-sm'>
+            Need help? Contact{' '}
+            <a
+              className='cursor-pointer text-sm text-green-600 hover:underline'
```
grod220, line 78: This green is not on brand. It probably should be teal.

apps/extension/src/routes/popup/enter-password.tsx
```diff
@@ -3,36 +3,87 @@ import { useState } from 'react';
...
+              target='_blank'
+              href='https://guide.penumbra.zone/main/extension.html'
```
grod220, line 80: This URL is wrong right?

apps/extension/src/routes/popup/enter-password.tsx
```diff
@@ -3,36 +3,87 @@ import { useState } from 'react';
...
-      >
-        <label>
-          Enter your password
-          <input
-            type='text'
-            value={inputTx}
-            onChange={e => {
-              setInputTxt(e.target.value);
+    <FadeTransition>
+      <Card className='w-[300px] p-2' gradient>
```
grod220, line 48: The popup is so small, we probably should not use cards unless it's an overlay perhaps.

I wonder if we should make the entire background the card gradient color 🤔

apps/extension/src/routes/popup/index.tsx
```diff
@@ -16,26 +27,52 @@ export const popupIndexLoader = async () => {
...
 };
 
 export const PopupIndex = () => {
-  const { all } = useStore(accountsSelector);
+  const navigate = usePopupNav();
+  const { clearSessionPassword, hashedPassword } = useStore(passwordSelector);
+
+  useEffect(() => {
+    if (!hashedPassword) navigate(PopupPath.ENTER_PASSWORD);
+  }, [hashedPassword]);
```
grod220, line 43: Remove. See loader above.

apps/extension/src/routes/popup/enter-password.tsx
Vadim121197: here we have local business logic that is not needed in other places, so I did not move it to slice

zpoken: here we have local business logic that is not needed in other contexts, so I didn't move it in the slice

apps/extension/src/routes/popup/enter-password.tsx
zpoken: we have no support, so we decided to add a manual

grod220: I think we should either link to our discord or remove this then

apps/extension/src/routes/popup/enter-password.tsx
grod220: I think it all relates to the password slice and can be moved there

apps/extension/src/routes/popup/index.tsx
zpoken: doesnt redirect when logout

grod220: Can you add a code comment, that the useEffect() is for that purpose?

apps/extension/src/components/password-input.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }));
+
+    const error = validation.find(i => i.variant === 'error' && i.check);
+    if (error) return error;
+
+    const warn = validation.find(i => i.variant === 'warn' && i.check);
+    if (warn) return warn;
+
+    return null;
+  }, [validations, passwordValue]);
```
grod220, line 41: Think this could be simpler:

```typescript
const priorityResult = useMemo(() => {
    const results = validations.filter(v => v.checkFn(passwordValue));
    const error = results.find(v => v.type === 'error');
    return error ? error : results.find(v => v.type === 'warn');
}, [validations, passwordValue]);
```

apps/extension/src/state/password.ts
```diff
@@ -35,10 +40,38 @@ export const createPasswordSlice =
...
+      setCorrectPassword: () => {
+        set(state => {
+          state.password.isCorrectPassword = true;
+        });
+      },
       isPassword: async password => {
         const locallyStoredPassword = await local.get('hashedPassword');
         return locallyStoredPassword === repeatedHash(password);
       },
+      isUnlock: async password => {
```
grod220, line 58: I would rename this to be `isPassword` and move that logic into here. I don't think that function is being used except here now.

apps/extension/src/state/password.ts
```diff
@@ -6,19 +6,24 @@ import { repeatedHash } from 'penumbra-crypto-ts';
...
 export const createPasswordSlice =
   (
     session: ExtensionStorage<SessionStorageState>,
     local: ExtensionStorage<LocalStorageState>,
   ): SliceCreator<PasswordSlice> =>
-  set => {
+  (set, get) => {
     return {
       hashedPassword: undefined,
+      isCorrectPassword: true,
```
grod220, line 26: Think this is risky having this default true. The component itself should just have;

`const result = isPassword(password)`. And the rest of the UI can react to that locally there.

packages/ui/styles/globals.css
```diff
@@ -42,12 +42,18 @@
     --rust: #8d5728;
   }
 
+  #popup-root {
+    width: 400px;
+    height: 600px;
+    @apply bg-card-radial bg-opacity-20;
```
grod220, line 48: This color isn't quite right. See the onboarding cart. It's lighter and more gray. You can compare to penumbra.zone's cards as well.

apps/extension/src/state/password.ts
grod220: I would remove `setCorrectPassword: () => void;`

zpoken: I don't understand! you wrote in the previous version that I should put it in slice

apps/extension/src/state/password.ts
zpoken: in the future, the ability to check the password when displaying the seed phrase will be required, as an example

grod220 (approved):

PR #18, opened 2023-08-22
Forgot password

zpoken:


apps/extension/src/state/persist.ts
```diff
@@ -25,6 +25,17 @@ export const customPersistImpl: Persist = f => (set, get, store) => {
...
+    chrome.storage.onChanged.addListener(changes => {
+      if (changes['hashedPassword']) {
+        const newValues = changes['hashedPassword'].newValue as { value: string } | undefined;
+        set(
+          produce((state: AllSlices) => {
+            state.password.hashedPassword = newValues ? newValues.value : undefined;
+          }),
+        );
+      }
+    });
```
grod220, line 38: I would do something like this:

```typescript
    // Setup listeners so that store can be synced with storage changes
    chrome.storage.onChanged.addListener((changes, area) => {
      if (area === 'local') {
        syncLocal(changes, set);
      }
      if (area === 'session') {
        syncSession(changes, set);
      }
    });


// Somewhere below
type Setter = (
  partial: (state: AllSlices) => Partial<AllSlices> | AllSlices,
  replace?: boolean | undefined,
) => void;

function syncLocal(changes: Record<string, chrome.storage.StorageChange>, set: Setter) {
  if (changes['accounts']) {
    const item = changes['accounts'].newValue as
      | StorageItem<LocalStorageState['accounts']>
      | undefined;
    set(
      produce((state: AllSlices) => {
        state.accounts.all = item ? item.value : [];
      }),
    );
  }
}

function syncSession(changes: Record<string, chrome.storage.StorageChange>, set: Setter) {
  if (changes['hashedPassword']?.newValue) {
    const item = changes['hashedPassword'].newValue as
      | StorageItem<SessionStorageState['hashedPassword']>
      | undefined;
    set(
      produce((state: AllSlices) => {
        state.password.hashedPassword = item ? item.value : undefined;
      }),
    );
  }
}
```

PR #19, opened 2023-08-24
Fix eslint

zpoken:


packages/crypto/tsconfig.json
```diff
@@ -2,7 +2,5 @@
   "extends": "tsconfig/base.json",
   "include": ["."],
   "exclude": ["dist", "build", "node_modules"],
-  "compilerOptions": {
-    "baseUrl": "."
-  }
+  "compilerOptions": {}
```
grod220, line 5: Think you can remove `"compilerOptions": {}` entirely here

packages/eslint-config-custom/index.js
```diff
@@ -13,6 +13,17 @@ module.exports = {
...
+      'warn',
+      {
+        whitelist: [
+          'bg\\-(card-radial|charcoal|teal|accent|background|teal/80|destructive|destructive/90)',
+          'text\\-(muted-foreground|accent-foreground|primary-foreground|destructive-foreground|primary|teal)',
+          'ring\\-(offset-background|ring)',
+          'border\\-(input|teal)',
+        ],
+      },
+    ],
```
grod220, line 26: Why is this necessary? This will be annoying to have to maintain.

grod220: When I remove this, I do not get any warning

grod220: By default, is `tailwindcss/no-custom-classname` off? When I remove these lines, I get no warnings. Are you adding in a new rule for extra protection?

zpoken: or do you not have a notification for this rule in your code? and a problem with my IDE

PR #20, opened 2023-08-24
Password recovery

zpoken:


apps/extension/src/routes/page/index.tsx
```diff
@@ -1,24 +1,74 @@
...
-      <Button onClick={() => navigate(PagePath.WELCOME)}>Go to onboarding</Button>
-    </div>
+    <FadeTransition>
+      <div className='absolute inset-0 flex w-screen items-center justify-center'>
+        <CompressedVideoLogo noWords className='w-[850px]' />
+      </div>
+      <Card className='w-[650px] p-6' gradient>
+        <CardHeader>
+          <CardTitle className='bg-gradient-to-r from-teal-400 via-neutral-300 to-orange-400 bg-clip-text text-6xl text-transparent opacity-80'>
+            Successfully login
```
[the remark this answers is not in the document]
grod220: Can we make a special page specifically for the end of the "forgot password" flow? Think it should have different text.

apps/extension/src/routes/page/login.tsx
```diff
@@ -0,0 +1,37 @@
+import { Button, Card, CardContent, CardDescription, CardHeader, CardTitle } from 'ui/components';
+import { FadeTransition, LoginForm } from '../../components';
+import { usePageNav } from '../../utils/navigate';
+import { PagePath } from './paths';
+
+export const Login = () => {
+  const navigate = usePageNav();
+
+  const gotoRestorePage = () => navigate(PagePath.RESTORE_PASSWORD);
```
grod220, line 9: Think this can be inline with the button

apps/extension/src/routes/page/onboarding/set-password.tsx
```diff
@@ -12,12 +12,21 @@ import { FadeTransition, PasswordInput } from '../../../components';
...
+  const { hashedPassword } = useStore(passwordSelector);
   const finalOnboardingSave = useOnboardingSave();
   const [password, setPassword] = useState('');
   const [confirmation, setConfirmation] = useState('');
+  console.log(window.location.hash);
+
+  useEffect(() => {
+    // redirect to main route after login from other tabs or popup, needed to clean up url
+    if (hashedPassword) window.location.replace(chrome.runtime.getURL('page.html'));
+  }, [hashedPassword]);
```
grod220, line 29: how could someone get to this page when they already have a password?

apps/extension/src/routes/page/onboarding/set-password.tsx
```diff
@@ -59,7 +68,7 @@ export const SetPassword = () => {
             disabled={password.length < 8 || password !== confirmation}
             onClick={() => {
               finalOnboardingSave(password);
-              navigate(PagePath.ONBOARDING_SUCCESS);
+              navigate(PagePath.INDEX);
```
grod220, line 71: Why would they go to index and not the confirmation page?

apps/extension/src/routes/page/page-layout.tsx
```diff
@@ -0,0 +1,40 @@
+import { Outlet } from 'react-router-dom';
+import { useStore } from '../../state';
+import { passwordSelector } from '../../state/password';
+import { useEffect } from 'react';
+import { PagePath } from './paths';
+import { usePageNav } from '../../utils/navigate';
+import { accountsSelector } from '../../state/accounts';
+
+const getLocationHash = (location: string, key: PagePath) =>
+  (location.replace('#', '') as PagePath) === key;
```
grod220, line 10: What does this do? This needs a code comment also. This feels like a non-standard way of redirecting.

apps/extension/src/routes/page/page-layout.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const PageLayout = () => {
```
grod220, line 12: Why should this logic wrap every single page? This whole component does not feel like the right approach.

apps/extension/src/routes/page/restore-password.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useStore } from '../../state';
+import { importSelector } from '../../state/seed-phrase/import';
+import { passwordSelector } from '../../state/password';
+import { useEffect } from 'react';
+
+export const RestorePassword = () => {
+  const { phrase } = useStore(importSelector);
+  const { hashedPassword } = useStore(passwordSelector);
+
+  const replaceUrl = () => window.location.replace(chrome.runtime.getURL('page.html'));
```
grod220, line 13: can be inline with button

apps/extension/src/routes/page/restore-password.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  useEffect(() => {
+    // redirect to index route after login from other tabs or popup, needed to clean up url
+    if (hashedPassword) replaceUrl();
+  }, [hashedPassword]);
```
grod220, line 18: Why would someone end up on this page if they already had a password?

apps/extension/src/routes/popup/index.tsx
```diff
@@ -33,35 +31,33 @@ export const PopupIndex = () => {
 
   const handleLogout = () => {
     clearSessionPassword();
-    navigate(PopupPath.ENTER_PASSWORD);
+    navigate(PopupPath.LOGIN);
   };
```
grod220, line 35: can be inline with button

apps/extension/src/routes/popup/login.tsx
```diff
@@ -0,0 +1,39 @@
+import { Button } from 'ui/components';
+import { FadeTransition, LoginForm } from '../../components';
+import { PagePath } from '../page/paths';
+
+export const Login = () => {
+  const gotoRestorePage = () =>
+    void (async function () {
+      await chrome.tabs.create({
+        url: `${chrome.runtime.getURL('page.html')}#${PagePath.RESTORE_PASSWORD}`,
```
grod220, line 9: Why do we have a hash and not a `/` that goes straight to that route?

apps/extension/src/routes/popup/popup-layout.tsx
```diff
@@ -1,13 +1,25 @@
...
+export const PopupLayout = () => {
+  const navigate = usePopupNav();
+  const { hashedPassword } = useStore(passwordSelector);
+
+  useEffect(() => {
+    hashedPassword
+      ? // redirect to index after login from tab
+        navigate(PopupPath.INDEX)
+      : // redirect to login after logout from tab
+        navigate(PopupPath.LOGIN);
```
grod220, line 17:
```diff
    hashedPassword
-    ? // redirect to index after login from tab
+   ? // password is in session, direct to index
        navigate(PopupPath.INDEX)
-     : // redirect to login after logout from tab
+     : // session storage expired, needs to re-login
        navigate(PopupPath.LOGIN);
```

apps/extension/src/routes/page/onboarding/set-password.tsx
Vadim121197: User open tab and enter password to popup

apps/extension/src/routes/page/restore-password.tsx
Vadim121197: User open tab and enter password to popup

apps/extension/src/routes/page/page-layout.tsx
Vadim121197: Why it does not feel like the right approach?

apps/extension/src/routes/page/onboarding/set-password.tsx
Vadim121197: There we have logic, what checking is you login

apps/extension/src/routes/page/page-layout.tsx
Vadim121197: There we have logic, what checking is you login

apps/extension/src/routes/page/login.tsx
```diff
@@ -0,0 +1,37 @@
...
+            Need help? Contact{' '}
+            <a
+              className='cursor-pointer text-sm text-teal-500 hover:underline'
+              target='_blank'
+              href='https://discord.com/channels/824484045370818580/1077672871251415141'
+            >
+              Penumbra Support
+            </a>
+          </p>
+        </CardContent>
```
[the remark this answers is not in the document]
zpoken: bottom as a separate block

apps/extension/src/routes/popup/login.tsx
zpoken: 'chrome-extension://gchndbcadfiekjgpbpjdmipdmmalakmc/page.html/restore-password' like this doesn`t work

zpoken: tab does not change the url, it always looks like this - chrome-extension://gchndbcadfiekjgpbpjdmipdmmalakmc/page.html

grod220: Ah, yes can you see if the hash router will work?

zpoken: hash router works!

apps/extension/src/routes/page/page-layout.tsx
zpoken: we need to wrap everything so that a user with an open page and popup can log in/out of the popup or page, and the page and popup states should change together! maybe you have another idea how to do it?

zpoken: User access to pages: 
1. ❌password ❌accounts =>  Welcome, Welcome-import, Welcome-set-password, welcome-generate, welcome-confirm-backup pages; 
2.  ❌password ✅accounts => Login, Restore password, Restore-password-set-password pages
3. ✅password ✅accounts => Index, welcome-success pages

grod220: Gotcha. See `pageIndexLoader`. I think there is overlap here. Let's implement this logic as a loader.

zpoken: And what about data synchronization between popup and tab?

grod220: Ah ok, I see what you mean. Ok, let's keep this as a hook then. More questions:

With hash router, we don't need this right?
```typescript
  useEffect(() => {
    // redirect to reset wallet after popup forgot password click
    if (getLocationHash(window.location.hash, PagePath.RESTORE_PASSWORD))
      navigate(PagePath.RESTORE_PASSWORD);
  }, []);
``` 

This line too right?

```typescript
if (getLocationHash(window.location.hash, PagePath.RESTORE_PASSWORD)) return;
```

Shouldn't it be like this?

```typescript
   if (!all.length) {
      navigate(PagePath.WELCOME);
    } else if (!hashedPassword) {
      navigate(PagePath.LOGIN);
    }
```

Because if someone has a password, we don't want to redirect them everytime to the index page.

zpoken: With hash router, we don't need this
```
 useEffect(() => {
    // redirect to reset wallet after popup forgot password click
    if (getLocationHash(window.location.hash, PagePath.RESTORE_PASSWORD))
      navigate(PagePath.RESTORE_PASSWORD);
  }, []);
```
and this
`if (getLocationHash(window.location.hash, PagePath.RESTORE_PASSWORD)) return;`

We should check all version
```
  if (!all.length) {
      navigate(PagePath.WELCOME);
    } else if (!hashedPassword) {
      navigate(PagePath.LOGIN);
    }else if(hashedPassword){
      navigate(PagePath.INDEX)}
```

zpoken: what differents between success and index pages?

grod220: Because page-layout wraps all pages, doesn't that mean if they have a password, they'll be guaranteed to always be redirected to index? Like what if they are in the settings page or doing something else. You don't want to redirect those users.

zpoken: yes, but we still have a situation where the user is on the login page, and here we should be redirected to the index when the login was performed from another context

grod220: That is an edge case I'd say. But what about the case I mentioned? That is far more common. How would we handle that if we are always redirect users to index?

zpoken: add this to useEffect with hashedPassword and isInit, add the value isInit, as an example, which will become true when we create the first account, because it's a bad idea to look at the length of the array of accounts (there will be many in the future)

grod220: I just don't see a case where a user has two login screens open. One in the popup and one in the page. When a user is logging in, they are 99% going to be doing so through the popup.

zpoken: We need to think about such issues, because there will be users who want to break something! our tester, for example)

zpoken: or refuse tab mode

grod220: This solution: https://github.com/penumbra-zone/web/pull/20#discussion_r1306034616, does not address the 99% case. Let's say a user on the popup clicks settings. They get directed to the page, but they have a password already and that means they get redirected to the index page. Essentially, this prevents users from getting to deeper pages.

zpoken:
```
export const pageRoutes: RouteObject[] = [
  {
    element: <PageLayout />,
```

```

export const PageLayout = () => {
  const { hashedPassword } = useStore(passwordSelector);

  useEffect(() => {
    console.log(hashedPassword);
  }, [hashedPassword]);

  return <Outlet />;
};
```
no, it works only when the hashedPassword value changes

grod220: In this case, let's drop this case:

```
if(hashedPassword) navigate(PagePath.INDEX)}
```

Adds complexity with almost no benefit.

Vadim121197: Is it possible to create a 3 layouts? for each case

grod220: Ah, if we care about that specific use case for the login page, let's just put the useEffect there and not in the layout component for every page.

Vadim121197: I'm sorry, I've been on the project for a long time now, and I use tab and popup only for develop and test new feature at the same time! You're right, new users will use 1 of the contexts, not 2 together

apps/extension/src/routes/popup/login.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <FadeTransition>
+      <div className='grid gap-4 p-6 text-white shadow-sm'>
+        <p className='text-center text-2xl font-semibold leading-none tracking-tight'>
+          Welcome back!
+        </p>
+        <p className='text-center text-sm text-muted-foreground'>A decentralized network awaits</p>
+        <div className='grid gap-4 p-6 pt-0'>
+          <LoginForm />
+          <Button
+            className='text-white'
```
grod220, line 16: This and the support text beneath should be a gray color like `text-muted-foreground`

apps/extension/src/routes/popup/login.tsx
```diff
@@ -0,0 +1,42 @@
+import { Button } from 'ui/components';
+import { FadeTransition, LoginForm } from '../../components';
+import { PagePath } from '../page/paths';
+
+export const Login = () => {
```
grod220, line 5: We should think about the design of this page later. It is missing a kind of branding element.

apps/extension/src/components/login-form.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setInputValue(e.target.value);
+    setEnteredIncorrect(false);
+  };
+
+  return (
+    <form
+      onSubmit={e => {
+        e.preventDefault();
+        handleUnlock();
+      }}
```
grod220, line 37: You should do:
```
onSubmit={handleUnlock}
```
And the function handles the event

apps/extension/src/routes/page/router.tsx
```diff
@@ -23,8 +21,19 @@ export const pageRoutes: RouteObject[] = [
...
+        path: PagePath.RESTORE_PASSWORD,
+        element: <RestorePasswordIndex />,
+        children: restorePasswordRoutes,
+        loader: noPasswordIndexLoader,
+      },
+      {
+        path: PagePath.LOGIN,
+        element: <Login />,
+        loader: noPasswordIndexLoader,
+      },
```
grod220, line 34: These loaders are another case we don't anticipate and I don't think we need to guard for.

apps/extension/src/components/login-form.tsx
```diff
@@ -0,0 +1,56 @@
...
+    })();
+  };
+
+  const handleChangePassword: InputProps['onChange'] = e => {
+    setInputValue(e.target.value);
+    setEnteredIncorrect(false);
+  };
+
+  return (
+    <form onSubmit={handleUnlock} className='grid gap-4'>
```
grod220, line 33: Your `handleUnlock` should take an event and preventdefault() right? Or does this not matter?

apps/extension/src/routes/page/onboarding/routes.tsx
```diff
@@ -14,21 +16,26 @@ export const onboardingRoutes = [
   {
     path: PagePath.GENERATE_SEED_PHRASE,
     element: <GenerateSeedPhrase />,
+    loader: onboardingIndexLoader,
```
grod220, line 19: Did you see my last comment on the `onboardingIndexLoader`?

apps/extension/src/routes/page/login.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const noPasswordIndexLoader = async () => {
+  const accounts = await localExtStorage.get('accounts');
+  if (!accounts.length) return redirect(PagePath.WELCOME);
+
+  const password = await sessionExtStorage.get('hashedPassword');
+  if (password) return redirect(PagePath.INDEX);
+
+  return null;
+};
```
grod220, line 17: This is unused right? Can delete.

apps/extension/src/routes/page/onboarding/routes.tsx
grod220: Think you should delete it

apps/extension/src/routes/page/router.tsx
zpoken: I don't understand why we have to remove noPasswordIndexLoader and onboardingIndexLoader! a user without them can enter links and be thrown to any link he shouldn't see! for example, a logged in user can go to the login page

apps/extension/src/routes/page/onboarding/routes.tsx
zpoken: I don't understand why we have to remove noPasswordIndexLoader and onboardingIndexLoader! a user without them can enter links and be thrown to any link he shouldn't see! for example, a logged user can go to the login page

apps/extension/src/routes/page/router.tsx
zpoken: I understood that you used keplr as a basis and you can add a few seed phrases to them

apps/extension/src/routes/page/onboarding/routes.tsx
zpoken: I understood that you used keplr as a basis and you can add a few seed phrases to them

zpoken: if we want to use the keplr approach then we should use tab mode only for creating a new account! all other functionality should be moved to popup! because now I made an example of metamask

apps/extension/src/components/login-form.tsx
grod220: I thought so, so you need to add it to `handleUnlock()`

apps/extension/src/routes/page/onboarding/routes.tsx
grod220: How is it possible a user ends up on these pages they shouldn't be on? Unless it's a security issue, I don't think we need to guard for events that are effectively not possible.

Whether we use tabs or popup should entirely be on the best UX possible for our users.

apps/extension/src/components/login-form.tsx
zpoken: handleUnlock used as button handler too

grod220: Oh, in that case, you should change the button to something like:

```
<button type="submit">Submit</button>
```

Right? And then `handleUnlock()` would only be in one place. And then you can do `e.preventDefault()` in `handleUnlock()`.

zpoken: i see

apps/extension/src/routes/page/onboarding/routes.tsx
zpoken: we need to decide how our extensions should work, we are currently looking at different prototypes (Metamask and keplr)

grod220 (approved):

PR #23, opened 2023-08-28
testing the receipt of different keys

zpoken:


packages/wasm/src/keys.test.ts
```diff
@@ -1,17 +1,87 @@
...
       }).not.toThrow();
     });
+
+    it('equal to penumbraspendkey1fcesjzswatt6g6hkr0jvu6svc8xl7zq3yzzxrnkaz470nvc039ts8c3e4d', () => {
+      const spendKey = generateSpendKey(seedPhrase);
+      expect(spendKey).equal(
+        'penumbraspendkey1fcesjzswatt6g6hkr0jvu6svc8xl7zq3yzzxrnkaz470nvc039ts8c3e4d',
+      );
+    });
+  });
```
grod220, line 30: We do not need to test implementation in these tests. We just need to validate types.

grod220 (approved):

PR #26, opened 2023-09-06
Settings

zpoken:


apps/extension/src/routes/popup/index.tsx
```diff
@@ -24,26 +26,21 @@ export const popupIndexLoader = async () => {
 };
 
 export const PopupIndex = () => {
```
grod220, line 28: Something changed, I think here, where we lost our secondary font kind. Looks like all text is now `Faktum`. We should figure out what did that. Less important text should be the secondary font.

apps/extension/src/routes/popup/popup-layout.tsx
```diff
@@ -1,10 +1,16 @@
...
 
 export const PopupLayout = () => {
+  const location = useLocation();
   return (
-    <div className='relative bg-charcoal'>
-      <div className='relative z-10'>{<Outlet />}</div>
-      <div className='absolute inset-0 z-0 rounded-2xl bg-card-radial p-6 opacity-20' />
+    <div className='relative flex flex-col bg-charcoal'>
+      <div className='relative z-10'>
+        {(location.pathname as PagePath) === PagePath.INDEX && <Header />}
```
grod220, line 10: If this only exists on that route, just add it to PopupIndex component instead

apps/extension/src/routes/popup/paths.ts
```diff
@@ -1,4 +1,11 @@
 export enum PopupPath {
   INDEX = '/',
   LOGIN = '/login',
+  SETTINGS = '/settings',
+  SETTINGS_INDEX = '/settings/',
```
grod220, line 5: Why have two? We should get rid of one

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+  {
+    title: 'Settings',
+    icon: <MixerHorizontalIcon className='h-5 w-5 text-foreground' />,
+    href: PopupPath.SETTINGS,
+  },
+  {
+    title: 'Contact Information',
+    icon: <PersonIcon className='h-5 w-5 text-foreground' />,
+    href: PopupPath.SETTINGS_CONTACTS,
+  },
```
grod220, line 24: `Contact information` is not something we'd have in the wallet. Think we can get rid of this.

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+];
+
+export const SettingSheet = () => {
+  const navigate = usePopupNav();
+  const { clearSessionPassword } = useStore(passwordSelector);
```
grod220, line 29: Let's put a space below

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <SizeIcon className='h-5 w-5 text-foreground' />
+          </Button>
+        </div>
+        <SheetFooter>
+          <Button
+            onClick={() => {
+              clearSessionPassword();
+              navigate(PopupPath.LOGIN);
+            }}
+          >
```
grod220, line 70: This button color is too bold for this action. Think we should have an outline button or a gray color that is more subtle.

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const SettingSheet = () => {
+  const navigate = usePopupNav();
+  const { clearSessionPassword } = useStore(passwordSelector);
+  return (
+    <Sheet>
+      <SheetTrigger>
+        <HamburgerMenuIcon className='h-6 w-6 hover:opacity-50' />
+      </SheetTrigger>
+      <SheetContent side='left'>
+        <SheetHeader className=''></SheetHeader>
```
grod220, line 36: If it's empty, do we still need it here?

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+    href: PopupPath.SETTINGS_CONTACTS,
+  },
+];
+
+export const SettingSheet = () => {
+  const navigate = usePopupNav();
+  const { clearSessionPassword } = useStore(passwordSelector);
+  return (
+    <Sheet>
+      <SheetTrigger>
```
grod220, line 32: It seems the sheet slides in nicely with an animation, but it does not slide out nicely. How do we get that?

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+              void (async function () {
+                await chrome.tabs.create({
+                  url: `${chrome.runtime.getURL('page.html')}#${PagePath.INDEX}`,
+                });
+              })()
+            }
+          >
+            <p className='text-foreground'>Expand View</p>
+            <SizeIcon className='h-5 w-5 text-foreground' />
+          </Button>
```
grod220, line 62: Don't think `exand view` is something we need anymore

apps/extension/src/routes/popup/settings/settings.tsx
```diff
@@ -0,0 +1,61 @@
...
+    title: 'Networks',
+    icon: <GlobeIcon className='h-5 w-5 text-foreground' />,
+    href: PopupPath.SETTINGS_NETWORKS,
+  },
+  {
+    title: 'Permission',
+    icon: <TextAlignLeftIcon className='h-5 w-5 text-foreground' />,
+    href: PopupPath.SETTINGS_PERMISSION,
+  },
+];
```
grod220, line 39: Add space below

apps/extension/src/routes/popup/settings/settings-advanced.tsx
```diff
@@ -0,0 +1,13 @@
+import { FadeTransition } from '../../../components';
+import { BackIcon } from 'ui/components';
+import { usePopupNav } from '../../../utils/navigate';
+
+export const SettingsAdvanced = () => {
```
grod220, line 5: Unfortunately, this design doesn't quite work so well. 

We have a settings sheet that pops out, but then it links to pages that have completely different backgrounds.

I think we should keep all settings within the sheet. If it is not enough space, we should make the sheet full width.

apps/extension/src/components/setting-sheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { usePopupNav } from '../utils/navigate';
+import { PopupPath } from '../routes/popup/paths';
+import { PagePath } from '../routes/page/paths';
+
+const links = [
+  {
+    title: 'Settings',
+    icon: <MixerHorizontalIcon className='h-5 w-5 text-foreground' />,
+    href: PopupPath.SETTINGS,
+  },
```
grod220, line 19: I'm not sure why we have a sheet with icons that link to another page with icons. Let's just keep all the icons on the sheet and not link to another page.

apps/extension/src/components/header.tsx
```diff
@@ -0,0 +1,13 @@
+import { AccountsSheet } from './accounts-sheet';
+import { SettingSheet } from './setting-sheet';
+
+export const Header = () => {
```
grod220, line 4: Let's call this SettingsHeader

apps/extension/src/components/header.tsx
grod220, line 1: In general, what do you think about the `components/` folder. Does it make sense? I wonder if we should just co-locate all these components next to the routes they are associated with. Hmm. Not sure, what do you think?

apps/extension/src/routes/popup/paths.ts
grod220: Why did I have two lol? Is this required? If not, let's get rid of both.

apps/extension/src/components/setting-sheet.tsx
grod220: Makes sense to me

apps/extension/src/components/header.tsx
grod220: It's not urgent, but we should think about this as times goes on to see if there are better methods

apps/extension/src/components/setting-sheet.tsx
grod220: We should only use tab mode for things that require larger screens. Think onboarding makes sense and password recovery. Not sure we need it for anything else.

apps/extension/src/routes/page/index.tsx
```diff
@@ -18,6 +18,8 @@ import { sessionExtStorage } from '../../storage/session';
 // Will redirect to onboarding if necessary.
 export const pageIndexLoader = async () => {
   const accounts = await localExtStorage.get('accounts');
+  console.log(accounts);
+
```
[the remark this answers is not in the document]
zpoken: +

apps/extension/src/components/setting-sheet.tsx
zpoken: but you asked in the fig if we need contact information

apps/extension/src/routes/popup/paths.ts
zpoken: it`s not required)

apps/extension/src/components/setting-sheet.tsx
zpoken: there - https://www.figma.com/file/iTnIhPxUm4ZleTRKWEn4GM/Penumbra?node-id=2426%3A7424&mode=dev

grod220: I'm confused, do we need it or not? We don't currently have contacts. Don't think it's a priority.

apps/extension/src/components/setting-sheet.tsx
zpoken: https://github.com/shadcn-ui/ui/issues/72. This issue is still open, but the solution is to add "data-[state=closed]:fill-mode-forwards data-[state=open]:fill-mode-forwards"  where animation use. maybe you will come across such a bug someday

apps/extension/src/components/setting-sheet.tsx
zpoken: Now we don't need it! Previously, we had the ability to send a transaction from the extension and we didn't remove it because we thought that maybe the opinion would change

apps/extension/src/components/header.tsx
zpoken: It's not urgent, then does your comment have any validity?

apps/extension/src/routes/popup/settings/settings-advanced.tsx
zpoken: What do you think that when reviewing PR, not to pay an attention to the design, features that we transfer from the previous version? At the moment, we do not have an approved design that I could transfer to our repository, I only transfer the logic from the previous version

apps/extension/src/components/header.tsx
grod220: It's more of a general questions we need to think about and not actionable unless we have a good idea

apps/extension/src/routes/popup/settings/settings-advanced.tsx
grod220: I suppose it's fine for now, but in general we should always be auditing design decisions that get added to the codebase to make sure it's the highest quality code

zpoken: I understand that, but design issues are not my responsibility until we approve it.

grod220:
Btw, when you are ready for me to re-review, please request it. Will wait for you 👍

apps/extension/src/components/setting-sheet.tsx
grod220: Let's address this after the designs are finalized

grod220 (approved):
Approved, but please fix those last things I mentioned before merging. The font issue is the most important.

grod220:
Did you see my comment? It wasn't addressed as requested and just merged. Was that a mistake?

grod220:
@zpoken? ⬆️

PR #38, opened 2023-09-20
Password crypto updates

grod220:
After pairing with @plaidfinch, found some helpful things to improve relating to encrypting the password:

- Using bcrypt (slow hashing algorithm)
- Only storing password in `chrome.storage.session` (in-memory storage)


packages/crypto/src/encryption.ts
```diff
@@ -1,26 +1,32 @@
...
 
-export const decrypt = (ciphertext: string, key: string) => {
-  const decrypted = AES.decrypt(ciphertext, key);
-  return decrypted.toString(Utf8);
+// If unsuccessful, `decryptedStr` will be an empty string.
+// This function coerces that to `undefined` to make it clear to the consumer.
+export const decrypt = (ciphertext: string, key: string): string | undefined => {
+  const decryptedArr = AES.decrypt(ciphertext, key);
+  const decryptedStr = decryptedArr.toString(Utf8);
+  return decryptedStr === '' ? undefined : decryptedStr;
```
grod220, line 31: Previously, there was an assumption that an incorrect decryption would throw. But that is not the case! It just returns `""`. So checking for that and returning `string | undefined` as a means of communicating success/failure.

packages/crypto/src/encryption.ts
```diff
@@ -1,26 +1,32 @@
-import SHA256 from 'crypto-js/sha256';
 import AES from 'crypto-js/aes';
 import Utf8 from 'crypto-js/enc-utf8';
+import { compare, hash } from 'bcryptjs';
 
-const STRETCH_ROUNDS = 5000;
+export const SALT = '$2a$10$iLpsQfxfYc3FKZZft82qTe';
```
grod220, line 5: We could use bcryptjs's `genSalt()` method, but that generates a random one each time. This makes it impossible to rehash to the same value to decrypt with. So, making a fixed salt. I got this from running `genSalt()` once locally.

cronokirby:
Also, after reading the cryptojs code, I confirmed my suspicion that the API doesn't use authenticated encryption which is not great. I wonder what can cause decryption to fail, but it's certainly possible for an adversary to muck with the bytes of the ciphertext and cause decryption to *not* fail.

grod220:
For Argon2, there appears to be single package available that works in the browser: https://github.com/antelle/argon2-browser. However, they wrote this package in a way is not compatible with the way we load .wasm files via `wasm-pack`. They appear to require unorthodox webpack bundling config.

Think the best option is to write helpers in our wasm crate and expose them like we do elsewhere. These two libraries appear to be frequently used:
- For hashing password: https://crates.io/crates/argon2
- For encrypting seedphrase: https://docs.rs/aes-gcm/latest/aes_gcm/

Do these look ok?

grod220:
@plaidfinch @cronokirby can you give another look? There is a lot of react stuff in here, but I'd say these are the most important:
- docs/custody.md
- packages/crypto/src/encryption.ts

docs/custody.md
```diff
@@ -0,0 +1,18 @@
+# Custody
+
+### Password
+
+When setting up for the first time, you'll need to set a password.
+
+This password is hashed via `PBKDF2`, see [hashPassword() function](../packages/crypto/src/encryption.ts). It utilizes a pseudorandom function along with a salt value and iteration. The use of a salt provides protection against pre-computed attacks such as rainbow tables, and the iteration count slows down brute-force attacks.
+
+This hashed key+salt is stored in `chrome.storage.session` used for deriving spending keys for wallets later.
```
grod220, line 9: Given a possible scenario of multiple wallets, it may be easier to keep storing the password in session storage to derive the spend key on the fly. If not, we can store spend keys in session storage instead. However, at the moment, we don't have account group ids setup yet, so it's hard to associate spend keys with wallets.

packages/crypto/src/encryption.test.ts
```diff
@@ -1,68 +1,91 @@
...
-  const password = 'correcthorsebatterystaple';
+vi.stubGlobal('crypto', webcrypto);
 
-  test('original message can be recovered', () => {
-    const encrypted = encrypt(secretMessage, password);
-    const decrypted = decrypt(encrypted, password);
-    expect(decrypted).toBe(secretMessage);
+describe('encryption', () => {
+  const password = 's0meUs3rP@ssword';
+  const seedPhrase = 'correct horse battery staple';
```
cronokirby, line 10: Is there no validation that the seed phrase is the right length in our code?

packages/crypto/src/encryption.ts
```diff
@@ -1,26 +1,93 @@
...
+    return false;
   }
-  return hash.toString();
 };
 
-export const encrypt = (message: string, key: string) => {
-  return AES.encrypt(message, key).toString();
+export const encrypt = async (
+  message: string,
+  initializationVector: Base64Str,
```
cronokirby, line 52: Being able to choose the IV is a bit of a footgun.

It seems that the rest of the code we have using this does a reasonable thing and generates a random IV, but we could instead simplify that code by having this encryption function generate a random IV and prepend it to the ciphertext, and then decryption would grab the IV from the ciphertext + IV combo directly.

This would be opaque, and just an internal detail of how this little library encodes ciphertexts.

cronokirby (changes requested):
Ok, so I think we can refine the crypto api we provide a lot to make misuse harder.

We may want to keep the current layer we have as an internal shim though, but IMO these functions are a bit prone to misuse as is.

Here's an idea of organization I have:
image

You have 3 new types:
- A type representing keys
- A type representing key "signatures", maybe calling them "KeyPrint" would make more sense. This would consist of the hash of the key (SHA256 is fine) along with the salt, in the clear. The idea is that a key print itself is not a key, but allows one to check that a password matches the one used to derive this key.
- A type for ciphertexts (which include the nonce / IV inside the ciphertext, opaquely)

Then you'd have 4 functions:
```ts
function create(password: Password): [Key, KeyPrint];
function recreate(password: Password, print: KeyPrint): Key | null;
function seal(key: Key, data: Data): Box;
function unseal(key: Key, box: Box): Data | null;
```

I think this makes using the API simpler, because then you just need to store the KeyPrint somewhere to be able to recreate the key with the password, but the KeyPrint itself is not sensitive data, unlike the current code which has a struct bundling the key itself with a salt, but the salt needs to be detached from the struct to be saved alone. Boxes also store the IV, simplifying the API for users.

packages/crypto/src/encryption.ts
```diff
@@ -1,26 +1,188 @@
...
-// In cryptography, key stretching techniques are used to make brute-force attacks more difficult.
-// The idea is to "stretch" the key (or in this case, the password) by applying a computationally intensive
-// operation to it multiple times. This makes each attempt to guess the password more time-consuming,
-// which in turn makes brute-force attacks less feasible.
-export const repeatedHash = (message: string) => {
-  let hash = SHA256(message);
-  for (let i = 0; i < STRETCH_ROUNDS; i++) {
-    hash = SHA256(hash.toString());
-  }
-  return hash.toString();
```
cronokirby, line 16: State of the art in 1987

cronokirby (approved):
Just a small simplification to isPassword, otherwise looks good to me :)

PR #48, opened 2023-09-28
Firebase hosting

zpoken:


apps/webapp/.firebase/hosting.cHVibGlj.cache
grod220, line 1: Think this can be in `.gitignore`

apps/webapp/next.config.js
```diff
@@ -1,18 +1,32 @@
...
+  const prodConfig = !isDev ? { output: 'export' } : {};
 
-    return config;
-  },
+  return {
+    ...prodConfig,
+    reactStrictMode: true,
+    transpilePackages: ['ui'],
+    images: {
+      unoptimized: true,
```
grod220, line 10: Why unoptimized?

apps/webapp/next.config.js
```diff
@@ -1,18 +1,32 @@
...
-    config.module.rules.push({
-      test: /\.mp4$/,
-      type: 'asset/resource',
-      generator: {
-        filename: 'videos/[hash][ext][query]',
-      },
-    });
+module.exports = () => {
+  const isDev = process.env.NODE_ENV !== 'production';
+  const prodConfig = !isDev ? { output: 'export' } : {};
```
grod220, line 3: Would do this a bit differently:

```javascript
  const isProd = process.env.NODE_ENV === 'production';

  /**
   * @type {import('next').NextConfig}
   */
  const nextJsConfig = {
    output: isProd ? 'export' : undefined,
    // and so on
  };

  return nextJsConfig;
```
Adding the type will help ensure we don't add the wrong config.

apps/webapp/tsconfig.json
```diff
@@ -1,8 +1,12 @@
 {
   "extends": "tsconfig/nextjs.json",
   "compilerOptions": {
-    "plugins": [{ "name": "next" }]
```
grod220, line 4: I would keep it in the slimmer format. Multiple lines is not necessary here.

firebase.json
grod220, line 1: Is it possible to have the firebase configs to be within `apps/webapp`? Just because at the top level of the monorepo it's not so concerned about how individual apps are published. In the future, this repo could have even more apps that need different ways to deploy.

package.json
```diff
@@ -10,14 +10,16 @@
     "test": "turbo run test",
     "clean": "turbo clean",
     "format": "prettier --write .",
-    "format-check": "prettier --check ."
+    "format-check": "prettier --check .",
+    "host": "turbo run build --filter=webapp && turbo run host --filter=webapp"
```
grod220, line 14: This line is more relevant to the specific app. I would not recommend putting this at the top level of the monorepo.

package.json
```diff
@@ -10,14 +10,16 @@
...
   },
   "devDependencies": {
     "@turbo/gen": "^1.10.14",
     "@typescript-eslint/eslint-plugin": "^6.7.3",
     "eslint": "^8.50.0",
     "eslint-config-custom": "workspace:*",
     "prettier": "^3.0.3",
-    "turbo": "^1.10.14"
+    "turbo": "^1.10.14",
+    "firebase-tools": "^12.6.0"
```
grod220, line 23: Same feedback on firebase tools.

.gitignore
```diff
@@ -34,3 +34,6 @@ yarn-error.log*
 
 # vercel
 .vercel
+
+# firebase
+.firebase
```
grod220, line 39: New line after `.firebase`

grod220 (commented):
Also, can you add a github action for deploying the webapp on master merges?

apps/webapp/next.config.js
zpoken: without this, image from domain (avatar.vercel.sh) doesnt show

grod220: This does not seem right. We don't want to hurt all of our images just for that one. Think this requires more investigation on how to get that avatar to show correctly. 

For example, you could do:

```typescript
import Image from 'next/image'
 
const UnoptimizedImage = (props) => {
  return <Image {...props} unoptimized />
}
```

if that avatar is giving problems.

.gitignore
zpoken: just add space?

package.json
zpoken: should we call pnpm host from apps/webapp? not from root

.gitignore
grod220: Screenshot 2023-09-28 at 12 17 14 PM

Yes, add a space below. You can see github giving a warning when you don't.

package.json
grod220: Yes, because the way the app is hosted is not a concern to the monorepo, but to that app specifically.

apps/webapp/next.config.js
zpoken: I'll add this as a separate component, in case you need it in the future

apps/webapp/.firebase/hosting.cHVibGlj.cache
zpoken: We'll need to create a separate firebase, which is currently my test account

grod220: We can use anything for now. When we are ready to publish to `app.testnet.penumbra.zone`, ideally we would just need to swap a few small things in the github action and we'd be good to go.

zpoken: i`ll add config from previous dapp

zpoken: now my personal card is connected to this account)

.github/workflows/deploy-firebase-dapp.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+    name: Deploy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v2
+
+      - name: Build static site
+        run: ./build-static-site
```
conorsch, line 24: The `build-static-site` script existed in the dapp repo (https://github.com/penumbra-zone/penumbra_dapp/blob/632df2679e03490a27f396d118eba11dbf79e202/build-static-site), but doesn't exist here. What is the equivalent command to export the static site docroot? Is it `npm run build`, which I see in the `package.json`? I also tried running that command from within the `apps/webapp/` subdir, but it's not clear to me where the HTML docroot is exported, if anywhere.

conorsch: From [discussion in Discord](https://discord.com/channels/824484045370818580/1044457038886998137/1156975423859732562), the answer is: 

```
# from repo root:
pnpm build
# docroot available at `./apps/webapp/dist/`
```

zpoken: @conorsch Is the logic ready for us to close this PR?

conorsch: As of https://github.com/penumbra-zone/web/pull/48/commits/80b10f2dda035a3f6aa347e98d46ea7dd5539084, yes, I think we should merge. However, I see a conflict: the `pnpm-lock.yaml` needs to be updated. I'm not sure how to do that, so I've not rebased this PR to resolve. As soon as the conflict is resolved, it's OK to merge this PR, and I'll follow up with next steps on the Firebase config.

grod220 (approved):

PR #60, opened 2023-10-01
Assert root valid error

zpoken:


packages/indexdb/src/indexed-db.ts
```diff
@@ -28,7 +28,10 @@ export class IndexedDb implements IndexedDbInterface {
         db.createObjectStore('tree_last_position');
         db.createObjectStore('tree_last_forgotten');
         db.createObjectStore('tree_commitments', { keyPath: 'commitment.inner' });
-        db.createObjectStore('tree_hashes', { keyPath: 'hash' });
+        db.createObjectStore('tree_hashes', {
+          autoIncrement: true,
+          keyPath: 'index',
+        });
```
grod220, line 34: Why does this solve the issue? Is it because tree hashes were having redundant keys before? How is it possible the tree hash can be redundant though?

Also, don't think you need to specify a keypath:

```
db.createObjectStore('tree_hashes', { autoIncrement: true });
```

Also, can you add a code comment as documentation why this field is necessary?

Vadim121197: Sometimes function with updates returns tree_hashes array with same hash! As example- you can see it, when sync progress is 200 blocks!

grod220: Ah, that's true. Same hash at different positions/heights. Just add this code comment and we'll be good to merge:
```
// Can be the same hash for different positions. `autoIncrement` makes the item key an incremented index
db.createObjectStore('tree_hashes', { autoIncrement: true });
```

PR #63, opened 2023-10-02
Dapp design [Send logic]

zpoken:


apps/webapp/shared/responsive-image.tsx
```diff
@@ -0,0 +1,17 @@
+// https://nextjs.org/docs/messages/next-image-upgrade-to-13
+import Image from 'next/legacy/image';
```
grod220, line 2: Why are we using the legacy image? We should be using the latest.

apps/webapp/shared/responsive-image.tsx
```diff
@@ -0,0 +1,17 @@
...
+  className?: string;
+}
+
+export const ResponsiveImage = ({ src, alt, className }: ResponsiveImageProps) => {
+  return (
+    <div className={cn('relative', className)}>
+      <Image src={src} alt={alt} layout='fill' objectFit='cover' />
+    </div>
+  );
+};
```
grod220, line 17: Why is it called `ResponsiveImage`? This may need some code comments/documentation. 

With `objectFit='cover'` isn't there a chance things could get cut off?

apps/webapp/app/send/types.ts
```diff
@@ -0,0 +1,18 @@
+export enum SendPageTab {
+  SEND = 'send',
+  RECEIVE = 'receive',
+  IBC = 'ibc',
+}
+
+export type TabHelper = Record<
```
grod220, line 7: Let's call this something like `SendTabMap` instead

apps/webapp/app/send/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  {
+    src: string;
+    label: string;
+  }
+>;
+
+export interface Asset {
+  name: string;
+  icon: string;
+}
```
grod220, line 18: As an exported type `Asset` name is too generic as it relates only to this page

apps/webapp/app/header/notifications.tsx
```diff
@@ -38,16 +38,14 @@ export const Notifications = () => {
   return (
     <Popover>
       <PopoverTrigger className='relative'>
-        <div className='relative'>
-          <div className='absolute right-[2px] top-[5px] h-[10px] w-[10px] rounded-full bg-sand'></div>
-          <Image src='/bell.svg' width={30} height={30} alt='bell' />
-        </div>
+        <div className='absolute right-[2px] top-[5px] z-10 h-[10px] w-[10px] rounded-full bg-sand'></div>
+        <ResponsiveImage src='/bell.svg' alt='Bell' className='h-[30px] w-[30px]' />
```
grod220, line 42:
```diff
- alt='Bell'
+ alt='notification bell'
```

apps/webapp/app/send/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+export const assets: Asset[] = [
+  {
+    name: 'ETH',
+    icon: '/test-asset-icon.svg',
+  },
+  {
+    name: 'BNB',
+    icon: '/test-asset-icon-2.svg',
+  },
+];
```
grod220, line 27: Where did this asset list come from before? Was it hard coded like this?

apps/webapp/app/send/send-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <div className='flex items-center justify-between'>
+        <div className='flex items-start gap-2'>
+          <ResponsiveImage src='/incognito.svg' alt='Incognito' className='h-5 w-5' />
+          <p className='font-bold'>Hide Sender from Recipient</p>
+        </div>
+        <Switch id='sender-mode' checked={hidden} onCheckedChange={checked => setHidden(checked)} />
+      </div>
+      <Button type='submit' variant='gradient' className='mt-4'>
+        Send
+      </Button>
```
grod220, line 54: Let's make this button disabled if all the fields are not filled. This may be worth a zustand store if it's getting complicated.

apps/webapp/shared/helper-card.tsx
```diff
@@ -0,0 +1,27 @@
+import { Card } from 'ui';
+import { ResponsiveImage } from './responsive-image';
+
+interface HelperCardProps {
+  src: string;
+  label: string;
+}
+
+export const HelperCard = ({ src, label }: HelperCardProps) => {
```
grod220, line 9: Instead of helper, we should have something more descriptive. I think `LearnMoreCard` may be better. Or `EduInfoCard`.

apps/webapp/app/send/send-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+      />
+      <InputToken
+        label='Amount to send'
+        placeholder='Enter an amount'
+        className='mb-1'
+        asset={asset}
+        setAsset={setAsset}
+        value={amount}
+        onChange={e => setAmount(e.target.value)}
+      />
```
grod220, line 38: At the moment, this is allowing letters as input. Think we need to make use of input's

```
<input type="number">
```

if we can to help with field validation

apps/webapp/app/send/send-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <InputBlock
+        label='Memo'
+        placeholder='Enter the text'
```
grod220, line 41:
```diff
- placeholder='Enter the text'
+ placeholder='Optional message'
```

apps/webapp/shared/input-token/select-token-modal.tsx
```diff
@@ -0,0 +1,72 @@
...
+                placeholder='Search asset...'
+              />
+            </div>
+            <div className='flex justify-between items-center font-headline text-base font-semibold mt-2'>
+              <p>Token name</p>
+              <p>Balance</p>
+            </div>
+            <div className='flex flex-col gap-2'>
+              {filteredAsset.map(asset => (
+                <DialogPrimitive.Close key={asset.name}>
```
grod220, line 52: This appears to have left/right margin. Think it can extend to the end.

apps/webapp/shared/input-token/select-token-modal.tsx
```diff
@@ -0,0 +1,72 @@
...
+  setAsset: Dispatch<SetStateAction<Asset>>;
+}
+
+export const SelectTokenModal = ({ asset, setAsset }: SelectTokenModalProps) => {
+  const [search, setSearch] = useState('');
+
+  const filteredAsset = useMemo(() => {
+    if (!search) return assets;
+    return assets.filter(asset => asset.name.toLowerCase().includes(search.toLowerCase()));
+  }, [search]);
```
grod220, line 21: Can you also sort this to where the assets with the most funds are at the top? And then if zero in the wallet, it can be alphabetical.

apps/webapp/shared/helper-card.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <Card gradient className='p-5 row-span-1 gap-2'>
+      <div className='flex gap-2'>
+        <ResponsiveImage src={src} alt='icons' className='w-8 h-8' />
+        <p className='text-2xl leading-9 font-bold bg-text-linear bg-clip-text text-transparent font-headline'>
+          {label}
+        </p>
+      </div>
+      <p className='text-muted-foreground'>
```
grod220, line 18: This needs more margin above to give it space from the headline.

apps/webapp/app/send/send-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+        e.preventDefault();
+      }}
+    >
+      <InputBlock
+        label='Recepient'
+        placeholder='Enter the address'
+        className='mb-1'
+        value={recepient}
+        onChange={e => setRecepient(e.target.value)}
+      />
```
grod220, line 29: We need some input validation that this is a valid penumbra address. Maybe there is some regex we can use to verify it.

apps/webapp/app/send/types.ts
zpoken: this is test data, in future we'll have another types

apps/webapp/app/header/notifications.tsx
zpoken: I just started working here! I'm adding states for this icon

apps/webapp/app/send/constants.ts
zpoken: it test data

apps/webapp/shared/input-token/select-token-modal.tsx
zpoken: ok, but it from figma

apps/webapp/shared/responsive-image.tsx
zpoken: I had an error related to this - https://nextjs.org/docs/messages/next-image-upgrade-to-13

apps/webapp/shared/responsive-image.tsx
zpoken: it's just a wrapper! maybe call it ImageWrapper?

apps/webapp/shared/responsive-image.tsx
zpoken: but there is new example - https://nextjs.org/docs/app/building-your-application/optimizing/images#fill-container! i`ll add this example

apps/webapp/app/send/send-form.tsx
zpoken: should we use zustand to store this data?

apps/webapp/app/send/send-form.tsx
grod220: Yes, that would be nice to have. That's probably why zustand may be easier to use for this as it makes testing easier as well.

apps/webapp/app/send/send-form.tsx
grod220: I'd probably say so. Like a new `send` slice.

apps/webapp/shared/responsive-image.tsx
grod220: But even `ImageWrapper` is too generic. This is a special kind of image wrapper that does ObjectFit. Perhaps something like: `FilledImage`.

apps/webapp/utils/validation.ts
```diff
@@ -0,0 +1,8 @@
+const BECH32_PREFIX = 'penumbrav2t';
+const ADDRESS_LENGTH = 146;
+
+export const validateAmount = (value: string, balance: number) =>
+  Boolean(value) && Number(value) > balance;
+
+export const validateRecepient = (value: string) =>
```
grod220, line 7:
```diff
- validateRecepient
+ validateRecipient
```

apps/webapp/utils/validation.ts
```diff
@@ -0,0 +1,8 @@
...
+  Boolean(value) && (value.length !== ADDRESS_LENGTH || !value.startsWith(BECH32_PREFIX));
```
grod220, line 8: is `Boolean(value)` necessary? I think that should be caught with `value.length` test right?

apps/webapp/utils/validation.ts
grod220: Also, can you write tests for this.

grod220: We should make a note for later to make use of this in the wasm crate:

```rust
impl std::str::FromStr for Address {
    type Err = anyhow::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        pb::Address {
            inner: bech32str::decode(s, bech32str::address::BECH32_PREFIX, bech32str::Bech32m)?,
            alt_bech32m: String::new(),
        }
        .try_into()
    }
}
```

and then expose it to be used here. That is probably the source of truth method.

apps/webapp/vite.config.ts
grod220, line 1: This looks unused. It is not required. Can remove.

apps/webapp/utils/validation.ts
zpoken: Знімок екрана 2023-10-03 о 18 28 46

without Boolean(value) the error is displayed immediately

apps/webapp/utils/format-number.ts
```diff
@@ -0,0 +1,5 @@
+export const formatNumber = (number: number) =>
```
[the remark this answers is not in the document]
zpoken: Знімок екрана 2023-10-03 о 18 40 26

like this)

apps/webapp/state/send.test.ts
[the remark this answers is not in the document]
zpoken: cool)

apps/webapp/app/send/send-form.tsx
```diff
@@ -0,0 +1,97 @@
...
+    setRecepient,
+    setMemo,
+    setHidden,
+  } = useStore(sendSelector);
+
+  useEffect(() => {
+    // Test logic!!!
+    setAsset(assets[0]!);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
zpoken: at the moment it should be, here the asset is set at initialisation, in the future we will load the list of assets from the extension and set it in the same way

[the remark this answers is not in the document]
grod220: Instead of `Test logic!!!`, think it would be better to add a code comment exactly what you mentioned here. Also, an exclamation point is something that could cause concern. Not necessary here.

apps/webapp/shared/filled-image.tsx
```diff
@@ -0,0 +1,23 @@
+import Image from 'next/image';
+import { cn } from 'ui/lib/utils';
+
+interface FilledImageProps {
+  src: string;
+  alt: string;
+  className?: string;
+}
+
+export const FilledImage = ({ src, alt, className }: FilledImageProps) => {
```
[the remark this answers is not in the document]
grod220: Instead of one long line, can you make this comment multiple lines so it doesn't require scrolling right?

apps/webapp/utils/validation.ts
```diff
@@ -0,0 +1,20 @@
...
+//   type Err = anyhow::Error;
+
+//   fn from_str(s: &str) -> Result<Self, Self::Err> {
+//       pb::Address {
+//           inner: bech32str::decode(s, bech32str::address::BECH32_PREFIX, bech32str::Bech32m)?,
+//           alt_bech32m: String::new(),
+//       }
+//       .try_into()
+//   }
+// }
```
grod220, line 17: Instead of pasting code, just point to the lines in github: https://github.com/penumbra-zone/penumbra/blob/8d1644620779ddfd961e58f0f4703318b3d08910/crates/core/keys/src/address.rs#L201-L211

grod220 (approved):
Looks good! Just have a few more things. After you make those changes, feel free to merge.

PR #81, opened 2023-10-05
Dapp design

zpoken:
pages: 
1.Dashboard : Assets, Transactions
2.Swap
3.Send: ibc


apps/webapp/app/swap/swap-form/index.tsx
grod220, line 1: Instead of being empty. We should start with default in/out denoms.

Screenshot 2023-10-06 at 9 59 46 AM

apps/webapp/app/swap/swap-form/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      onSubmit={e => {
+        e.preventDefault();
+      }}
+    >
+      <div className='relative flex flex-col gap-2'>
+        <SwapInput
+          placeholder='Enter an amount'
+          asset={pay.asset}
+          setAsset={setAsset(SwapToken.PAY)}
+          value={pay.amount}
```
grod220, line 24: Currently not able to enter in two amounts:
https://github.com/penumbra-zone/web/assets/16624263/5d81f793-57da-46ee-a31a-684b788a997e

apps/webapp/app/swap/swap-form/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        disabled={
+          !pay.asset ||
+          !pay.amount ||
+          !receive.asset ||
+          !receive.amount ||
+          !!Object.values(validationErrors).find(Boolean)
+        }
+      >
+        {validationErrors.pay ? `Insufficient ${pay.asset?.name} balance` : 'Swap'}
+      </Button>
```
grod220, line 91: When I enter an amount at the top and click the ↕️ button, I get an `undefined` error in the button

Screenshot 2023-10-06 at 10 04 56 AM

grod220: Or maybe because `Select Asset` is on top now? With default denoms for both, we shouldn't have `Select Asset` anymore.

apps/webapp/state/ibc.ts
```diff
@@ -0,0 +1,52 @@
+import { AllSlices, SliceCreator } from '.';
+import { Asset, Chain } from '../types/asset';
+import { validateAmount } from '../utils';
+
+export interface IbcValidationFields {
+  amount: boolean;
+}
+
+export interface IbcSlice {
```
grod220, line 9: Let's call this `IbcSendSlice`. In the future, we may have more IBC-related slices.

apps/webapp/app/send/ibc-form.tsx
```diff
@@ -0,0 +1,81 @@
...
+              ]
+            : undefined
+        }
+      />
+      <div className='flex flex-col gap-3 rounded-lg border bg-background px-4 pb-5 pt-3'>
+        <p className='text-base font-bold'>Chain</p>
+        <Select
+          value={chain?.name ?? ''}
+          onValueChange={e => setChain(chains.find(i => i.name === e))}
+        >
```
grod220, line 48: Screenshot 2023-10-06 at 10 16 28 AM

We need to revamp this design. 
- It kinda blends in with the item. Perhaps it should be a dark gray.
- On hover, the background of that row should be highlighted
- I don't think it's wide enough. It should be the width of the thing being selected.
- If the item is already selected, we can just remove it from the drop down list (at the moment we have that check)

apps/webapp/app/send/ibc-form.tsx
```diff
@@ -0,0 +1,81 @@
...
+        e.preventDefault();
+      }}
+    >
+      <InputToken
+        label='Amount to send'
+        placeholder='Enter an amount'
+        className='mb-1'
+        asset={asset}
+        setAsset={setAsset}
+        value={amount}
```
grod220, line 25: Screenshot 2023-10-06 at 10 20 46 AM

Is the font of the amount correct? 🤔

grod220 (commented):
Given we already have the balances query ready, it would be nice to integrate it so it shows actual balances. It doesn't have to be this PR, but maybe the next one.

apps/webapp/app/dashboard/assets-table.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <TableBody>
+        {assets.map(i => (
+          <TableRow key={i.name}>
+            <TableCell>
+              <div className='flex items-center gap-4'>
+                <FilledImage src={i.icon} alt='Asset' className='h-6 w-6' />
+                <p className='text-base font-bold'>{i.name}</p>
+              </div>
+            </TableCell>
+            <TableCell className='text-center'>100%</TableCell>
```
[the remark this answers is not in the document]
zpoken: This will change when we have a reflection of real balances

apps/webapp/app/swap/swap-form/index.tsx
zpoken: this is currently a feature) ideally, the pay field should change automatically if the user changes the receive field and vice versa! so I'm still cleaning it up

apps/webapp/app/send/ibc-form.tsx
zpoken: I am waiting for the ui kit from Daria

apps/webapp/app/send/ibc-form.tsx
zpoken: I am waiting for the ui kit from Daria

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
zpoken: will we count % of the total number in usdc?

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
zpoken: https://www.figma.com/file/iTnIhPxUm4ZleTRKWEn4GM/Penumbra?node-id=2861%3A4350&mode=dev 
should it be the same here?

apps/webapp/app/swap/swap-form/index.tsx
zpoken: possible case when the user will have no assets, or only 1

apps/webapp/app/swap/swap-form/index.tsx
zpoken: will we be able to display the pay value when the user changes the receive value?

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
grod220: Correct. And the box should only grow to the size of the transactions inside of it. That design it looks far too big for those two transactions.

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
grod220: yes that's right

apps/webapp/app/swap/swap-form/index.tsx
grod220: Ah! That's right. Do you know if penumbra supports both exact out and exact in? Forgot about these. 🤔 

If we only support exact in, then we should not allow users to edit the second box at all.

apps/webapp/app/swap/swap-form/index.tsx
grod220: Yes it's possible that user cannot make a swap given they don't have any assets. But it's still nicer to show two assets pre-populated there. Osmosis does the same.

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
zpoken: I will do it when we connect the balances and all the tokens are equal to $1

apps/webapp/app/swap/swap-form/index.tsx
zpoken: then we need to get all the tokens. I remember you talking about this with your team earlier

apps/webapp/app/swap/swap-form/index.tsx
zpoken: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1alpha1#penumbra.core.component.dex.v1alpha1.SimulationService.SimulateTrade - it returns the value how much we can receive, but we can also get how much we need to pay, right?

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
zpoken: balances query doesnt return asset name and decimals! asset query is not working yet?

apps/webapp/app/swap/swap-form/index.tsx
grod220: That simulation endpoint implies we only support _exact in_. This means we provide the asset id + amount as input and just the asset id as an output. It then returns how much we can expect to receive.

apps/webapp/app/dashboard/assets-table.tsx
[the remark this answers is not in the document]
grod220: Yes, correct, the asset query still needs to be implemented. Maybe your next PR can be to do that + implement real balances in this UI.

apps/webapp/app/swap/swap-form/index.tsx
zpoken: if we pass the value from the second field to the input and the value from the first field to the output?Will it not work?

zpoken: 1 penumbra = 15 ugm
15 ugm = 1 penumbra
is it right?

grod220: > 1 penumbra = 15 ugm
This will be a dynamic number and always changing given market conditions.

zpoken: I gave this as an example

grod220: > if we pass the value from the second field to the input and the value from the first field to the output?Will it not work?

Not sure exactly sure what you mean, but if we only allow exact in, this means when we click that arrow, the amount should stay in the first input and not move to the second. We should not be able to edit the second input.

zpoken: Знімок екрана 2023-10-06 о 15 33 33

user input pay : input - firtst input value and output - second input value

Знімок екрана 2023-10-06 о 15 35 11

user input receive: input - second input value and output - first input value
Знімок екрана 2023-10-06 о 15 35 32

zpoken: Will it work?

grod220: @erwanor need some assistance here. Do I understand this right?

[This protobuf definition](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1alpha1#penumbra.core.component.dex.v1alpha1.SimulationService.SimulateTrade) says that you need to provide:
- input amount
- input asset id
- output asset id

And you'll receive a [SwapExecution](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1alpha1#penumbra.core.component.dex.v1alpha1.SwapExecution).

This will include the original input amount (right? the one we added ourselves) but now also include the output amount. Also, Penumbra only supports exact in right?

apps/webapp/app/swap/swap-form/index.tsx
zpoken: and I didn't pay attention to something, when you click on the replace button, do we have to change asset and amount?

grod220: Actually, we should copy the way Osmosis does this. If you click the button to change the in/out assets, it uses the estimated out amount/denom as the new denom in/amount.

zpoken: and the question of assets is open

Vadim121197: @grod220

grod220: What's the question? That you are missing an official assets list?

zpoken: i don' t have official assets list

grod220: The frontend will need to hardcode an assets list regardless that has things like token images and such. I'd recommend just doing that.

grod220: and support the case that there will be tokens that aren't in this list. Someone will always be able to IBC a token we've never seen before.

zpoken: Знімок екрана 2023-10-06 о 17 08 28

i know only this assets

zpoken: e will have the same type for assets as in IndexedDB, only instead of Uint8Array we will have string? + icon which will be optional

grod220: Yes, in indexeddb we have:

```
  assets: {
    key: Uint8Array;
    value: DenomMetadata;
  };
```

The frontend can query for this. However, it doesn't provide icons. So in theory, we'll need another asset list regardless to hold those. And we also need to support the case that assets will show up that are not in indexeddb.

zpoken: I'll just create an array with the asserts in packages/constants, where all 4 asserts will be, and then I can do PR for new asserts

apps/webapp/app/swap/swap-form/index.tsx
erwanor: @grod220 that's exactly it, and we only have support for exact in, though it's possible to build an exact out api on top of `SimulateTrade`. I can detail how async.

zpoken:
@grod220 I'll connect the display of real tokens, can you see what changes you would make?

apps/webapp/shared/select-token-modal.tsx
```diff
@@ -1,42 +1,29 @@
...
 
   return (
     <Dialog>
-      <DialogTrigger asChild>
-        <div className='flex items-center gap-2 bg-light-brown px-6 py-[6px] rounded-lg cursor-pointer'>
-          <FilledImage src={asset.icon} alt='Asset' className='w-6 h-6' />
-          <p className='text-base font-bold text-light-grey'>{asset.name}</p>
+      <DialogTrigger disabled={!assets.length}>
+        <div className='flex items-center justify-center gap-2 bg-light-brown w-[100px] h-9 rounded-lg'>
+          {asset.icon && <FilledImage src={asset.icon} alt='Asset' className='w-6 h-6' />}
```
grod220, line 25: I'm not sure if we can ever rely upon `asset.icon`. But I suppose it's best for now. Later, we'll want to add a default state if we don't have the icon.

apps/webapp/app/dashboard/assets-table.tsx
```diff
@@ -0,0 +1,66 @@
...
+                  {asset.icon && <FilledImage src={asset.icon} alt='Asset' className='h-6 w-6' />}
+                  <p className='text-base font-bold'>{asset.display}</p>
+                </div>
+              </TableCell>
+              <TableCell className='text-center'>{formatNumber(asset.percentageOf)}%</TableCell>
+              <TableCell className='text-center'>
+                <div className='flex flex-col'>
+                  <p>$1,577.15</p>
+                  <p className='text-[15px] font-normal leading-[22px] text-red'>
+                    -$15.19 (-0.95%)
```
grod220, line 47: If we have `test_usd` on testnet. We can use that price to calculate the actual values for this.

apps/webapp/utils/format-number.ts
```diff
@@ -1,11 +1,22 @@
 //  This function takes a number and formats it for display with exactly two decimal places.
 // If the number is an integer or has fewer than two decimal places, it will be displayed with exactly two decimal places. For example, if number is 2000, it will be formatted as "2,000.00".
 // If the number is a floating-point number with more than two decimal places, it will be rounded to exactly two decimal places. For example, if number is 2001.1, it will be formatted as "2,001.10" because it is rounded to two decimal places.
-export const formatNumber = (number: number): string =>
+export const formatNumber = (number: number): string => {
```
grod220, line 4: If there aren't any "cents" it should not show anything. For example:

Right now we have `8,372.00`. But given there aren't cents, we should just have `8,372`

apps/webapp/app/header/header.tsx
```diff
@@ -16,12 +16,10 @@ export const Header = () => {
         <Notifications />
         <NetworksPopover triggerClassName='px-[9px]' />
         <div className='ml-1 flex items-center gap-3 rounded-lg border px-5 py-[7px]'>
-          <UnoptimizedImage
-            src='https://avatar.vercel.sh/rauchg'
-            alt='icon'
-            width={20}
-            height={20}
-            className='rounded-full'
+          {/* TODO add name in hardcore valur  */}
```
grod220, line 19:
```diff

- {/* TODO add name in hardcore valur  */}
+ {/* TODO: replace hardcoded value  */}
```

apps/webapp/utils/calculate-balance.ts
```diff
@@ -0,0 +1,9 @@
+import { Asset } from 'penumbra-constants';
+import { LoHi } from 'penumbra-types';
+
+export const calculateBalance = (loHi: LoHi, asset: Asset): number => {
+  // found exponent from denomUnits by display name
+  const exponent = asset.denomUnits.find(denom => denom.denom === asset.display)?.exponent;
+
+  return (Number(loHi.lo) + 2 ** 64 * Number(loHi.hi)) / (exponent ? 10 ** exponent : 1);
+};
```
grod220, line 9: In `packages/types/src/lo-hi.ts`, we have `joinLoHi`. Let's use that. If that doesn't cut it, let's just add core logic that is missing there. Putting it in a package let's other apps/packages make use of it.

apps/webapp/app/send/ibc-form.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { FilledImage } from '../../shared';
+import { useStore } from '../../state';
+import { ibcSelector } from '../../state/ibc';
+import { calculateBalance, validateAmount } from '../../utils';
+import { chains } from './constants';
+const InputToken = dynamic(() => import('../../shared/input-token'), {
+  ssr: false,
+});
+
+export default function IbcForm() {
```
grod220, line 15: https://github.com/penumbra-zone/web/assets/16624263/a457a37b-920a-4cf1-b3e8-c0f950988cfa

There appears to be a flicker for this component.

apps/webapp/app/send/send-form.tsx
```diff
@@ -1,33 +1,57 @@
...
+      return;
+    }
+
+    const loHi: LoHi = {
+      lo: selectedAsset.balance?.amount?.lo ?? 0n,
+      hi: selectedAsset.balance?.amount?.hi ?? 0n,
+    };
+
+    setAssetBalance(calculateBalance(loHi, asset));
+  }, [data, end, asset, setAssetBalance]);
```
grod220, line 54: This looks like redundant code to the other code block. Can you extract this to a hook so it's re-used?

apps/webapp/app/swap/asset-statistics.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Asset } from 'penumbra-constants';
+import { useStore } from '../../state';
+import { swapSelector } from '../../state/swap';
+
+export const AssetStatistics = () => {
+  const { pay, receive } = useStore(swapSelector);
+
+  const [firstAsset, setFirstAsset] = useState<
+    (Asset & { price: number; '24h': number }) | undefined
+  >(undefined);
```
grod220, line 17: If it starts undefined, you don't need to pass it to the function fyi.

apps/webapp/app/swap/swap-form/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+
+    const loHi: LoHi = {
+      lo: selectedAsset.balance?.amount?.lo ?? 0n,
+      hi: selectedAsset.balance?.amount?.hi ?? 0n,
+    };
+
+    setAssetBalance(calculateBalance(loHi, pay.asset));
+  }, [data, end, pay.asset, setAssetBalance]);
```
grod220, line 42: Looks like this is used in three places?

apps/webapp/hooks/balances.ts
```diff
@@ -0,0 +1,24 @@
...
+    () =>
+      client.balances({
+        accountFilter: {
+          account,
+        },
+        // TODO receive asset balance by id
+        // extension receive inner as  {1:12,2:34.....}, need format Uint8Array
+        // assetIdFilter: {
+        // 	inner: base64ToUint8Array(assetId)
+        // }
```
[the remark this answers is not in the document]
zpoken: this is for the future, at the moment it doesn't work for 1 asset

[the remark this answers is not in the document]
zpoken: the extension does not receive the correct format

apps/webapp/app/dashboard/assets-table.tsx
zpoken: how we will know what the price is for each token?

apps/webapp/hooks/sorted-asset.ts
```diff
@@ -0,0 +1,73 @@
...
+import { useMemo } from 'react';
+import { uint8ArrayToBase64 } from 'penumbra-types';
+import { calculateBalance } from '../utils';
+
+export interface AssetBalances {
+  balance: number;
+  usdcValue: number;
+}
+
+export type AssetWithBalances = Asset & AssetBalances;
```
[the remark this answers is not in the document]
zpoken:
```
interface AssetWithBalance {
  denomMetadata: Pick<Asset, 'display' | 'icon' | 'penumbraAssetId'>;
  balance: AssetBalance;
}
```

it ok? @grod220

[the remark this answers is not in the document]
grod220: Yes, sounds good to me

apps/webapp/app/dashboard/assets-table.tsx
zpoken: @grod220

grod220: Let's leave this as another task. Ignore for now. We can use the swap estimates to do this later.

apps/webapp/utils/calculate-balance.ts
grod220: Yes, please transfer that to the hi/lo package and write tests for it. Also, please include documentation for what that formula is doing. Also, don't do `Number(loHi.lo)` and coerce it to a number in the first step. Do all the math as BigInt and at the very end you can coerce to a number. If you do it too early, you'll lose precision.

apps/webapp/app/send/ibc-form.tsx
zpoken: This is due to dynamic imports! looking for a solution

grod220: Ah ok. It may be trickier to fix than I thought. Feel free to skip this if it seems too difficult for now.

zpoken:
@grod220 error from extension `"ConnectError: [unknown] TypeError: Cannot read properties of undefined (reading 'accountFilter')"` when try to receive asset balance

grod220:
I'm not getting that error, can you confirm this is still an issue when you re-install the extension?

zpoken:
@grod220  Do you have asset on your balance?because i can not check is balance show right! bot ignore my message on discord channel)

grod220:
This seed phrase always has funds: https://rustdoc.penumbra.zone/main/penumbra_chain/test_keys/constant.SEED_PHRASE.html

zpoken:
Знімок екрана 2023-10-11 о 17 01 21

then asset doesnt save
@grod220

zpoken:
everything works well

apps/webapp/app/send/page.tsx
```diff
@@ -3,12 +3,18 @@
...
+const IbcForm = dynamic(() => import('./ibc-form'), {
+  ssr: false,
+});
+const SendForm = dynamic(() => import('./send-form'), {
+  ssr: false,
+});
 
 export default function Page() {
-  const [tab, setTab] = useState<SendPageTab>(SendPageTab.SEND);
+  const [tab, setTab] = useState<SendPageTab>(SendPageTab.IBC);
```
grod220, line 17: `SendPageTab.SEND` should be the default

apps/webapp/hooks/sorted-asset.ts
```diff
@@ -0,0 +1,91 @@
...
+import { useBalances } from './balances';
+import { useMemo } from 'react';
+import { Asset, calculateLoHiExponent, uint8ArrayToBase64 } from 'penumbra-types';
+
+export interface AssetBalance {
+  amount: number;
+  usdcValue: number;
+}
+
+// export type AssetWithBalances = Omit<Asset, ''> & AssetBalances;
```
grod220, line 11: Remove

apps/webapp/hooks/balances.ts
```diff
@@ -0,0 +1,24 @@
...
+      }),
+    [],
```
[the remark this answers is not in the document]
grod220: Still missing `[account]` in the deps

apps/webapp/hooks/balances.ts
```diff
@@ -0,0 +1,21 @@
...
+        },
+        // TODO receive asset balance by id
+        // extension receive inner as  {1:12,2:34.....}, need format Uint8Array
+      }),
+    [],
+  );
+
+  const { data, end, error } = useStreamQuery(balances);
+
+  return { data, end, error };
```
grod220, line 20: Combine this to just:

`return { data, end, error } = useStreamQuery(balances);`

grod220 (approved):
Few comments. After that, feel free to merge 👍

Issue #70, opened 2023-10-03
Clear cache


grod220:

zpoken:
Clear cache steps:

- [ ] stop the sync process
      - [ ]

zpoken:
@grod220  Can you check if these steps are correct? I've already done the 1st and 2nd point; for the 3rd we need a method with wasm; 4 I understand that so far only 1 value we have to change

grod220:
For the viewServer, you should just need to re-initialize that class. I don't think you need a free method. Just throw the old class away. For `networks custom values`, not sure you need to clear anything there.

The rest seems right to me 👍

zpoken:
@grod220 can you check if I'm moving in the right direction-  https://github.com/penumbra-zone/web/compare/main...70-clear-cache

grod220:
Yes, that looks like the right direction to me. If you open a PR, I can give more specific feedback. Does it work?

zpoken:
`Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["sync-blocks","https://grpc.testnet.penumbra.zone"]`

```
Error: ["sync-blocks","https://grpc.testnet.penumbra.zone"] data is undefined
    at Object.onSuccess (vendor.js:145256:19)
    at resolve
```
works, but show errors from console this hooks -useSyncProgress

PR #111, opened 2023-10-14
70 clear cache

zpoken:


apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
```diff
@@ -26,7 +27,14 @@ export const SettingsClearCache = () => {
...
           </div>
-          <Button variant='gradient' size='lg' className='w-full'>
+          <Button
+            variant='gradient'
+            size='lg'
+            className='w-full'
+            onClick={() => {
+              void (async () => await swClient.clearCache())();
+            }}
+          >
```
grod220, line 37: `onClick={void swClient.clearCache}`

apps/extension/src/routes/service-worker/extension/sync.ts
```diff
@@ -1,9 +1,13 @@
-import { SwMessage, SwMessageHandler } from './types';
 import { services } from '../../../service-worker';
+import { SwMessage, SwMessageHandler } from './types';
 
-export type SyncBlocksMessage = SwMessage<'SYNC_BLOCKS', undefined, Promise<void>>;
+export type SyncBlocksMessage = SwMessage<'SYNC_BLOCKS' | 'CLEAR_CACHE', undefined, Promise<void>>;
```
grod220, line 4: Messages should not share definitions like that. Should create a new one for `ClearCacheMessage`.

packages/query/src/block-processor/index.ts
```diff
@@ -95,6 +95,10 @@ export class BlockProcessor {
     // await this.assertRootValid(result.height);
   }
 
+  stopSyncProgress() {
```
grod220, line 98:
```diff
- stopSyncProgress
+ stopSync
```

packages/query/src/block-processor/index.ts
```diff
@@ -136,6 +140,8 @@ export class BlockProcessor {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
+      console.log(block);
+
```
grod220, line 144: Remove

packages/storage/src/indexed-db/index.ts
```diff
@@ -138,6 +138,14 @@ export class IndexedDb implements IndexedDbInterface {
     return this.db.get('TRANSACTIONS', source.inner);
   }
 
+  clear() {
+    const objectStoreNames = Object.values(this.db.objectStoreNames);
+
+    objectStoreNames.forEach(storeName => {
+      void (async () => await this.db.clear(storeName))();
+    });
```
grod220, line 146: This should not be a void. This means the db's can't be guaranteed to be cleared before trying to resync. Make `clear()` an async function and await each `this.db.clear()`.

packages/storage/src/indexed-db/index.ts
```diff
@@ -138,6 +138,14 @@ export class IndexedDb implements IndexedDbInterface {
     return this.db.get('TRANSACTIONS', source.inner);
   }
 
+  clear() {
```
grod220, line 141: This is a good candidate for a test. Just initialize, assert the tables exist, then call clear and assert they are all gone.

Also, because indexeddb requires a browser and does not exist in node, you'll have to use `pnpm run run-browser-tests` in package.json to run it.

grod220 (commented):
Also, any progress fixing this: https://github.com/penumbra-zone/web/issues/70#issuecomment-1763112766?

packages/storage/src/indexed-db/index.ts
zpoken: I still have a question here, shouldn't we delete the entire indexedDB and then reinitialize it? now you added the logic that the database name includes the chainID
@grod220

apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
zpoken: it doesn't work!on click do nothing

zpoken: ` onClick={() => void swClient.clearCache()}` works

zpoken:
@grod220 What can cause errors? https://github.com/penumbra-zone/web/issues/70#issuecomment-1763112766

packages/storage/src/indexed-db/index.ts
grod220: Will that make a difference than what you are doing here clearing all the tables?

grod220:
Here no? https://github.com/penumbra-zone/web/blob/73f584a1c2f48f8a3ce8df6972d057eb00bc7268/apps/extension/src/hooks/last-block-synced.ts#L22C1-L25

zpoken:
yes! What is this code for? we do not expect any changes from this

packages/storage/src/indexed-db/index.ts
grod220: If the RPC changes the chain, it will start a new db. But if someone just clicks clear cache, they have the same chain id and therefore the same db.

zpoken: Знімок екрана 2023-10-16 о 12 03 15

I understand, but here they will accumulate over time

grod220: I'm confused what your question is. Are you saying we should clear all databases if someone clicks clear cache? That's fine.

zpoken:
i try to delete this code and it works! but it may become inactive after a while and then it is probably better to leave this code

zpoken:
How can I make sure that this error is not displayed?

grod220:
The error says: `Query data cannot be undefined`, but `await swClient.syncBlocks()` returns undefined. What happens is that that async function runs basically indefinitely because the sync never stops. So if it ends, it returns undefined, which is what stopping it does. So useQuery() needs to get a value back that is not `undefined`. Just do something like this:

```typescript
  // Wake up service worker so block syncing can resume
  useQuery({
    queryKey: ['sync-blocks', grpcEndpoint],
    queryFn: async () => {
      await swClient.syncBlocks();
      return true; // react-query cannot return undefined
    },
  });
```

zpoken:
i see) thanks!

packages/storage/src/indexed-db/index.ts
zpoken: @grod220 
```

 it('object store should be empty after clear', async () => {
      const db = await IndexedDb.initialize(testInitialProps);
      await db.saveSpendableNote(newNote);
      expect((await db.getAllNotes()).length).toBe(1);

      await db.saveAssetsMetadata(denomMetadataA);
      expect((await db.getAllAssetsMetadata()).length).toBe(1);

      const tx = {
        blockHeight: 10000n,
        id: new NoteSource(),
        tx: new Transaction(),
        perspective: new TransactionPerspective(),
        view: new TransactionView(),
      };

      await db.saveTransactionInfo(tx);
      expect((await db.getAllTransactions()).length).toBe(1);

      await db.clear();
      expect((await db.getAllNotes()).length).toBe(0);
      expect((await db.getAllAssetsMetadata()).length).toBe(0);
      expect((await db.getAllTransactions()).length).toBe(0);
    });
```

Will it not be enough?

zpoken: we will add other objects as they are realized!

grod220: Yes, that's fine

zpoken: but I can't check if there are values for some store, (TREE_LAST_POSITION, TREE_LAST_FORGOTTEN ....) for them there is no get implemented... function (or I should check getStateCommitmentTree for this data)

grod220: No, what you have there is enough I'd say

grod220 (approved):

PR #112, opened 2023-10-16
72 reveal secure data

zpoken:


apps/extension/src/state/wallets.ts
```diff
@@ -35,6 +38,26 @@ export const createWalletsSlice =
         await local.set('wallets', [newWallet.toJson(), ...wallets]);
         return newWallet;
       },
+      getSeedPhrase: async () => {
+        const passwordKey = get().password.key;
+        const key = await Key.fromJson(passwordKey!);
+
+        const encryptedSeedPhrase = get().wallets.all[0]?.custody.encryptedSeedPhrase;
```
grod220, line 45: See bottom of this file. We should use `getActiveWallet()` instead of manually getting [0] wallet

grod220 (commented):
good job with tests!

apps/extension/src/state/wallets.ts
```diff
@@ -35,6 +38,30 @@ export const createWalletsSlice =
...
+        const key = await Key.fromJson(passwordKey!);
+
+        const activeWallet = getActiveWallet(get());
+
+        if (!activeWallet) return [];
+
+        const encryptedSeedPhrase = activeWallet.custody.encryptedSeedPhrase;
+        const unsealeSeedPhrase = await key.unseal(encryptedSeedPhrase);
+
+        return unsealeSeedPhrase?.split(' ') ?? [];
```
grod220, line 52: Making `[]` to mean `no seed phrase` is not the best approach. We should either make it return `string[] | undefined` or just `string[]` and throw errors if someone attempts to call this without the necessary requirements. This is what it would look like:

```typescript
      getSeedPhrase: async () => {
        const passwordKey = get().password.key;
        if (!passwordKey) throw new Error('no password set');

        const key = await Key.fromJson(passwordKey);
        const activeWallet = getActiveWallet(get());
        if (!activeWallet) throw new Error('no wallet set');

        const decryptedSeedPhrase = await key.unseal(activeWallet.custody.encryptedSeedPhrase);
        if (!decryptedSeedPhrase) throw new Error('Unable to decrypt seed phrase with password');

        return decryptedSeedPhrase.split(' ');
      },
```

grod220 (approved):
Make that last change and you can merge

apps/extension/src/state/wallets.ts
```diff
@@ -35,6 +38,41 @@ export const createWalletsSlice =
...
+        // const key = await Key.fromJson(passwordKey!);
+
+        // const activeWallet = getActiveWallet(get());
+
+        // if (!activeWallet) return [];
+
+        // const encryptedSeedPhrase = activeWallet.custody.encryptedSeedPhrase;
+        // const unsealeSeedPhrase = await key.unseal(encryptedSeedPhrase);
+
+        // return unsealeSeedPhrase?.split(' ') ?? [];
```
grod220, line 52: @zpoken It looks like this was not cleaned up. We shouldn't commit commented out code. Please remove on next PR.

PR #154, opened 2023-10-27
Tweaks to UI

hdevalence:
So far I just changed the assets list page to avoid implying that testnet assets have any monetary value.

I'm not sure if this is the best way to go, or if it would be better to change the data representation so that the `usdcValue` were an option. That way, we'd be able to distinguish between "no known value" and "value zero" in a principled way.  Longer term, there will definitely be assets on the chain that don't have a known USDC value, so it seems good to figure out how we want to model that.


hdevalence:
There's still work to do on the info card text but the other tweaks might be good to merge, if they seem useful.

PR #159, opened 2023-10-28
[WIP] Add a basic transaction view

hdevalence:


hdevalence:
This adds a view transaction functionality like the previous code. I tried to factor it into components but I'm not that good at typescript so I've probably made a bit of a mess of things. The results are reasonable though: 
image
image


There's also a sketch of how we can support a top-level toggle between "your view" and "public view" that would allow users to easily see what data is hidden from others.

grod220 (approved):

PR #171, opened 2023-10-30
Show labels for encrypted data

hdevalence:
The structure is actually publicly visible, so we should render it as such.


hdevalence:
View with these changes:

Screen Shot 2023-10-30 at 12 43 10 PM
Screen Shot 2023-10-30 at 12 42 59 PM

grod220 (approved):

PR #175, opened 2023-10-31
Update web for testnet63 changes

turbocrime:
addresses

* wasm crate version bump
* nct-sct rename penumbra-zone/penumbra@167e8e5d1eaee3d238f39de1646804647cde964a
* prefix change penumbra-zone/penumbra#1320

### DIFF
--- a/apps/webapp/state/send.test.ts
+++ b/apps/webapp/state/send.test.ts
@@ -88,7 +88,7 @@ describe('Send Slice', () => {
 
   describe('setRecipient and validate', () => {
     const rightAddress =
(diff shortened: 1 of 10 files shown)


hdevalence:
Screen Shot 2023-10-30 at 9 23 13 PM
Unsure if I'm not reloading the extension properly or if there's a bug here

turbocrime:
`pnpm install` to make sure you're pulling the new wasm

remove, build, reinstall extension

hdevalence:
I removed the extension entirely and reloaded it (thanks for the suggestion @turbocrime) and that error went away.

grod220 (approved):

hdevalence:
I was able to make transactions!

Issue #162, opened 2023-10-28
Fill in Receive UI in "Send" page


hdevalence:
The `/send` page has a toggle between "send" and "receive".  The Receive side isn't filled in yet.

I think the best way to do it is to reuse some components from the extension popup and from the send tab, combining these two elements:

<img width="383" alt="Screen Shot 2023-10-28 at 1 24 14 PM" src="https://github.com/penumbra-zone/web/assets/44879/9f84c1a4-1e76-4548-9491-831fd2792321">
<img width="382" alt="Screen Shot 2023-10-28 at 1 24 39 PM" src="https://github.com/penumbra-zone/web/assets/44879/d40e9c6e-c27d-4804-97a5-1d797c065d26">

The behavior of the toggle should be as follows:

- When the toggle is switched to "off", the displayed address should be set to the default address for the selected account, using [`AddressByIndex`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.AddressByIndex).
- When the toggle is switched to "on", the displayed address should be set to a fresh ephemeral address for the selected account, using [`EphemeralAddress`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.EphemeralAddress)

This means that a user can generate fresh ephemeral addresses by flipping the toggle on and off repeatedly, so we don't have to have a separate UI for that.

Along the way, we'll probably want to extract the address picker into a common UI component. One thing to consider when doing that is whether there's a way to allow users to input a numeric account (so they don't have to click through left and right buttons a zillion times), but without blowing up the complexity of the UI.  One idea in that regard could be to have the "Address #X" text secretly be an input field, so that clicking on it allows editing and typing in a number directly. This way we retain the simplicity of clicking through a few accounts while allowing users to specify arbitrary account indexes.

hdevalence:
dupe of #142

turbocrime:
confirmed in meeting:

send form address selection component indicates address index of the transaction's funding source. the ui may need to use this address index to query available funds.

this selected address index becomes the `source` field of `TransactionPlannerRequest`. fee change return address will match the address indicated by this index, in our implementation.

the `sender` field of `MemoPlaintext` provided to `TransactionPlannerRequest` in our implementation will match the address indicated by this index, or be an ephemeral address generated from the source index.

[...]

Issue #161, opened 2023-10-28
Send dialog should allow specifying a source account / planning should use sources


hdevalence:
Currently, the Send dialog doesn't allow specifying a source account. I'm also not sure whether the planning is actually scoped by account, as it should be, but in any case the UI blocks it.

Here's the state of my balances:
<img width="764" alt="Screen Shot 2023-10-28 at 1 15 14 PM" src="https://github.com/penumbra-zone/web/assets/44879/8b6939da-81d9-4128-a27a-d44d3876e309">
The dialog says I only have 8 gm available, the balance in account 0. When I try to send 1000gm, I get:
<img width="413" alt="Screen Shot 2023-10-28 at 1 15 00 PM" src="https://github.com/penumbra-zone/web/assets/44879/dc7016c0-501a-4b90-899b-41e9b2f2850b">

I think a good UI here would be to have a picker for "Source" above the "Recipient" box, so that the user first selects the source account (or accepts the default selection, the UI could be slightly different so that it's de-emphasized as an "input", like the picker in the popup:

<img width="388" alt="Screen Shot 2023-10-28 at 1 19 55 PM" src="https://github.com/penumbra-zone/web/assets/44879/275bcafa-8e0e-4284-85eb-8e61ef25831e">

We should also check that the extension's implementation of the `TransactionPlanner` respects source control.

hdevalence:
From the point of view of the frontend, I think that all that has to happen here is setting the source field of the TransactionPlannerRequest. If the wasm view server doesn't respect it, that's a bug that can be fixed independently.

Valentine1898:
I've split this issue into three different ones 
- [ ] https://github.com/penumbra-zone/penumbra/issues/3333 - add source control support to wasm planner 
- [ ] https://github.com/penumbra-zone/web/issues/243 - TransactionPlannerRequest handler must consider source
- [ ] https://github.com/penumbra-zone/web/issues/244 - User should be able to choose the account from which he is going to spend funds when sending (UI issue)

Valentine1898:
Closed as done after  https://github.com/penumbra-zone/web/pull/266 merged

PR #180, opened 2023-11-01
Send form work

### LINKED ISSUE #160: Memo handling is incorrect
Steps to reproduce:

1. Using `pcli` (perhaps with a different wallet), generate an address.
2. Use the web extension's "send" dialog to send to that address. Add a memo text and click "Hide Sender from Recipient".
3. Use `pcli v tx HASH` to view the transaction.
4. See that the memo text is discarded rather than included in the transaction.
5. Use `pcli tx send 1penumbra --to RETURN_ADDRESS` to see that the address specified as the return address is invalid.

turbocrime:
edit: updated with some more work that was closely aligned
 
fixes #160 but i don't think this is ready

initially, the send page provided undefined memo fields to the planner request. the page didn't send it, and handleTxPlannerReq just skipped the memo plan step. this is apparently a valid way to plan a transaction. i don't have a good wasm/rust debug flow yet but i believe this results in a zeroed MemoPlaintext.sender and zero-length MemoPlaintext.text in the final transaction on the chain.

the send form now always supplies a return address, and at least an empty-string memo.

ephemeraladdressrequest is unimplemented #181 so that part doesn't work, but it's on the way to also addressing parts of #162

### DIFF
--- a/apps/webapp/app/send/send-form.tsx
+++ b/apps/webapp/app/send/send-form.tsx
@@ -7,7 +7,7 @@ import { amountToBig, sendSelector, sendValidationErrors } from '../../state/sen
 import { useToast } from '@penumbra-zone/ui/components/ui/use-toast';
 import { isPenumbraAddr } from '@penumbra-zone/types';
 import { useSendBalance } from '../../hooks/send-balance';
-import { useAddress, useEphemeralAddress, } from '../../hooks/address';
+import { useAddress, useEphemeralAddress } from '../../hooks/address';
 import { bech32Address } from '@penumbra-zone/types';
 import { useEffect, useMemo, useState } from 'react';
 import { cn } from '@penumbra-zone/ui/lib/utils';
@@ -20,7 +20,7 @@ const Send = () => {
     asset,
     recipient,
     memoText,
-    memoSender,
+    //memoSender,
     setAmount,
     setAsset,
     setRecipient,
@@ -36,7 +36,7 @@ const Send = () => {
 
   // TODO: enable address selection
   const defaultSender = useAddress(0);
-  const ephemeralSender = useEphemeralAddress(0, { enabled: hidden })
+  const ephemeralSender = useEphemeralAddress(0, { enabled: hidden });
 
   const sender = useMemo(() => {
     if (!hidden && defaultSender.data) return bech32Address(defaultSender.data);
--- a/apps/webapp/app/send/send-form.tsx
+++ b/apps/webapp/app/send/send-form.tsx
@@ -9,7 +9,7 @@ import { isPenumbraAddr } from '@penumbra-zone/types';
 import { useSendBalance } from '../../hooks/send-balance';
 import { useAddress, useEphemeralAddress, } from '../../hooks/address';
 import { bech32Address } from '@penumbra-zone/types';
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { cn } from '@penumbra-zone/ui/lib/utils';
 import InputToken from '../../shared/input-token';
 
@@ -20,10 +20,12 @@ const Send = () => {
     asset,
     recipient,
     memoText,
+    memoSender,
     setAmount,
     setAsset,
     setRecipient,
     setMemoText,
+    setMemoSender,
     sendTx,
     txInProgress,
   } = useStore(sendSelector);
@@ -41,6 +43,10 @@ const Send = () => {
     if (hidden && ephemeralSender.data) return bech32Address(ephemeralSender.data);
     return '';
   }, [hidden, defaultSender, ephemeralSender]);
+
+  useEffect(() => {
+    setMemoSender(sender);
+  }, [sender, setMemoSender]);
 
   return (
     <form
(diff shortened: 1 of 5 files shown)


packages/router/src/grpc/view-protocol-server/tx-planner.ts
```diff
@@ -42,6 +43,7 @@ export const handleTxPlannerReq = async (
   }
 
   if (req.memo) {
+    req.memo.sender ??= refundAddr;
     planner.memo(req.memo);
```
turbocrime, line 47: if planner.memo is skipped, the transaction can successfully plan and authorize. this agrees with TransactionPlannerRequest, but is that the intended behavior?

if planner.memo is called with a MemoPlaintext that contains a "text" but no "sender", it fails, despite MemoPlaintext.sender being optional.

apps/webapp/state/send.ts
```diff
@@ -97,8 +97,9 @@ const getExponent = (asset: TempAsset): number => {
   return match?.exponent ?? 0;
 };
 
-const planWitnessBuildBroadcast = async ({ amount, recipient, asset }: SendSlice) => {
+const planWitnessBuildBroadcast = async ({ amount, memoText, recipient, asset }: SendSlice) => {
   const req = new TransactionPlannerRequest({
+    memo: { text: memoText },
```
turbocrime, line 102: memo sender should probably be defined here instead of in view-protocol-server

apps/webapp/state/send.test.ts
```diff
@@ -68,20 +69,6 @@ describe('Send Slice', () => {
     });
   });
```
turbocrime, line 71: some of these tests seem to be tautological

turbocrime:
ephemeral address is now populated by the extension thanks to f61b090e1674379085fbb3b35aebf5f5b01a2cb1 so this now works. proceeding to cleanup

Screenshot 2023-11-02 at 11 40 26

turbocrime:
returning funds as in #160 and then spending those funds seems to work. cleaning up for merge

turbocrime:
cc @grod220 ping for review

hdevalence:
I'm wondering if the UI changes make sense in light of https://github.com/penumbra-zone/web/issues/161 to allow selecting the source account.

I'm not sure it's meaningful to show the actual address that will be used, since it's random data. I don't think it conveys any additional information beyond the state of the toggle — the user can see that an address is there, but they have no way to interpret it.

(It's not meaningfully confirmatory, either, because the frontend code shouldn't be relied on for that purpose. We will be adding an approval pop up to the extension that displays the transaction plan before authorization, and that popup should render the TransactionPlan in a way that shows the user what they are authorizing and what data they will be revealing to their counterparty or to the public. So the user should be able to see that a one-time address was used in that confirmation dialog.)

grod220 (commented):
Now seeing the UI, I can see that the sender field is a bit confusing. Think the source address ticket is probably the thing that will add clarity for users here.

apps/webapp/app/send/send-form.tsx
```diff
@@ -15,27 +19,49 @@ const Send = () => {
...
+  const sender = useMemo(() => {
+    if (!hidden && defaultSender.data) return bech32Address(defaultSender.data);
+    if (hidden && ephemeralSender.data) return bech32Address(ephemeralSender.data);
+    return '';
+  }, [hidden, defaultSender, ephemeralSender]);
+
+  useEffect(() => {
+    setMemoSender(sender);
+  }, [sender, setMemoSender]);
+
```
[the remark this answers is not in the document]
turbocrime: when i wrote it this way i was thinking:

* avoid generating the ephemeral address until it's required, to minimize requests. it looks like this one is local and might be computationally simple.
* ephemeral addresses should not be reused. if we pull an address from the store, how do we ensure the address is new and specific to this transaction? we're just going to have to trigger a request.
* we should show the actual selected address to the user. there's a conceptual symmetry here, as the user must provide a destination address. in the case of a non-ephemeral sender, it is additionally confirming and helps the user understand "i'm transmitting *my* address" in a correlateable way

if we don't display the address to the user, the points are all eliminated and it's fine to just set the page controls into the store like before, putting off the requests to sendTx. in discussion i think we had talked specifically about showing the address, so that's the approach i took.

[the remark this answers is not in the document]
grod220: == Update ==
A good deal of refactoring of data fetching has just been merged. We essentially have two paths for retrieving data:
- Loaders (preferred)
  - Fetches on route changes and does not load the component until data is available. We now have a bunch of places in the code base where this is used.
  - Sometimes it may seem like this is not an option, but we often have more opportunities to change/add to the URL path than we realize.
- In-component useQuery hooks
  - If changing the route (in response to user actions) is not really possible or the right flow, we can always just "reactively" fetch using react-query. However, we should be careful not to fall into the same pit we had before where there were like +5 hooks composing with each other. If we go this route, let's try our best to use a single useQuery hook and compose within the async fetch function. 
  - In this PR's case, I think we likely are in this bucket

turbocrime:
all of this work was obsoleted by #230 #185 and is ultimately blocked by penumbra-zone/penumbra#3333 

will open a new pr at some point lol

PR #189, opened 2023-11-01
149 tabs should be routes

zpoken:


apps/webapp/app/ibc/constants.ts
grod220, line 1: This should be in the constants package

grod220 (commented):
- There appears to be a fade-in transition that is jarring. Please remove.
- The url does not match the nested nature of the design. Think we should have this:
  - `/dashboard` (goes to assets tab)
  - `/dashboard/transactions`
  - `/send` (goes to send tab)
  - `/send/receive` (kinda weird wording, but the best we have)
- Using nested routes will allow nextjs to not have to re-render the header

apps/webapp/hooks/typesafe-pathname.ts
```diff
@@ -0,0 +1,7 @@
+import { usePathname } from 'next/navigation';
+
+export const useTypesafePathname = <T>() => {
```
grod220, line 3:
```diff
- useTypesafePathname
+ useTypedPathname
```

apps/webapp/app/header/navbar.tsx
```diff
@@ -1,45 +1,51 @@
 'use client';
 
 import Link from 'next/link';
-import { usePathname } from 'next/navigation';
 import { DappPath } from './paths';
 import { cn } from '@penumbra-zone/ui/lib/utils';
+import { useTypesafePathname } from '../../hooks/typesafe-pathname';
 
 const links = [
```
grod220, line 8: Please create an interface for this. Also, make `subLinks` an optional field so you don't need to do `subLinks: []` a lot.

apps/webapp/app/send/send-form.tsx
```diff
@@ -13,7 +13,7 @@ const InputToken = dynamic(() => import('../../shared/input-token'), {
   ssr: false,
 });
 
-export default function SendForm() {
+const Send = () => {
```
grod220, line 16: Given this is dynamically imported, don't think you need to dynamically import `InputToken` above right?

grod220 (approved):
After those last changes, you can merge

apps/webapp/app/send/send-form.tsx
Vadim121197: Yes

PR #215, opened 2023-11-02
added NOTES docs

Valentine1898:


packages/types/src/indexed-db.ts
```diff
@@ -95,8 +95,15 @@ export interface PenumbraDb extends DBSchema {
       nullifier: Base64Str; // Jsonified<SpendableNoteRecord['nullifier']['inner']>
     };
   };
+  // Note is part of a SpendableNote
+  // The rust view service uses relation DB, so there for each SpendableNote
+  // there is a corresponding Note record (linked by noteCommitment)
+  // IndexedDb is not a relation DB, so we can just store a SpendableNote with a nested Note in one SPENDABLE_NOTES table,
+  // so we don't use the NOTES table to store each Note,
+  // but only when we don't have a SpendableNote generated yet but already have a Note
```
grod220, line 103: I'm having a hard time understanding this. It looks like the wasm crate uses this table to store swap outputs? Why don't we call this table `SWAP_NOTES`?

Valentine1898: The object stored in this table is of type `Note` (swap outputs are of type `Note`). Besides, we actually repeat the database structure from pcli now. 
But in general, I think it's a good idea to rename this table to `SWAP_NOTES` or `ADVICES` to avoid confusion between `NOTES` and `SPENDABLE_NOTES`

grod220: If it's of type: `noteCommitment`, shouldn't the key be?: 
```
Jsonified<Note['note']['address']['inner']>
```

I am not familiar with what `advices` are. Are these only related to swaps?

Valentine1898: I think we should globally decide whether we will completely repeat the DB structure from rust view server or offer our own structure for extension view service. 

In the `NOTES` table, we store notes that we have already detected but cannot yet spend. In practice, this only occurs with swap. We detect a swap, get output notes from it, but these notes cannot be spent until we detect a swap claim. 
We also call these notes **advices**, because from the point of view of the block scanner we save some data (notes) that in the future will help us to correctly process the swap claim and get spendable notes.  The term advices is also used in the rust view service 

Renaming `NOTES` to `SWAP_NOTES` has two disadvantages 
1. We are moving away from the database structure that is used in the rust view service. 
2. In the future, theoretically, there may be other types of two-stage transactions that will generate notes that cannot be spent yet, and the `SWAP_NOTES` naming will lose relevance 

Regarding the key 
My variant of `Jsonified<Note['noteCommitment']['inner']>` is not correct because there is no `noteCommitment` field in the Note type. The `Jsonified<Note['address']['inner']>` variant is not correct either. 
The key is of type `StateCommitment` (same as the `noteCommitment` field for `SPENDABLE_NOTES`), but it is not part of the stored object.

grod220: So is the `NOTES` table consistent with the Rust code and what you recommend keeping it as?

If those keys aren't right, can you fix the comment so it's correct?

Valentine1898: The `NOTES` table is consistent with the rust code, we store the same object (`Note`), with the same key type (`StateCommitment`)  as is done in rust. 

Yes, I've already changed the comment regarding the key.

grod220 (approved):

PR #220, opened 2023-11-04
use toFixed instead toString

zpoken:


grod220 (commented):
Please write a test for this that failed before and now passes due to this fix

grod220 (approved):

PR #225, opened 2023-11-08
130 add remaining indexeddb tests

Valentine1898:


packages/storage/src/indexed-db/indexed-db.test.ts
grod220, line 1: This file is quite large. Can you either move the test data to another file, or split up the file into multiple tests with test data next to it.

grod220 (commented):
I don't think this is every table right? We need tests for:

LAST_BLOCK_SYNCED
TREE_LAST_POSITION
TREE_LAST_FORGOTTEN
TREE_HASHES
TREE_COMMITMENTS
FMD_PARAMETERS
TRANSACTIONS
ASSETS
SPENDABLE_NOTES
NOTES
SWAPS

Valentine1898:
Yes, I've covered all functions

grod220:
Ah ok. In that case, address the comment on the test file length and we'll be good

grod220 (approved):

Issue #132, opened 2023-10-25
IBC out from UI: Part 1


grod220:
The wasm crate now supports building an IBC transaction. This task is implementing the IBC out in the UI + service worker.

Should re-enable IBC tab on /send page

conorsch:
We're working on a point release of ibc-types later this week to unblock the integration. 

Also, be aware that we just merged https://github.com/penumbra-zone/penumbra/pull/3284 which renames some IBC proto fields. These changes will be live on preview momentarily, and land on testnet as of 64.

PR #249, opened 2023-11-17
Ics20Withdraw Part 1

grod220:
Closes https://github.com/penumbra-zone/web/issues/132

- Adds new IbcClient to the service worker. This is using the _old_ routing. Hopefully can kill this later.
- Handles the killer clientStates protobuf Any type
- Fixes an issue where we weren't calling to/fromJson on the messages sent to/from the webapp and extension. Also uses new typeregistry to allow serializing protobuf Any's


packages/transport/src/types.ts
```diff
@@ -51,30 +58,48 @@ export const isUnaryResponse = <S extends ServiceType>(
...
 ): UnaryResponse<S> => {
   return {
     type: OUTGOING_GRPC_MESSAGE,
     sequence: req.sequence,
-    requestTypeName: req.requestTypeName,
     serviceTypeName: req.serviceTypeName,
-    result,
+    requestTypeName: req.requestTypeName,
+    responseTypeName: result.getType().typeName,
+    result: result.toJson({ typeRegistry }),
```
grod220, line 69: @turbocrime fyi

When attempting to send a protobuf `Any` to the frontend, it was throwing json serialization issues. Turns out, we weren't doing `.json()` ourselves and relying upon chrome's internal methods to do this. We didn't notice until now because our serialization requires a registry now.

PR #250, opened 2023-11-20
162 fill in receive UI in send page

zpoken:


grod220:
Can you fix the git conflicts and then I'll review?

zpoken:
ready for review @grod220

apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -34,31 +33,23 @@ export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> =>
...
-            className='w-4'
+      <div className='my-32'>
+        {account && (
+          <SelectAccount
+            previous={previous}
+            next={next}
+            setIndex={setIndex}
+            account={account}
+            ephemeral={ephemeral}
+            setEphemeral={setEphemeral}
```
grod220, line 50: In the popup, upon every block that gets synced, it appears the ephemeral address gets refreshed. Is that on purpose? Think it should stay the same between blocks.

apps/extension/src/state/accounts.ts
```diff
@@ -47,12 +53,19 @@ export const selectedAccount = (state: AllSlices): Account | undefined => {
...
+    ? getEphemeralByIndex(active.fullViewingKey, index)
+    : getAddressByIndex(active.fullViewingKey, index);
   const bech32Addr = bech32Address(addr);
 
   return {
     address: bech32Addr,
-    preview: getShortAddressByIndex(active.fullViewingKey, state.accounts.index),
-    index: state.accounts.index,
+    preview: ephemeral
+      ? bech32Addr.slice(0, 33) + '…'
```
grod220, line 67: This isn't ideal. We should replace `get_short_address_by_index` with `get_short_address` so it can be used generically. Maybe @Valentine1898, you can assist with that?

And then on the next wasm crate deployment, we can update this line.

apps/webapp/src/components/send/receive.tsx
```diff
@@ -1,10 +1,40 @@
...
+      const address = ephemeral
+        ? await getEphemeralAddress(Number(index))
+        : await getAddressByIndex(Number(index));
+      setAccount({
+        address,
+        preview: address.slice(0, 33) + '…',
+        index,
+      });
+    })();
+  }, [index, ephemeral]);
```
grod220, line 24: We should move this into the ReceiveSlice

apps/webapp/src/components/send/receive.tsx
```diff
@@ -1,10 +1,40 @@
...
 export const Receive = () => {
+  const { index, ephemeral, setIndex, previous, next, setEphemeral } = useStore(receiveSelector);
+
+  const [account, setAccount] = useState<Account | undefined>();
+
+  useEffect(() => {
+    void (async () => {
+      const address = ephemeral
+        ? await getEphemeralAddress(Number(index))
+        : await getAddressByIndex(Number(index));
```
grod220, line 17: Isn't `index` already a Number?

apps/webapp/src/fetchers/address.ts
```diff
@@ -29,3 +32,19 @@ export const getAddresses = async (accounts: (number | undefined)[]): Promise<In
...
     }, {});
 };
+
+export const getAddressByIndex = async (account: number): Promise<string> => {
+  const req = new AddressByIndexRequest();
+  req.addressIndex = new AddressIndex({ account });
+  const res = await viewClient.addressByIndex(req);
+
+  return bech32Address(res.address!);
+};
```
grod220, line 42: There can be re-use with the function above:

```typescript
export const getAddresses = async (accounts: (number | undefined)[]): Promise<IndexAddrRecord> => {
  const allReqs = accounts.map(getAddressByIndex);

  const responses = await Promise.all(allReqs);
  return responses
    .map((address, i) => {
      return {
        index: accounts[i] ?? 0,
        address,
      };
    })
    .reduce<IndexAddrRecord>((acc, curr) => {
      acc[curr.index] = curr.address;
      return acc;
    }, {});
};

export const getAddressByIndex = async (account: number | undefined): Promise<string> => {
  const req = new AddressByIndexRequest();
  if (account) req.addressIndex = new AddressIndex({ account });
  const res = await viewClient.addressByIndex(req);
  return bech32Address(res.address!);
};
```

apps/webapp/src/fetchers/address.ts
```diff
@@ -29,3 +32,19 @@ export const getAddresses = async (accounts: (number | undefined)[]): Promise<In
...
+
+export const getEphemeralAddress = async (account: number): Promise<string> => {
+  const req = new EphemeralAddressRequest();
+  req.addressIndex = new AddressIndex({ account });
+  const res = await viewClient.ephemeralAddress(req);
+
+  return bech32Address(res.address!);
+};
```
grod220, line 50: can shorten:

```typescript
export const getEphemeralAddress = async (account: number): Promise<string> => {
  const req = new EphemeralAddressRequest({ addressIndex: { account } });
  const res = await viewClient.ephemeralAddress(req);
  return bech32Address(res.address!);
};

```

apps/webapp/src/state/receive.ts
```diff
@@ -0,0 +1,43 @@
...
+  return {
+    index: 0,
+    ephemeral: false,
+    previous: () => {
+      const current = get().receive.index;
+      set(state => {
+        if (current > 0) {
+          state.receive.index = current - 1;
+        }
+      });
```
grod220, line 22: Shouldn't the if clause be on the outside?

```typescript
      if (current > 0) {
        set(state => {
          state.receive.index = current - 1;
        });
      }
```

grod220:
There seems to be a flicker when the select-asset dialog closes. Is it related to this PR?

https://github.com/penumbra-zone/web/assets/16624263/53940b9a-eb3d-491b-8a0e-fa7cab1c278e

Vadim121197:
No, but I can fix it there

apps/webapp/src/state/receive.ts
zpoken: yes, but i reuse it from there - https://github.com/penumbra-zone/web/blob/main/apps/extension/src/state/accounts.ts#L24

apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -33,32 +33,36 @@ export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> =>
...
+    next,
+    previous,
+    setIndex,
+    setEphemeral,
+    setSelectedAccount,
+  } = useStore(accountsSelector);
+
+  useEffect(() => {
+    setSelectedAccount();
+  }, [ephemeral, index, setSelectedAccount]);
```
grod220, line 49: We should not require a useEffect. Please move this logic into the accounts slice.

grod220: In fact, this is pointing to the fact that `selectedAccount` is really "derived" data and should be a selector.

apps/extension/src/state/accounts.ts
```diff
@@ -38,21 +41,32 @@ export const createAccountsSlice: SliceCreator<AccountsSlice> = (set, get) => {
...
+      set(state => {
+        state.accounts.selectedAccount = {
+          address: bech32Addr,
+          preview: ephemeral
+            ? bech32Addr.slice(0, 33) + '…'
+            : getShortAddressByIndex(active.fullViewingKey, index),
+          index,
+        };
+      });
+    },
```
grod220, line 68: Move this into a selector function instead of a setter like this

apps/webapp/vite.config.ts
```diff
@@ -4,4 +4,8 @@ import react from '@vitejs/plugin-react-swc';
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react()],
+  test: {
+    globals: true,
+    environment: 'jsdom',
+  },
```
grod220, line 10: What specifically requires this? It should have a code comment that explains why it's necessary if so.

apps/webapp/src/components/send/receive.tsx
```diff
@@ -1,10 +1,36 @@
...
+    setIndex,
+    previous,
+    next,
+    setEphemeral,
+    setSelectedAccount,
+  } = useStore(receiveSelector);
+
+  useEffect(() => {
+    void setSelectedAccount();
+  }, [index, ephemeral, setSelectedAccount]);
```
grod220, line 20: Same comment as the other. This should not be necessary.

apps/webapp/src/state/receive.ts
```diff
@@ -0,0 +1,60 @@
...
+  ephemeral: boolean;
+  selectedAccount: Account | undefined;
+  previous: () => void;
+  next: () => void;
+  setIndex: (index: number) => void;
+  setEphemeral: (ephemeral: boolean) => void;
+  setSelectedAccount: () => Promise<void>;
+}
+
+export const createReceiveSlice = (): SliceCreator<ReceiveSlice> => (set, get) => {
```
grod220, line 16: Isn't this slice completely redundant to the accounts slice? Why are we repeating this code here and not just using the accounts slice?

grod220: Think we should consider combining these

apps/extension/src/state/accounts.ts
zpoken: because of this, there was a problem that after processing the block, the ephemeral address changed

apps/webapp/src/state/receive.ts
zpoken: it`s not clear for me!more details possible

grod220: `apps/extension/src/state/accounts.ts`
`apps/webapp/src/state/receive.ts`

These are basically duplicates of each other. Why not build a `AccountsSelectorComponent`, which has state management built in locally, so we don't need duplicates here?

apps/extension/src/state/accounts.ts
grod220: Why does do new blocks re-render this component? Regardless, I think that means we need to re-architect this component so re-renders do not influence it like that.

grod220: Think a React.memo or useMemo would help with this re-render issue

apps/webapp/src/state/receive.ts
zpoken: without zustand? just use useState and useMemo? we can't use zustand in ui package

grod220: Sure, why not?

zpoken: `setSelectedAccount` has the same logic, but uses different methods to get the address (`getAddressByIndex, getEphemeralAddress` functions): webapp - from the extension, extension - wasm library

zpoken: I think it makes sense to move everything except selectedAccount to the local state of AccountsSelectorComponent

grod220: Ok, sounds good to me

packages/ui/state/account.ts
grod220, line 1: Let's not have zustand as a dependency in the UI package (unless we must). Can you do this with just useState?

apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -33,32 +38,31 @@ export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> =>
...
+    const bech32Addr = bech32Address(addr);
+
+    return {
+      address: bech32Addr,
+      preview: ephemeral
+        ? bech32Addr.slice(0, 33) + '…'
+        : getShortAddressByIndex(active.fullViewingKey, index),
+      index,
+    };
+  };
```
grod220, line 58: This logic should be a selector function that lives inside /state/wallets.

```typescript
const getAccount = useStore(accountAddrSelector);
```

apps/webapp/src/components/send/receive.tsx
```diff
@@ -1,10 +1,20 @@
...
 export const Receive = () => {
+  const getAccount = async (index: number, ephemeral: boolean) => {
+    const address = ephemeral ? await getEphemeralAddress(index) : await getAddressByIndex(index);
+
+    return {
+      address,
+      preview: address.slice(0, 33) + '…',
+      index,
+    };
+  };
```
grod220, line 13: This should be defined _outside_ the component above

apps/extension/src/routes/popup/home/index.tsx
zpoken: do I need to store the value of the function (selectedAccount) in /state/wallets?

grod220: what do you mean?

zpoken: i add fucntion to state.wallets which return account address:
```
getAccountAddr: (index: number, ephemeral: boolean) => {
        const active = getActiveWallet(get());
        if (!active) return;

        const addr = ephemeral
          ? getEphemeralByIndex(active.fullViewingKey, index)
          : getAddressByIndex(active.fullViewingKey, index);
        const bech32Addr = bech32Address(addr);

        return {
          address: bech32Addr,
          preview: ephemeral
            ? bech32Addr.slice(0, 33) + '…'
            : getShortAddressByIndex(active.fullViewingKey, index),
          index,
        };
      },
```
Should i save this addresses to WalletsSlice? whether the user will be able to select the account index and when closing/opening the extension, he will be see the one he specified index, not 0 index

grod220: No, not within the slice. It's a selector:

```typescript
export const walletsSelector = (state: AllSlices) => state.wallets;
export const getActiveWallet = (state: AllSlices) => state.wallets.all[0];

export const accountAddrSelector = (state: AllSlices) => (index: number, ephemeral: boolean) => {
  const active = getActiveWallet(state);
  if (!active) return;

  const addr = ephemeral
    ? getEphemeralByIndex(active.fullViewingKey, index)
    : getAddressByIndex(active.fullViewingKey, index);
  const bech32Addr = bech32Address(addr);

  return {
    address: bech32Addr,
    preview: ephemeral
      ? bech32Addr.slice(0, 33) + '…'
      : getShortAddressByIndex(active.fullViewingKey, index),
    index,
  };
};

=======

  const getAccount = useStore(accountAddrSelector);

  return (
    <div className='relative flex h-full flex-col items-stretch justify-start bg-left-bottom px-[30px]'>
      <div className='absolute bottom-[50px] left-[-10px] -z-10 h-[715px] w-[900px] overflow-hidden bg-logo opacity-10' />
      <IndexHeader />
      <div className='my-32'>
        <SelectAccount getAccount={getAccount} />
      </div>
      <BlockSync />
    </div>
  );
};


```

zpoken: ok, thanks

packages/ui/components/ui/select-account.tsx
```diff
@@ -1,79 +1,152 @@
...
-                  className='mb-[3px] h-6 py-[2px] font-headline text-xl font-semibold leading-[30px]'
-                  onChange={e => {
-                    const value = Number(e.target.value);
-                    const valueLength = e.target.value.replace(/^0+/, '').length;
+  useEffect(() => {
+    void (async () => {
+      const account = await getAccount(index, ephemeral);
+      setAccount(account);
+    })();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
grod220, line 30: Why not put `getAccount` as a dep? Also, we shouldn't have eslint-disable comments without an additional comment on why it's there. Please add code comment to why it's necessary.

zpoken: this updates the ephemeral address of each block

PR #253, opened 2023-11-22
252 need to clear indexed db during version migration

Valentine1898:
We'll get a display bug if we don't execute 
```typescript
void localExtStorage.set('lastBlockSynced', 0); 
``` 
The synс progress bar will display the old value of `lastBlockSynced`, while in fact sync started from 0 block

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -42,6 +42,13 @@ export class IndexedDb implements IndexedDbInterface {
 
     const db = await openDB<PenumbraDb>(dbName, dbVersion, {
       upgrade(db: IDBPDatabase<PenumbraDb>) {
+
+        // delete existing ObjectStores before re-creating them
+        // all existing indexed-db data will be deleted when version is increased
+        for (const objectStoreName of db.objectStoreNames) {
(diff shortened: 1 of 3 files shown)


grod220 (commented):
Can you write a test that failed before and now passes because of this code change?

packages/storage/src/indexed-db/index.ts
```diff
@@ -23,6 +23,7 @@ import {
   AssetId,
   DenomMetadata,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
+import { localExtStorage } from '../chrome';
```
grod220, line 26: Instead of a direct import, it may be better to have storage or a function passed into the constructor

grod220: This will also fix your test issue

Valentine1898:
I found a simpler solution than calling 
```typescript
void localExtStorage.set('lastBlockSynced', 0);
``` 
We just need to sync local `lastBlockSynced` with the current `lastBlockSynced` value from `indexed-db `

grod220 (commented):
Can you still write a test for this?

Valentine1898:
Sure, added test

PR #254, opened 2023-11-22
198 store swap/swap claim pairs

Valentine1898:


packages/storage/src/indexed-db/index.ts
```diff
@@ -63,7 +64,9 @@ export class IndexedDb implements IndexedDbInterface {
 
         // TODO: To implement
```
grod220, line 65: Is this todo still relevant? 🤔

packages/storage/src/indexed-db/index.ts
```diff
@@ -207,4 +215,23 @@ export class IndexedDb implements IndexedDbInterface {
       txs.add({ table: 'SPENDABLE_NOTES', value: n.toJson() });
     }
   }
+
+  private addNewSwaps(txs: IbdUpdates, swaps: SwapRecord[]): void {
+    for (const n of swaps) {
+      if (!n.swapCommitment?.inner) throw new Error('swapCommitment not present');
```
grod220, line 221: Do we need this check? Looks like we can still call n.toJson() regardless

packages/storage/src/indexed-db/index.ts
```diff
@@ -207,4 +215,23 @@ export class IndexedDb implements IndexedDbInterface {
...
+
+  async getSwapByNullifier(nullifier: Nullifier): Promise<SwapRecord | undefined> {
+    const key = uint8ArrayToBase64(nullifier.inner);
+    const json = await this.db.getFromIndex('SWAPS', 'nullifier', key);
+    if (!json) return undefined;
+    return SwapRecord.fromJson(json);
+  }
+
+  async saveSwap(swap: SwapRecord) {
+    if (!swap.swapCommitment?.inner) throw new Error('swapCommitment not present');
```
grod220, line 234: Same feedback here. Do we need to validate it like this?

grod220 (approved):
Good job! Left a few comments.

packages/storage/src/indexed-db/index.ts
Valentine1898: I agree, `NOTES` does not require a change, and `SWAPS` is done
Removed todo

packages/storage/src/indexed-db/index.ts
Valentine1898: I agree that this check is redundant, I added it only because a similar one is implemented for notes. 
Removed such checks for both swaps and notes

grod220:
Sounds good, feel free to merge

PR #257, opened 2023-11-27
133 add coin icons to assets

Vadim121197:


grod220 (commented):
Instead of two components, can we just have one `<Identicon>` and it accepts a prop for either `gradient | solid`?

The areas it's called `IdenticonColor` we should call `identiconSolid`.

packages/ui/components/ui/identicon/generate.ts
```diff
@@ -19,3 +19,25 @@ export const generateGradient = (str: string) => {
     toColor,
   };
 };
+
+export const generateColor = (str: string) => {
```
grod220, line 23: Change this to `generateSolidColor`

grod220 (approved):

Vadim121197:
Знімок екрана 2023-11-27 о 17 32 23

@grod220

Vadim121197:
i can not merge

grod220:
Screenshot 2023-11-27 at 3 44 21 PM

Just noticed that `gm` and `gn` coins have the same color. Given they have different names, this should not have happened right?

Vadim121197:
Знімок екрана 2023-11-28 о 13 37 02

@grod220

grod220:
Ah ok, bad luck then I suppose

PR #260, opened 2023-11-28
0.64 preview updates

grod220:
Updating to latest wasm + bufbuild types. This is in the attempts to fix build issues we've had since the last testnet release.

Given this breaks the experience for current users, I am disabling the firebase deploy github action until we are ready for the next testnet release.

In dev mode, we should move to this in the meantime:

```diff
export const testnetConstants: Config = {
-  grpcEndpoint: 'https://grpc.testnet.penumbra.zone',
+  grpcEndpoint: 'https://grpc.testnet-preview.penumbra.zone',
  indexedDbVersion: 16,
  usdcAssetId: 'reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg=',
};
```


packages/wasm/src/build.ts
grod220, line 1: @TalDerei fyi

PR #261, opened 2023-11-28
store lpnft

Valentine1898:


Valentine1898:
If we detected a new asset but could not get its metadata from grpc, we store the generated metadata using the bech32 string prefixed with `passet`  as base denom
https://github.com/penumbra-zone/web/blob/288ac1d180a8b815e50716258dc5385a8707cfa7/packages/query/src/block-processor/index.ts#L122

But at the same time, it will cause a serialization error if we try to pass generated asset to wasm 
https://github.com/penumbra-zone/penumbra/blob/f08d72db8d8aa22dbd8bdd4b371a9ba983897646/crates/core/asset/src/asset/denom_metadata.rs#L91

```
Error:   denom metadata has mismatched penumbra asset ID
``` 

If we send an Open LP position transaction, we get this scanning error script
1. Detected new notes with unknown asset
2. Grpc query `denomMetadata()` did not return asset information 
3. We stored the generated `DenomMetadata` for the asset in indexed-db
4. We also found our transaction in block, and tried to get `TransactionInfo` for it via wasm
5. Wasm while executing `transactionInfo()` read our asset from indexed-db but failed to serialize it and threw error `denom metadata has mismatched penumbra asset ID `

Valentine1898:
It seems that the issue `denom metadata has mismatched penumbra asset ID` is related to the fact that in extension we get TransactionInfo once during synchronization, while in rust view service TransactionInfo is generated every time a transaction retrieval request is executed

Valentine1898:
A simple solution is that we should store LPNFT before calling `transactionInfo() `

packages/query/src/block-processor/transactions.ts
```diff
@@ -67,6 +84,32 @@ export class Transactions {
...
+        await this.indexedDb.saveAssetsMetadata(this.viewServer.getLpNftDenom(position, closed));
+
+        const withdrawn = new PositionState({ state: PositionState_PositionStateEnum.WITHDRAWN });
+        await this.indexedDb.saveAssetsMetadata(this.viewServer.getLpNftDenom(position, withdrawn));
+
+        const claimed = new PositionState({ state: PositionState_PositionStateEnum.CLAIMED });
+        await this.indexedDb.saveAssetsMetadata(this.viewServer.getLpNftDenom(position, claimed));
+      }
+    }
+  }
```
grod220, line 112: What do you think about this as an alernative?

```typescript
    for (const a of tx.body.actions) {
      if (a.action.case !== 'positionOpen' || a.action.value.position === undefined) {
        continue;
      }

      const positionStates = [
        PositionState_PositionStateEnum.OPENED,
        PositionState_PositionStateEnum.CLOSED,
        PositionState_PositionStateEnum.WITHDRAWN,
        PositionState_PositionStateEnum.CLAIMED,
      ];

      for (const state of positionStates) {
        const positionState = new PositionState({ state });
        const metadata = this.viewServer.getLpNftDenom(a.action.value.position, positionState);
        await this.indexedDb.saveAssetsMetadata(metadata);
      }
```

packages/query/src/block-processor/transactions.ts
```diff
@@ -67,6 +84,32 @@ export class Transactions {
       }
     }
   }
+
+  // LPNFT asset IDs won't be known to the chain, so we need to pre-populate them in local
```
grod220, line 88: Could we do this casing instead? `LpNft`?

packages/query/src/block-processor/transactions.ts
```diff
@@ -67,6 +84,32 @@ export class Transactions {
...
+  // registry based on transaction contents.
+  // Necessary to save LPNFT denoms for 'closed' 'withdrawn' and 'claimed' as soon as we detect a positionOpen action
```
grod220, line 90: Can you add context in the comment on why it's needed?

packages/wasm/src/view-server.ts
```diff
@@ -73,4 +79,12 @@ export class ViewServer implements ViewServerInterface {
     const raw = validateSchema(ScanResultSchema, this.wasmViewServer.flush_updates());
     return parseScanResult(raw);
   }
+
+  getLpNftDenom(position: Position, positionState: PositionState): DenomMetadata {
```
grod220, line 83: should we call this: `getLpNftMetadata`?

packages/query/src/block-processor/transactions.ts
```diff
@@ -49,6 +62,10 @@ export class Transactions {
 
       if (noteSourcePresent(this.all, noteSource)) {
         const transaction = decodeTx(txBytes);
+
+        // We must store LPNFT denoms before calling transactionInfo()
```
grod220, line 66: can we add a bit more on why?

packages/query/src/block-processor/transactions.ts
Valentine1898: Yes, it makes sense to avoid code duplication

grod220 (approved):

PR #263, opened 2023-11-29
implement IndexByAddress

Vadim121197:


packages/wasm/src/keys.ts
```diff
@@ -28,6 +32,23 @@ export const getEphemeralByIndex = (fullViewingKey: string, index: number): Addr
...
+    }),
+    is_controlled_address(fullViewingKey, address),
+  );
+
+  const uintArray = base64ToUint8Array(res.randomizer);
+
+  return new AddressIndex({
+    account: res.account ?? 0,
+    randomizer: uintArray,
+  });
```
grod220, line 49: I think we can do AddressIndex.fromJson(res) right?

packages/wasm/src/keys.ts
```diff
@@ -28,6 +32,23 @@ export const getEphemeralByIndex = (fullViewingKey: string, index: number): Addr
   return new Address({ inner: uintArray });
 };
 
+export const getIndexByAddress = (fullViewingKey: string, address: string): AddressIndex => {
+  const res = validateSchema(
+    z.object({
+      account: z.number().optional(),
+      randomizer: z.string(),
+    }),
+    is_controlled_address(fullViewingKey, address),
```
grod220, line 41: Is appears to return a `Option<pb::AddressIndex>`. Does this mean that the result could potentially be undefined? If so, that zod type is not right.

packages/wasm/src/keys.test.ts
```diff
@@ -77,4 +79,16 @@ describe('keys', () => {
       }).not.toThrow();
     });
   });
+
+  describe('getIndexByAddress()', () => {
+    it('does not raise zod validation error', () => {
```
grod220, line 84: Given it's an `Option<pb::AddressIndex>`, can you provide a case that returns `undefined`?

packages/types/src/address.ts
```diff
@@ -13,3 +13,9 @@ export const bech32Address = (addr: Address): string =>
 
 export const isPenumbraAddr = (addr: string): boolean =>
   addr.length === BECH32_ADDRESS_LENGTH && addr.startsWith(`${BECH32_PREFIX}1`);
+
+// need to decode address penumbra147mfall.. to Uint8Array (used for IndexByAddressRequest)
+export const bech32ToUint8Array = (addr: string): Uint8Array => {
+  const decodeAddress = bech32m.decode(addr, BECH32_ADDRESS_LENGTH);
+  return new Uint8Array(bech32m.fromWords(decodeAddress.words));
+};
```
grod220, line 21: this doesn't seem to be used anywhere

packages/router/src/grpc/view-protocol-server/index-by-address.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const isIndexByAddressRequest = (req: ViewReqMessage): req is IndexByAddressRequest => {
+  return req.getType().typeName === IndexByAddressRequest.typeName;
+};
+
+export const handleIndexByAddressReq = async (
+  req: IndexByAddressRequest,
+): Promise<IndexByAddressResponse> => {
+  const wallets = await localExtStorage.get('wallets');
+  const address = bech32Address(req.address!);
```
grod220, line 18: Instead of using `!` let's have a type guard above it:

```typescript
if (!req.address) throw new Error('no address given in request')

```

packages/wasm/src/keys.ts
Vadim121197: if account = 0, then return only randomizer

grod220: But let's say you pass an address into `is_controlled_address` that isn't connected with that fullViewingKey, does it return undefined?

packages/wasm/src/keys.ts
Vadim121197: ok, but you use```
 new AddressIndex({
    account: res.account ?? 0,
    randomizer: uintArray,
  });
``` too
<img width="526" alt="Знімок екрана 2023-11-29 о 21 05 09" src="https://github.com/penumbra-zone/web/assets/52686175/3a2faa93-e9d2-4131-836a-bb9c91a31a81">

Comment (Vadim121197) [reply packages/wasm/src/keys.ts:49]: ok, but you use 
```
new AddressIndex({
    account: res.account ?? 0,
    randomizer: uintArray,
  })
```

File: packages/types/src/address.ts (hunk 4 above)
Comment (Vadim121197) [reply packages/types/src/address.ts:21]: yes, only for test `getAddressByIndex`

File: packages/wasm/src/keys.ts (hunk 1 above)
Comment (grod220) [reply packages/wasm/src/keys.ts:49]: But we don't need to manually deserialize it if the `fromJson` function is given to us

Comment (Vadim121197) [reply packages/wasm/src/keys.ts:49]:
```
 return new AddressIndex().fromJson({
    account: res.account ?? 0,
    randomizer: res.randomizer,
  });
```
this work, AddressIndex.fromJson(res) get types warnings, probably because account can be undefined

packages/wasm/src/keys.ts
Vadim121197: res as undefined

Vadim121197: i see, you mean is_controlled_address can return undefined

Vadim121197: what we should return if is_controlled_address return undefined?

Vadim121197: throw new Error('no address given in request')?

grod220 (approved):
@Vadim121197 Have a look at the changes I pushed. Please test out as well. If it looks good to you, merge.

PR #264, opened 2023-11-30
200 notesrequest

Vadim121197:


packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const handleNotesReq = async function* (
+  req: NotesRequest,
+  services: ServicesInterface,
+): AsyncIterable<NotesResponse> {
+  const { indexedDb } = await services.getWalletServices();
+  const allNotes = await indexedDb.getAllNotes();
+  const wallets = await localExtStorage.get('wallets');
+
+  if (
```
grod220, line 22: Go for something like this:

```typescript
// If wallet id passed in req, ensure it's in storage
async function assertWalletIdMatches(req: NotesRequest) {
  const wallets = await localExtStorage.get('wallets');
  if (
    req.walletId?.inner &&
    !wallets.find(wallet => wallet.id === uint8ArrayToString(req.walletId!.inner))
  ) {
    throw new Error('Invalid account ID');
  }
}

export const handleNotesReq = async function* (
  req: NotesRequest,
  services: ServicesInterface,
): AsyncIterable<NotesResponse> {
  await assertWalletIdMatches(req);
```

packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -0,0 +1,61 @@
...
+    req.walletId?.inner &&
+    !wallets.find(wallet => wallet.id === uint8ArrayToString(req.walletId!.inner))
+  ) {
+    throw new Error('Invalid account ID');
+  }
+
+  const filteredNotes = allNotes
+    // filter by assetId
```
grod220, line 30: Don't think we need these comments, doesn't add that much value

packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -0,0 +1,61 @@
...
+    .filter(note => (req.assetId ? note.note?.value?.assetId?.equals(req.assetId) : note))
```
grod220, line 31: Should it be like this for all the filters? 

```diff
- .filter(note => (req.assetId ? note.note?.value?.assetId?.equals(req.assetId) : note))
+ .filter(note => (req.assetId ? note.note?.value?.assetId?.equals(req.assetId) : true))
```

packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -0,0 +1,61 @@
...
+    // filter by addressIndex
+    .filter(note => (req.addressIndex ? note.addressIndex?.equals(req.addressIndex) : note))
+    // filter by heightSpent
+    .filter(note => (!req.includeSpent ? !note.heightSpent : note));
+
+  let response: SpendableNoteRecord[];
+
+  if ((req.amountToSpend && (!req.assetId || req.includeSpent)) ?? !req.amountToSpend) {
```
grod220, line 39: This logic is quite confusing to me. Think we should add more clarity to this and add more code comments.

grod220: Think this works:

```typescript


// If wallet id passed in req, ensure it's in storage
async function assertWalletIdMatches(req: NotesRequest) {
  const wallets = await localExtStorage.get('wallets');
  if (
    req.walletId?.inner &&
    !wallets.find(wallet => wallet.id === uint8ArrayToString(req.walletId!.inner))
  ) {
    throw new Error('Invalid account ID');
  }
}

export const handleNotesReq = async function* (
  req: NotesRequest,
  services: ServicesInterface,
): AsyncIterable<NotesResponse> {
  await assertWalletIdMatches(req);

  const allNotes = await indexedDb.getAllNotes();

  const { assetId, addressIndex, includeSpent, amountToSpend } = req;

  let spent = new Amount();

  yield* allNotes
    .filter(n => (assetId ? n.note?.value?.assetId?.equals(assetId) : true))
    .filter(n => (addressIndex ? n.addressIndex?.equals(addressIndex) : true))
    .filter(n => (!includeSpent ? !n.heightSpent : true))
    .filter(n => {
      // If set, stop returning notes once the total exceeds this amount.
      // Ignored if `assetId` is unset or if `includeSpent` is set.
      if (!(amountToSpend && assetId && !includeSpent)) return true;

      const noteAmount = n.note?.value?.amount ?? new Amount();
      spent = addAmounts(spent, noteAmount);

      return joinLoHiAmount(spent) <= joinLoHiAmount(amountToSpend);
    })
    .map(noteRecord => new NotesResponse({ noteRecord }));
};

```

packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+// If wallet id passed in req, ensure it's in storage
+async function assertWalletIdMatches(req: NotesRequest) {
+  const wallets = await localExtStorage.get('wallets');
+  if (
+    req.walletId?.inner &&
+    !wallets.find(wallet => wallet.id === uint8ArrayToString(req.walletId!.inner))
+  ) {
+    throw new Error('Invalid account ID');
```
grod220, line 26: Can you change this error to read:

```
walletId does not match walletIds in storage
```

grod220 (approved):

PR #266, opened 2023-12-01
244 user should be able to choose the account from which he is going to spend funds when sending UI issue

Vadim121197:


apps/webapp/src/components/send/send-form.tsx
```diff
@@ -1,49 +1,64 @@
...
+    useStore.setState(state => {
+      state.send.account = balancesByAccount[0]?.address ?? '';
+    });
+  }
+
+  if (balancesByAccount[0]?.balances.length) {
+    useStore.setState(state => {
+      state.send.asset = balancesByAccount[0]?.balances[0];
+    });
+  }
```
grod220, line 28: Can you add a code comment on what this is for? I'm assuming you want to initialize with defaults in form?

apps/webapp/src/state/send.ts
```diff
@@ -96,26 +91,31 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
   };
 };
 
-const getExponent = (asset: TempAsset): number => {
-  const match = asset.denomUnits.find(u => u.denom === asset.display);
-  return match?.exponent ?? 0;
-};
+const getExponent = (asset: AssetBalance): number => asset.denom.exponent;
```
grod220, line 94: If this is the function, we don't need it and can put `asset.denom.exponent` below

apps/webapp/src/state/send.ts
```diff
@@ -133,13 +133,11 @@ const planWitnessBuildBroadcast = async ({ amount, recipient, asset }: SendSlice
   return uint8ArrayToHex(id.hash);
 };
 
-export const amountToBig = (asset: TempAsset, assetBalance: Amount) => {
-  const exponent = asset.denomUnits.find(d => d.denom === asset.display)!.exponent;
-  return fromBaseUnitAmount(assetBalance, exponent);
-};
+export const amountToBig = (asset: AssetBalance) =>
```
grod220, line 136: Same with this function. Can just move `fromBaseUnitAmount(...)` where it's used

apps/webapp/src/state/send.ts
```diff
@@ -96,26 +91,31 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
...
-  const match = asset.denomUnits.find(u => u.denom === asset.display);
-  return match?.exponent ?? 0;
-};
+const getExponent = (asset: AssetBalance): number => asset.denom.exponent;
+
+const planWitnessBuildBroadcast = async ({ amount, recipient, asset, account }: SendSlice) => {
+  if (typeof account === 'undefined') throw new Error('no selected account');
+  if (!asset) throw new Error('no selected asset');
+
+  const source = await getIndexByAddress(account);
```
grod220, line 100: I don't think it's necessary to make the request if, in the dialog, the index is selected along with the asset. We should already know the index. Let's change the SendSlice to this:


```typescript
interface Selection {
  address: string;
  accountIndex: number;
  asset: AssetBalance;
}

export interface SendSlice {
  selection: Selection | undefined;
  ...

```

This way, if the user selected something, we always have access to the things that come with the selection without needing further requests.

apps/webapp/src/components/send/ibc/ibc-form.tsx
```diff
@@ -1,33 +1,15 @@
...
-import InputToken from '../../shared/input-token.tsx';
-import { amountToBig } from '../../../state/send.ts';
-import { useLoaderData } from 'react-router-dom';
-import { AssetBalance } from '../../../fetchers/balances.ts';
-import { uint8ArrayToBase64 } from '@penumbra-zone/types';
-import { Amount } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/num/v1alpha1/num_pb';
 import { ChainSelector } from './chain-selector.tsx';
 import { InputBlock } from '../../shared/input-block.tsx';
 
+// TODO: Re-implementing ibc form, see send form
```
grod220, line 8: Could you follow up with a second PR to implement this for IBC as well? (as well as the validations plz)

apps/webapp/src/components/shared/asset-icon.tsx
```diff
@@ -1,13 +1,18 @@
-import { Asset } from '@penumbra-zone/types';
+import { assets } from '@penumbra-zone/constants';
 import { Identicon } from '@penumbra-zone/ui';
+import { useMemo } from 'react';
+
+export const AssetIcon = ({ name }: { name: string }) => {
+  const icon = useMemo(() => {
+    return assets.find(i => i.display === name)?.icon;
+  }, [name]);
```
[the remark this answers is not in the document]
Vadim121197:
```
{icon ? (
        <img className='h-6 w-6 rounded-full' src={icon} alt='Asset icon' />
      ) : (
        <Identicon name={name} size={24} className='rounded-full' type='solid' />
      )}
```
if icon undefined then show Identicon

apps/webapp/src/state/send.ts
Vadim121197: we do not have a randomize value for AddressIndex! or do they not need it?

grod220: If we need to pass a source address, I don't think we need randomizer. The randomizer is only relevant for ephemeral addresses. Please test though.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -73,11 +78,13 @@ export default function InputToken({
             value && 'text-muted-foreground',
           )}
         >
-          ${displayAmount(Number(value) * asset.price)}
+          ${displayAmount(Number(value) * (asset?.usdcValue ?? 0))}
```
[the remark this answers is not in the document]
Vadim121197:
```
const normalize =
  (metadata: AssetsResponse[], indexAddrRecord: IndexAddrRecord | undefined) =>
  (res: BalancesResponse): NormalizedBalance => {
    const index = res.account?.account ?? 0;
    const address = indexAddrRecord?.[index] ?? '';

    const { display, exponent, amount } = getDenomAmount(res, metadata);

    return {
      denom: { display, exponent },
      assetId: res.balance!.assetId!,
      amount,
      //usdcValue: amount * 0.93245, // TODO: Temporary until pricing implemented
      usdcValue: 0, // Important not to imply that testnet balances have any value
      account: { index, address },
    };
  };
```
i can add price there

apps/webapp/src/state/send.ts
Vadim121197: works! i'll delete this

Issue #233, opened 2023-11-09
Implement real sync detection


grod220:
At the moment, it's a fake sleep timer: https://github.com/penumbra-zone/web/blob/efbed4a417606a9e92f134b981d1f955e7ef287e/packages/router/src/grpc/view-protocol-server/broadcast.ts#L22-L23

It should be waiting until it is scanned in a block

Valentine1898:
https://github.com/penumbra-zone/web/blob/02d7a61afa04bb596ce25d799749b37caab266a4/packages/router/src/grpc/view-protocol-server/note-by-commitment.ts#L19
we should add await_detection handling for NoteByCommitmentRequest after resolving this issue

PR #273, opened 2023-12-06
added note by commitment request handler

Valentine1898:


packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const handleNoteByCommitmentReq = async (
+  req: NoteByCommitmentRequest,
+  services: ServicesInterface,
+): Promise<NoteByCommitmentResponse> => {
+  const { indexedDb } = await services.getWalletServices();
+  if (!req.noteCommitment) return new NoteByCommitmentResponse();
+
+  // TODO  req.await_detection processing
```
grod220, line 19: We have an issue for await detection somewhere. Can you add a todo in that issue to implement this?

packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
```diff
@@ -0,0 +1,25 @@
...
+export const isNoteByCommitmentRequest = (req: ViewReqMessage): req is NoteByCommitmentRequest => {
+  return req.getType().typeName === NoteByCommitmentRequest.typeName;
+};
+
+export const handleNoteByCommitmentReq = async (
+  req: NoteByCommitmentRequest,
+  services: ServicesInterface,
+): Promise<NoteByCommitmentResponse> => {
+  const { indexedDb } = await services.getWalletServices();
+  if (!req.noteCommitment) return new NoteByCommitmentResponse();
```
grod220, line 17: I feel like we should throw an error. Or does the Rust implementation send an empty response like this?

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -191,6 +191,15 @@ describe('IndexedDb', () => {
...
     });
+
+    it('should be able to set/get by commitment', async () => {
+      const db = await IndexedDb.initialize({ ...generateInitialProps() });
+
+      await db.saveSpendableNote(newNote);
+      const noteByCommitment = await db.getNoteByCommitment(newNote.noteCommitment!);
+
+      expect(newNote.equals(noteByCommitment)).toBeTruthy();
+    });
```
grod220, line 202: Can you put an undefined case as well?

packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
Valentine1898: Rust imlementation throws an error. I noticed that in one more request we return an empty response where we should throw an error
https://github.com/penumbra-zone/web/blob/89d4ec182ec5946410bea0b057e45afa00be6552/packages/router/src/grpc/view-protocol-server/tx-info-by-hash.ts#L18

packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
Valentine1898: Yes, I was planning to add a TODO link to https://github.com/penumbra-zone/web/issues/233 after the merge

grod220 (approved):

PR #275, opened 2023-12-06
Vite update

Vadim121197:


Vadim121197:
what is the difference between pnpm build locally and github actions?

Valentine1898:
It looks like the difference between local build and github actions build is in the handling of `Buffer -> buffer `cases

Vadim121197:
I tried to find an analog of the buffer library, but buffer was updated 3 years ago! https://www.npmjs.com/package/buffer-lite tried to use this library and everything works! maybe you know another library?

jessepinho:
Superseded by #391

PR #278, opened 2023-12-07
version bump buf deps

### LINKED ISSUE #276: version bump buf deps
requested in review

turbocrime:
keyValue is no longer available on AppQuerier

this only affected assertRootValid in block-processor, which was not specifically used anywhere

Fixes #276


packages/query/src/block-processor/index.ts
```diff
@@ -126,20 +122,6 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-  private async assertRootValid(blockHeight: bigint): Promise<void> {
-    const sourceOfTruth = await this.querier.app.keyValue(`sct/anchor/${blockHeight}`);
-    const inMemoryRoot = this.viewServer.getSctRoot();
-
-    if (!decodeSctRoot(sourceOfTruth).equals(inMemoryRoot)) {
-      throw new Error(
-        `Block height: ${blockHeight}. Wasm root does not match remote source of truth. Programmer error.`,
-      );
-    }
-  }
```
grod220, line 141: It's actually important for us to keep this. In dev mode, we use this on occasion to assert that the filtered state tree we are creating is valid. We do this by querying a root hash and comparing it to our own. Think we need to hunt down where someone moved this query to 🤔

turbocrime: the type is now located here

https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.storage.v1alpha1#penumbra.storage.v1alpha1.KeyValueResponse.Value

but i don't think this is accessible via any service we currently implement.

there seem to be some replacement endpoints on the query service

https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.app.v1alpha1#penumbra.core.app.v1alpha1.QueryService

turbocrime: there are commits today so i think it's in motion

turbocrime:
it's worth considering managing these versions at the workspace root rather than individually in every workspace

packages/query/src/block-processor/index.ts
grod220: Hi @erwanor 👋 , looks like this query was removed: https://github.com/penumbra-zone/penumbra/commit/2b48cf12ff35720c433999f4a0135d18a029fc8b. Is there another place we can make a query to validate our local tree?

erwanor: > Hi @erwanor 👋 , looks like this query was removed: [penumbra-zone/penumbra@2b48cf1](https://github.com/penumbra-zone/penumbra/commit/2b48cf12ff35720c433999f4a0135d18a029fc8b). Is there another place we can make a query to validate our local tree?

The KV RPCs are in the storage query service. I didn't realize this was used here, will flag this next time, sorry about that and thanks for tracking me down!

turbocrime:
@grod220 does https://github.com/penumbra-zone/web/pull/278#discussion_r1418223514 indicate version bump depends on a new storage query service

PR #279, opened 2023-12-07
Support for tx approvals

### LINKED ISSUE #235: Transaction approval UI
See https://github.com/penumbra-zone/web/pull/126 for UI inspiration

grod220:
Closes #235

Adds tx approvals to web app. Requires writing a new internal msg types, handlers, routing, etc. Given we already have std router, we can borrow a lot from there.

Approval UI isn't amazing. But we can improve this later.


turbocrime:
this might seem a little more complex than a simple awaited postmessage but it nicely handles a few null attitudes and sidesteps react's inability to act normal

Issue #267, opened 2023-12-01
Load proving keys from disk


grod220:
Currently, at runtime, we are fetching the proving keys to build: https://github.com/penumbra-zone/web/blob/main/packages/wasm/src/build.ts#L44-L54.

This means we'll incur some network fetch cost at this stage of the transaction. We should try to avoid this.

Options:

- Bundle proving keys within `@penumbra-zone/wasm-bundler` NPM package: https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/publish/run.ts
  - At the moment, we are running `wasm-pack` to create the npm package and simply publishing it.
  - We could not just put the bin files in the package. For consumers, that is not a bundle-friendly format, we should provide something more helpful.
  - We'd need to create a wrapper NPM package that would expose helpful loadProvingKey() utilities (that ideally would be saved as a Uint8Array instead of a bin file) as well as re-export the functions from the output of wasm-pack. This is not trivial though. May require some webpack magic.
- In the extension, add a webpack build step that downloads the latest proving keys and saves it locally
   - In this case, it could be a bin file if we were to add a loader to our webpack config or it could be saved as another format too
   - The files+loading utilities would then be bundled up with our normal extension package
   - May be the simpler path

PR #283, opened 2023-12-07
refactor: fetch proving keys

TalDerei:
References #267

Bundling the proving keys with the extension package incurs a one-time network cost (during build) to download the binary files. This is preferable to paying the cost at runtime.


apps/extension/src/utils/download-proving-keys.mjs
grod220, line 1: Let's convert this to a typescript file. That means we'll need a different way to run it. Think we can use https://www.npmjs.com/package/ts-node and do `ts-node <file>` and it should work.

packages/wasm/src/build.ts
```diff
@@ -28,38 +37,29 @@ export const witness = (txPlan: TransactionPlan, sct: StateCommitmentTree): Witn
...
-];
-
-const loadProvingKeys = async () => {
-  const promises = provingKeys.map(async ({ file, keyType }) => {
-    const response = await fetch(`${githubSourceDir}${file}`);
-    if (!response.ok) throw new Error(`Failed to fetch ${file}`);
-
-    const buffer = await response.arrayBuffer();
-    wasmLoadProvingKey(buffer, keyType);
-  });
```
grod220, line 51: I'm personally more of a fan of having this array and just looping over it like so. It really only looks like you need to do:

```diff
- const response = await fetch(`${githubSourceDir}${file}`);
+ const response = await fetchBinaryFile(`${githubSourceDir}${file}`);
```

packages/wasm/src/build.ts
```diff
@@ -18,6 +18,15 @@ import {
   witness as wasmWitness,
 } from '@penumbra-zone/wasm-bundler';
 
+export async function fetchBinaryFile(filename: string) {
```
grod220, line 21: Fetch feels a bit like it's making a network call. Could we do something like `loadLocalBinary`?

apps/extension/src/utils/download-proving-keys.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+const provingKeyFiles = [
+  { file: 'spend_pk.bin' },
+  { file: 'output_pk.bin' },
+  { file: 'swap_pk.bin' },
+  { file: 'swapclaim_pk.bin' },
+  { file: 'nullifier_derivation_pk.bin' },
+  { file: 'delegator_vote_pk.bin' },
+  { file: 'undelegateclaim_pk.bin' },
+];
```
grod220, line 16: Could we re-use one array instead of having this in two places? Perhaps we can move https://github.com/penumbra-zone/web/pull/283/files#diff-792d69d65ae8bb0bcbeeec13edcf4c7aded7bf621432261ac326705f69d88d29L34-L42 to the types pack and both consumers of this can share it.

apps/extension/src/utils/download-proving-keys.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+const binaryFilesDir = path.join('dist/bin');
+
+const downloadProvingKeys = async () => {
+  // Check if the bin directory already exists
+  if (!fs.existsSync(binaryFilesDir)) {
```
grod220, line 22: The inverse guard would be nicer imo. 

```diff
- if (!fs.existsSync(binaryFilesDir)) {
+ if (fs.existsSync(binaryFilesDir)) return
```

apps/extension/src/utils/download-proving-keys.mjs
```diff
@@ -0,0 +1,44 @@
...
+      if (!response.ok) throw new Error(`Failed to fetch ${file}`);
+
+      const buffer = await response.arrayBuffer();
+
+      const outputPath = path.join(binaryFilesDir, file);
+      if (!fs.existsSync(outputPath)) {
+        const fileStream = fs.createWriteStream(outputPath, { flags: 'a' });
+        fileStream.write(Buffer.from(buffer));
+        fileStream.end();
+        console.log(`${file}`, 'downloaded to:', outputPath);
```
grod220, line 36: Can we add: 
```diff
- console.log(`${file}`, 'downloaded to:', outputPath);
+ console.log(`Proving key ${file} downloaded to ${outputPath}`);
```

apps/extension/src/utils/download-proving-keys.mjs
TalDerei: that's a good idea, changed

apps/extension/src/utils/download-proving-keys.mjs
```diff
@@ -0,0 +1,44 @@
...
+    fs.mkdirSync(binaryFilesDir, { recursive: true });
+
+    const promises = provingKeyFiles.map(async ({ file }) => {
+      const response = await fetch(`${githubSourceDir}${file}`);
+      if (!response.ok) throw new Error(`Failed to fetch ${file}`);
+
+      const buffer = await response.arrayBuffer();
+
+      const outputPath = path.join(binaryFilesDir, file);
+      if (!fs.existsSync(outputPath)) {
```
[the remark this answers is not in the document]
TalDerei: we don't actually need this clause

apps/extension/src/utils/download-proving-keys.mjs
TalDerei: changed

apps/extension/src/utils/download-proving-keys.mjs
```diff
@@ -0,0 +1,44 @@
...
+  { file: 'swapclaim_pk.bin' },
+  { file: 'nullifier_derivation_pk.bin' },
+  { file: 'delegator_vote_pk.bin' },
+  { file: 'undelegateclaim_pk.bin' },
+];
+
+const binaryFilesDir = path.join('dist/bin');
+
+const downloadProvingKeys = async () => {
+  // Check if the bin directory already exists
```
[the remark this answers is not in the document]
TalDerei: added

packages/wasm/src/build.ts
TalDerei: changed

apps/extension/src/utils/download-proving-keys.mjs
TalDerei: This didn't seem to work. I was running into strange `ts-node` configuration issues in the extension package when dealing with `path` and `fs` imports. I modified the package.json with `"type": "module"` and was with different settings in the tsconfig.json, but I kept running into `cannot find module 'path' or its corresponding type declarations`. 

could you take a stab at it?

packages/wasm/src/build.ts
TalDerei: I added a `utils.ts` file in the `wasm` package with this pattern.

apps/extension/src/utils/download-proving-keys.mjs
grod220: There were a couple complexities with esm modules. In the end, was easier to just use tsx that has bundler assistance built in.

grod220 (approved):

PR #294, opened 2023-12-12
show chain id

Vadim121197:


apps/extension/src/clients/index.ts
```diff
@@ -1,12 +0,0 @@
...
-import { createChannelTransport, ChannelClientLabel } from '@penumbra-zone/transport';
-import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-runtime/client-connection-manager';
-
-export const grpcClient = createPromiseClient(
-  ViewProtocolService,
-  createChannelTransport(
-    ViewProtocolService,
-    ClientConnectionManager.init(ChannelClientLabel.Extension),
-  ),
-);
```
grod220, line 12: It looks like you found a workaround, but this should be able to work. There is something deeper wrong. Can you do more debugging to see why the popup is not able to use this grpc client?

Vadim121197: Ok

turbocrime: sorry, this looks like my fault. missed this internal client init in one of the refactors during review. intra-extension transport init requires one more step. i'll correct this in a moment

turbocrime:
Screenshot 2023-12-12 at 15 34 57

turbocrime (approved):
corrected init, cleaned up problematic unused checks in the background manager. this is working now

Issue #293, opened 2023-12-11
On onboarding, internal error


grod220:
To replicate, install the extension fresh and go to web app. Without having gone through the onboarding phase, the web app does not load and offers unclear error.

<img width="799" alt="Screenshot 2023-12-11 at 3 54 16 PM" src="https://github.com/penumbra-zone/web/assets/16624263/d4511bee-71b0-48e4-a889-403e7e42abf3">

turbocrime:
there is a behavior in place now in #295 that delays requests until connection is certain

https://github.com/penumbra-zone/web/blob/a846fabaddbabca25e315348dfcb8bb584358b0b/packages/transport/src/create.ts#L100

additionally, errors from the services now transmit through the client, and surface in the page

<img width="624" alt="Screenshot 2023-12-12 at 19 13 42" src="https://github.com/penumbra-zone/web/assets/134443988/694606fe-243e-47d9-a927-7d346a6957c7">

i think that's "fixed" for the purposes of this issue but the webapp still does not understand the situation of "extension is present but i can't talk to it yet"

Issue #287, opened 2023-12-11
More descriptive errors in background connection manager


turbocrime:

turbocrime:
fixed in #295

PR #295, opened 2023-12-12
more robust error propagation and handling

### LINKED ISSUE #280: Service worker error propogation
Related to https://github.com/penumbra-zone/web/pull/255

As the service worker handles an incoming grpc request, if in handling it raises, this error does not propagate back to the web app.

Easiest way to test is to pull down https://github.com/penumbra-zone/web/pull/279, and decline a tx approval request. It throws in the service worker but the client hangs indefinitely.

turbocrime:
fixes #280 #287 and should make debugging any failures a bit nicer

### DIFF
--- a/packages/transport/src/chrome-runtime/background-connection-manager.ts
+++ b/packages/transport/src/chrome-runtime/background-connection-manager.ts
@@ -17,6 +17,7 @@ import { JsonValue } from '@bufbuild/protobuf';
 import { ChromeRuntimeStreamSink } from './stream';
 import { ConnectError, Code as ConnectErrorCode } from '@connectrpc/connect';
 import { errorToJson } from '@connectrpc/connect/protocol-connect';
+import { isDisconnectedPortError } from './errors';
 
 interface BackgroundConnection {
   type: ChannelClientLabel;
@@ -155,13 +156,16 @@ export class BackgroundConnectionManager {
           if (subPort.name !== channel) return;
           chrome.runtime.onConnect.removeListener(subListener);
           connection.streams.set(channel, acont);
-          subPort.onDisconnect.addListener(() => acont.abort(channel));
+          subPort.onDisconnect.addListener(() => acont.abort(`Disconnected: ${channel}`));
           void stream
             .pipeTo(new WritableStream(new ChromeRuntimeStreamSink(subPort)), {
               signal: acont.signal,
             })
             .catch((error: unknown) => {
-              if (error !== channel) throw error;
+              // This throw won't reach the client.  The failure has already
+              // been transmitted in-channel by the stream sink if possible, and
+              // the client source will fail in a meaningful way.
+              if (!isDisconnectedPortError(error)) throw error;
             })
             .finally(() => {
               connection.streams.delete(channel);
--- a/packages/transport/src/chrome-runtime/background-connection-manager.ts
+++ b/packages/transport/src/chrome-runtime/background-connection-manager.ts
@@ -10,11 +10,11 @@ import {
   TransportError,
   TransportStream,
   TransportState,
+  TransportEvent,
 } from '../types';
 
 import { JsonValue } from '@bufbuild/protobuf';
 import { ChromeRuntimeStreamSink } from './stream';
-import { isDisconnectedPortError } from './errors';
 import { ConnectError } from '@connectrpc/connect';
 import { errorToJson } from '@connectrpc/connect/protocol-connect';
 
@@ -32,6 +32,7 @@ type UnconditionalServiceAccessFn = (
   x: JsonValue,
 ) => Promise<JsonValue | ReadableStream<JsonValue>>;
 
+// TODO: conditional service access
 type UnconditionalServiceAccess = Record<string, UnconditionalServiceAccessFn>;
 
 /**
@@ -106,6 +107,7 @@ export class BackgroundConnectionManager {
         return;
       }
 
+      // create connection record
       const connection: BackgroundConnection = {
         tab,
         type: clientType as ChannelClientLabel,
@@ -115,16 +117,17 @@ export class BackgroundConnectionManager {
         tlsChannelId,
         documentId,
       };
+      originConnections?.set(clientId, connection);
 
       /**
-       * This method is used to manage a sub-channel, encapsulating a response
-       * stream. A TransportInitChannel representing the channel is returned for
-       * transport to the client. The client should respond by opening a
-       * connection bearing this name.
+       * This method is used to manage a response sub-channel, encapsulating a
+       * response stream. A TransportInitChannel representing the channel is
+       * returned for transport to the client. The client should respond by
+       * opening a connection bearing this name.
        *
        * @param subStream ReadableStream to be transmitted via sub-channel
        */
-      const initSubChannel = ({ requestId, stream }: TransportStream): TransportInitChannel => {
+      const responseSubChannel = ({ requestId, stream }: TransportStream): TransportInitChannel => {
         const [channel] = nameChannel(ChannelSubLabel.ServerStream);
         const acont = new AbortController();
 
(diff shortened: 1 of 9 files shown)


turbocrime:
fixes #293

packages/transport/src/chrome-runtime/background-connection-manager.ts
```diff
@@ -72,105 +77,163 @@ export class BackgroundConnectionManager {
...
+        // TODO: revisit this
+        /*
+        setTimeout(() => {
+          if (chrome.runtime.onConnect.hasListener(subListener)) {
+            // handler still hasn't activated, or it would be gone
+            chrome.runtime.onConnect.removeListener(subListener);
+            acont.abort('Timed out subchannel activation');
+          }
+        }, 1000);
+        */
```
grod220, line 176: Can we either delete this or give more info in the TODO as to what this is about and why it needs revisiting?

packages/transport/src/chrome-runtime/background-connection-manager.ts
```diff
@@ -72,105 +77,163 @@ export class BackgroundConnectionManager {
...
+          void stream
             .pipeTo(new WritableStream(new ChromeRuntimeStreamSink(subPort)), {
               signal: acont.signal,
             })
-            .catch(e => {
-              if (e !== subName) throw e;
+            .catch((error: unknown) => {
+              if (error !== channel) {
+                // TODO: handling errors in this promise chain is problematic.
+                // we're in a callback, and can't reliably answer the requester.
```
grod220, line 150: But isn't the postMessage below answering the requestor? When you say it's not reliable, does that mean it only sends sometimes?

packages/transport/src/chrome-runtime/background-connection-manager.ts
```diff
@@ -72,105 +77,163 @@ export class BackgroundConnectionManager {
...
+
+        const { requestId, message: request } = transportRequest;
+
+        /*
+        const serviceTimeout = new Promise<void>((_, reject) =>
+          setTimeout(reject, 1000, 'Timed out service request'),
+        );
+
+        const transportResponse = Promise.race([serviceEntry(request), serviceTimeout])
+        */
```
grod220, line 198: Delete?

packages/transport/src/chrome-runtime/background-connection-manager.ts
```diff
@@ -72,105 +77,163 @@ export class BackgroundConnectionManager {
...
+                error: errorToJson(ConnectError.from(error), undefined),
+              }) as TransportError,
+          );
+
+        transportResponse
+          .then(extResponse => transPort.postMessage(extResponse))
+          .catch(e => {
+            // TODO: abort requests
+            if (!isDisconnectedPortError(e)) throw e;
+          });
```
grod220, line 219: These can be combined right? Something like:

```typescript
serviceEntry(request)
          .then(response => {
            const extResponse =
              response instanceof ReadableStream
                ? initSubChannel({ requestId, stream: response })
                : ({ requestId, message: response } as TransportMessage);
            transPort.postMessage(extResponse);
          })
          .catch((error: unknown) => {
            const tErr = { requestId, error: String(error) } as TransportError;
            if (!isDisconnectedPortError(tErr)) throw error;
          });
```

packages/transport/src/chrome-runtime/client-connection-manager.ts
```diff
@@ -129,20 +128,25 @@ export class ClientConnectionManager {
...
+            const sourcePort = chrome.runtime.connect({ includeTlsChannelId: true, name: subName });
+            const stream = new ReadableStream(new ChromeRuntimeStreamSource(sourcePort));
+            clientPort.postMessage({ requestId, stream }, [stream]);
+          } else throw Error('Unimplemented response kind');
+        } else if (isTransportError(ev)) clientPort.postMessage(ev);
+        else throw Error('Unknown transport from service');
+      } catch (error) {
+        console.error('Error in service listener', error, ev);
+        clientPort.postMessage({ error: String(error) });
+      }
```
grod220, line 149: This multi-conditional block is a bit hard to read. Not sure how to improve though 🤔

packages/transport/src/create.ts
```diff
@@ -45,60 +53,99 @@ const makeAnyServiceImpl: CreateAnyServiceImpl = <S extends ServiceType>(
...
+      callbackError ??= ConnectError.from(error);
+      throw callbackError;
     }
   });
 
-  const request = (msg: AnyMessage) => {
+  const request = async (msg: AnyMessage) => {
+    if (callbackError) throw callbackError;
+    await providerAck;
+    // TODO: timeout ack? webapp needs to handle error states
```
grod220, line 101: I think a timeout would be helpful. Can be done later I supppose.

packages/transport/src/types/channel-messages.ts
```diff
@@ -1,78 +1,77 @@
...
   connected: boolean;
+  reason?: JsonValue;
 }
 
-export interface TransportMessage extends TransportEvent {
+export interface TransportError extends Partial<TransportEvent> {
+  error: JsonValue;
+}
+
+// transport actual
```
grod220, line 14: What's this mean?

packages/transport/src/types/channel-messages.ts
```diff
@@ -1,78 +1,77 @@
...
+
+export type TransportData = TransportMessage | TransportStream | TransportInitChannel;
+
+export interface TransportEvent {
   requestId: ReturnType<typeof crypto.randomUUID>;
+  responseOrigin?: string; // currently unused
```
grod220, line 20: Should we delete then?

packages/transport/src/chrome-runtime/background-connection-manager.ts
```diff
@@ -5,21 +5,33 @@ import {
...
-  tlsChannelId: string;
+  documentId: string | undefined;
+  tlsChannelId: string | undefined;
 }
 
+type UnconditionalServiceAccessFn = (
+  x: JsonValue,
+) => Promise<JsonValue | ReadableStream<JsonValue>>;
+
+type UnconditionalServiceAccess = Record<string, UnconditionalServiceAccessFn>;
```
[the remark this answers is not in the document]
grod220: Sorry, was this addressed? Or it's just a temporary todo to fix later?

packages/transport/src/create.ts
```diff
@@ -45,60 +54,108 @@ const makeAnyServiceImpl: CreateAnyServiceImpl = <S extends ServiceType>(
...
+  let callbackError: Error | undefined;
+
+  let connected: boolean | undefined;
+
+  let connectionState: (a: boolean, r?: JsonValue) => void;
+  const providerAck = Promise.race([
+    new Promise<void>((_, reject) =>
+      setTimeout(
+        reject,
+        1000,
```
grod220, line 72: Can we make this an constant at the top of the file (e.g. `const REQUEST_TIMEOUT = 1000`)? or perhaps a part of a config that gets passed in?

packages/transport/src/types/channel-messages.ts
```diff
@@ -1,78 +1,76 @@
...
-  channed: unknown,
-): channed is StreamChannelChunk | StreamChannelEnd =>
-  typeof channed === 'object' &&
-  channed !== null &&
-  'sequence' in channed &&
-  typeof channed.sequence === 'number' &&
-  ('value' in channed || 'done' in channed);
+// guards
+
+const obj = (o: unknown): o is object => typeof o === 'object' && o !== null;
```
grod220, line 50: maybe `isObj()`?

grod220 (approved):
Just had a few more comments.

Also, in general, not all of my comments need responses, but some it would be valuable to reply to.

packages/transport/src/chrome-runtime/background-connection-manager.ts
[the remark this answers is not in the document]
turbocrime: ah, sorry, i thought i had replied to this. this is todo.

turbocrime:
ugh i was using the vscode ui to reply to things and it looks like it dropped all my comments? so that's why a bunch of things were resolved with no obvious interaction

PR #299, opened 2023-12-13
don't remove init listener after activation

### LINKED ISSUE #297: [Bug] Other wallet extensions block transport between the webapp and extension
If the user has other wallet extensions installed (like Leap wallet or Namada) then webapp cannot setup the penumbra extension transport. 
The user just gets a black screen instead of webapp. 
image

turbocrime:
[Bug] Other wallet extensions block transport between the webapp and extension
Fixes #297

obviously intended to be inside the conditional, but i guess there is not really any reason to deactivate it


Valentine1898:
That solved the issue

Valentine1898 (approved):

grod220 (approved):

PR #300, opened 2023-12-13
implemented ExportFullViewingKey

Vadim121197:


packages/router/src/grpc/view-protocol-server/tx-planner.ts
```diff
@@ -7,18 +7,21 @@ import {
   Address,
   AddressIndex,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1alpha1/keys_pb';
-import { localExtStorage } from '@penumbra-zone/storage';
 import { ViewReqMessage } from './router';
-import { ServicesInterface } from '@penumbra-zone/types';
+import { ServicesInterface, uint8ArrayToString } from '@penumbra-zone/types';
+import { handleExportFullViewingKeyReq } from '../custody/export-full-viewing-key';
```
grod220, line 12: All of these should not be calling `handleExportFullViewingKeyReq` directly but making a grpc request to custody service

Valentine1898: `ViewService` cannot exist without the `FVK`. `FVK` must be passed during `ViewService` initialization, I don't think we have to request `FVK` from custody at runtime to get `FVK`.

Valentine1898: Rust viewer service stores the `full_viewing_key` in sqlite

grod220: > FVK must be passed during ViewService initialization

This may be a helpful thing to do, but here's nothing in the grpc service implementation that requires this.

> I don't think we have to request FVK from custody at runtime to get FVK

We hope that our view service implementation will one day be use by external partners. This means, in the implemented methods we shouldn't be attempting to grab global storage. Other folks may implement their custody methods completely different. This also means that the custody service may change wallets on us (and therefore the FVK changes). This would make it less beneficial to store the FVK at initialization.

I think the best solution is the simplest. Just swap the lines of code where we are reaching directly into localstorage and replace them with grpc calls to the grpc service for the FVK.

Valentine1898: But this solution allows that `FVK` can change after `View service` initialization? Changing `FVK` means that all scan results stored in `indexed-db` will not be up to date

hdevalence: The original intent for adding that method to the custody service was to be able to support a use case like a hardware wallet, where the root keys are stored on device and never exported.

I would actually suggest that we don't want to be using the custody RPC here. As a general guideline, I think we should think of the custody service as being a proxy for the custodian — in our case, the user — and having the user-in-the-loop for processing custody requests. If the view service is constantly requesting an FVK export, how does that work for a custody service backed by a hardware wallet?

I think it's cleaner to think of the custody service as being at a higher level of capability, a level of capability that's powerful enough that we want to have a human in the loop, and to think of the view service as being a lower level of capability, which we feel more comfortable giving automatic access to.

I don't understand the point about storage. The view server needs to store data anyways. Storing the FVK doesn't seem different than storing the view data it saves during scanning.

grod220: I suppose my point is around storage abstraction and how calling the custody service could provide that. For instance, let's take this line:
https://github.com/penumbra-zone/web/blob/main/packages/router/src/grpc/view-protocol-server/address.ts#L16

To fulfill this request, it relies upon the specific way we have stored the seed phrase in local storage. But what if we want say Leap to be able to use our ViewService implementation? They likely are handling seed phrase storage in a completely different way. They may not even be using local storage at all. So this implementation would not work for them.

However, if we replaced this line of code to a call to the custody service, the method of acquiring a FVK is abstracted and our implementation would work generically.

But that leads to another question, _when_ should this call be made? If we never anticipate a different FVK for the full-life of the ViewService, then we can call it at initialization. If there's a chance it could change, I think it's more appropriate at runtime.

cronokirby: I think of changing a custody service like changing which hardware wallet is plugged into your computer: the wallet could be completely different, and so the data in the view service doesn't make any sense at that point. Because of this, it makes more sense to call out to the custody service at the time we initialize the view service, and then cache the FVK thereafter.

hdevalence: @grod220 Wasn't the point of segmenting the data by wallet ID so that there could be multiple instances of view server data, one for each FVK? The FVK will never change for the entire lifetime of the wallet/chain. Supporting multiple wallets will mean modally switching which database is in use. There's nothing special about the FVK vs other view data. I'd Leap or whoever wanted to import this code they'd need to be able to store and retrieve notes too.

grod220: All of that is true, but I'm not sure how that relates to the core of this issue of this PR. It isn't the ViewServer database (indexeddb) but the CustodyService database (local storage). We store seed phrases in a specific way and this schema will not be shared by anyone else. Instead of a global dependency like this, Lucas' point sounds like the right approach: 

>  it makes more sense to call out to the custody service at the time we initialize the view service, and then cache the FVK thereafter.

grod220 (approved):

PR #301, opened 2023-12-13
Await detection impls

### LINKED ISSUE #233: Implement real sync detection

grod220:
Closes #233

Given we already have database subscriptions in place, this was much easier than I thought.

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/nullifier-status.ts
+++ b/packages/router/src/grpc/view-protocol-server/nullifier-status.ts
@@ -4,41 +4,50 @@ import {
   SpendableNoteRecord,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 import { ViewReqMessage } from './router';
-import { ServicesInterface } from '@penumbra-zone/types';
+import { IndexedDbInterface, ServicesInterface } from '@penumbra-zone/types';
 import { assertWalletIdMatches } from './utils';
+import type { Nullifier } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/sct/v1alpha1/sct_pb';
 
 export const isNullifierStatusRequest = (req: ViewReqMessage): req is NullifierStatusRequest => {
   return req.getType().typeName === NullifierStatusRequest.typeName;
 };
+
+async function noteOrSwapSpent(
+  indexedDb: IndexedDbInterface,
+  nullifier: Nullifier,
+): Promise<boolean> {
+  const noteByNullifier = await indexedDb.getNoteByNullifier(nullifier);
+  const swapByNullifier = await indexedDb.getSwapByNullifier(nullifier);
+
+  // The 'heightSpent' and 'heightClaimed' fields will never be undefined,
+  // so we compare to 0n assuming it is impossible to spend nullifier in block 0
+  const noteSpent = noteByNullifier ? noteByNullifier.heightSpent !== 0n : false;
+  const swapSpent = swapByNullifier ? swapByNullifier.heightClaimed !== 0n : false;
+  return noteSpent || swapSpent;
+}
 
 export const handleNullifierStatusReq = async (
   req: NullifierStatusRequest,
   services: ServicesInterface,
 ): Promise<NullifierStatusResponse> => {
   await assertWalletIdMatches(req.walletId);
-
-  const { indexedDb } = await services.getWalletServices();
   if (!req.nullifier) return new NullifierStatusResponse();
 
-  const noteByNullifier = await indexedDb.getNoteByNullifier(req.nullifier);
-  const swapByNullifier = await indexedDb.getSwapByNullifier(req.nullifier);
+  const { indexedDb } = await services.getWalletServices();
+  const spent = await noteOrSwapSpent(indexedDb, req.nullifier);
 
-  // The 'heightSpent' and 'heightClaimed' fields will never be undefined,
-  // so we compare to 0n assuming it is impossible to spend nullifier in block 0
-  const noteSpent = noteByNullifier ? noteByNullifier.heightSpent !== 0n : false;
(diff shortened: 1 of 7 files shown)


packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -29,5 +28,18 @@ export const handleNullifierStatusReq = async (
...
+  }
+
+  // Wait until our DB encounters a new note with this nullifier
+  const subscription = indexedDb.subscribe('SPENDABLE_NOTES');
+  for await (const update of subscription) {
+    const note = SpendableNoteRecord.fromJson(update.value);
+    if (note.nullifier?.equals(req.nullifier)) break;
+  }
+
+  return new NullifierStatusResponse({ spent: true });
```
Valentine1898, line 44: With `awaitDetection`, we should only return a response when the nullifier is spent, not when it is detected.

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -29,5 +28,18 @@ export const handleNullifierStatusReq = async (
   const noteSpent = noteByNullifier ? noteByNullifier.heightSpent !== 0n : false;
   const swapSpent = swapByNullifier ? swapByNullifier.heightClaimed !== 0n : false;
 
-  return new NullifierStatusResponse({ spent: noteSpent || swapSpent });
+  if (noteSpent || swapSpent) {
+    return new NullifierStatusResponse({ spent: true });
+  } else if (!req.awaitDetection) {
+    return new NullifierStatusResponse({ spent: false });
+  }
```
Valentine1898, line 35: There's an inconsistency here with the rust view service, I missed it when I implemented this method. 
The thing is that rust just ignores nullifiers from swap, so we should ignore them too.

hdevalence: That might be a bug in the rust view server?

Valentine1898: I'm not sure, but I think it would generally be useful to track swap nullifier to make sure the swap is claimed.

Valentine1898: But at the same time, the response of this method is `spent: true/false` but swap nullifier cannot be spent. 
So I'm leaning towards the fact that there is no bug in rust, and this method is only responsible for getting the status of note nullifier only

grod220: Indeed, the Rust code is not looking in the swaps table it seems:

https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/storage.rs#L510-L519

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
grod220: Oh! Good catch. Will fix.

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
grod220: Also, NullifierStatusResponse says "spent", but I suppose the consumer would know that in the case of swaps, that would mean "claimed"?

hdevalence: The SwapClaim "spends" the nullifier in the sense that it consumes it and it can't be claimed again. Since we moved to generic nullifiers (nullifiers are generated for arbitrary state commitments, notes, swaps, etc) "spending" a nullifier refers to consuming it, generally. I'd say this is a bug in the Rust view server and it probably causes problems.

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -1,33 +1,54 @@
...
 
-  // The 'heightSpent' and 'heightClaimed' fields will never be undefined,
-  // so we compare to 0n assuming it is impossible to spend nullifier in block 0
-  const noteSpent = noteByNullifier ? noteByNullifier.heightSpent !== 0n : false;
-  const swapSpent = swapByNullifier ? swapByNullifier.heightClaimed !== 0n : false;
+  // Wait until our DB encounters corresponding spent note
+  const subscription = indexedDb.subscribe('SPENDABLE_NOTES');
```
Valentine1898, line 46: We should also subscribe to `SWAPS` and handle updates the same way we do for notes

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -1,33 +1,54 @@
...
+  for await (const update of subscription) {
+    const note = SpendableNoteRecord.fromJson(update.value);
+    if (note.nullifier?.equals(req.nullifier) && (await noteOrSwapSpent(indexedDb, note.nullifier)))
```
Valentine1898, line 49: It looks like `(await noteOrSwapSpent(indexedDb, note.nullifier))` can simply be replaced with `note.heightSpent !== 0n` 
Once we've made sure that nullifier matches note, then we don't need to check swaps

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
grod220: You're totally right. Turns out this was trickier than I thought 😅 . Had to combine async generators. Added a bunch of tests.

Valentine1898: Oh, I thought it's possible to just create two subscriptions and handle updates separately, yes it's definitely more complicated.

Valentine1898 (approved):
Everything's perfect

PR #302, opened 2023-12-14
Don't immediately init transport connection, wait until use.

### LINKED ISSUE #292: No extension = no load
When the extension is not installed, the web app shows a blank screen. We should still load the UI, but show the connection button in the top right that directs user to install the extension.

Screenshot 2023-12-11 at 3 52 26 PM

turbocrime:
fixes #292 

Screenshot 2023-12-13 at 23 05 56

### DIFF
--- a/apps/extension/src/clients/index.ts
+++ b/apps/extension/src/clients/index.ts
@@ -8,7 +8,7 @@ import {
 } from '@penumbra-zone/transport';
 import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-runtime/client-connection-manager';
 
-export const getPenumbraPort = (serviceTypeName: string) => {
+export const getPenumbraPort = async (serviceTypeName: string) => {
   const { port1: port, port2: transferPort } = new MessageChannel();
   const initPort = ClientConnectionManager.init(ChannelClientLabel.Extension);
   initPort.postMessage(
@@ -19,10 +19,10 @@ export const getPenumbraPort = (serviceTypeName: string) => {
     } as InitChannelClientMessage,
     [transferPort],
   );
-  return port;
+  return Promise.resolve(port);
 };
 
 export const grpcClient = createPromiseClient(
   ViewProtocolService,
-  createChannelTransport(ViewProtocolService, getPenumbraPort(ViewProtocolService.typeName)),
+  createChannelTransport(ViewProtocolService, getPenumbraPort),
 );
--- a/apps/extension/src/clients/index.ts
+++ b/apps/extension/src/clients/index.ts
@@ -8,7 +8,7 @@ import {
 } from '@penumbra-zone/transport';
 import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-runtime/client-connection-manager';
 
-export const getPenumbraPort = async (serviceTypeName: string) => {
+export const getPenumbraPort = (serviceTypeName: string) => {
   const { port1: port, port2: transferPort } = new MessageChannel();
   const initPort = ClientConnectionManager.init(ChannelClientLabel.Extension);
   initPort.postMessage(
@@ -19,7 +19,7 @@ export const getPenumbraPort = async (serviceTypeName: string) => {
     } as InitChannelClientMessage,
     [transferPort],
   );
-  return Promise.resolve(port);
+  return port;
 };
 
 export const grpcClient = createPromiseClient(
--- a/apps/extension/src/clients/index.ts
+++ b/apps/extension/src/clients/index.ts
@@ -8,6 +8,8 @@ import {
 } from '@penumbra-zone/transport';
 import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-runtime/client-connection-manager';
 
+import { typeRegistry } from '@penumbra-zone/types/src/registry';
+
 export const getPenumbraPort = (serviceTypeName: string) => {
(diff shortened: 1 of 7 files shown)


turbocrime:
oops, had extra work in the branch

apps/webapp/src/components/header/notifications.tsx
```diff
@@ -1,16 +1,24 @@
...
-  const { data, error } = useStream(syncStream);
+  const [statusStream, setStatusStream] = useState<AsyncIterable<StatusStreamResponse>>();
+  const { data, error } = useStream(statusStream);
+
+  useEffect(() => {
+    void (async () => {
+      const { viewClient } = await import('../../clients/grpc');
+      setStatusStream(viewClient.statusStream({}));
+    })();
+  }, []);
```
turbocrime, line 21: this should probably be a custom fetcher? i wasn't sure how to get `useStream` to behave, and i'm pretty sure `useEffect` will fire the import often than it should.

turbocrime:
i don't see the appropriate layout in the repo

apps/webapp/src/components/header/notifications.tsx
turbocrime: yeag

Screenshot 2023-12-13 at 23 28 07

apps/extension/src/clients/index.ts
```diff
@@ -19,10 +19,10 @@ export const getPenumbraPort = (serviceTypeName: string) => {
     } as InitChannelClientMessage,
     [transferPort],
   );
-  return port;
+  return Promise.resolve(port);
```
grod220, line 22: I don't think we need to force this to be an async function. We can just pass it as synchronous.

turbocrime:
init properly delays

packages/transport/src/create.ts
```diff
@@ -68,53 +71,59 @@ export const createChannelTransport = (
...
-    } catch (error) {
-      const err = ConnectError.from(error);
-      callbackError ??= err;
-      throw err;
-    }
-  });
+  let port: MessagePort | undefined;
+
+  const connect = async () =>
+    getPort(s.typeName).then((port: MessagePort) => {
```
grod220, line 77: Can we use async/await here? Or actually, if it's a syncronous function, we can just call it.

grod220 (commented):
Nice

apps/webapp/src/clients/penumbra-port.ts
```diff
@@ -13,7 +13,7 @@ declare global {
   }
 }
 
-export const getPenumbraPort = (serviceTypeName: string) => {
+export const getPenumbraPort = async (serviceTypeName: string) => {
```
grod220, line 16: Not super related to this PR, but if this is the canonical way that web apps will be connecting to our extension, we should expose this in our own transport package so consumers don't have to think about this.

grod220:
Also, if this fixes the dynamic import problem, would be nice to have those fixed here too 🙏

packages/transport/src/create.ts
turbocrime: it's important to delay calling until first request, that's why it's wrapped in a function

turbocrime: previously, just calling here was what caused tests to fail static import, even when the function was async

calling on creation caused init handshake at creation. the tests were noticing the rejected promise that hadn't surfaced to the caller

so wrapping it with this function and then calling from request with the conditional ensured it wasn't firing on actual creation

apps/webapp/src/clients/penumbra-port.ts
turbocrime: would like to make package-level distinctions between

* penumbra-developed connectrpc transport
* penumbra-developed connectrpc service adapters
* penumbra package providing a client to penumbra services

and we should talk about the precise organization and locations of those boundaries

packages/transport/src/create.ts
grod220: My comment is less so about the wrapper `connect()` function. It's more so that we can do this:

```typescript
 const connect = async () => {
    const port = getPort(s.typeName);
    ...
```

apps/extension/src/clients/index.ts
```diff
@@ -24,5 +26,10 @@ export const getPenumbraPort = (serviceTypeName: string) => {
 
 export const grpcClient = createPromiseClient(
   ViewProtocolService,
-  createChannelTransport(ViewProtocolService, getPenumbraPort(ViewProtocolService.typeName)),
+  createChannelTransport({
+    defaultTimeoutMs: 10000,
+    serviceType: ViewProtocolService,
+    getPort: getPenumbraPort,
+    jsonOptions: { typeRegistry },
```
grod220, line 33: Let's make this optional

packages/transport/src/create.ts
```diff
@@ -56,65 +56,83 @@ const makeAnyServiceImpl: CreateAnyServiceImpl = <S extends ServiceType>(
...
-      else throw Error('Unknown transport item');
-    } catch (error) {
-      const err = ConnectError.from(error);
-      callbackError ??= err;
-      throw err;
-    }
-  });
+  let port: MessagePort | undefined;
+
+  const connect = () =>
```
grod220, line 83: Let's leave a code comment about how this prevents lazy loading of client

grod220 (approved):

Issue #146, opened 2023-10-25
Implement offscreen API method [web workers]


grod220:
We need a way to spawn web workers in the service worker. This requires using the offscreen api to open a separate window, sending it messages, and that window spawning web workers.

See https://github.com/penumbra-zone/web/tree/offscreen-web-worker for an example.

This will allow us to be able to use multi-threading (for building/proving) in the extension context.

TalDerei:
1. Get WASM build + proving key methods to work in current service worker. 
2. Get the web-worker to work in a service worker context (launching invisible windows). 
3. Get the build methods to work in the web-worker.

PR #255, opened 2023-11-23
connection management refactor

turbocrime:
# current status: working afaict

pulled out of concept testbed and started repo integration.

grafting in to the existing router was a bit more work than i anticipated,
because it is tightly integrated with the old transport.

but here's some visible activity now

## what

this is a draft, beginning to integrate a chunk of the work i've done to
refactor message routing. client connection management, and a simplified client
are the main features.

hopefully landing connection management alone will be simpler than trying to
replace routing at the same time.

### connection management

clients should be isolated from each other and conduct communications privately.

* a content-script connection manager listens for init events and handles
  page-side channel negotiation.

* a background connection manager authenticates clients and integrates with the
  router.

background origin auth is currently a stub. a temporary adapter truncates my
routing work, and simply feeds into the existing service routers.

### channel transport

new client conducts all communication over channels. init is pluggable, so
different contexts such as intra-extension transport may be simplified.

### awkward

existing code to convey types was overcomplicated. much of this still remains in
the existing router, as it was tightly integrated with the old message
transport.

## message flow

```mermaid
sequenceDiagram
    box Page
        participant PromiseClient
        participant Transport as ChannelTransport
    end

    box Injected
        participant ContentScript as ConnectionManager
    end

    box Extension Background
        participant Background as ConnectionManager
        participant Service as ConnectRouter
    end

    Note left of PromiseClient: Init
    PromiseClient -> Transport: construct
    Transport -->>+ ContentScript: InitChannelTransport MessagePort
    ContentScript ->>+ Background: chrome.runtime.connect Port

    Note right of Background: OriginRegistry
    Background -> Service: router entry

    rect seashell
    Note left of PromiseClient: unary
    PromiseClient -) Transport: unary rpc
    Transport ->> ContentScript: MessagePort TransportMessage call
    ContentScript -->> Background: chrome.runtime.Port TransportMessage call
    Background -)+ Service: request routed, awaited
    Service --) Service: Proxy or Local
    Service ->- Background: response resolved
    Background -->> ContentScript: chrome.runtimePort TransportMessage response
    ContentScript ->> Transport: MessagePort TransportMessage response
    Transport -> PromiseClient: unary resolved
    end

    rect seashell
    Note left of PromiseClient: serverstream
    PromiseClient -) Transport: serverstream rpc
    Transport ->> ContentScript: MessagePort TransportMessage call
    ContentScript -->> Background: chrome.runtime.Port TransportMessage call
    Background -)+ Service: request routed, awaited
[...]

Issue #113, opened 2023-10-17
Split up AuthorizeAndBuild


grod220:
Currently, the grpc method for building requires authorization beforehand: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.AuthorizeAndBuild.

This is an issue as building can be done at the same time the user is viewing/signing their transaction. This will reduce the _percieved_ building time. The rust code already supports this (TransactionPlan::build returns the UnauthTransaction). 

WitnessAndBuild may need to be altered as well: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.WitnessAndBuild.

Valentine1898:
Additionally, it makes sense to add `UnauthTransaction` to proto 
Since there is supposed to be a wasm function that should return `UnauthTransaction`

TalDerei:
This is a consideration further down the line, but it may be prudent to start thinking about potential DoS attack vectors? For instance, can a user can open many connections that preemptively start building the transaction proofs -- without ever completing the signing phase of the transaction. Regardless, this optimistic proving approach seems promising from a UX perspective.

grod220:
> potential DoS attack vectors

This work is always client-side though. So wouldn't that be the user DoS'ing themselves?

grod220:
Superseded by [#324](https://github.com/penumbra-zone/web/issues/324)

PR #303, opened 2023-12-14
Offscreen API

TalDerei:
References #146 

This PR supports building transactions in the web in parallel using web-workers! Additionally, the `serial` build method has been deprecated. 

The output console logs can be monitored by navigating to `chrome://extensions`, clicking `details`, and then `inspect view`.


TalDerei:
Tackling some internal complications with spawning workers in a service-worker context, but should be ready for review soon.

TalDerei:
@grod220 building the actions in parallel in the offscreen document and passing the prebuilt vector of actions back to the extension is working, but calls to the `build_parallel` method is failing. It's throwing "`Error: invalid type: byte array, expected a base64 string`". Think it's a protobuf related error?

grod220:
Ah, you've run into our classic wasm serialization issue. We've each run into this independently 😅 . Think we should document this somewhere. 

When crossing the javascript->wasm boundary, you can pass javascript objects. However, it's not a guarantee that serde can deserialize it into structs. For that to be successful, it has to be in json format ([protobuf rules though](https://protobuf.dev/programming-guides/proto3/#json)). Bufbuild provides the helper `.toJson()` on our proto-classes to get in this specific format. So try doing this for each argument like so:

```typescript
const transaction = build_parallel(
    batchActions.map(action => action.toJson()), 
    req.transactionPlan.toJson(), 
    witnessData.toJson(), 
    req.authorizationData.toJson()
  )
```

TalDerei:
Benchmarks on my machine (`2020 M1 Mac, 8-core CPU and GPU`) for `2 spend` and `2 output` actions: 
**- Parallel transaction: 12s build + 6s broadcast = ~18s total** 
**- Serial transaction: 30s build + 6s broadcast = ~36s total**

The broadcasting operation `viewClient.broadcastTransaction` is currently taking around `1/3` of the transaction time for constructing a transaction in parallel. 

In terms of outstanding work before this PR can be merged, we should consider load-balancing the workload distributed to the web-workers when the number of `ActionPlan` exceeds the number of logical CPU cores on the device. 

Additionally, @turbocrime we should also implement a fallback mechanism for serial builds in cases where web workers are not supported. One idea is to introduce a global variable like `webWorkersSupported` and check against that. Where should this ideally be included considering the changes in #255?

hdevalence:
The "broadcast" step in the web code is currently a hardcoded 6s timeout. Two points:

1. The hardcoded timeout is a stand-in for the logic we have in the Rust view server, where we await detection of the transaction during sync, closing the loop on the broadcast (not just waiting for "did the server say it was accepted" but "did it end up in a block we scanned"). This means the time will be between ~0 and ~5s, depending on when the transaction was broadcast relative to the block interval.
2. For this reason, I think it makes sense to expose this to users as a different phase ("Building Transaction..." vs "Broadcasting Transaction...") in the UI.

@grod220 is currently working on replacing the hardcoded timeout with detection logic.

hdevalence:
**Going from 30s build time to 12s build time is amazing!**

If we can also cut down some of the wasm overhead, we're stacking improvements that get us pretty close to the ideal scenario of being able to build the transaction in the time it takes the user to review the transaction in the approval dialog.

TalDerei:
> The "broadcast" step in the web code is currently a hardcoded 6s timeout. Two points:
> 
> 1. The hardcoded timeout is a stand-in for the logic we have in the Rust view server, where we await detection of the transaction during sync, closing the loop on the broadcast (not just waiting for "did the server say it was accepted" but "did it end up in a block we scanned"). This means the time will be between ~0 and ~5s, depending on when the transaction was broadcast relative to the block interval.
> 2. For this reason, I think it makes sense to expose this to users as a different phase ("Building Transaction..." vs "Broadcasting Transaction...") in the UI.
> 
> @grod220 is currently working on replacing the hardcoded timeout with detection logic.

ah I was wondering why it was hard-coded specifically to 6s!

TalDerei:
> If we can also cut down some of the wasm overhead, we're stacking improvements that get us pretty close to the ideal scenario of being able to build the transaction in the time it takes the user to review the transaction in the approval dialog.

Agreed #113 shouldn't be blocked for much longer.

hdevalence:
> In terms of outstanding work before this PR can be merged, @turbocrime we should implement a fallback mechanism for serial builds in cases where web workers are not supported. One idea is to introduce a global variable like webWorkersSupported and check against that. 

In what case would web workers not be supported?

TalDerei:
> In what case would web workers not be supported?

 It's generally a good idea to check for web worker support (https://www.w3schools.com/html/html5_webworkers.asp). I referenced this chart (https://caniuse.com/webworkers), yet it suggests that 98% of users wouldn't have an issue.

TalDerei:
@hdevalence If we don't implement the check, there's no reason to keep the `serial` build method in the wasm crate or web code, which means we can further reduce the WASM binary size from `8.5 MB` to `6.2 MB`.

hdevalence:
From the caniuse.com page, we can unconditionally use web workers, since we're only targeting Chrome

turbocrime:
there is some work in PR already fixing broadcast timeout. agree don't need feature detection. will take a close look for review soon

TalDerei:
The build is broken after commit `a87f5c51a2e63463b8136e6dcd0cebf8fc4ef9a7`. I was trying to satisfy the strict linter and broke something, but planning on fixing soon.

turbocrime:
> We can further consider load-balancing the workload distributed to the web-workers using a queue when the number of `ActionPlan` exceeds the number of logical CPU cores on the device. Alternatively, spawning more web-workers than the available cores leads to excessive context switching and resource contention, which degrades the performance.

we could potentially use [chrome.system.cpu](https://developer.chrome.com/docs/extensions/reference/api/system/cpu) but that requires additional permission

TalDerei:
> we could potentially use [chrome.system.cpu](https://developer.chrome.com/docs/extensions/reference/api/system/cpu) but that requires additional permission

I usually use [navigator.hardwareConcurrency](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency)

turbocrime:
i think my conception of the 'routes' directory is that it is specifically referring to react/ui routes, and this is kind of a standalone thing. the 'offscreen' directory could move up a level, and we might even want to have webpack bundle the entry js directly into offscreen.html

apps/extension/src/routes/offscreen/window-management.ts
```diff
@@ -0,0 +1,52 @@
...
+async function hasDocument() {
+  // Check all windows controlled by the service worker if one of them is the offscreen document
+  const matchedClients = await sw.clients.matchAll();
+  for (const client of matchedClients) {
+    if (client.url.endsWith(OFFSCREEN_DOCUMENT_PATH)) {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
TalDerei: there isn't, I removed `closeOffscreenDocument`

apps/extension/src/routes/offscreen/window-management.ts
```diff
@@ -0,0 +1,52 @@
...
+  req: WitnessAndBuildRequest,
+  witness: WitnessData,
+  fullViewingKey: string,
+  key_type: string[],
+): Promise<Action[]> => {
+  if (!(await hasDocument())) {
+    await chrome.offscreen.createDocument({
+      url: OFFSCREEN_DOCUMENT_PATH,
+      reasons: [chrome.offscreen.Reason.WORKERS],
+      justification: 'Spawn web workers',
```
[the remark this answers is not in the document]
TalDerei: changed justification to `"Spawn web workers from offscreen document"`

apps/extension/src/routes/web-worker/router.ts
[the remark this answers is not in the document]
TalDerei: moved logic to a single `web-worker.ts` file under `offscreen` directory

packages/types/src/internal-msg/shared.ts
```diff
@@ -2,6 +2,7 @@ export interface InternalMessage<Type extends string, Req, Res> {
   type: Type;
   request: Req;
   response: Res;
+  target: string;
```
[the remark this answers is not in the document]
TalDerei: good catch, forgot to remove this. We only need `type` and can check against that.

apps/extension/src/routes/offscreen/window-management.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+  }
+
+  const result = await offscreenClient.buildAction(req, witness, fullViewingKey, key_type);
```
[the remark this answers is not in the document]
TalDerei: wouldn't `if ('error' in result) throw result.error;` already suffice?

packages/router/src/grpc/view-protocol-server/witness-build.ts
```diff
@@ -4,8 +4,9 @@ import {
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 import { ViewReqMessage } from './router';
 import { ServicesInterface } from '@penumbra-zone/types';
-import { build, witness } from '@penumbra-zone/wasm-ts';
+import { build_parallel, witness } from '@penumbra-zone/wasm-ts';
 import { localExtStorage } from '@penumbra-zone/storage';
+import { handleOffscreenAPI } from '../../../../../apps/extension/src/routes/offscreen/window-management';
```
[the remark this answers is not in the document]
TalDerei: what's the best way to simplify the path here? I can't add `chrome-extension` as a dependency to the `packages` directory.

apps/extension/src/routes/offscreen/build.ts
```diff
@@ -0,0 +1,80 @@
...
+  // Destructure the data object to get individual fields
+  const { transactionPlan, actionPlan, witness, fullViewingKey, key_type } = jsonReq;
+
+  // Array to store promises for each worker
+  const workerPromises: Promise<Action>[] = [];
+
+  // Spawn web workers
+  for (let i = 0; i < actionPlan.length; i++) {
+    workerPromises.push(
+      spawnWorker(transactionPlan, actionPlan[i]!, witness, fullViewingKey, key_type[i]!),
```
grod220, line 65: If you use a `for of...` loop, you don't need to have this array indexing thing.

grod220: If `actionPlan[i]` is paired with `key_type[i]`, it would be nice to have a data structure that combines them both so they are always together and the array length/order doesn't matter.

apps/extension/webpack/webpack.common.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     'service-worker': path.join(srcDir, 'service-worker.ts'),
     'injected-connection-manager': path.join(srcDir, 'injected-connection-manager.ts'),
     'injected-penumbra-global': path.join(srcDir, 'injected-penumbra-global.ts'),
+    offscreen: path.join(srcDir, 'routes/offscreen/router.ts'),
```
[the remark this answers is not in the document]
TalDerei: noticed this as well. I'll play around with the config

apps/extension/src/routes/offscreen/window-management.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  Action,
+  WitnessData,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
+
+const sw = self as unknown as ServiceWorkerGlobalScope;
+
+const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
+
+export const handleOffscreenAPI = async (
```
grod220, line 12: It feels a bit weird that the those wanting to do the build action, need to call a offscreen-management function. I think it would be better if all this wrapped `sendOffscreenMessage`:
https://github.com/penumbra-zone/web/pull/303/files#diff-a3ef69bb05e35d0d370d757db6a9d3b23d1f9e9c57e9972407c135614af9ffebR62-R70

And then, the caller just needs to be concerned with what function they want to call, aka buildAction().

apps/webapp/src/state/send.ts
```diff
@@ -128,12 +128,20 @@ const planWitnessBuildBroadcast = async ({
   const { data: authorizationData } = await custodyClient.authorize({ plan });
   if (!authorizationData) throw new Error('no authorization data in response');
 
+  // Start timer
+  const startTime = performance.now(); // Record start time
+
```
grod220, line 133: Can remove these timers

packages/wasm/src/build.ts
```diff
@@ -38,3 +44,17 @@ export const build = async (
   );
   return Transaction.fromJsonString(JSON.stringify(result));
 };
+
+export const build_parallel = async (
```
grod220, line 48: camelCase plz

apps/extension/src/routes/offscreen/build.ts
```diff
@@ -0,0 +1,80 @@
...
+  transactionPlan: TransactionPlan,
+  actionPlan: ActionPlan,
+  witness: WitnessData,
+  fullViewingKey: string,
+  key_type: string,
+): Promise<Action> => {
+  return new Promise((resolve, reject) => {
+    const worker = new Worker(new URL('../web-worker/router.ts', import.meta.url));
+
+    // Set up event listener to recieve messages from the web worker
```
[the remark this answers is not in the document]
TalDerei: I considered using comlink, but it wasn't working properly! Let's stick to message-passing for simplicity. 

In terms of memory leaks, I've considered this and made some local changes that terminate the worker (`worker.terminate()`) and event listener (`worker.removeEventListener('message', handleMessage)`) to clean up the resources associated with the workers lifecycle.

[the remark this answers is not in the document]
TalDerei: we actually already terminate the worker with `close()` inside `web-worker.ts `, but you're right, I still need to handle the event listener to prevent potential memory leaks.

apps/extension/src/routes/offscreen/build.ts
```diff
@@ -0,0 +1,80 @@
...
+    worker.addEventListener(
+      'message',
+      function (e) {
+        resolve(e.data as Action);
+      },
+      false,
```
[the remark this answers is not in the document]
TalDerei: yes, removed

apps/webapp/src/state/send.ts
TalDerei: accidentally committed! I already removed it locally

packages/router/src/grpc/view-protocol-server/witness-build.ts
```diff
@@ -24,10 +25,18 @@ export const handleWitnessBuildReq = async (
   const witnessData = witness(req.transactionPlan, sct);
 
   const wallets = await localExtStorage.get('wallets');
+
   const { fullViewingKey } = wallets[0]!;
 
-  const transaction = await build(
-    fullViewingKey,
+  const action_types = [];
```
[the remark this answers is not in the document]
TalDerei: `action.action.case` is a string, so we can simply do `const action_types: string[] = [];` and `handleOffScreenAPI` expects a [string[]](https://github.com/penumbra-zone/web/blob/offscreen-api/apps/extension/src/routes/offscreen/window-management.ts#L16). This is necessary for the worker to determine what action to build.

apps/extension/src/routes/offscreen/build.ts
```diff
@@ -0,0 +1,80 @@
...
+): Promise<Action> => {
+  return new Promise((resolve, reject) => {
+    const worker = new Worker(new URL('../web-worker/router.ts', import.meta.url));
+
+    // Set up event listener to recieve messages from the web worker
+    worker.addEventListener(
+      'message',
+      function (e) {
+        resolve(e.data as Action);
+      },
```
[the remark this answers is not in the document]
TalDerei: changed

apps/extension/src/routes/offscreen/types.ts
```diff
@@ -0,0 +1,70 @@
...
+export type OffscreenResponse = InternalResponse<ActionBuildMessage>;
+
+export type ActionBuildMessage = InternalMessage<
+  'ACTION_AND_BUILD',
+  {
+    transactionPlan: TransactionPlan;
+    actionPlan: ActionPlan[];
+    witness: WitnessData;
+    fullViewingKey: string;
+    key_type: string[];
```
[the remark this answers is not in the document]
TalDerei: changed

packages/wasm/src/build.ts
TalDerei: changed

apps/extension/src/routes/web-worker/router.ts
```diff
@@ -0,0 +1,78 @@
...
+async function execute_worker(
+  transactionPlan: JsonValue,
+  actionPlan: JsonValue,
+  witness: JsonValue,
+  fullViewingKey: string,
+  key_type: string,
+) {
+  console.log('web worker running...');
+
+  // Dynamically load wasm module
```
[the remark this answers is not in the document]
TalDerei: there isn't a performance benefit, instead the wasm module needs to be dynamically loaded directly from inside the worker. Doing `import { load_proving_key, build_action } from '@penumbra-zone/wasm-bundler'` for instance at the top of the file won't work. WASM loading time is a critical factor in the performance of the web-worker, which is why it was critical to reduce the binary size.

apps/extension/src/routes/web-worker/router.ts
```diff
@@ -0,0 +1,78 @@
...
+  const penumbraWasmModule = await import('@penumbra-zone/wasm-bundler');
```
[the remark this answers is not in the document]
TalDerei: good call! I'll add `build_action` to the `/wasm` package and `WasmBuildActionSchema`schema to `/types`

packages/router/src/grpc/view-protocol-server/witness-build.ts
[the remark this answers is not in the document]
turbocrime: ah, youre right, not just a path thing.

current method is extend stdclient. see the popup approval method for an example of how to await another part of the extension. right now it is admittedly complex, and is probably changing

ill see if i can give you a solution, but i wont work much today/tomorrow, internet at my location is not very available, its fine to keep for now

apps/extension/src/routes/offscreen/window-management.ts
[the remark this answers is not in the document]
turbocrime: Something like "Manages Penumbra transaction WASM workers"

apps/extension/src/routes/offscreen/build.ts
TalDerei: changed to `for (const [i, action] of actionPlan.entries())`

apps/extension/src/routes/offscreen/window-management.ts
TalDerei: that's much cleaner, modified!

apps/extension/src/routes/web-worker/router.ts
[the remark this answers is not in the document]
TalDerei: can you help fix the action [schema](https://github.com/penumbra-zone/web/pull/303/commits/5e933cb44f843b0ecc21ffb23393e62a5438f134#diff-72930539f567ea480df7b7b112392ca41a00b0a9b00b9918a0e7ad25fcdf1eeeR53)?

apps/extension/src/routes/offscreen/window-management.ts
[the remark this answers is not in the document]
TalDerei: changed

apps/extension/src/routes/web-worker/router.ts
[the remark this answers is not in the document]
grod220: Would do something like this:

```typescript
export const WasmActionSchema = z.object({
  action: z.object({
    value: z.unknown(),
    case: z.string(),
  }),
});

export const buildActionParallel = (
  txPlan: TransactionPlan,
  actionPlan: ActionPlan,
  witnessData: WitnessData,
  fullViewingKey: string,
): Action => {
  const result = validateSchema(
    WasmActionSchema,
    build_action(txPlan, actionPlan, fullViewingKey, witnessData),
  );
  return Action.fromJson(result);
};

```

apps/extension/webpack/webpack.common.js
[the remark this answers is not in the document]
TalDerei: I think we can make this change separately. I'll create an issue for it.

apps/extension/src/routes/web-worker/router.ts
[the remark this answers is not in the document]
TalDerei: hmm this doesn't seem to work. It's throwing
```
Argument of type '{ action: { case: string; value?: unknown; }; }' is not assignable to parameter of type 'JsonValue'.
  Type '{ action: { case: string; value?: unknown; }; }' is not assignable to type 'JsonObject'.
    Property 'action' is incompatible with index signature.
      Type '{ case: string; value?: unknown; }' is not assignable to type 'JsonValue'.
        Type '{ case: string; value?: unknown; }' is not assignable to type 'JsonObject'.
          Property 'value' is incompatible with index signature.
            Type 'unknown' is not assignable to type 'JsonValue'.
```

and changing the return type to `Action.fromJsonString(JSON.stringify(result))` doesn't work. It seems like it's failing at `validateSchema` and the actions aren't being built.

TalDerei:
Average `build + broadcast` time is **~15s** on my machine.

turbocrime:
hey, i think have quite a lot of feedback on the event and message flow, so i'm just going to add some commits to this pr. we should pair when possible

TalDerei:
> hey, i think have quite a lot of feedback on the event and message flow, so i'm just going to add some commits to this pr. we should pair when possible

I paired with Gabe yesterday, and am currently refactoring the underlying types as `JsonValue`. Let's pair later about the event and message flow.

apps/extension/src/routes/web-worker/router.ts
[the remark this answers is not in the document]
TalDerei:
```
const result = validateSchema(
    WasmActionSchema,
    build_parallel(txPlan.toJson(), txPlan.actions[actionId]?.toJson(), fullViewingKey, witnessData.toJson()),
  );
```
  
  where 
```
  export const WasmActionSchema = z.object({
  action: z.object({
    value: z.unknown(),
    case: z.string(),
  }),
});
```

still doesn't seem to work.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,47 @@
...
+        fullViewingKey,
+        length: arg.transactionPlan!.actions.length,
+      },
+    }),
+};
+
+export const sendOffscreenMessage = async <T extends OffscreenMessage>(
+  req: InternalRequest<T>,
+): Promise<InternalResponse<ActionBuildMessage>> => {
+  await chrome.offscreen.createDocument({
```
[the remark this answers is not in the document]
turbocrime: it also occurs to me that crosstalk between two tabs building transactions could occur. will have some more notes on this in a moment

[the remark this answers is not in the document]
turbocrime: well, maybe that's wrong, it is an awaited response

apps/extension/src/web-worker.ts
```diff
@@ -1,48 +1,51 @@
 import {
   Action,
-  ActionPlan,
   TransactionPlan,
   WitnessData,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
-import { WebWorkerMessagePayload } from './types';
+import { JsonValue } from '@bufbuild/protobuf/dist/esm';
```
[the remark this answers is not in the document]
TalDerei: changed to `import { JsonValue } from '@bufbuild/protobuf'`

apps/extension/tsconfig.json
```diff
@@ -1,6 +1,11 @@
 {
   "extends": "tsconfig/extension.json",
-  "include": ["src/**/*.ts", "src/**/*.tsx", "vite.config.ts"],
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "vite.config.ts",
+    "../../packages/router/src/grpc/offscreen-client.ts"
+  ],
```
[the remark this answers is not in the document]
TalDerei: reverted

packages/wasm/src/build.ts
```diff
@@ -41,16 +41,21 @@ export const buildParallel = (
...
 
 export const buildActionParallel = (
   txPlan: TransactionPlan,
-  actionPlan: ActionPlan,
   witnessData: WitnessData,
   fullViewingKey: string,
+  actionId: number,
 ): Action => {
-  const result = build_action(txPlan, actionPlan, fullViewingKey, witnessData) as Action;
+  const result = build_action(
```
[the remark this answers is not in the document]
TalDerei: changed

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,47 @@
...
+    url: OFFSCREEN_DOCUMENT_PATH,
+    reasons: [chrome.offscreen.Reason.WORKERS],
+    justification: 'spawn web workers from offscreen document',
```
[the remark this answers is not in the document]
TalDerei: it didn't get dropped, we just didn't change it. Changed to `"Manages Penumbra transaction WASM workers"`

apps/extension/src/offscreen.ts
```diff
@@ -0,0 +1,80 @@
...
+  responder: (r: JsonValue) => void,
+) => {
+  // Destructure the data object to get individual fields
+  const { transactionPlan, witness, fullViewingKey, length } = jsonReq.request;
+
+  // Spawn web workers
+  const workerPromises: Promise<JsonValue>[] = [];
+  for (let i = 0; i < length; i++) {
+    workerPromises.push(spawnWorker(transactionPlan, witness, fullViewingKey, i));
+  }
```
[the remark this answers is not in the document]
TalDerei: But `transactionPlan` is a `JsonValue`, so we'd need to convert out of that type to perform the mapping you describe.

packages/wasm/src/build.ts
```diff
@@ -24,17 +30,32 @@ export const witness = (txPlan: TransactionPlan, sct: StateCommitmentTree): Witn
...
+  witnessData: WitnessData,
+  fullViewingKey: string,
+  actionId: number,
+): Action => {
+  const result = build_action(
+    txPlan.toJson(),
+    txPlan.actions[actionId]?.toJson(),
+    fullViewingKey,
+    witnessData.toJson(),
+  ) as Action;
```
turbocrime, line 58: this cast seems incorrect. we are calling the wasm build_action directly and i don't believe it's returning a real Action, but some json that hopefully represents an action.

turbocrime: https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/build.rs#L43 right?

apps/extension/src/web-worker.ts
```diff
@@ -0,0 +1,82 @@
...
+    actionId,
+  } = data;
+
+  // Deserialize payload
+  const transactionPlan = TransactionPlan.fromJson(transactionPlanJson);
+  const witness = WitnessData.fromJson(witnessJson);
+
+  void executeWorker(transactionPlan, witness, fullViewingKey, actionId).then(action => {
+    self.postMessage(action);
+  });
```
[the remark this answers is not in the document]
TalDerei: I thought about this. @grod220 mentioned we should be using `fromJson`. Can we keep it as a `JsonValue` for the entire build process?

apps/extension/src/web-worker.ts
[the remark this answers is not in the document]
TalDerei: changed

packages/wasm/src/build.ts
TalDerei: confirmed the output from calling `buildAction` is returning the JsonValue of `Action`.

apps/extension/webpack/webpack.common.js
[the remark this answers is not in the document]
TalDerei: can you open an issue?

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
TalDerei: what was wrong with the former detection [method](https://github.com/penumbra-zone/web/pull/318/commits/c29f8f6783bca29645dcff5d87ebc13542eb3a93#diff-da83cef549a23dfcccff524ac5206ac90d96d058a42a7183bd22a59d9451aefdR76)?

I was referencing this [example](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/functional-samples/cookbook.offscreen-dom/background.js).

packages/types/src/transaction/schema.ts
```diff
@@ -48,3 +48,10 @@ export const WasmBuildSchema = z.object({
     transactionParameters: z.unknown(),
   }),
 });
+
+export const WasmActionSchema = z.object({
+  action: z.object({
+    value: z.unknown(),
+    case: z.string(),
+  }),
+});
```
turbocrime, line 57: this is how actions actually come out of wasm, a 'record' where the keyname is of ActionCase. i don't know how to get zod to understand the union type

```suggestion
//type ActionCase = string & Action['action']['case'];
export const WasmActionSchema = z.record(z.unknown());
```

turbocrime: ping @grod220

apps/extension/webpack/webpack.common.js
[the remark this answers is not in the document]
turbocrime: #319

apps/extension/src/web-worker.ts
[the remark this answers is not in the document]
turbocrime: currently the behavior of the wasm-ts wrapper is to validate zod schema and deserialize with fromJson, and this works.

but, in the wasm-task worker, we're immediately re-serializing and posting it to the parent offscreen, which also immediately needs it serialized to post it back to the caller of the offscreen work.

i'm still of the opinion that we're doing excessive ser/de, but let's just go for consistency, and do it like the other wrapper methods. this can be a separate issue.

packages/wasm/src/build.ts
TalDerei: This is tricky. Some actions have the `authSig` field (spend action), while others do not (output action). `authSig` is currently set as a placeholder value I believe

Screenshot 2023-12-22 at 1 53 40 PM

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
turbocrime: interesting. the examples seem inconsistent. the doc suggests this method, and one of the examples does use it

https://github.com/GoogleChrome/chrome-extensions-samples/blob/578c684d3e27325d513f924bf079705c5f7a767d/functional-samples/sample.tabcapture-recorder/service-worker.js#L16-L21

but, in mentioning the `getContexts` api, the doc links to your example, and i do see one other example also using the method you identified.

i'm inclined to prefer the `chrome.runtime.getContexts` api, instead of using the browser's `Client` api, as the documentation suggests. also, using the chrome api would avoid declaring a global or casting the global object, which feels incorrect.

packages/wasm/src/utils.ts
```diff
@@ -10,10 +9,7 @@ export const loadLocalBinary = async (filename: string) => {
...
 
-export const loadProvingKeys = async () => {
-  const promises = provingKeys.map(async ({ file, keyType }) => {
-    const response = await loadLocalBinary(file);
-    wasmLoadProvingKey(response, keyType);
-  });
-  await Promise.all(promises);
+export const loadProvingKey = async (file: string, keyType: string) => {
+  const response = await loadLocalBinary(file);
+  wasmLoadProvingKey(response, keyType);
```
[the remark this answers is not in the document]
TalDerei: We already have a [map](https://github.com/penumbra-zone/web/blob/main/packages/types/src/proving-keys.ts) under the `type` package, I'll modify it to use it.

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
TalDerei: It properly detected and threw an error when multiple offscreen documents were attempted to open.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  const result = (await chrome.runtime.sendMessage({
+    ...req,
+  })) as InternalResponse<ActionBuildMessage>;
+  if ('error' in result) throw new Error('failed to build action');
+
+  // Close offscreen document
+  await chrome.offscreen.closeDocument();
```
turbocrime, line 44: we should probably not be killing the offscreen when it's done - multiple callers could be talking to the same `offscreen.html`. there won't be crosstalk in the responses, because direct responses are awaited. but it *is* possible that one handler may kill an offscreen doc that another handler is currently pending.

leaving the doc alive should be ok for now. it maintains one message listener, and kills workers when it's done with them, so it should be very lightweight. let's remove this line and i'll make an issue on it

```suggestion
```

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
TalDerei: `chrome.runtime.getContexts` was only added in Chrome 116 https://developer.chrome.com/docs/extensions/reference/api/offscreen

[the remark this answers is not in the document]
TalDerei: > i'm inclined to prefer the `chrome.runtime.getContexts` api,

chrome.runtime.getContexts was only added in Chrome 116 https://developer.chrome.com/docs/extensions/reference/api/offscreen

packages/router/src/grpc/offscreen-client.ts
TalDerei: Is it good practice to generally keep the offscreen API document open? I'm not sure

TalDerei: Actually, seems like we need to close the document, no? Initiating a new transaction throws `"Only a single offscreen document may be created"`

turbocrime: it isn't! but offscreen may simultaneously perform work for multiple callers. it's unclear how to identify that we no longer need the offscreen document.

one solution would be to make offscreen responsible for its own termination, but this seems incorrect.

maybe attach a counter to offscreenClient that is incremented/decremented when handlers launch and resolve, and kill when it's zero. there is still a race condition there, but it will prevent killing other callers' work in most cases.

apps/extension/src/offscreen.ts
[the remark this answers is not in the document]
turbocrime:
```ts
  const { transactionPlan, witness, fullViewingKey } = request;

  const hasActionPlanJsonArray = (x: JsonValue): x is JsonObject & { actions: JsonValue[] } =>
    typeof x == 'object' && x != null && 'actions' in x && Array.isArray(x['actions']);

  if (!hasActionPlanJsonArray(transactionPlan)) throw new Error('No action plans');

  const actionsWork = transactionPlan.actions.map((_, i) =>
    spawnWorker(transactionPlan, witness, fullViewingKey, i),
  );
  ```
  
  like that or preferably `.map((action) =>` without index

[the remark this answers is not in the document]
Comment (turbocrime) [reply apps/extension/src/offscreen.ts:35]: ideally make it part of the type guard we use to check in our listener, before we call buildActionHandler, then we have the type before this fn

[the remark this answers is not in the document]
Comment (turbocrime) [reply apps/extension/src/offscreen.ts:35]: this is nitpicking, sorry, resolve if you'd like

File: packages/router/src/grpc/offscreen-client.ts (hunk 18 above)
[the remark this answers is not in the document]
Comment (turbocrime) [reply packages/router/src/grpc/offscreen-client.ts:31]: i think it is okay to target latest chrome. iirc we are already using other features from 118 for things like popup lifecycle

File: packages/wasm/src/build.ts (hunk 24 above)
Comment (turbocrime) [reply packages/wasm/src/build.ts:58]: ok, looked this up for my own clarity: the transaction as returned is unauthorized, and so `authSig` is *expected* to be absent at this stage of the plan/build/auth flow, and this is not a failure.

https://github.com/penumbra-zone/penumbra/blob/1eb74b7c78e114759a270a358e36a7b11cf117eb/crates/wasm/src/build.rs#L41C39-L41C39

but i would like to issue the wasm crate to not define `authSig`, rather than define junk.

Comment (TalDerei) [reply packages/wasm/src/build.ts:58]: > ok, looked this up for my own clarity: the transaction as returned is unauthorized, and so `authSig` is _expected_ to be absent at this stage of the plan/build/auth flow, and this is not a failure.

If you look at Henry's comment in the [split-build](https://github.com/penumbra-zone/penumbra/pull/3289) PR, you'll notice that defining dummy `authSig` was an intentional design choice, although it may be unappealing.

File: apps/extension/src/routes/web-worker/router.ts (hunk 17 above)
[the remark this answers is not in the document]
Comment (turbocrime) [reply apps/extension/src/routes/web-worker/router.ts]: missed this thread but i had a comment here

https://github.com/penumbra-zone/web/pull/303#discussion_r1435354349

File: apps/extension/src/offscreen.ts [hunk 29]
```diff
@@ -0,0 +1,80 @@
...
+
+  // Wait for promises to resolve and construct response format
+  void Promise.all(workerPromises).then(batchActions => {
+    responder({
+      type: 'BUILD_ACTION',
+      data: batchActions,
+    });
```
grod220, line 42: Can you type this with: 

```typescript
    responder({
      type: 'BUILD_ACTION',
      data: batchActions,
    } satisfies OffscreenResponse);
```

When you do, you may notice that there seems to be a typing issue?

apps/extension/src/offscreen.ts
```diff
@@ -0,0 +1,80 @@
...
+    worker.addEventListener('message', onWorkerMessage);
+    worker.addEventListener('error', onWorkerError);
+
+    // Send data to web worker
+    worker.postMessage({
+      transactionPlan,
+      witness,
+      fullViewingKey,
+      actionId,
+    });
```
grod220, line 78: Please add `satisfies` typing to this. Is it supposed to be `OffscreenRequestPayload`?

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+  WitnessData,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
+import { JsonValue } from '@bufbuild/protobuf';
+
+interface WasmTaskInput {
+  transactionPlan: JsonValue;
+  witness: JsonValue;
+  fullViewingKey: string;
+  actionId: number;
+}
```
grod220, line 13: This should be a shared type between the requestor and handler. So this probably needs exported somewhere?

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+  actionId: number,
+): Promise<Action> {
+  console.log('web worker running...');
+
+  // Dynamically load wasm module
+  const penumbraWasmModule = await import('@penumbra-zone/wasm-ts');
+
+  // Conditionally read proving keys from disk and load keys into WASM binary
+  const actionKey = transactionPlan.actions[actionId]?.action.case;
+  await penumbraWasmModule.loadProvingKey(actionKey!);
```
grod220, line 47: Instead of `!` let's throw an error. We'll get better messaging if something goes wrong:

```typescript
  const actionKey = transactionPlan.actions[actionId]?.action.case;
  if (!actionKey) throw new Error('No action key provided');

  await penumbraWasmModule.loadProvingKey(actionKey);
```

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  // Build action according to specification in `TransactionPlan`
+  const action = penumbraWasmModule.buildActionParallel(
+    transactionPlan,
+    witness,
+    fullViewingKey,
+    actionId,
+  );
+
+  return action;
```
grod220, line 57: can inline the return:

```typescript
return penumbraWasmModule.buildActionParallel(transactionPlan, witness, fullViewingKey, actionId);
```

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+
+self.addEventListener('message', workerListener);
+
+async function executeWorker(
+  transactionPlan: TransactionPlan,
+  witness: WitnessData,
+  fullViewingKey: string,
+  actionId: number,
+): Promise<Action> {
+  console.log('web worker running...');
```
grod220, line 40: Let's get rid of the console log

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+    actionId,
+  } = data;
+
+  // Deserialize payload
+  const transactionPlan = TransactionPlan.fromJson(transactionPlanJson);
+  const witness = WitnessData.fromJson(witnessJson);
+
+  void executeWorker(transactionPlan, witness, fullViewingKey, actionId).then(action => {
+    self.postMessage(action);
+  });
```
grod220, line 29: can shorten:

```typescript

  void executeWorker(transactionPlan, witness, fullViewingKey, actionId).then(self.postMessage);
```

apps/extension/src/wasm-task.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+self.addEventListener('message', workerListener);
+
+async function executeWorker(
+  transactionPlan: TransactionPlan,
+  witness: WitnessData,
+  fullViewingKey: string,
+  actionId: number,
+): Promise<Action> {
```
grod220, line 39: This is not JSON serialized. Think this type needs to be `JsonValue`.

grod220: Also, in general, `JsonValue` does not communicate much. I think we should try to use `Jsonified<Action>` for the sake of code documentation.

apps/extension/src/wasm-task.ts
grod220, line 1: The label `WasmTask` feels a bit too generic. Maybe `WasmBuildAction` instead?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,47 @@
...
+export const sendOffscreenMessage = async <T extends OffscreenMessage>(
+  req: InternalRequest<T>,
+): Promise<InternalResponse<ActionBuildMessage>> => {
+  await chrome.offscreen.createDocument({
+    url: OFFSCREEN_DOCUMENT_PATH,
+    reasons: [chrome.offscreen.Reason.WORKERS],
+    justification: 'Manages Penumbra transaction WASM workers',
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
grod220, line 37: this seems to make the linter happy:

```typescript

  const result = (await chrome.runtime.sendMessage({
    ...req,
  })) satisfies InternalResponse<ActionBuildMessage>;
```

packages/router/src/grpc/offscreen-client.ts
grod220: Yeah, I feel like we can just leave it open. When the service worker spins down, it should kill the offscreen too (I think?).

packages/types/src/indexed-db.ts
```diff
@@ -118,7 +118,7 @@ export interface PenumbraDb extends DBSchema {
 // @ts-expect-error Meant to be a marker to indicate it's json serialized.
 //                  Protobuf values often need to be as they are json-deserialized in the wasm crate.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-type Jsonified<T> = JsonValue;
+export type Jsonified<T> = JsonValue;
```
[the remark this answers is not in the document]
grod220: The issue is that `JsonValue` does not provide any documentation. It's essentially an `any`. This type is a helpful alias. I think we should implement this basically wherever we reasonably can. 

And if so prolific in our repo, we should move this out of the `indexed-db.ts` file to somewhere else.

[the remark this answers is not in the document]
TalDerei: moving it to `internal-msg/shared` in the `/types` package

apps/extension/src/wasm-task.ts
TalDerei: maybe it should be a generic `web-worker.ts`? We might decide to move more functionality to the web-worker in the future, and it should be named as such.

apps/extension/src/wasm-task.ts
TalDerei: removed

apps/extension/src/wasm-task.ts
TalDerei: This doesn't work. Instead we had to do

```
void executeWorker(transactionPlan, witness, fullViewingKey, actionPlanIndex).then(action =>
    self.postMessage(action.toJson()),
  );
```

apps/extension/src/wasm-task.ts
TalDerei: changed

apps/extension/src/wasm-task.ts
TalDerei: changed

apps/extension/src/wasm-task.ts
TalDerei: changed to `return penumbraWasmModule.buildActionParallel(transactionPlan, witness, fullViewingKey, actionId).toJson()` and returning `Jsonified<Action>` as you suggested.

apps/extension/src/wasm-task.ts
TalDerei: L moved this to the `internal-msg/offscreen.ts` in the `/types` package

apps/extension/src/offscreen.ts
TalDerei: L incorporated this change

packages/wasm/src/build.ts
TalDerei: Resolved using `zod record`

apps/extension/src/offscreen.ts
TalDerei: outdated, see updated flow

apps/extension/src/wasm-task.ts
TalDerei: added error handling

packages/types/src/transaction/schema.ts
grod220: Zod can do unions: https://zod.dev/?id=unions. However, the spirit of this effort isn't so much to re-type everything (and have to keep it updated as protobufs change), but have some base type validation to provide some safety from the wasm code from sending back something completely different and rugging the frontend. So in short, it's your call if you want to enumerate all of these or just expect an object.

packages/router/src/grpc/offscreen-client.ts
TalDerei: we can't keep it open. The user won't be able to send another transaction directly after. 

> Actually, seems like we need to close the document, no? Initiating a new transaction throws `"Only a single offscreen document may be created"`

If we keep the offscreen document open, this is still the case. The user can't issue a transaction directly after the first one.

TalDerei: > it isn't! but offscreen may simultaneously perform work for multiple callers. it's unclear how to identify that we no longer need the offscreen document.

@turbocrime I don't see in what scenario there would be multiple callers (at least for now). Let's keep `await chrome.offscreen.closeDocument();` for now since it's good practice, and we can address this in a separate issue.

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
TalDerei: It's strangely not recognizing `chrome.runtime.getContexts`. @turbocrime can you implement the detection system you're satisfied with?

apps/extension/src/wasm-task.ts
TalDerei: changed to `WasmBuildActionInput`

TalDerei:
I think we're close to be able to merge once the last few comments are addressed.

packages/router/src/grpc/offscreen-client.ts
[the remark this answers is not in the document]
turbocrime: lol. we're using a third-party types package which doesn't provide it

https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/66593

and google's official chrome-types also fails to provide it

https://github.com/search?q=repo%3AGoogleChrome%2Fchrome-types+Context&type=code

TalDerei:
The offscreen lifecycle management is sufficient for now. In a separate issue, we'll allow servicing multiple simultaneous requests.

apps/extension/src/offscreen.ts
```diff
@@ -0,0 +1,83 @@
...
+  const { transactionPlan, witness, fullViewingKey } = request;
+
+  return transactionPlan.actions.map((_, i) =>
+    spawnWorker(transactionPlan, witness, fullViewingKey, i),
+  );
+};
+
+const spawnWorker = (
+  transactionPlan: JsonObject & { actions: Jsonified<Action>[] },
+  witness: JsonObject,
```
grod220, line 51: Ah interesting. I haven't used `JsonObject` before. It's something we probably should use more often of than `JsonValue` I'd imagine.

apps/extension/src/wasm-task.ts
grod220: Think this would apply to `wasm-task.ts` too no? I don't think we'd re-use this for another purpose besides the build action.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+  OffscreenMessage,
+} from '@penumbra-zone/types/src/internal-msg/offscreen';
+
+const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
+
+let active = 0;
+
+const activateOffscreen = async () => {
+  // @ts-expect-error: no types available yet
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
grod220, line 19: Don't think this eslint code comment is needed, just the ts-expect-error one

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+  const getContexts = chrome.runtime.getContexts as (x: unknown) => Promise<unknown[]>;
```
grod220, line 20: Can we add some rough types here? Something like:

```typescript
  interface ContextsRequest {
    contextTypes: string[];
    documentUrls: string[];
  }

  interface Context {
    // ..
  }

  // @ts-expect-error: types to import not available yet
  const getContexts = chrome.runtime.getContexts as (
    request: ContextsRequest,
  ) => Promise<Context[]>;

  const offscreenExists = getContexts({
    contextTypes: ['OFFSCREEN_DOCUMENT'],
    documentUrls: [OFFSCREEN_DOCUMENT_PATH],
  }).then(contexts => contexts.length > 0);
```

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  if (active && (await offscreenExists)) active++;
+  else {
+    active++;
+    await chrome.offscreen.createDocument({
+      url: OFFSCREEN_DOCUMENT_PATH,
+      reasons: [chrome.offscreen.Reason.WORKERS],
+      justification: 'Manages Penumbra transaction WASM workers',
+    });
+  }
```
grod220, line 34: 1 - we should await getContexts above
2 - we shouldn't combine the no-brackets style with the brackets style of the if condition

```typescript
  const offscreenExists = await getContexts({...

  if (active && offscreenExists) {
    active++;
  } else {
    active++;
    await chrome.offscreen.createDocument({
      url: OFFSCREEN_DOCUMENT_PATH,
      reasons: [chrome.offscreen.Reason.WORKERS],
      justification: 'Manages Penumbra transaction WASM workers',
    });
  }
```

grod220: Hmm, actually, would this simplify the logic further?

```typescript
  const offscreenExists = await getContexts({...

  if (!active || !offscreenExists) {
    await chrome.offscreen.createDocument({
      url: OFFSCREEN_DOCUMENT_PATH,
      reasons: [chrome.offscreen.Reason.WORKERS],
      justification: 'Manages Penumbra transaction WASM workers',
    });
  }

  active++;
```

packages/types/src/internal-msg/offscreen.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type ActionBuildMessage = InternalMessage<
+  'BUILD_ACTION',
+  ActionBuildRequest,
+  ActionBuildResponse
+>;
+
+export type OffscreenMessage = ActionBuildMessage;
+export type OffscreenRequest = InternalRequest<OffscreenMessage>;
+export type OffscreenResponse = InternalResponse<OffscreenMessage>;
```
grod220, line 13: `OffscreenResponse` is not used. Is it supposed to be?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const sendOffscreenMessage = async <T extends OffscreenMessage>(
+  req: InternalRequest<T>,
+): Promise<InternalResponse<ActionBuildMessage>> => {
+  try {
+    await activateOffscreen();
+    return await chrome.runtime.sendMessage<InternalRequest<T>, InternalResponse<T>>(req);
+  } finally {
+    releaseOffscreen();
```
grod220, line 49: Does this need to be a void function given the surrounding environment is async? We could await this. Is it a void so the release doesn't slow down the return of the response?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+  active--;
+  if (!active) void chrome.offscreen.closeDocument();
+};
+
+const sendOffscreenMessage = async <T extends OffscreenMessage>(
+  req: InternalRequest<T>,
+): Promise<InternalResponse<ActionBuildMessage>> => {
+  try {
+    await activateOffscreen();
+    return await chrome.runtime.sendMessage<InternalRequest<T>, InternalResponse<T>>(req);
```
grod220, line 47: I've never done a try+finally with a return in the try block. Does finally get called after the function returns?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+  }
+};
+
+const releaseOffscreen = () => {
+  active--;
+  if (!active) void chrome.offscreen.closeDocument();
+};
+
+const sendOffscreenMessage = async <T extends OffscreenMessage>(
```
grod220, line 42: Can we add some code comments about the lifecycle of the offscreen window? So the next dev can understand what's happening below?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -0,0 +1,71 @@
...
+    type: 'BUILD_ACTION',
+    request: {
+      transactionPlan: arg.transactionPlan!.toJson() as ActionBuildRequest['transactionPlan'],
+      witness: witness.toJson() as ActionBuildRequest['witness'],
+      fullViewingKey,
+    },
+  });
+  if ('error' in buildRes) throw new Error(String(buildRes.error));
+  const actions = buildRes.data.map(a => Action.fromJson(a));
+  return actions;
```
grod220, line 68: Can shorten:
```typescript
return buildRes.data.map(a => Action.fromJson(a));
```

grod220 (approved):
Great job! Have some remaining points from me. Really excited to get this in though. After this (and Christmas), we should do a release and get users on the new fast builds.

packages/router/src/grpc/offscreen-client.ts
TalDerei: removed

apps/extension/src/offscreen.ts
TalDerei: https://stackoverflow.com/questions/61803709/what-is-the-difference-between-a-jsonvalue-and-a-jsonobject

packages/types/src/internal-msg/offscreen.ts
TalDerei: It's not after the refactor, removed.

packages/router/src/grpc/offscreen-client.ts
TalDerei: shortened

packages/router/src/grpc/offscreen-client.ts
TalDerei: added `ContextsRequest`. We don't know what the return type of `getContexts` is, hence `Promise<unknown[]>`.

packages/router/src/grpc/offscreen-client.ts
TalDerei: yes, these are equivalent. simplified

packages/router/src/grpc/offscreen-client.ts
TalDerei: The finally block is guaranteed to be executed, whether an exception is thrown in the try block or not. The finally block will be called before the function returns.

packages/router/src/grpc/offscreen-client.ts
TalDerei: It doesn't seem to affect the build time, but I don't see a reason to make the offscreen cleanup a synchronous operation by calling await. Or maybe it doesn't make a difference between the finally block has to be called regardless before the function returns.

TalDerei: The offscreen API documentation mentions `await chrome.offscreen.closeDocument()`, so we should do the same. I changed it to await.

packages/router/src/grpc/offscreen-client.ts
TalDerei: added comments how the offscreen lifecycle is constructed.

apps/extension/src/wasm-task.ts
TalDerei: changed to `wasm-build-action.ts`

PR #306, opened 2023-12-15
Retain lost promise racer

grod220:
Thanks to @turbocrime for finding this one. Follow up from: https://github.com/penumbra-zone/web/pull/301. @Valentine1898 fyi.

Previously, the promise that "lost" the race was dropped. This would mean that the slower detected swap/note would not be checked again on the next iteration. `mergeAsyncGenerators` has been updated to retain the slower.

gene-wilder-you-lose

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/nullifier-status.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/nullifier-status.test.ts
@@ -9,6 +9,7 @@ import { ServicesInterface, stringToUint8Array } from '@penumbra-zone/types';
 import { WalletId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1alpha1/keys_pb';
 import { Nullifier } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/sct/v1alpha1/sct_pb';
 import { assertWalletIdMatches } from './utils';
+import { JsonValue } from '@bufbuild/protobuf';
 
 const mockAssertWalletId = assertWalletIdMatches as Mock;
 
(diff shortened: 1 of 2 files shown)


packages/router/src/grpc/view-protocol-server/nullifier-status.test.ts
```diff
@@ -228,8 +229,68 @@ describe('handleNullifierStatusReq', () => {
...
+      })
+      .mockResolvedValueOnce({
+        value: { value: nonMatchingNote.toJson(), table: 'SPENDABLE_NOTES' },
+      })
+      .mockResolvedValueOnce({
+        value: { value: nonMatchingNote.toJson(), table: 'SPENDABLE_NOTES' },
+      })
+      .mockResolvedValueOnce({
+        value: { value: nonMatchingNote.toJson(), table: 'SPENDABLE_NOTES' },
+      });
```
grod220, line 252: Adding more spendable notes, triggered a failure with the last method

packages/router/src/grpc/view-protocol-server/nullifier-status.test.ts
```diff
@@ -228,8 +229,68 @@ describe('handleNullifierStatusReq', () => {
...
+    const req = new NullifierStatusRequest({
+      nullifier: matchingNullifier,
+      awaitDetection: true,
+    });
+
+    const res = await handleNullifierStatusReq(req, mockServices as unknown as ServicesInterface);
+    expect(res.spent).toBe(true);
+  });
+
+  test('await losing the race does gets picked up next loop', async () => {
```
grod220, line 263: Adding a test to explicitly guard for this

turbocrime:
i think it would be conceptually cleaner to iterate each generator independently, instead of merging them. give me a moment and i'll suggest something

grod220:
> i think it would be conceptually cleaner to iterate each generator independently, instead of merging them

I would be nice to have two for await loops, but one will block the other until it's finished. And in this case, it's infinite given its just waiting for indexeddb updates. Unless you have another idea?

turbocrime:
im gonna say streams. im gonna do it im gonna say streams

* function accepts a list of the iterators, and a function to use for confirmation.
* returns a promise of its resolution.
* internally, converts iterators to streams, and void pipeTos them to sinks that have the resolution callback.
* voided pipes are running in parallel. confirmation function runs async in the sink. they both go as fast as they'd like.
* resolution of the promise cancels the streams

but its 7pm and im gonna go get dinner, so this lgtm

turbocrime (approved):

grod220:
If you can beat `mergeAsyncGenerators` and do it in 18 lines of code or less, I'm all game ha. Otherwise, I think the web streams API is a good candidate for an upcoming "Knowledge Share" meeting.

PR #321, opened 2023-12-28
Change terminology to IBC Deposit Address

hdevalence:
Cf https://github.com/penumbra-zone/penumbra/issues/3381#issuecomment-1869311071

### DIFF
--- a/apps/webapp/src/components/send/send-form.tsx
+++ b/apps/webapp/src/components/send/send-form.tsx
@@ -1,4 +1,4 @@
-import { Button, Switch } from '@penumbra-zone/ui';
+import { Button } from '@penumbra-zone/ui';
 import { useStore } from '../../state';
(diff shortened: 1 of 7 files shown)


hdevalence:
Build failures seem unrelated to the content of the PR, maybe someone else could pick this up.

turbocrime (approved):
test was using old props. updated, seems ok

but "ibc deposit address" feels like opaque jargon

hdevalence:
> but "ibc deposit address" feels like opaque jargon

Yeah, I think that's OK. Probably better to have jargon that's [precise](https://without.boats/blog/names-and-scuba/) than to make up a word people can project their meaning onto. If people don't use them because they don't know what it means that seems fine (per motivation in https://github.com/penumbra-zone/penumbra/issues/3381#issuecomment-1869311071)

Issue #324, opened 2023-12-28
start building transactions optimistically while the approval popup is displayed


hdevalence:
we can hide the latency of proving by starting to build the transaction optimistically while the user is looking at the approval popup. if we get the proving fast enough (faster than the time to review the tx plan) we may be able to have "apparently instant" building.

https://twitter.com/hdevalence/status/1738713884138336306

turbocrime:
currently, the building toast blocking time includes broadcast and confirmation

TalDerei:
References #113

hdevalence:
> currently, the building toast blocking time includes broadcast and confirmation

there's not a way to change this with the current view service rpc, right? should we change it so that it does streaming responses, for the different stages (returning a message when built, broadcasted, confirmed, etc)?

grod220:
> there's not a way to change this with the current view service rpc, right?

We now can use the `AuthorizeAndBuild` method which would achieve the original intention of this issue.

> should we change it so that it does streaming responses, for the different stages

I think this would be good regardless. At the moment, we have a generic loading spinner on the web app. It doesn't give a ton of information to the user on what actually is happening (or what stage something actually errored).

turbocrime:
after reading this issue the other day i did some work to split the broadcast time to a separate toast, i'll make a PR. i thought about streaming responses back from offscreen while we worked on that and i think that is simply achievable too, but i didnt bring it up in the interest of limiting scope.

build progress would require a spec change if we want to inform the client, but maybe it makes sense to display transaction build progress in the extension icon or popup, which would require no spec change.

it seems like some build steps require authorization data? is that correct? if we want to optimistically build while pending interactive approval, we could just provide the auth, and make broadcast the approval-gated step. that would require no spec change, but that seems unsafe since auth data flows all the way back to the app.

with #282 , we possibly could keep auth in-extension at all times, and use context or an 'interceptor' to inject authorizationData for approved requests.

hdevalence:
We can and should change the view rpc definitions to allow giving progress info. However, we can't easily do it at the moment because I made a large set of consensus breaking proto changes (hopefully the last set before mainnet) and so the current protos aren't compatible with the current testnet. (We've had this problem before but shouldn't have it again past the 15th of January).

PR #327, opened 2023-12-30
broadcast toast

turbocrime:
slight motion towards transaction progress indication. discussed in #324

### DIFF
--- a/apps/webapp/src/components/shared/toast-content.tsx
+++ b/apps/webapp/src/components/shared/toast-content.tsx
@@ -1,23 +1,107 @@
+import {
+  AuthorizeAndBuildResponse,
+  WitnessAndBuildResponse,
+  BroadcastTransactionResponse,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { shorten } from '@penumbra-zone/types';
 import { ToastAction } from '@penumbra-zone/ui/components/ui/toast';
 import { ToastFnProps } from '@penumbra-zone/ui/components/ui/use-toast';
-import { Grid } from 'react-loader-spinner';
+import { Bars, Grid, Watch } from 'react-loader-spinner';
 import { Link } from 'react-router-dom';
 
-export const loadingTxToast: ToastFnProps = {
-  main: (
-    <div className='flex items-center justify-center gap-4'>
-      <span>Transaction build in progress</span>
-      <Grid height='20' width='20' color='white' />
-    </div>
-  ),
-  duration: Infinity, // Building could potentially take a long time. Expected to be dismissed manually.
+export const buildingTxToast = (
+  status?: (AuthorizeAndBuildResponse | WitnessAndBuildResponse)['status'],
+): ToastFnProps => {
+  let progress: undefined | number;
+
+  switch (status?.case) {
+    case undefined:
+      progress = 0;
+      break;
+    case 'buildProgress':
+      progress = status.value.progress;
+      break;
+    case 'complete':
+      progress = 1;
+      break;
+    default:
+      console.warn('Unknown build status', status);
+      break;
+  }
+
+  return {
+    duration: Infinity,
+    title: 'Building Transaction',
+    main: (
+      <div className='flex items-center justify-center gap-4'>
+        <span>
+          Transaction building...
+          {progress != null ? `${Math.round(progress * 100)}%` : null}
+        </span>
+        <Grid height='20' width='20' color='white' />
+      </div>
+    ),
+  };
 };
 
-export const successTxToast = (txHash: string): ToastFnProps => ({
-  main: 'Transaction success 🎉',
+export const broadcastingTxToast = (
+  txHash: string,
+  status?: BroadcastTransactionResponse['status'],
+): ToastFnProps => {
+  let main: JSX.Element | undefined = undefined;
+  let variant: 'default' | 'success' | undefined = undefined;
+  switch (status?.case) {
+    case undefined:
+      variant = 'default';
+      main = (
+        <div className='flex items-center justify-center gap-4'>
+          <span>Emitting transaction...</span>
+          <Bars height='20' width='20' color='white' />
+        </div>
+      );
+      break;
+    case 'broadcastSuccess':
+      variant = 'default';
+      main = (
+        <div className='flex items-center justify-center gap-4'>
+          <span>Waiting for confirmation...</span>
+          <Watch height='20' width='20' color='white' />
+        </div>
+      );
+      break;
+    case 'confirmed': // about to dismiss the toast
+      variant = 'success';
+      break;
+    default:
+      console.warn('Unknown broadcast status', status);
+      break;
+  }
+
+  return {
+    duration: Infinity,
+    title: 'Broadcasting Transaction',
+    variant,
+    subText: shorten(txHash, 8),
+    main,
+  };
+};
+
+export const successTxToast = (
+  txHash: string,
+  detectionHeight?: bigint | undefined,
+): ToastFnProps => ({
+  duration: Infinity,
+  title: 'Confirmed Transaction',
   variant: 'success',
   subText: shorten(txHash, 8),
+  main: (
+    <div className='flex items-center justify-center gap-4'>
+      <span>
+        Transaction appeared on chain
+        {detectionHeight ? <>at height ${detectionHeight}</> : null}
+      </span>
+    </div>
+  ),
   action: (
     <Link to={`/tx/${txHash}`}>
       <ToastAction className='border-transparent bg-teal-800' altText='See transaction details'>
@@ -28,6 +112,7 @@ export const successTxToast = (txHash: string): ToastFnProps => ({
 });
 
 export const errorTxToast = (error: unknown): ToastFnProps => ({
+  duration: Infinity,
   variant: 'error',
   main: 'Error with transaction',
   subText: <p className='break-all'>{String(error)}</p>,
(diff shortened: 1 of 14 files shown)


apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -14,6 +14,20 @@ export const loadingTxToast: ToastFnProps = {
...
+export const broadcastingTxToast = (): ToastFnProps => ({
+  main: 'Broadcasting transaction',
+  variant: 'default',
+  //subText: shorten(txHash, 8),
+  action: (
+    <Link to='/dashboard/transactions'>
+      <ToastAction className='border-transparent bg-teal-800' altText='See confirmed transactions'>
+        Pending confirmation...
+      </ToastAction>
+    </Link>
```
turbocrime, line 26: any suggestions for toast content?

hdevalence: maybe "Waiting for confirmation..."?

apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -14,6 +14,20 @@ export const loadingTxToast: ToastFnProps = {
   duration: Infinity, // Building could potentially take a long time. Expected to be dismissed manually.
 };
 
+export const broadcastingTxToast = (): ToastFnProps => ({
+  main: 'Broadcasting transaction',
+  variant: 'default',
```
grod220, line 19: Soon we'll have design resources to help us think through and polish the flow of this a bit more. I suspect that toasts aren't the way to go with these given how quickly they are dismissed between stages. A table this is probably better:

```
[x] Authorizing
[-] Building
[ ] Broadcasting
[ ] Waiting for confirmation
```

But this is for a later discussion when the design team is onboarded.

apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -14,6 +14,20 @@ export const loadingTxToast: ToastFnProps = {
...
+  //subText: shorten(txHash, 8),
```
grod220, line 20: Can remove

apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -14,6 +14,20 @@ export const loadingTxToast: ToastFnProps = {
...
+  action: (
```
grod220, line 21: Don't think we should have an action button for this stage. It too quickly disappears and it goes to a table that isn't responsive when a new transaction comes in.

Given the short length of time, we should just have another spinner like: https://loading.io/spinner/radio/-alien-connection-signal-radio-rss-wifi-information-transmit-sync-transfer with the new text.

apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -14,6 +14,20 @@ export const loadingTxToast: ToastFnProps = {
   duration: Infinity, // Building could potentially take a long time. Expected to be dismissed manually.
 };
 
+export const broadcastingTxToast = (): ToastFnProps => ({
```
grod220, line 17: Given we aren't passing props, we can just make this an object and not a function

apps/webapp/src/components/shared/toast-content.tsx
turbocrime: toasts provide an "update" function that could be used to feed status into a sustained toast, or we could have some other ui. i think personally i'd like to see it tracked in the header

turbocrime:
resurrecting this

packages/types/src/internal-msg/offscreen.ts
turbocrime, line 1: wasm worker no longer has a separate input message. this could probably be improved

apps/webapp/src/components/shared/toast-content.tsx
```diff
@@ -86,7 +86,10 @@ export const broadcastingTxToast = (
   };
 };
 
-export const successTxToast = (txHash: string, detectionHeight?: bigint): ToastFnProps => ({
+export const successTxToast = (
+  txHash: string,
+  detectionHeight?: bigint | undefined,
```
jessepinho, line 91: `exactOptionalPropertyTypes` is disabled now! so you don't need this 🥳

jessepinho (commented):
I haven't reviewed the PR fully but left a couple comments, since it looks like it's still in progress (right?). As for the toast functionality, that shouldn't be a problem — either you or I will have to rebase off the other's PR, but shouldn't be too painful.

turbocrime:
i think this is ready to land

apps/webapp/src/components/shared/toast-content.tsx
turbocrime: gonna defer removal of optional |undefineds to another issue

apps/webapp/src/state/swap.ts
```diff
@@ -55,28 +55,38 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
+        // @see https://protocol.penumbra.zone/main/zswap/swap.html#claiming-swap-outputs
+
+        // start to build swap claim
+        const swapCommitment = getSwapCommitment(swapTx);
+
+        const swapClaimPlan = await plan(assembleSwapClaimRequest(swapCommitment));
+        const swapClaimTx = await witnessBuild(update, { transactionPlan: swapClaimPlan });
+
+        // after swap broadcast success, emit the claim
+        await swapBroadcast;
```
jessepinho, line 83: Ohhh smart! allows the build of one to start while the other is still being broadcast 👏🏻

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -15,59 +15,70 @@ const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
...
-  if (!active || !offscreenExists) {
-    await chrome.offscreen.createDocument({
-      url: chrome.runtime.getURL(OFFSCREEN_DOCUMENT_PATH),
-      reasons: [chrome.offscreen.Reason.WORKERS],
-      justification: 'Manages Penumbra transaction WASM workers',
-    });
-  }
-
-  active++;
+  if (!active++ || (await noOffscreen))
```
jessepinho, line 24: Could we move the `await` back up above?

```suggestion
  if (!active++ || noOffscreen)
```

I'd also consider renaming `noOffscreen` to be a bit more descriptive so that we can understand what it means at a glance — e.g., `offscreenWorkerDoesNotExist` or something

jessepinho: Hm, I'd also consider incrementing `active` _inside_ the `if` block. This line is packing a lot into it, which makes it a little harder to grok. It's:
- Checking that there are 0 active offscreen contexts
- Incrementing the count of offscreen contexts
- `await`ing `noOffscreen` and then checking its value

May just be my preference, but I feel like an `if` statement is generally clearest if it's just evaluating existing values, rather than assigning new ones.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -15,59 +15,70 @@ const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
...
+      });
 };
 
+/**
+ * Decrement and close if there is no remaining activity.
+ */
 const releaseOffscreen = async () => {
-  active--;
-  if (!active) await chrome.offscreen.closeDocument();
+  if (!--active) await chrome.offscreen.closeDocument();
```
jessepinho, line 42: Same comment here — no need to pack too much into an `if` statement — these can be separate lines to aid human-readability.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -15,59 +15,70 @@ const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
...
-    type: 'BUILD_ACTION',
-    request: {
-      transactionPlan: transactionPlan.toJson() as ActionBuildRequest['transactionPlan'],
-      witness: witness.toJson() as ActionBuildRequest['witness'],
-      fullViewingKey,
-    },
+  cancel: PromiseLike<never>,
+): Promise<Action>[] => {
+  const active = activateOffscreen();
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
jessepinho, line 62:
```suggestion
```

Per my VSCode at least, this isn't needed.

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -15,59 +15,70 @@ const OFFSCREEN_DOCUMENT_PATH = '/offscreen.html';
...
-    await releaseOffscreen();
-  }
-};
+): Promise<InternalResponse<ActionBuildMessage>> =>
+  chrome.runtime.sendMessage<InternalRequest<T>, InternalResponse<T>>(req);
 
-const buildAction = async (
+/**
+ * Build actions in parallel, in an offscreen window where we can run wasm.
+ * @param cancel Promise that rejects if the build should be cancelled, usually auth denial.
```
jessepinho, line 52: This pattern is a bit strange to me — can't you just return a promise in this function that `reject`s on failure? Or am I misunderstanding the logic flow here? (Is `cancel` a way to remotely stop `buildActions`, sort of like an `AbortController`'s `signal`? And if so, why not use an `AbortController`? Partly asking just for my own benefit.)

jessepinho (approved):
see comments — no blockers

packages/router/src/grpc/offscreen-client.ts
jessepinho: Per our convo on Discord, we're not using `AbortController` because you can't use `Promise.race()` with it

packages/router/src/grpc/offscreen-client.ts
turbocrime: we discussed this while pairing on discord and i agree this does need more explanation but it's this way for a reason. would like to get off the rebase treadmill so i'll issue this and handle it later

PR #329, opened 2023-12-30
form validation for memo

### LINKED ISSUE #323: send dialog and plumbing does not handle over-long memos
1. generate a 512-byte memo plaintext, this is longer than the limit (the address and the text together are 512 bytes)
```
Python 3.9.6 (default, Nov 10 2023, 13:38:27)
[Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'aaaaaaa ' * (512//8)
```
2. paste into memo field in send dialog
3. tx approval popup appears, click approve
4. tx build fails with inscrutable error message
Screenshot 2023-12-28 at 2 30 58 PM

two problems here:

1. form should be doing input validation on memo length
2. errors should be propagated back to user

turbocrime:
form validation for memo

### DIFF
--- a/apps/webapp/src/state/send.ts
+++ b/apps/webapp/src/state/send.ts
@@ -15,6 +15,7 @@ import { Selection } from './types';
 import { MemoPlaintext } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import { viewClient, custodyClient } from '../clients/grpc';
 import { getAddressByIndex } from '../fetchers/address.ts';
+import { BECH32_ADDRESS_LENGTH } from '@penumbra-zone/types';
 
 export interface SendSlice {
   selection: Selection | undefined;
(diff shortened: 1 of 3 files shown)


apps/webapp/src/state/send.ts
```diff
@@ -148,16 +149,19 @@ export const validateAmount = (asset: AssetBalance, amount: string): boolean =>
...
 export const sendValidationErrors = (
   asset: AssetBalance | undefined,
   amount: string,
   recipient: string,
+  memo?: string,
 ): SendValidationFields => {
   return {
     recipientErr: Boolean(recipient) && !isPenumbraAddr(recipient),
     amountErr: !asset ? false : validateAmount(asset, amount),
+    memoErr: new TextEncoder().encode(memo).length > 512 - BECH32_ADDRESS_LENGTH,
```
turbocrime, line 164: is this length calc correct

turbocrime:
i only updated the validation function, but it might be nice to provide an indication of maximum length, or use an input attribute to actually limit the field.

turbocrime:
error happens in wasm. so handler is failing to rethrow as ConnectError, and, not validating before calling wasm

turbocrime:
the js shim provided by wasm-bundler is not rethrowing the error. this is the source of the error flow disconnect. the error received by the implementation is literally a webassembly 'unreachable'

we can try/catch around every wasm call, to address this. but a correct solution should involve fixing wasm crate to rethrow specific errors

apps/webapp/src/state/send.ts
grod220: According to Henry's issue it's `address+memo <= 521 bytes`. So this looks right. Can we add a code comment right above?

```diff
+ // The memo plus the address cannot exceed 512 bytes
memoErr: new TextEncoder().encode(memo).length > 512 - BECH32_ADDRESS_LENGTH,
```

grod220 (approved):

grod220:
Also, I'd change the PR description so it doesn't close the issue on merge because the Rust code still needs changed too.

apps/webapp/src/state/send.ts
turbocrime: turns out this was incorrect, at least according to how wasm throws. it is using an 80-byte address in the memo and will throw between a 432-char -> 433-char memo length

PR #331, opened 2024-01-02
Require login before tx popup

grod220:
After the user approves the popup, if they are not logged in, an error is thrown. This PR moves that check up so that the user doesn't get an approval popup without being logged in first.


turbocrime (approved):
updated after #282 merge

PR #346, opened 2024-01-04
wrap impls to rethrow

turbocrime:
this is a temp workaround. we could also 

1) do nothing. it's not the end of the world to default to concealed errors
2) manually try/catch in impls
3) [PR connectrpc](https://github.com/connectrpc/connect-es/compare/main...turbocrime:connect-es:internal-errors)

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -22,7 +22,12 @@ import { BackgroundConnectionManager } from '@penumbra-zone/transport/src/chrome
 import { createProxyImpl } from '@penumbra-zone/transport/src/proxy';
 import { connectChromeRuntimeAdapter } from '@penumbra-zone/transport/src/chrome-runtime/adapter';
 
-import { ConnectRouter, createContextValues, createPromiseClient } from '@connectrpc/connect';
+import {
+  ConnectError,
+  ConnectRouter,
+  createContextValues,
+  createPromiseClient,
+} from '@connectrpc/connect';
 import { createGrpcWebTransport } from '@connectrpc/connect-web';
 
 // this is a remote service we proxy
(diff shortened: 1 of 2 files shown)


apps/extension/src/service-worker.ts
```diff
@@ -71,10 +76,39 @@ const ibcImpl = createProxyImpl(
...
+        try {
+          const x = (v as (...args: unknown[]) => unknown)(...args);
+          if (x instanceof Promise)
+            return (x as Promise<unknown>).catch(e => {
+              throw ConnectError.from(e);
+            });
+          return x;
+        } catch (e) {
+          throw ConnectError.from(e);
+        }
```
turbocrime, line 102: this probably fails to rethrow errors that occur mid-stream

packages/transport/src/chrome-runtime/adapter.ts
```diff
@@ -134,12 +134,7 @@ export const connectChromeRuntimeAdapter = (
         Object.entries(handler) as [keyof UniversalHandler, unknown][],
       ) as { [k in keyof UniversalHandler]: UniversalHandler[k] };
       // wrap the handler function to generate and apply context
-      const wrappedFn: UniversalHandlerFn = req =>
-        opt.createRequestContext(req).then(handlerFn, error => {
-          // it's convenient to log and rethrow as ConnectError here
-          console.warn('Handler Error', handler.name, error);
-          throw ConnectError.from(error);
-        });
+      const wrappedFn: UniversalHandlerFn = req => opt.createRequestContext(req).then(handlerFn);
```
turbocrime, line 137: if we choose 'do nothing', we should still at least merge this removal of this ineffective rethrow

grod220 (approved):
Definitely in favor of propagating errors vs concealing them

PR #347, opened 2024-01-04
add custodyClient to context

turbocrime:
cc @TalDerei


turbocrime:
this is pretty much just for example, if you'd actually like custodyClient to be within the context just apply these changes to #350

TalDerei:
looks good to me!

PR #350, opened 2024-01-04
Refactor Stage 2: HandlerContext, Document Control, Internal Messaging

turbocrime:
Expanded refactor after #282

Moves document lifecycle management into extension.

Abstracted most extension document interaction into context - storage interaction remains to be refactored.

Deletes the stdRouter, in favor of document-specific handlers more closely associated with the functionality.


packages/transport/src/chrome-runtime/adapter.ts
```diff
@@ -134,12 +134,7 @@ export const connectChromeRuntimeAdapter = (
         Object.entries(handler) as [keyof UniversalHandler, unknown][],
       ) as { [k in keyof UniversalHandler]: UniversalHandler[k] };
       // wrap the handler function to generate and apply context
-      const wrappedFn: UniversalHandlerFn = req =>
-        opt.createRequestContext(req).then(handlerFn, error => {
-          // it's convenient to log and rethrow as ConnectError here
-          console.warn('Handler Error', handler.name, error);
-          throw ConnectError.from(error);
-        });
+      const wrappedFn: UniversalHandlerFn = req => opt.createRequestContext(req).then(handlerFn);
```
turbocrime, line 137: this was ineffective

apps/extension/src/hooks/last-block-synced.ts
```diff
@@ -22,7 +21,8 @@ export const useSyncProgress = () => {
   useQuery({
     queryKey: ['sync-blocks', grpcEndpoint],
     queryFn: async () => {
-      await internalSwClient.syncBlocks();
+      const syncing: boolean = await chrome.runtime.sendMessage('PENUMBRA_SYNC_BLOCKS');
```
grod220, line 24: This feels like a worse dev experience / less type safe. We should encapsulate Std messages in one place somewhere. I expect we will need more Std messages as time goes by too.

apps/extension/src/service-worker.ts
```diff
@@ -52,18 +54,25 @@ const servicesConfig = {
...
+    case 'PENUMBRA_SYNC_BLOCKS':
+      void services.tryToSync().then(
+        () => sendResponse(true),
+        () => sendResponse(false),
+      );
+      return true; // async handler
+    default:
+      return; // unhandled
+  }
+});
```
turbocrime, line 75: this is a little simplistic and should soon be refactored to, say, an enum, and a convenience promise rejector. but now internal extension communication is not mediated through the router package at all, the router package is solely for rpc services.

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -24,7 +23,8 @@ export const SettingsRPC = () => {
       try {
         await querier.chainParams();
         await setGRPCEndpoint(rpc);
-        await internalSwClient.clearCache();
+        const cleared: boolean = await chrome.runtime.sendMessage('PENUMBRA_CLEAR_CACHE');
```
grod220, line 26: Manually annotating the types here is also not so great. The consumers of these should get the typing automatically. If not, it will not stay in sync if changed.

apps/extension/src/service-worker.ts
```diff
@@ -52,18 +54,25 @@ const servicesConfig = {
...
+        () => sendResponse(true),
+        () => sendResponse(false),
+      );
+      return true; // async handler
+    case 'PENUMBRA_SYNC_BLOCKS':
+      void services.tryToSync().then(
+        () => sendResponse(true),
+        () => sendResponse(false),
+      );
+      return true; // async handler
```
grod220, line 71: Is there a way we can do this without every method doing .then() + sendResponse? Maybe a wrapper for this? Think the old std router did that.

packages/router/src/grpc/custody/authorize.ts
```diff
@@ -26,7 +26,9 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!decryptedSeedPhrase)
     throw new ConnectError('Unable to decrypt seed phrase with password', Code.Unauthenticated);
 
-  await approveReq(req);
+  await getTxApproval(req).catch(no => {
+    throw ConnectError.from(no, Code.PermissionDenied);
+  });
```
grod220, line 31: Is that required? Why can't we just let its own error bubble up?

packages/types/src/std-route.ts
```diff
@@ -1,59 +0,0 @@
...
-
-// List all service worker messages that are allowed to be called from dapp
-export const allowedDappMessages: SwRequestMessage['type'][] = ['PING'];
-
-export type SyncBlocksMessage = InternalMessage<'SYNC_BLOCKS', undefined, Promise<void>>;
-export interface PongResponse {
-  ack: string;
-}
-export type PingMessage = InternalMessage<'PING', string, PongResponse>;
-export type ClearCacheMessage = InternalMessage<'CLEAR_CACHE', undefined, Promise<void>>;
```
grod220, line 59: There were benefits of having the types be like this that address my previous comments

packages/router/src/grpc/custody/authorize.ts
turbocrime: throwing as connecterror provides some informative detail codes. and until #346 it won't bubble otherwise

apps/extension/src/hooks/last-block-synced.ts
turbocrime: i agree we should have something convenient, this was a shortcut to demonstrate removing stdclient was imminently possible

i held stdclient removal back from routing-refactor because i knew this would come up. a simpler interface can be developed in this pr.

packages/transport/src/chrome-runtime/adapter.ts
turbocrime: another pr merged this change

apps/extension/src/routes/popup/settings/settings-rpc.tsx
turbocrime: using a 'blockCacheControl' interface now

packages/router/src/grpc/custody/authorize.ts
turbocrime: merged, but i'd like to keep this as the error code adds detail

packages/types/src/std-route.ts
turbocrime: revised

TalDerei:
Is there generally any downsides to using a Context mechanism for managing and sharing state across multiple levels of the component tree? More specifically, what's happening internally (simply messaging passing?) that enables the `custodyClient` to be callable by the `viewClient` in a context setting?

packages/router/src/ctx.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// these context values provide broad access to storage, to maintain existing
+// impl. they are too tightly coupled to our extension and should be refactored
+export const servicesCtx = createContextKey({} as ServicesInterface);
+export const extLocalCtx = createContextKey({} as ExtensionStorage<LocalStorageState>);
+export const extSessionCtx = createContextKey({} as ExtensionStorage<SessionStorageState>);
+
+// these are good examples: simple functions, using types from spec
+export type AssertWalletIdFn = (walletId?: WalletId) => Promise<void>;
+export type GetTxApprovalFn = (authReq: AuthorizeRequest) => Promise<void>;
```
TalDerei, line 22: Why are we putting simple functions inside of context?

turbocrime: assertwalletid is removed, in version update. so forget about that one

the implementations are hopefully reused by developers of third party apps. context allows interfacing with the 'host' of the implementation. currently, the interface is not as abstract as it should be - but, the idea is, interactive approval is a host implementation detail, of whatever app is hosting the service, so we provide a simple, typed interface to interact with it.

turbocrime:
> Is there generally any downsides to using a Context mechanism for managing and sharing state across multiple levels of the component tree? More specifically, what's happening internally (simply messaging passing?) that enables the `custodyClient` to be callable by the `viewClient` in a context setting?

context should be used to abstract interfaces of the extension that the implementation needs to access. context should not be used to share state directly, because context objects are not reused - they are created new for every activation of a handler. but, context may be used to inject state known to the adapter or router at request time, or to provide a queryable interface to state.

context is provided by a function parameter to the adapter. our context creation function is in service-worker.ts

https://github.com/penumbra-zone/web/blob/34cd33bfc1a2112e4c21240f576fa33da1f8165b/apps/extension/src/service-worker.ts#L114-L129

it's used in adapter.ts to inject state. the adapter knows nothing of our extension architecture, it's a generic 'chrome runtime to connectrpc' adapter, and we're *configuring it with our function def in service-worker*. there are comments in adapter.ts that may be informative.

https://github.com/penumbra-zone/web/blob/34cd33bfc1a2112e4c21240f576fa33da1f8165b/packages/transport/src/chrome-runtime/adapter.ts

turbocrime:
it's also worth understanding that contextValues is not the only thing contained within `HandlerContext`

https://github.com/connectrpc/connect-es/blob/cc6662442d2502f04dceb155d85c4b0208017840/packages/connect/src/implementation.ts#L69

it provides features for interacting with / gaining information about the request itself, and respecting things like timeouts and aborts, which are managed by the adapter

PR #351, opened 2024-01-04
Fix proving key path

TalDerei:


apps/extension/src/utils/download-proving-keys.ts
```diff
@@ -4,7 +4,7 @@ import * as path from 'path';
 import { provingKeys } from '@penumbra-zone/types/src/proving-keys';
 
 const githubSourceDir =
-  'https://github.com/penumbra-zone/penumbra/raw/main/crates/crypto/proof-params/src/gen/';
+  'https://github.com/penumbra-zone/penumbra/raw/v0.64.1/crates/crypto/proof-params/src/gen/';
```
cronokirby, line 7:
```suggestion
  'https://github.com/penumbra-zone/penumbra/tree/v0.64.1/crates/crypto/proof-params/src/gen/';
```

cronokirby:
Oh wait, I think `raw` might actually have been correct for the files themselves; not sure what was failing earlier.

TalDerei:
> Oh wait, I think `raw` might actually have been correct for the files themselves; not sure what was failing earlier.

I committed your suggestion prematurely, it seems to be failing now. `raw` should be correct for accessing the direct, unprocessed contents of the file.

grod220 (approved):

PR #353, opened 2024-01-05
Hide the copy button for one-time addresses

### LINKED ISSUE #173: AddressViewComponent should hide copy button for visible one-time addresses
See https://x.com/hdevalence/status/1719109465851793570  for motivation.

jessepinho:
🚨 Note that hundreds of the line changes are just from `pnpm-lock.yaml`

## Background
Per #173, we should hide the copy button for one-time addresses, since they should never be reused and thus don't need to be copied.

## In this PR
- Hide the copy button for one-time addresses
- Set up test infrastructure with Vitest + JSDom + React Testing Library for the `ui` package. This adds a bit of scope to the PR, but I figured it'd be worth it so that we can start writing tests for all of our UI components. Let me know what y'all think of the setup.

Closes #173

### DIFF
--- a/packages/ui/components/ui/tx/view/address-view.tsx
+++ b/packages/ui/components/ui/tx/view/address-view.tsx
@@ -24,17 +24,19 @@ export const AddressViewComponent = ({
 
   const accountIndex =
     view.addressView.case === 'visible' ? view.addressView.value.index?.account : undefined;
-  const isRandomized =
+  const isOneTimeAddress =
     view.addressView.case === 'visible'
-      ? !view.addressView.value.index?.randomizer.every(v => v === 0) //   Randomized if the randomizer is not all zeros.
+      ? !view.addressView.value.index?.randomizer.every(v => v === 0) // Randomized (and thus, a one-time address) if the randomizer is not all zeros.
       : undefined;
+
+  copyable = isOneTimeAddress ? false : copyable;
 
   return (
     <div className='flex'>
       {accountIndex !== undefined ? (
         <div className='flex items-baseline gap-2'>
           <Identicon name={encoded} size={14} className='rounded-full' type='gradient' />
-          {isRandomized ? (
+          {isOneTimeAddress ? (
             <span className='font-bold'>One-time Address for Account #{accountIndex}</span>
           ) : (
             <span className='font-bold'>Account #{accountIndex}</span>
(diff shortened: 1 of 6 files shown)


packages/ui/components/ui/tx/view/address-view.tsx
```diff
@@ -24,17 +24,19 @@ export const AddressViewComponent = ({
 
   const accountIndex =
     view.addressView.case === 'visible' ? view.addressView.value.index?.account : undefined;
-  const isRandomized =
+  const isOneTimeAddress =
```
jessepinho, line 27: Changed to `isOneTimeAddress` since it makes the reassignment of `copyable` on line 32 below make more sense.

packages/ui/package.json
```diff
@@ -26,23 +27,29 @@
     "djb2a": "^2.0.0",
     "framer-motion": "^10.16.16",
     "lucide-react": "^0.294.0",
+    "react-dom": "^18.2.0",
```
jessepinho, line 30: Required as a peer dependency of `@testing-library/react`.

packages/ui/components/ui/tx/view/address-view.tsx
```diff
@@ -47,7 +49,7 @@ export const AddressViewComponent = ({
         <CopyToClipboard
           text={encoded}
           label={
-            <div>
+            <div data-testid='AddressView__CopyIcon'>
```
jessepinho, line 52: Do we have strong feelings one way or the other about test IDs? I avoid them whenever possible, but couldn't think of a better way to test that the copy icon is present.

grod220: I don't think it's a big deal, but in theory you can just query the container right?

```typescript
const { container } = render(<AddressViewComponent />);
const svgSelector = 'svg.h-4.w-4.text-muted-foreground.hover\\:opacity-50';
const divWithSvg = container.querySelector(`div > ${svgSelector}`);
expect(divWithSvg).toBeDefined() // or something like this
```

grod220 (approved):

packages/ui/components/ui/tx/view/address-view.tsx
jessepinho: Hm... that'd be pretty brittle though. If we end up deciding to go with `h-8` in the future, these tests would break for reasons unrelated to the test's purpose. If it's no big deal to you either way, I'll leave it as-is.

PR #354, opened 2024-01-05
Rename undelegateclaim proving key to convert

cronokirby:
See: https://github.com/penumbra-zone/penumbra/pull/3579


grod220 (approved):
@cronokirby with the `penumbra-zone/web` PR!? 🙌 

Build will fail until the other PR is merged. Please merge that first, re-run the github tasks on this one, and then you should be good to merge here too 👍

cronokirby:
Oh wait, now that we've pinned the tag for proving keys to the old release, this won't work. We should just keep this around and merge when when we update the proving keys.

turbocrime:
part of #357

PR #355, opened 2024-01-05
Add a feedback button to the header of the web app

### LINKED ISSUE #221: Frontend site should steer users to the web-ext-feedback channel
At this stage, getting comments and feedback from users is really valuable. We should add some kind of UI to the frontend site that prompts users for feedback with an invite to the web-ext-feedback channel. 

At some later point we may want to remove this but for now I think it'd be a good way to collect feedback.

jessepinho:
## In this PR
- Add a feedback button to the web app, which links to an invite the `#web-ext-feedback` Discord channel.
- Tweak the layout of the header so that, instead of having a fixed gap between header links, it is responsive to the width of the window.

screencap

Closes #221

### DIFF
--- a/apps/webapp/src/components/header/header.tsx
+++ b/apps/webapp/src/components/header/header.tsx
@@ -14,8 +14,9 @@ import Notifications from './notifications.tsx';
 import { PagePath } from '../metadata/paths.ts';
 import { TabletNavMenu } from './tablet-nav-menu.tsx';
 
-// Infinite-expiry invite link, provided by Henry, to the #web-ext-feedback
-// channel.
+// Infinite-expiry invite link to the #web-ext-feedback channel. Provided by
+// Henry (@hdevalence) and thus tied to his Discord account, so reach out to him
+// if there are any problems with this link.
 const WEB_EXT_FEEDBACK_DISCORD_CHANNEL = 'https://discord.gg/XDNcrhKVwV';
 
 export const Header = () => {
--- a/apps/webapp/src/components/header/header.tsx
+++ b/apps/webapp/src/components/header/header.tsx
@@ -5,6 +5,7 @@ import Notifications from './notifications.tsx';
 import { TabletNavMenu } from './tablet-nav-menu.tsx';
 import { LayoutLoaderResult } from '../layout.tsx';
 import { NetworksPopover } from '@penumbra-zone/ui';
+import { MessageWarningIcon } from '../../icons/message-warning.tsx';
 import { MobileNavMenu } from './mobile-nav-menu.tsx';
 
 export const Header = () => {
@@ -27,18 +28,22 @@ export const Header = () => {
         </Link>
       </div>
       <Navbar />
-      <div className='flex w-full items-center justify-between md:w-auto md:gap-6 xl:gap-4'>
+      <div className='flex w-full items-center justify-between gap-4 md:w-auto md:gap-6 xl:gap-4'>
         <div className='order-1 block md:hidden'>
           <MobileNavMenu />
         </div>
         <TabletNavMenu />
+
+        <div className='order-3 flex items-center justify-center md:order-none'>
+          <MessageWarningIcon />
+        </div>
 
         {result.isInstalled ? (
           <>
             <div className='order-3 flex items-center justify-center md:order-none'>
               <Notifications />
             </div>
-            <div className='order-2 md:order-none'>
+            <div className='order-2 flex grow justify-center md:order-none'>
               <NetworksPopover name={result.chainId} />
             </div>
           </>
--- a/apps/webapp/src/components/header/header.tsx
+++ b/apps/webapp/src/components/header/header.tsx
@@ -21,7 +21,7 @@ export const Header = () => {
   const result = useLoaderData() as LayoutLoaderResult;
 
   return (
-    <header className='z-10 flex w-full flex-col items-center justify-between px-6 md:h-[82px] md:flex-row md:px-12'>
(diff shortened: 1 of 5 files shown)


apps/webapp/src/components/header/header.tsx
```diff
@@ -1,11 +1,21 @@
...
+import {
+  NetworksPopover,
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@penumbra-zone/ui';
 import Notifications from './notifications.tsx';
+import { PagePath } from '../metadata/paths.ts';
 import { TabletNavMenu } from './tablet-nav-menu.tsx';
```
jessepinho, line 15: all of these import changes are just alphabetically ordering them (except for the `Tooltip*` imports)

apps/webapp/src/components/header/header.tsx
```diff
@@ -1,11 +1,21 @@
...
-import { LayoutLoaderResult } from '../layout.tsx';
-import { NetworksPopover } from '@penumbra-zone/ui';
-import { MobileNavMenu } from './mobile-nav-menu.tsx';
+
+const WEB_EXT_FEEDBACK_DISCORD_CHANNEL =
+  'https://discord.com/channels/824484045370818580/1077672871251415141';
```
jessepinho, line 18: Note that this is just the public link of the channel, which I got by right-clicking the channel and choosing "Copy link":
image

If it should be an invite link rather than the public link, someone other than me should create the invite link, since I'm only a contractor.

apps/webapp/src/components/header/header.tsx
```diff
@@ -27,18 +37,36 @@ export const Header = () => {
         </Link>
       </div>
       <Navbar />
-      <div className='flex w-full items-center justify-between md:w-auto md:gap-6 xl:gap-4'>
+      <div className='flex w-full items-center justify-between gap-4 md:w-auto md:gap-6 xl:gap-4'>
```
jessepinho, line 40: We need a small gap for mobile, since we now have two icons right next to each other: the feedback icon and the bell icon

apps/webapp/src/components/header/header.tsx
hdevalence: Try https://discord.gg/XDNcrhKVwV

hdevalence: That link should be an infinite-expiry invite to the channel

jessepinho: Ah, thanks @hdevalence !

apps/webapp/src/components/header/header.tsx
```diff
@@ -1,17 +1,28 @@
...
+// Infinite-expiry invite link, provided by Henry, to the #web-ext-feedback
+// channel.
+const WEB_EXT_FEEDBACK_DISCORD_CHANNEL = 'https://discord.gg/XDNcrhKVwV';
 
 export const Header = () => {
   const result = useLoaderData() as LayoutLoaderResult;
 
   return (
-    <header className='z-10 flex w-full flex-col items-center justify-between px-6 md:h-[82px] md:flex-row md:px-12'>
+    <header className='z-10 flex w-full flex-col items-center justify-between px-6 md:h-[82px] md:flex-row md:gap-12 md:px-12'>
```
jessepinho, line 25: Added a gap so that the nav bar would have some space between it and the other elements on narrower screens

apps/webapp/src/components/header/navbar.tsx
```diff
@@ -7,25 +7,25 @@ export const Navbar = () => {
   const pathname = usePagePath();
 
   return (
-    <nav className='hidden gap-4 xl:flex'>
+    <nav className='hidden gap-4 xl:flex xl:grow xl:justify-between max-w-2xl'>
```
jessepinho, line 10: The changes in this file make the nav bar a bit more flexible for smaller screen sizes.

apps/webapp/src/components/header/header.tsx
grod220: If we don't have a prettier or eslint command to do this automatically, count on this order not being maintained ;)

apps/webapp/src/components/header/header.tsx
```diff
@@ -1,17 +1,28 @@
...
+  TooltipTrigger,
+} from '@penumbra-zone/ui';
 import Notifications from './notifications.tsx';
+import { PagePath } from '../metadata/paths.ts';
 import { TabletNavMenu } from './tablet-nav-menu.tsx';
-import { LayoutLoaderResult } from '../layout.tsx';
-import { NetworksPopover } from '@penumbra-zone/ui';
-import { MobileNavMenu } from './mobile-nav-menu.tsx';
+
+// Infinite-expiry invite link, provided by Henry, to the #web-ext-feedback
```
grod220, line 17: Think we can axe the `provided by Henry` part

grod220 (approved):

apps/webapp/src/components/header/header.tsx
jessepinho: My thinking was that we might want to know whose account this link is tied to in the future, right? If not though, I'll remove that.

grod220: Ah. I'm not sure 🤷‍♂️. Feel free to add that context to the code comment and it should be good to keep the reference to Henry.

jessepinho: OK will do.

Issue #352, opened 2024-01-05
Handle Tendermint's `res.log` property if it exists


jessepinho:
When we [broadcast a transaction](https://github.com/penumbra-zone/web/blob/main/packages/query/src/queriers/tendermint.ts#L29-L30) via `TendermintQuerier#broadcastTx`, we currently just return the transaction hash (via `res.hash`). That means that all the other properties of `res` are lost — including `res.log`, which may contain a useful error or other message.

This ticket is to log `res.log` to the console if it's truthy, and to throw it if it's a known error message. Per @grod220's proposal below:

```ts
// Should add more errors as we discover them
const knownErrors = [
  "proof did not verify",
  "is not a valid field element",
]

  async broadcastTx(params: Uint8Array) {
    const req = new BroadcastTxSyncRequest({ params });
    const { hash, log } = await this.client.broadcastTxSync(req);

    if (log) {
      if (knownErrors.some(e => log.includes(e)) {
        throw new Error(log)
      } else {
        console.error(log)
      }
    }

    return hash;
  }
```

grod220:
It's sort of a sad behavior that if there is a tendermint error, it shows up as a string in that response log versus raising.

I think we should consider having some kind of logic to help work around this. Otherwise, like in your case with the `output proof did not verify` error, it gets swallowed. Proposal:

```typescript
// Should add more errors as we discover them
const knownErrors = [
  "proof did not verify",
  "is not a valid field element",
]

  async broadcastTx(params: Uint8Array) {
    const req = new BroadcastTxSyncRequest({ params });
    const { hash, log } = await this.client.broadcastTxSync(req);
    if (knownErrors.some(e => log.includes(e)) {
       throw new Error(log)
    }
    return hash;
  }

```

jessepinho:
That sounds perfect. I'll add your proposal to the original ticket description so whoever takes it knows what to do with it.

Is `is not a valid field element` an actual error we might get from Tendermint?

grod220:
I took that error from the source code for the verify step in Rust: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/shielded-pool/src/output/proof.rs#L163

PR #357, opened 2024-01-05
Update to testnet-preview-v0.65.0

grod220:
# edit by L

ready for review, you can probably ignore the long thread of progress reports and just go check out the diff and comments there


packages/types/src/proving-keys.ts
```diff
@@ -5,5 +5,5 @@ export const provingKeys = [
   { keyType: 'swap_claim', file: 'swapclaim_pk.bin' },
   { keyType: 'nullifier_derivation', file: 'nullifier_derivation_pk.bin' },
   { keyType: 'delegator_vote', file: 'delegator_vote_pk.bin' },
-  { keyType: 'undelegate_claim', file: 'undelegateclaim_pk.bin' },
+  { keyType: 'undelegate_claim', file: 'convert_pk.bin' },
```
grod220, line 8: See: https://github.com/penumbra-zone/web/pull/354

apps/extension/src/utils/download-proving-keys.ts
```diff
@@ -3,8 +3,9 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { provingKeys } from '@penumbra-zone/types/src/proving-keys';
 
+// TODO: When v0.65.0 launches, change this back to a git tag path
 const githubSourceDir =
-  'https://github.com/penumbra-zone/penumbra/raw/v0.64.1/crates/crypto/proof-params/src/gen/';
+  'https://github.com/penumbra-zone/penumbra/raw/main/crates/crypto/proof-params/src/gen/';
```
grod220, line 8: Unfortunately, we have to temporarily revert this https://github.com/penumbra-zone/web/pull/351

packages/query/src/queriers/tendermint.ts
```diff
@@ -27,6 +27,7 @@ export class TendermintQuerier implements TendermintQuerierInterface {
     // the fullnode", not "wait for the tx to be included on chain.
     const req = new BroadcastTxSyncRequest({ params });
     const res = await this.client.broadcastTxSync(req);
+    console.log(res.log);
```
grod220, line 30: Currently an issue with the effect hash: https://discord.com/channels/824484045370818580/1044457038886998137/1192940867024072795. Only seen when logged here. Relates to: https://github.com/penumbra-zone/web/issues/352

apps/extension/src/utils/download-proving-keys.ts
grod220: Make sure to remove your `/dist/` folder so these re-download

packages/router/src/grpc/view-protocol-server/notes.test.ts
```diff
@@ -274,6 +270,7 @@ describe('Notes request handler', () => {
   });
 });
 
+// TODO: Regenerate new test data, these are not the right types
```
grod220, line 273: Perhaps we should use `new SpendableNoteRecord()` instead of `.fromJson()` as it makes it easier to see which fields are invalid.

packages/query/src/queriers/tendermint.ts
turbocrime: afaict this is only present if there's something wrong - i converted it to a throw, which propagates

turbocrime:
rebased to tip of main

packages/types/src/indexed-db.ts
```diff
@@ -84,7 +84,7 @@ export interface PenumbraDb extends DBSchema {
     value: FmdParameters;
   };
   TRANSACTIONS: {
-    key: NoteSource['inner'];
+    key: CommitmentSource_Transaction['id'];
```
turbocrime, line 87: should this be `TransactionId['inner']`? are these equivalent?

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -255,7 +253,7 @@ describe('IndexedDb', () => {
       await db.saveTransactionInfo(transactionInfo);
 
       const savedTransaction = await db.getTransaction(
-        new NoteSource({ inner: transactionInfo.id!.hash }),
+        new CommitmentSource_Transaction({ id: transactionInfo.id!.inner }),
```
turbocrime, line 256: okay, they are equivalent? is there a reason we're using CommitmentSource_Transaction to index transactions rather than the semantically clearer TransactionId?

hdevalence: indexing by (nullable) transaction ID seems the cleanest, the point of the CommitmentSource is to be able to tell whether funds come from a transaction or not.

eventually we will want to figure out the best way to index other kinds of sources (particularly IBC transfers) but for the moment the tx hash seems like the cleanest thing

turbocrime:
it seems like:
1) scanning works properly
2) building works properly
3) encoding works properly
4) broadcast fails with "binding signature failed to verify: Invalid signature." https://github.com/penumbra-zone/web/blob/2f04e49746dd7cf6af352c583ee3cf84d044303f/packages/query/src/queriers/tendermint.ts#L29-L30

notably, the length of the uint8array going in there has changed

believed to be effect hash related but i don't know enough to investigate

possibly contributing:

* source on SpendableNoteRecord is now a CommitmentSource rather than a NoteSource 
* keyType / key file mapping. the convert pk is still referred to by undelegate_claim but afaict that's correct?
* binary serialization. some of these rust types do manual binary serialization, some of them use prost to do it. confirmed in #360 that our bufbuild binary serialization and rust binary serialization produced identical results for transactions, so probably not. there may be some binary serialization options that tendermint remote is expecting that we've failed to apply

hdevalence:
The error that the binding signature failed to verify is interesting, and suggests something related to the proto encodings.

There are two kinds of signatures use in a Penumbra transaction, spend auth signatures and binding signatures. Spend auth signatures sign over the effect hash, a hash of the transaction's "effecting data", which includes the commitments etc but does not include proofs and signatures. Spend auth signatures are made using (randomizations of) the spending key.

The binding signature, on the other hand, signs over the entire transaction body (the "auth hash" is just the hash of the whole transaction body) and is computed using the sum of all of the actions' value commitments: https://protocol.penumbra.zone/main/protocol/value_commitments.html#binding-signatures The signature is "binding" in the sense that it can only be created with knowledge of the blinding factors for the value commitments for each action, preventing replay of actions across transactions. 

A failure to verify an honestly generated binding signature could be caused in one of three ways:

1. A mismatch in the computation of the key used to create the binding signature. This seems unlikely, because none of this code changed.
2. A mismatch in the computation of the auth hash the binding signature signs. This also seems unlikely, because (I believe) this code also hasn't changed.
3. A mismatch in the encoding of the transaction body fed into the auth hash computation. We need deterministic protobuf encoding (Cosmos ADR027), did lifting this out of wasm have any effect here?

turbocrime:
if deterministic serialization is necessary, that's very likely the problem.

we can replace the encodeTx from wasm but, as i said, i confirmed i was getting identical binary output, and both encodeTx and the buf toBinary were failing the same way.

is ADR027 implemented in the rust repo? i can't find any explicit references to it, and encodeTx from wasm simply used prost for binary serialization

TalDerei:
> is ADR027 implemented in the rust repo?

https://docs.cosmos.network/main/build/architecture/adr-027-deterministic-protobuf-serialization

turbocrime:
interestingly, main is working on the current testnet, using the bufbuild toBinary.

maybe we are getting lucky with the current transaction structure, and bufbuild's default binary writer happens to meet the cosmos rules, which seem mostly common-sense.

here's bufbuild's binary format https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf/src/private/binary-format-proto3.ts

it seems like the most possible thing to fail, if this is the cause, would be default values. what fields could be involved there?

i'll restore using wasm to encode, but that was also broken, and binary-identical.

turbocrime:
proto3 so all fields are packed

TalDerei:
@turbocrime It seems like the [payload](https://github.com/penumbra-zone/web/blob/2b6b7f49e2846a49a9269aa7df7e2ed8d257b5c0/packages/query/src/queriers/tendermint.ts#L25-L26) being passed in has size `2253` bytes, which is smaller than the payload, `2768` bytes, in main. Could this factor in?

turbocrime:
possibly. right now i'm investigating zod schema and possible deserialization failure. do you have an example of a well-formed transaction and a correct encoding

hdevalence:
> is ADR027 implemented in the rust repo? i can't find any explicit references to it, and encodeTx from wasm simply used prost for binary serialization

The prost generated code obeys the ADR027 rules (which are pretty basic stuff like "write fields in numerical order" "always use the shortest varint encoding" "don't use maps" etc).

turbocrime:
i couldn't identify anything specifically wrong with the zod schema, but removing zod results in a successful broadcast.

tx appears on chain and is viewable in webapp, if i navigate to the hash url. but page continues to await confirmation, and it doesn't show up in tx listing.

turbocrime:
Screenshot 2024-01-06 at 22 46 52
yeag

turbocrime:
https://github.com/penumbra-zone/web/blob/2b6b7f49e2846a49a9269aa7df7e2ed8d257b5c0/packages/storage/src/indexed-db/index.ts#L58

possibly contributing to missing tx listing

turbocrime:
> * keyType / key file mapping. the convert pk is still referred to by undelegate_claim but afaict that's correct?

ok maybe https://github.com/penumbra-zone/penumbra/commit/c8c7b41480524c64cd9ed521115b16ee6cf54590

turbocrime:
i think this is going to involve a db schema change so i'm looking into that today

turbocrime:
getting close, tx sources coming out of wasm are malformed, maybe due to idb change. starting work on a patch for penumbra repo to address this. would be cool to pair at some point

Screenshot 2024-01-09 at 13 28 06

turbocrime:
i'm questioning the utility of zod, which actually obscured this failure by silently truncating the incoming objects, and the strange typeless and haphazard serialization we perform at the idb and wasm boundaries. i played with wasm and tsify some this weekend and it was very possible to emit types from wasm - we should be typing the wasm interface this way, and then these disconnects would be immediately obvious. i didn't try importing ts types to rust but it does seem like that might be a good way to define idb types.

i am going to go for the minimal changeset to achieve version bump. but soon we should make an effort to define these interfaces in a single location, instead of writing pb types, writing types in the rust repo, writing duplicate rust types in the wasm crate, writing storage types for idb in both rust and ts, writing the wasm crate types again in typescript, and also duplicating a bunch of those with zod.

turbocrime:
> getting close, tx sources coming out of wasm are malformed, maybe due to idb change. starting work on a patch for penumbra repo to address this. would be cool to pair at some point

ah, maybe this is intentional, i guess some tx are supposed to have no id

https://github.com/penumbra-zone/penumbra/blob/29b7cb2afde3a664bcbcbbe33e544b43a802ee83/crates/bin/pcli/src/command/view/balance.rs#L98-L99

turbocrime:
current status:
* transactions can be built and broadcast
* transactions on the chain cannot be detected, because blocks are not processed correctly in wasm
* incoming compact block commitment source ids need to be 'rehydrated'

turbocrime:
current status:

paired with @erwanor to understand flow of commitments/nullifiers/ids and i think i have a way to match nullifiers, rehydrate sources, and detect broadcasts in typescript. draft for review soon

turbocrime:
Screenshot 2024-01-12 at 14 04 58

turbocrime:
need to clean up but this can build, broadcast, and detect broadcast on tip of main penumbra

turbocrime:
cleanup todo

- split resolveNullifiers into smaller steps
  1. resolve nullifiers
      - identify items present in idb to which nullifiers apply
      - if those items' source commitment is a transaction, check presence of id
      - if id is not present, collect into list
  2. fetch fullblock?
      - if wasm advises
      - if swap or note records present on this block
      - if resolved nullifiers for any record with no source id
      - log any disagreement
  3. assign sources from fullblock
      - resolved swap/notes nullifiers should match tx action nullifiers on this block
- restore zod
- update zod schema
- minimize idb schema change
- write idb migration?
- remove logspam everywhere

hdevalence:
I don't have full context but I'm not sure writing migrations is necessary. Since all the data is recoverable via syncing, rather than spending time writing a migration it should be possible to just clear the cache and resync with the new schema

turbocrime:
i just assumed it was standard to write a migration when changing schema. but now that you mention it i don't see any migrations around, and there's nothing we might lose

turbocrime:
okay, this *was* working, but something's strange now

block 1254 contains something interesting for the test wallet, it was one of the earlier tx built with the new wasm so maybe something funny happened

i started re-applying zod schema and at some point it broke the tx builds - that same 'binding signature failed' error is back

but, as i start to remove zod again, eventually if i try to make a transaction, it complains that a nullifier is spent twice, initially on block 1254

im sleepy and its late and the last few commits are a mess

turbocrime:
the tx on 1254. maybe nothing is up with it but depending on block processor it sometimes tries to spend the relevant nullifier twice

```json
{
  "height": "1254",
  "id": {
    "inner": "m2IguTPFA+uEZOsgWcatcmV7zVeAE+MssxB76mlar4I="
  },
  "transaction": {
    "body": {
      "actions": [
        {
          "output": {
            "body": {
              "notePayload": {
                "noteCommitment": {
                  "inner": "PuqvSawII53FcB8PPirwK6NHGlppkdtpNE5rbmPTmxA="
                },
                "ephemeralKey": "JIgzjDrK9Ucyy0fwwzI/PFUhC5o/nFFEksorTzZ/mQI=",
                "encryptedNote": {
                  "inner": "emYaGArj/Hy0tQjRTt/QaLJ9s6yC1tePqFyz6Bq5/uUELZ7Y9YH5FWetps+8N4N/LIRqFSSeIeWeL/KSebdqta1vpH+gJSspmoNaAQu4xHUUgljIspYN2WTAMdVfIufw/Mqji570AZaAkrtq0spFCvSbaztrCpHd99S/+0dViQf0eC8+cYiYB2CfFEKkbUKwe+40kS9M+jOf4K1ogPRn60mfj95N4yaG72gDLfMlZhE="
                }
              },
              "balanceCommitment": {
                "inner": "fgN1sFAYz3A4kay/BdYjdEyEJtLcpyrJd9AUeXe/Vg0="
              },
              "wrappedMemoKey": "TcuVTlestzR4hQmux7/Ro2vTvvDjX7qJyLNa5F4R6l68vMM97Vinlv7h5y4fXrO8",
              "ovkWrappedKey": "EvgZ3LTn7gPxQ8jSK6GdsXhyP9nr8RKErKGpUJ/lgfuAud6QuA8z26c4iHwyHYCo"
            },
            "proof": {
              "inner": "0sKx90nwDQoASJiaQQhBFjJS8TjvgxQ866r5fWOPfWBNgNzh9fl+s+baUZhPE1gA6WKQxNB/3eAS7hRyeTImGs65FKkYFsrEzj/gjGxxZ23TkgtOUE3LYWmo2nBLJJIAzPx623PC2fJ2B2zxSnVNop4G7wlI4zssMluUTC5wh0TrFWD4jnL0gcJ53+v0SJqAe4YxuvvKrIaxNPBLiRTneucgeH4rV5H1es5hC3Yvd8AzKnD18/UYrmCRzKU5Y34B"
            }
          }
        },
        {
          "spend": {
            "body": {
              "balanceCommitment": {
                "inner": "atcCoway2xt+qGOXWqoH82gscukNtwKnWxHOXW2JKQA="
              },
              "nullifier": {
                "inner": "JkbS8lQEHrhez9RyWhQjZojkPXpAUnwjjXtE/S5T4Ak="
              },
              "rk": {
                "inner": "xORRwwdWxASC6eTL+laRaB4rd3YoTQTYS1WqlIW3/wE="
              }
            },
            "authSig": {
              "inner": "MNDlVxO/OnbppplUM7AvA+dOgXNQv0SV8rm9hyEFiAy4PdZoksmuredi7ixghb+U26sVL1Kc3wlfiWFD2iP2AQ=="
            },
            "proof": {
              "inner": "rV4nrGsNYASEHxghCGeEpFa1uB9uHEByjUWj93JT9zOVs65NZxtdWJkyKtPaWyMAdSNzQIu5L7ofdF1giUb6RHHiaBwQabOhVICNXeLxuk92kgIe9TEjNNFum5rsf5QAwEZInCqcxA0VBW8+G/NBOgW5QbX0Ah6IO560HCo/4Bm+9i+o4MxAyA3gGtXYgUiAaCq0Zy5i3vGnYxbLO4Dhg/69KhhzLZMs3RpMREGhC929v+BBJTGwXxynhinD61cA"
            }
          }
        },
        {
          "spend": {
            "body": {
              "balanceCommitment": {
                "inner": "dBjc4BjIGnQI8BC1wWVQbiUYZNE8u+Trx9WCacDJVBE="
              },
              "nullifier": {
                "inner": "23iVZL9Spqo07OB9a9CFP0ScHs3Lu5N0WgR60WZCQAw="
              },
              "rk": {
                "inner": "WlqRtOlWi7SR2Jw9Pciq73l36Zhz+XPUgOMmbz3+jAI="
              }
            },
            "authSig": {
              "inner": "niMSUhoRkB5eV1gq+0Bac7XLjss3hnhsBGFHdJkhuAzW54SEDeyrGrV30ZLsFNww7WG+AcVmysKsXsDGI0gwAA=="
            },
            "proof": {
              "inner": "smKin/8aCJRF8WAyW+THg39vX/8goAEPAT7JJM+nxGjzFvvdYa4tUoqgnbq8sbiAdnScT5ODF0+4gnvrF2Fbr1UuT++hjHHxEXwr0A9R1wCc+g3QAtEdcnstwKvCGAEBeY9vKAdhCDuAckCQOB/KAWyGEj/HvFfyVmGZLB3eTbXxvZ2/mygaIkYwTm+c3fsAYi7EYhG1HmslKlFiDrep2fQqOVhraTKE3I6cwRGRYB90fscZbVWRryMycexPXnaA"
            }
          }
        },
        {
          "output": {
            "body": {
              "notePayload": {
                "noteCommitment": {
                  "inner": "Qb+bmkBvpTjVpMbPQ50BUoeQ1SmO1xoLAXduro4o8wc="
                },
                "ephemeralKey": "TMLePwzW/z0Qe4n1HQkqLjahTP3EvXfQJ6Kun/nwxQk=",
                "encryptedNote": {
                  "inner": "VICzCF8eu0/tI5Op4yY84l8rhvRjw9VmHYbO7FUBCQFp+q5dhcr4gPLUO/4KfWtIvVRDW48FbmQ3op04KE8O1fxBB35IE6N3lsi896A5RWEMftW9oVpzZdOWFqV4hymSnCoY8dgVctI2dU0MT63QN0xDSqMJoU9e0JcWDIFX/4cYAvmKrebd4Q4+qY/Z6EcpW59qwv4HNpB8rG/11lGI+fGvDgbI+DGSODlWS0cKno8="
                }
              },
              "balanceCommitment": {
                "inner": "Ci/36nqINZKNOWbHwFlEA2nywKa0QIV45SLdcacKhw8="
              },
              "wrappedMemoKey": "kr+PoJFXx5kW8mZ3qAlLPXU1UAYBT+fTE8AuQH65CEzfS1GXzh8V4428Y5ukCYEx",
              "ovkWrappedKey": "j+dUjplkWXkrCcA0pAPf2QmsjPllkxp6zXu4FZkjhZuBN3hSqZHLvzOrH1Y51mHZ"
            },
            "proof": {
              "inner": "HHoOtJsXWx0iPImpaEl5o0VyYSEcpnhderv9YR1uq12lf59PhLf+512p9e98aacBubCST4ei1GRWgTeVjMuOniXSHYKkgcCy7E+tnanuYEifz3K6vEJr+v7WPTMbkaEBLkFPPoMoqezqTSuZ/kbyH8Lmv3FS3ZFtdg+zTRcloKmN73MJ5kJKrI/v1nvVIkSBTGL6ZyVgwbcToLDS8nPBXjWpc1/TxGtM+mZVLj/JSZwVrOQiYDMG16rqXhjdfGyB"
            }
          }
        }
      ],
      "transactionParameters": {
        "chainId": "penumbra-testnet-deimos-a46a0dfd",
        "fee": {
          "amount": {}
        }
      },
      "detectionData": {
        "fmdClues": [
          {
            "inner": "dJNBt0+23wBxo94mb8vTkkxrzjVlXosAqBSARPA9Vg4UpbYtt6Ll6QOGtJDUE869rFUuLg351R2lfuuD6MSAAwAAAAA="
          },
          {
            "inner": "KMpd1GfpoGSTEHDakphPqIU01tjHRqgXeEqNbqIiXgzZ0oqu33909PaxmAI1gRkelKekn71xAYDffPClxGXwAgAAAAA="
          }
        ]
      },
      "memo": {
        "inner": "1MErAgZfbqZgJ4Ch/9kWt4CMbQKwtQj9xiXrkymYt+uQref6PLzWJSMYhFLrGA0fbDaZ1BfqSATbnVt/Id3W7QFioDOY1OeRAVGh6u+hd0z7ZqQ6GSQcQweVvcZmuZou4Oqjeol0sZdSQHftuD3VcMW/UUaBZTH58pFogaR/QS97A3bBhmw/95gqVI/V+DYxiq06s6OQCO/5j8bb09B2Z60lVPf5CHSEFsrhHkQsz00AFZEVLbLy6Bj39s19R1/8Fwsggo3TCyjwCLX5yD5+/YVVIiGYeSF3rsJG6bC3vd3jq1KVTJ9asJNMx64puJcwRannLMf+qCRNTxd0WdQSIs8iKuDG2SQmT4pcWyb0Q13g06mh5hKpYNqZUd/pJmO/FG3MB3k+DQUewTZnOpzy+Tq/DOjyjoglUpKfZY8ZY1nzDeY1ZgKgSLLyqGGhqLF4OVD25zEvQgNyuaQdajwpKes7cE0rQUV5m8uOl0bxen6PEHfufUfOrOIy0o+RoD/uMC5Aev04HEQdyeG8ghoWq88dT5tnI+L8tXIfUMiKjOC2PPoeWXDh+qDQMLHTvT65eVBtdC/dBxwkFodd7Ba9YRcfeIcl6a6tudZdRkVhdHVxqOEDhPYZVN0lBD9tq7iziBWBEpcxpl3gryaYIKEb/qszTz9gnqYGq9t1A9+9uTjp89DeKcyDNhKKpD6jKvy3"
      }
    },
    "bindingSig": {
      "inner": "aMh/dvCYW80AqgJp8Bne/YO3tu4UwcFVVm80BE2hEwPfNGQv3klMWc3EAv+tuB/sG3E6v45nWzLdD6CcSXSUAw=="
    },
    "anchor": {
      "inner": "PDu/xV1Bf3Aqqg48f39hwBQm3q+8Qb3KKGRWKmbZWAM="
    }
  },
  "perspective": {
    "payloadKeys": [
      {
        "payloadKey": {
          "inner": "J/m+DiTM1lr5nIaNqPcWRq4o1BOOKRMzFcKfADxdyC0="
        },
        "commitment": {
          "inner": "Qb+bmkBvpTjVpMbPQ50BUoeQ1SmO1xoLAXduro4o8wc="
        }
      },
      {
        "payloadKey": {
          "inner": "HU5bdnhu/p10M3uRJ6M70Yx3d7X4CLRXyU9VaVB/wn0="
        },
        "commitment": {
          "inner": "PuqvSawII53FcB8PPirwK6NHGlppkdtpNE5rbmPTmxA="
        }
      }
    ],
    "spendNullifiers": [
      {
        "nullifier": {
          "inner": "JkbS8lQEHrhez9RyWhQjZojkPXpAUnwjjXtE/S5T4Ak="
        },
        "note": {
          "value": {
            "amount": {
              "lo": "1001882102603448320",
              "hi": "27105"
            },
            "assetId": {
              "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
            }
          },
          "rseed": "mAU06Am3l4IK16V2UfOY/E6NitYOXqvCgG4vCKx1l6o=",
          "address": {
            "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
          }
        }
      },
      {
        "nullifier": {
          "inner": "23iVZL9Spqo07OB9a9CFP0ScHs3Lu5N0WgR60WZCQAw="
        },
        "note": {
          "value": {
            "amount": {
              "lo": "4877702122287661056",
              "hi": "27159"
            },
            "assetId": {
              "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
            }
          },
          "rseed": "mtzHWKVJy4DAEfEYIJ+KnMIq5dJvIuof2ZJiHIKCs44=",
          "address": {
            "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
          }
        }
      }
    ],
    "addressViews": [
      {
        "visible": {
          "address": {
            "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
          },
          "index": {},
          "walletId": {
            "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
          }
        }
      }
    ],
    "denoms": [
      {
        "denomUnits": [
          {
            "denom": "test_usd",
            "exponent": 18
          },
          {
            "denom": "wtest_usd"
          }
        ],
        "base": "wtest_usd",
        "display": "test_usd",
        "penumbraAssetId": {
          "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
        }
      }
    ],
    "transactionId": {
      "inner": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
    }
  },
  "view": {
    "bodyView": {
      "actionViews": [
        {
          "output": {
            "visible": {
              "output": {
                "body": {
                  "notePayload": {
                    "noteCommitment": {
                      "inner": "PuqvSawII53FcB8PPirwK6NHGlppkdtpNE5rbmPTmxA="
                    },
                    "ephemeralKey": "JIgzjDrK9Ucyy0fwwzI/PFUhC5o/nFFEksorTzZ/mQI=",
                    "encryptedNote": {
                      "inner": "emYaGArj/Hy0tQjRTt/QaLJ9s6yC1tePqFyz6Bq5/uUELZ7Y9YH5FWetps+8N4N/LIRqFSSeIeWeL/KSebdqta1vpH+gJSspmoNaAQu4xHUUgljIspYN2WTAMdVfIufw/Mqji570AZaAkrtq0spFCvSbaztrCpHd99S/+0dViQf0eC8+cYiYB2CfFEKkbUKwe+40kS9M+jOf4K1ogPRn60mfj95N4yaG72gDLfMlZhE="
                    }
                  },
                  "balanceCommitment": {
                    "inner": "fgN1sFAYz3A4kay/BdYjdEyEJtLcpyrJd9AUeXe/Vg0="
                  },
                  "wrappedMemoKey": "TcuVTlestzR4hQmux7/Ro2vTvvDjX7qJyLNa5F4R6l68vMM97Vinlv7h5y4fXrO8",
                  "ovkWrappedKey": "EvgZ3LTn7gPxQ8jSK6GdsXhyP9nr8RKErKGpUJ/lgfuAud6QuA8z26c4iHwyHYCo"
                },
                "proof": {
                  "inner": "0sKx90nwDQoASJiaQQhBFjJS8TjvgxQ866r5fWOPfWBNgNzh9fl+s+baUZhPE1gA6WKQxNB/3eAS7hRyeTImGs65FKkYFsrEzj/gjGxxZ23TkgtOUE3LYWmo2nBLJJIAzPx623PC2fJ2B2zxSnVNop4G7wlI4zssMluUTC5wh0TrFWD4jnL0gcJ53+v0SJqAe4YxuvvKrIaxNPBLiRTneucgeH4rV5H1es5hC3Yvd8AzKnD18/UYrmCRzKU5Y34B"
                }
              },
              "note": {
                "value": {
                  "knownDenom": {
                    "amount": {
                      "lo": "1000000000000000000"
                    },
                    "denom": {
                      "denomUnits": [
                        {
                          "denom": "test_usd",
                          "exponent": 18
                        },
                        {
                          "denom": "wtest_usd"
                        }
                      ],
                      "base": "wtest_usd",
                      "display": "test_usd",
                      "penumbraAssetId": {
                        "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
                      }
                    }
                  }
                },
                "rseed": "1jG59yLaE7oAci2+aVMeBjHLDv4tEdBC/usdjisfMUA=",
                "address": {
                  "visible": {
                    "address": {
                      "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
                    },
                    "index": {},
                    "walletId": {
                      "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
                    }
                  }
                }
              },
              "payloadKey": {
                "inner": "DPWYHMZ/8t9qzeCkNKdAHTb7LTYO7gn4oqkqOTLy9cA="
              }
            }
          }
        },
        {
          "spend": {
            "visible": {
              "spend": {
                "body": {
                  "balanceCommitment": {
                    "inner": "atcCoway2xt+qGOXWqoH82gscukNtwKnWxHOXW2JKQA="
                  },
                  "nullifier": {
                    "inner": "JkbS8lQEHrhez9RyWhQjZojkPXpAUnwjjXtE/S5T4Ak="
                  },
                  "rk": {
                    "inner": "xORRwwdWxASC6eTL+laRaB4rd3YoTQTYS1WqlIW3/wE="
                  }
                },
                "authSig": {
                  "inner": "MNDlVxO/OnbppplUM7AvA+dOgXNQv0SV8rm9hyEFiAy4PdZoksmuredi7ixghb+U26sVL1Kc3wlfiWFD2iP2AQ=="
                },
                "proof": {
                  "inner": "rV4nrGsNYASEHxghCGeEpFa1uB9uHEByjUWj93JT9zOVs65NZxtdWJkyKtPaWyMAdSNzQIu5L7ofdF1giUb6RHHiaBwQabOhVICNXeLxuk92kgIe9TEjNNFum5rsf5QAwEZInCqcxA0VBW8+G/NBOgW5QbX0Ah6IO560HCo/4Bm+9i+o4MxAyA3gGtXYgUiAaCq0Zy5i3vGnYxbLO4Dhg/69KhhzLZMs3RpMREGhC929v+BBJTGwXxynhinD61cA"
                }
              },
              "note": {
                "value": {
                  "knownDenom": {
                    "amount": {
                      "lo": "1001882102603448320",
                      "hi": "27105"
                    },
                    "denom": {
                      "denomUnits": [
                        {
                          "denom": "test_usd",
                          "exponent": 18
                        },
                        {
                          "denom": "wtest_usd"
                        }
                      ],
                      "base": "wtest_usd",
                      "display": "test_usd",
                      "penumbraAssetId": {
                        "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
                      }
                    }
                  }
                },
                "rseed": "mAU06Am3l4IK16V2UfOY/E6NitYOXqvCgG4vCKx1l6o=",
                "address": {
                  "visible": {
                    "address": {
                      "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
                    },
                    "index": {},
                    "walletId": {
                      "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
                    }
                  }
                }
              }
            }
          }
        },
        {
          "spend": {
            "visible": {
              "spend": {
                "body": {
                  "balanceCommitment": {
                    "inner": "dBjc4BjIGnQI8BC1wWVQbiUYZNE8u+Trx9WCacDJVBE="
                  },
                  "nullifier": {
                    "inner": "23iVZL9Spqo07OB9a9CFP0ScHs3Lu5N0WgR60WZCQAw="
                  },
                  "rk": {
                    "inner": "WlqRtOlWi7SR2Jw9Pciq73l36Zhz+XPUgOMmbz3+jAI="
                  }
                },
                "authSig": {
                  "inner": "niMSUhoRkB5eV1gq+0Bac7XLjss3hnhsBGFHdJkhuAzW54SEDeyrGrV30ZLsFNww7WG+AcVmysKsXsDGI0gwAA=="
                },
                "proof": {
                  "inner": "smKin/8aCJRF8WAyW+THg39vX/8goAEPAT7JJM+nxGjzFvvdYa4tUoqgnbq8sbiAdnScT5ODF0+4gnvrF2Fbr1UuT++hjHHxEXwr0A9R1wCc+g3QAtEdcnstwKvCGAEBeY9vKAdhCDuAckCQOB/KAWyGEj/HvFfyVmGZLB3eTbXxvZ2/mygaIkYwTm+c3fsAYi7EYhG1HmslKlFiDrep2fQqOVhraTKE3I6cwRGRYB90fscZbVWRryMycexPXnaA"
                }
              },
              "note": {
                "value": {
                  "knownDenom": {
                    "amount": {
                      "lo": "4877702122287661056",
                      "hi": "27159"
                    },
                    "denom": {
                      "denomUnits": [
                        {
                          "denom": "test_usd",
                          "exponent": 18
                        },
                        {
                          "denom": "wtest_usd"
                        }
                      ],
                      "base": "wtest_usd",
                      "display": "test_usd",
                      "penumbraAssetId": {
                        "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
                      }
                    }
                  }
                },
                "rseed": "mtzHWKVJy4DAEfEYIJ+KnMIq5dJvIuof2ZJiHIKCs44=",
                "address": {
                  "visible": {
                    "address": {
                      "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
                    },
                    "index": {},
                    "walletId": {
                      "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
                    }
                  }
                }
              }
            }
          }
        },
        {
          "output": {
            "visible": {
              "output": {
                "body": {
                  "notePayload": {
                    "noteCommitment": {
                      "inner": "Qb+bmkBvpTjVpMbPQ50BUoeQ1SmO1xoLAXduro4o8wc="
                    },
                    "ephemeralKey": "TMLePwzW/z0Qe4n1HQkqLjahTP3EvXfQJ6Kun/nwxQk=",
                    "encryptedNote": {
                      "inner": "VICzCF8eu0/tI5Op4yY84l8rhvRjw9VmHYbO7FUBCQFp+q5dhcr4gPLUO/4KfWtIvVRDW48FbmQ3op04KE8O1fxBB35IE6N3lsi896A5RWEMftW9oVpzZdOWFqV4hymSnCoY8dgVctI2dU0MT63QN0xDSqMJoU9e0JcWDIFX/4cYAvmKrebd4Q4+qY/Z6EcpW59qwv4HNpB8rG/11lGI+fGvDgbI+DGSODlWS0cKno8="
                    }
                  },
                  "balanceCommitment": {
                    "inner": "Ci/36nqINZKNOWbHwFlEA2nywKa0QIV45SLdcacKhw8="
                  },
                  "wrappedMemoKey": "kr+PoJFXx5kW8mZ3qAlLPXU1UAYBT+fTE8AuQH65CEzfS1GXzh8V4428Y5ukCYEx",
                  "ovkWrappedKey": "j+dUjplkWXkrCcA0pAPf2QmsjPllkxp6zXu4FZkjhZuBN3hSqZHLvzOrH1Y51mHZ"
                },
                "proof": {
                  "inner": "HHoOtJsXWx0iPImpaEl5o0VyYSEcpnhderv9YR1uq12lf59PhLf+512p9e98aacBubCST4ei1GRWgTeVjMuOniXSHYKkgcCy7E+tnanuYEifz3K6vEJr+v7WPTMbkaEBLkFPPoMoqezqTSuZ/kbyH8Lmv3FS3ZFtdg+zTRcloKmN73MJ5kJKrI/v1nvVIkSBTGL6ZyVgwbcToLDS8nPBXjWpc1/TxGtM+mZVLj/JSZwVrOQiYDMG16rqXhjdfGyB"
                }
              },
              "note": {
                "value": {
                  "knownDenom": {
                    "amount": {
                      "lo": "4879584224891109376",
                      "hi": "54264"
                    },
                    "denom": {
                      "denomUnits": [
                        {
                          "denom": "test_usd",
                          "exponent": 18
                        },
                        {
                          "denom": "wtest_usd"
                        }
                      ],
                      "base": "wtest_usd",
                      "display": "test_usd",
                      "penumbraAssetId": {
                        "inner": "reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
                      }
                    }
                  }
                },
                "rseed": "26VWYad+8G4M9KXeFSsovoZOpXTOtAdhp3ofqkd/1ec=",
                "address": {
                  "visible": {
                    "address": {
                      "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
                    },
                    "index": {},
                    "walletId": {
                      "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
                    }
                  }
                }
              },
              "payloadKey": {
                "inner": "DPWYHMZ/8t9qzeCkNKdAHTb7LTYO7gn4oqkqOTLy9cA="
              }
            }
          }
        }
      ],
      "transactionParameters": {
        "chainId": "penumbra-testnet-deimos-a46a0dfd",
        "fee": {
          "amount": {}
        }
      },
      "detectionData": {
        "fmdClues": [
          {
            "inner": "dJNBt0+23wBxo94mb8vTkkxrzjVlXosAqBSARPA9Vg4UpbYtt6Ll6QOGtJDUE869rFUuLg351R2lfuuD6MSAAwAAAAA="
          },
          {
            "inner": "KMpd1GfpoGSTEHDakphPqIU01tjHRqgXeEqNbqIiXgzZ0oqu33909PaxmAI1gRkelKekn71xAYDffPClxGXwAgAAAAA="
          }
        ]
      },
      "memoView": {
        "visible": {
          "ciphertext": {
            "inner": "1MErAgZfbqZgJ4Ch/9kWt4CMbQKwtQj9xiXrkymYt+uQref6PLzWJSMYhFLrGA0fbDaZ1BfqSATbnVt/Id3W7QFioDOY1OeRAVGh6u+hd0z7ZqQ6GSQcQweVvcZmuZou4Oqjeol0sZdSQHftuD3VcMW/UUaBZTH58pFogaR/QS97A3bBhmw/95gqVI/V+DYxiq06s6OQCO/5j8bb09B2Z60lVPf5CHSEFsrhHkQsz00AFZEVLbLy6Bj39s19R1/8Fwsggo3TCyjwCLX5yD5+/YVVIiGYeSF3rsJG6bC3vd3jq1KVTJ9asJNMx64puJcwRannLMf+qCRNTxd0WdQSIs8iKuDG2SQmT4pcWyb0Q13g06mh5hKpYNqZUd/pJmO/FG3MB3k+DQUewTZnOpzy+Tq/DOjyjoglUpKfZY8ZY1nzDeY1ZgKgSLLyqGGhqLF4OVD25zEvQgNyuaQdajwpKes7cE0rQUV5m8uOl0bxen6PEHfufUfOrOIy0o+RoD/uMC5Aev04HEQdyeG8ghoWq88dT5tnI+L8tXIfUMiKjOC2PPoeWXDh+qDQMLHTvT65eVBtdC/dBxwkFodd7Ba9YRcfeIcl6a6tudZdRkVhdHVxqOEDhPYZVN0lBD9tq7iziBWBEpcxpl3gryaYIKEb/qszTz9gnqYGq9t1A9+9uTjp89DeKcyDNhKKpD6jKvy3"
          },
          "plaintext": {
            "returnAddress": {
              "visible": {
                "address": {
                  "inner": "r7ae/+8Q9d3QdaAs66/GAAbYBo/Am59nYWeIBU7REchE3LYtFPa1EHW2Lo1KZcRWuXzO/cM54CLSFnv2iArQnxjrlJnTB4nGnuLdFtCY9vc="
                },
                "index": {},
                "walletId": {
                  "inner": "8rEdDhdZani3OEcE5ETRDR9pGEUl6pL/jmiOAmd4KwM="
                }
              }
            },
            "text": "this shit better work"
          }
        }
      }
    },
    "bindingSig": {
      "inner": "aMh/dvCYW80AqgJp8Bne/YO3tu4UwcFVVm80BE2hEwPfNGQv3klMWc3EAv+tuB/sG3E6v45nWzLdD6CcSXSUAw=="
    },
    "anchor": {
      "inner": "PDu/xV1Bf3Aqqg48f39hwBQm3q+8Qb3KKGRWKmbZWAM="
    }
  }
}
```

turbocrime:
ok, working again. bringing zod back brought back failures. zod schema as used do not identify problems - they conceal them - so i removed that validation rather than update it.

Valentine1898:
A little outside perspective
- Because of refactoring `block-processor/index.ts` -> `block-processor.ts` it is inconvenient to track changes, maybe it would be better to make refactoring a separate PR? 
- It seems that when wasm creates advices, we don't need to fetch fullblock since we will do it when wasm using advices creates a new `SpendableNoteRecord`.  In practice, this means that when we swap, wasm saves advices, and when we swap claim wasm using advices creates a `SpendableNoteRecord`. If my reasoning about advices is correct, then we don't need to make any changes to wasm. 
- We just need to increase the version of `indexed-db` to make correct changes to its structure. Our migration between versions is always the same - we just delete all tables and create them anew (and start synchronization from the beginning)

turbocrime:
current wip pushed

refactor of block processor is generally necessary, because the old block processor's structure was based on assumptions about information available at each step, which has changed in a fundamental way now that records no longer identify source tx id

achieved update without changing wasm

bumping idb version because
1) the internal structure of SpendableNoteRecord has changed
2) some records, though json serialized, indicated they contained a full type
3) use of uint8array stringification was inconsistent

working more on idb schema today, to figure out if everything is correct there

Valentine1898:
Changing the internal structure of stored data usually does not require bumping idb version, but it does require data cleanup and re-synchronization. 
Only changing table names, keys and indexes requires bumping idb version. In this PR bumping is unavoidable because the `keyPath` for the `TRANSACTIONS` table has changed

packages/types/src/proving-keys.ts
turbocrime: i haven't been able to identify any errors, but: is it correct that the filename changes, but keyType remains the same?

TalDerei: keyType should be named `convert` according to the changes in https://github.com/penumbra-zone/penumbra/pull/3579/files

turbocrime:
remaining: updated tests and test data

packages/router/src/grpc/view-protocol-server/notes.test.ts
turbocrime: i think using the constructor would be worse, unless we have a systematic way to create data. many of these fields are binary. i was able to just update the json

packages/query/src/block-processor.ts
turbocrime, line 1: largest change of this PR is a refactored block processor. old block processor made assumptions about available information in its structure and control flow that were no longer valid, and it was necessary to refactor.

i condensed it to one file and tried to make the flow obvious. many comments in the main syncAndStore method, and sub methods

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,371 @@
...
+        Array.from(sourcelessRecords.values()).map(x => x.toJsonString()),
+      );
+
+      // probably scanning too often?
+      // are some of these equivalent?
+      const fullScanReasons =
+        Boolean(compactBlock.nullifiers.length) ||
+        Boolean(compactScan?.newNotes.length) ||
+        Boolean(compactScan?.newSwaps.length) ||
+        Boolean(sourcelessRecords.size);
```
turbocrime, line 161: confirm?

packages/query/src/queriers/cnidarium.ts
turbocrime, line 1: maybe a rename wasn't strictly necessary lol

packages/router/src/grpc/view-protocol-server/broadcast-transaction.ts
```diff
@@ -13,20 +12,18 @@ export const broadcastTransaction: Impl['broadcastTransaction'] = async (req, ct
...
 
-  // Wait until our DB encounters a new note with this hash
+  // Wait until DB records a new transaction with this id
   if (req.awaitDetection) {
-    for await (const update of subscription) {
-      const note = SpendableNoteRecord.fromJson(update.value);
-      if (note.source?.equals(new NoteSource({ inner: hash }))) break;
+    for await (const { value } of subscription) {
+      const update = TransactionInfo.fromJson(value);
+      if (id.equals(update.id)) return { id, detectionHeight: update.height };
```
turbocrime, line 24: no longer awaiting a note, but maybe we could go back. it now awaits write of the full transactioninfo to the db which occurs moments later.

would a spendablenoterecord with a genesis source fail to be detected by the old spendablenote await?

packages/types/src/asset.ts
```diff
@@ -29,13 +27,6 @@ export interface Asset {
   icon?: string;
 }
 
-export const WasmDenomMetadataSchema = z.object({
-  base: z.string(),
-  denomUnits: z.unknown(),
-  penumbraAssetId: InnerBase64Schema,
-  display: z.string(),
-});
```
turbocrime, line 37: just using DenomMetadata.fromJson alone now, which provides more informative validation

packages/types/src/indexed-db.ts
turbocrime, line 1: switched `Jsonified` to `JsonValue` which is synonymous because `Jsonified` does not actually provide any type information. and using `Base64Str` as keys doesn't actually enforce anything, so it's just `string` now

other than that, mostly edited comments. notable actual changes:

`TRANSACTIONS` table is now `TRANSACTION_INFO` to indicate the actual type contained

future, we should be storing `PlainMessage<T>` and we should be using byte arrays to index where possible, not strings.

packages/types/src/state-commitment-tree.ts
```diff
@@ -52,31 +52,15 @@ export const SctUpdatesSchema = z.object({
...
 
-export const parseScanResult = (r: RawScanResult): ScanResult => {
-  return {
-    height: BigInt(r.height), // TODO: Should see if wasm crate can pass bigint instead
-    sctUpdates: r.sct_updates,
-    newNotes: r.new_notes.map(n => SpendableNoteRecord.fromJsonString(JSON.stringify(n))),
-    newSwaps: r.new_swaps.map(s => SwapRecord.fromJsonString(JSON.stringify(s))),
-  };
-};
-
```
turbocrime, line 79: just performing parsing in flushUpdates now

Valentine1898:
I have tested different transaction types, and noticed that extension does not save incoming transactions
1. List of transactions in extension 
image
2. List of transactions in pcli 
image
3. Indexed-db dump 
image
4. pcli sqllite dump 
image

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+      // - decrypts new swaps
+      // - updates idb with advice
+      //const shouldStoreProgress = await this.viewServer.scanBlock(compactBlock);
+      const wasmWantsFlush = await this.viewServer.scanBlock(compactBlock);
+
+      // flushing is slow, avoid it until
+      // - wasm says
+      // - every 1000th block
+      // - every block at tip
+      const flushReasons =
```
jessepinho, line 114: Nitpick, but this variable is a boolean representing whether or not to flush — not a variable containing the reasons for flushing. I'd rename it to something like `should Flush`.

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+      );
+
+      console.log(
+        'sourcelessRecords',
+        Array.from(sourcelessRecords.values()).map(x => x.toJsonString()),
+      );
+
+      // probably scanning too often?
+      // are some of these equivalent?
+      const fullScanReasons =
```
jessepinho, line 157: Same thing here — I'd rename to `shouldRunFullScan` or similar

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+      }
+
+      // - nullifiers on this block may match stored notes or swaps
+      // - queries idb for note/swap records
+      // - updates resolved height of record
+      // - returns any touched records with unknown source
+      // identified records may be old, or they may have just landed. wasm
+      // should flush every time we encounter new records, so they are ready in
+      // idb for this resolution.
+      const sourcelessRecords = await this.resolveNullifiers(
```
Valentine1898, line 145: `resolveNullifiers()` should not return `sourcelessRecords`, only a list of `spent nullifiers`. For notes that we have spent, we should not override source 
https://github.com/penumbra-zone/penumbra/blob/2b0408879aaef969aade53cb93cff1dad69b99ce/crates/view/src/worker.rs#L146

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+        continue;
+      }
+
+      const id = await sha256Hash(sourceTx.toBinary());
+      // TODO: is there a race condition here? should we query the record again?
+      record.source = new CommitmentSource({
+        source: { case: 'transaction', value: { id } },
+      });
+      const idString = uint8ArrayToHex(id);
+      relevantTx.set(idString, sourceTx);
```
Valentine1898, line 291: We correctly define relevant by nullifier transactions, but we should not overwrite the notes or swap source in this case, only save the transaction 
https://github.com/penumbra-zone/penumbra/blob/2b0408879aaef969aade53cb93cff1dad69b99ce/crates/view/src/worker.rs#L146

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+        recordNeedsSourceTxId.set(uint8ArrayToHex(nullifier.inner), record);
+    }
+
+    return recordNeedsSourceTxId;
+  }
+
+  // TODO: can we use compactBlock.nullifiers?
+  async resolveSources(
+    sourcelessRecords: Map<string, SwapRecord | SpendableNoteRecord>,
+    blockTransactions: Transaction[],
```
Valentine1898, line 252: in resolveSources() we have to pass 
1. spent_nullifiers 
2. new_notes 
3. new_swaps
4. blockTransactions

Perhaps the function should be named differently?

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+      record.source = new CommitmentSource({
+        source: { case: 'transaction', value: { id } },
+      });
+      const idString = uint8ArrayToHex(id);
+      relevantTx.set(idString, sourceTx);
+
+      if (record instanceof SpendableNoteRecord) await this.indexedDb.saveSpendableNote(record);
+      else if (record instanceof SwapRecord) await this.indexedDb.saveSwap(record);
+    }
+
```
Valentine1898, line 296: We missed the definition of transactions relevant by commitment. 
We should compare the commitment from `new_notes` and `new_swaps` with the commitment in `tx.actions` 
https://github.com/penumbra-zone/penumbra/blob/2b0408879aaef969aade53cb93cff1dad69b99ce/crates/view/src/worker.rs#L155C42-L155C42 
In this case we must define source for `new_notes` and `new_swaps`

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,367 @@
...
+        await this.indexedDb.saveSwap(record);
+      }
+
+      // we want to recover ids for transaction sources that don't have an id,
+      // represented by an empty Uint8Array
+      if (
+        record.source?.equals({ source: { case: 'transaction', value: { id: new Uint8Array() } } })
+      )
+        recordNeedsSourceTxId.set(uint8ArrayToHex(nullifier.inner), record);
+    }
```
Valentine1898, line 244: This condition should be removed, we should return spent nullifiers regardless of source

packages/query/src/block-processor.ts
turbocrime: changed, still wip

packages/query/src/block-processor.ts
turbocrime: now resolve nullifiers does not modify source, but it does modify spend/claim height.

returns spent nullifiers

packages/query/src/block-processor.ts
turbocrime: no longer assigning source tx id

packages/query/src/block-processor.ts
turbocrime: refactored, no longer updates record sources

turbocrime:
new commit addressing comments by @Valentine1898, please confirm correctness. still need test data

turbocrime:
fails to identify tx at height 2378 on current chain, going to lunch

packages/wasm/package.json
```diff
@@ -7,19 +7,19 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@bufbuild/protobuf": "^1.5.1",
+    "@bufbuild/protobuf": "^1.6.0",
     "@penumbra-zone/types": "workspace:*",
-    "@penumbra-zone/wasm-bundler": "0.64.0",
-    "@penumbra-zone/wasm-nodejs": "0.64.0",
+    "@penumbra-zone/wasm-bundler": "file:../../../penumbra/crates/wasm/publish/bundler/penumbra-zone-wasm-bundler-0.64.1.tgz",
```
TalDerei, line 12: change to `0.64.1-preview-3`

turbocrime:
just kidding, i forgot to reset idb. it does detect 2378. review pls

packages/wasm/package.json
turbocrime: does it certainly work? there's new commits since then

Valentine1898:
We're still not processing commitments correctly
I'll be back soon with a more detailed description of what we need to do

Valentine1898:
> We're still not processing commitments correctly I'll be back soon with a more detailed description of what we need to do

But it seems to work, although it's not done the same way as on rust

packages/wasm/package.json
TalDerei: actually it doesn't. I can run `pnpm install` with `file:../../../penumbra/crates/wasm/publish/bundler/penumbra-zone-wasm-bundler-0.64.1.tgz`

turbocrime:
yes - the logic is split between wasm crate scan and typescript block processor, but in rust code it's more integrated.

the old block processor performed fewer functions, and i suspect did not detect these tx either. i was formerly rehydrating/assigning commitment sources but now no longer doing that.

packages/wasm/package.json
TalDerei: It doesn't, we'll need to publish `0.64.1-preview-4` from main.

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,390 @@
...
+
+      if (!record) continue;
+
+      records.set(stateCommitment, record);
+    }
+
+    if (commitments?.length) console.log('resolveCommitments', commitments.length, records.size);
+
+    return records;
+  }
```
Valentine1898, line 228: This is a very suboptimal way to get commitments from the current block
The point is that `flush_updates() `returns `new_notes` and `new_swaps` where commitments are defined. 
Instead of using `sctUpdates.store_commitments` , get notes and swap records from indexed-db we can just use commitments from `new_swaps` and `new_notes`

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,390 @@
...
+      console.log('resolveNullifiers', nullifiers.length, spentNullifiers.size);
+
+    return spentNullifiers;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async identifyRelevantTransactions(
+    blockSpent: Nullifier[],
+    blockCommitted: StateCommitment[],
+    blockTxs: Transaction[],
```
Valentine1898, line 268: Now there is no problem with transaction detection, but we can also achieve this without giving up rehydrating. 
We pass `blockCommitted: StateCommitment[]`, but we can pass `new_swaps` and `new_notes` instead 

I think the name `identifyRelevantTransactionsAndRehydrate()` would better represent what should be happening in this function 

And handle commitments like this 
```
for (const newNote of newNotes) {
        if (txCommitments.includes(newNote.noteCommitment)) {
          newNote.source = new CommitmentSource({
            source: { case: 'transaction', value: { id: tx } }
          });
          relevantTx.push(tx);
        }
      };

      for (const newSwap of newSwaps) {
        if (txCommitments.includes(newSwap.noteCommitment)) {
          newSwap.source = new CommitmentSource({
            source: { case: 'transaction', value: { id: tx } } }
          });
          relevantTx.push(tx);
        }
      };
```
This is just an illustrative example, I think this will lead to other changes and it could be done better

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,390 @@
...
+      // - every block at tip
+      const flushReasons = {
+        wasmWantsFlush,
+        interval: compactBlock.height % 1000n === 0n,
+        new: compactBlock.height > latestBlockHeight,
+      };
+
+      let compactScan;
+      if (Object.values(flushReasons).some(Boolean)) {
+        compactScan = this.viewServer.flushUpdates();
```
Valentine1898, line 119: I've ignored refactoring before, but why `compactScan`?   
flushUpdates() returns updates, and updates are stored in this variable

packages/query/src/block-processor.ts
```diff
@@ -0,0 +1,390 @@
...
+        'updated commitments': commitmentRecords.size,
+        'new notes': compactScan?.newNotes.length,
+        'new swaps': compactScan?.newSwaps.length,
+      };
+
+      if (Object.values(fullScanReasons).some(Boolean)) {
+        console.log('full scan', compactBlock.height, fullScanReasons);
+
+        // this is a network query
+        const fullBlock = await this.querier.app.txsByHeight(compactBlock.height);
```
Valentine1898, line 159: I think it should not be `fullBlock`, this variable stores the list of transactions of the block. Perhaps `blockTxs`?

Valentine1898 (changes requested):

packages/query/src/block-processor.ts
turbocrime: as far as i can tell, sctUpdates.store_commitments is equivalent to commitments in new_notes and new_swaps.

i am now using commitments from new_swaps and new_notes, but do you have an example or can you explain where store_commitments is suboptimal?

Valentine1898: We can avoid additional indexed-db calls 
```
const record =
        (await this.indexedDb.getSpendableNoteByCommitment(stateCommitment))) ??
        (await this.indexedDb.getSwapByCommitment(stateCommitment));

      if (!record) continue;
```

packages/query/src/block-processor.ts
turbocrime: passing new_swaps or new_notes seems incorrect because they have already been written to db, and possibly re-fetched and re-written by nullifier detection step. if we just reuse the object here we might destroy the re-write. and none of this is atomic, so we must fetch again anyway. they are indexed by commitment

the block processor is already complex, so i would like to keep functions as simple as possible and perform a single action. i've written a separate function for rehydration.

packages/query/src/block-processor.ts
turbocrime: renamed

packages/query/src/block-processor.ts
turbocrime: ok

packages/query/src/block-processor.ts
Valentine1898: I don't think we can just leave out the source rehydrate

Valentine1898: Maybe it makes sense to rewrite the block processor so that notes and swaps are saved only once at the end?

Valentine1898: It seems to me that we can't detect and spend a note or swap in the same block. So we don't have to worry about `new_notes` or `new_swaps` being overwritten when nullifiers are processed, and then overwritten again when source rehydrate occurs

packages/query/src/block-processor.ts
```diff
@@ -103,89 +115,126 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+        for (const { noteCommitment, source } of flush.newNotes)
+          blockCommitments.set(
+            uint8ArrayToHex(noteCommitment!.inner),
+            source!.equals(blankTxSource),
+          );
+        for (const { swapCommitment, source } of flush.newSwaps)
+          blockCommitments.set(
+            uint8ArrayToHex(swapCommitment!.inner),
+            source!.equals(blankTxSource),
+          );
```
turbocrime, line 164: this seems completely equivalent to compactBlock.sctUpdates.state_commitments

packages/query/src/block-processor.ts
```diff
@@ -103,89 +115,126 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-          fullBlock.transactions,
+        const { relevantTx, sourceTx } = await this.identifyTransactions(
+          allCommitments,
+          transactions,
         );
 
-        console.warn('relevantTx', relevantTx.length);
+        console.warn('relevantTx', relevantTx.size, 'sourceTx', sourceTx.size);
+
+        await this.saveSourceIds(sourceTx);
```
turbocrime, line 199: rehydrating source ids is its own function

packages/query/src/block-processor.ts
```diff
@@ -103,89 +115,126 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+            source!.equals(blankTxSource),
+          );
       }
 
       // nullifiers on this block may match notes or swaps from db
-      // mark them as spent/claimed, return what was used
-      const spentNullifiers = await this.resolveNullifiers(
+      // - update idb, mark as spent/claimed
+      // - return records' commitments
+      // - with a boolean to indicate if we should recover source id
```
turbocrime, line 170: as far as i can tell: records retrieved this way always contain sources already

packages/query/src/block-processor.ts
```diff
@@ -8,8 +8,9 @@ import { sha256Hash } from '@penumbra-zone/crypto-web';
 import {
   BlockProcessorInterface,
   IndexedDbInterface,
-  StoreCommitment,
   ViewServerInterface,
+  hexToUint8Array,
+  uint8ArrayToHex,
```
turbocrime, line 13: previous versions of this PR simply matched on objects, but now we are performing more map lookups and using more equivalent objects from different sources. javascript cannot compare uint8 for equality so they must all be stringified

turbocrime:
> I discovered another potential issue after refactoring The point is that we should always process nullifiers, even if there are no `flushReasons` This seems to be the most complicated testnet update ever

nullifiers are indeed always processed 👍

turbocrime:
much of this complexity is due to the split logic between wasm scanning and the ts processor

packages/query/src/block-processor.ts
turbocrime: > Maybe it makes sense to rewrite the block processor so that notes and swaps are saved only once at the end?

i was trying to avoid changing the flow in/out of idb, as some of the later steps, some in wasm, reference things written to idb

packages/query/src/block-processor.ts
turbocrime: ok 👍 but that was written not believing that the lists are equivalent - will store_commitments ever contain something we don't have? primarily i'm getting this list to identify relevant tx. if we can just use store_commitments directly to scan tx this is better than both approaches

Valentine1898:
> > I discovered another potential issue after refactoring The point is that we should always process nullifiers, even if there are no `flushReasons` This seems to be the most complicated testnet update ever
> 
> nullifiers are indeed always processed 👍

Yes, indeed, my bad

packages/query/src/block-processor.ts
Valentine1898: Yes, I agree, the in/out of idb flow really does not look transparent because of wasm

packages/query/src/block-processor.ts
Valentine1898: I'll take some time to test the latest changes of block-processor and get back to  discussion/review

TalDerei:
Screenshot 2024-01-18 at 10 40 12 AM
Screenshot 2024-01-18 at 10 40 26 AM
Screenshot 2024-01-18 at 12 26 21 PM

@Valentine1898 the testnet-preview chain reset, but any ideas why dumping the sqlite contents displays a different ordering?

Valentine1898:
> Screenshot 2024-01-18 at 10 40 12 AM Screenshot 2024-01-18 at 10 40 26 AM Screenshot 2024-01-18 at 12 26 21 PM
> @Valentine1898 the testnet-preview chain reset, but any ideas why dumping the sqlite contents displays a different ordering?

No idea, I thought maybe pcli was using `ORDER BY height `when fetching from sqllite, but not really

turbocrime:
>but any ideas why dumping the sqlite contents displays a different ordering?

the sql dump is in lexicographical order by hash

packages/types/src/transaction/classify.ts
turbocrime, line 1: todo revert this file's changes before merge, they're for testing visibility

packages/query/src/block-processor.ts
turbocrime: @Valentine1898 any insight to this? will a list generated this way, and the contents of sctUpdates.state_commitments ever differ?

packages/query/src/block-processor.ts
turbocrime: this feels more complex than the combined detect tx / hydrate source as requested in another comment so i will combine

packages/query/src/block-processor.ts
turbocrime: i believe these commitments are also never present on the current block so i will stop collecting them

packages/query/src/block-processor.ts
Valentine1898: The idea is that we don't need to create a 
`Map<CommitmentHexString, NeedsSourceId>()` , we just need to use `newNotes` and `newSwaps` when processing transactions and redefine the source for them. 
 There is also no reason for `resolveNullifiers` to return `Map<CommitmentHexString, NeedsSourceId>()`. When we spend a `nullifier` we should never override `source` (even if it suddenly turns out to be undefined). If we find a spent nullifier in a transaction we should only save that transaction.

Valentine1898: https://github.com/penumbra-zone/penumbra/blob/2b0408879aaef969aade53cb93cff1dad69b99ce/crates/view/src/worker.rs#L141 

This code on rust (lines 141-175) does everything we do, but simpler. At the same time we have all the same data as rust:
`new_notes`
`new_swap`
`spent_nullifiers` ( it is assumed that resolveNullifiers() will just return a list of nullifiers)

Valentine1898 (approved):

Valentine1898:
Block processing works correctly. 
I still think we can improve the code related to rehydrate and transaction detection if we go the same way as in Rust (pass `new_notes` and `new_swaps` directly to rehydrate instead of creating `Map<CommitmentHexString, NeedsSourceId>()`.
@turbocrime if you think we have objective reasons to create `Map<CommitmentHexString, NeedsSourceId>()` then I'm willing to agree and won't block PR. 

Do we need to wait for Gabe to get his confirmation regarding the zod removal?

hdevalence:
We shouldn't wait on Gabe, if he'd like further changes when he's back we can do those then but for now I think the most important thing is to get this merged, it's been much bigger than expected.

turbocrime:
i was primarily creating the string maps for js object equivalency reasons. it only matters if we bring state commitments from multiple places into one collection. if we certainly don't need commitments from anywhere but records on the wasm scan, this can be simpler.

i have tried to maintain the idb state of the previous block processor whenever performing an equivalent step - e.g. if a table is written, and later queried, i tried to not use the previously-written record again because it is possibly involved in the query. if that is certainly not possible in the case of records and nullifiers, this is simpler.

i've seen worker.rs but
- it was unclear if the information from wasm is really equivalent to filtered_block
- worker.rs does not seem concerned with modifying spend heights
- it was unclear if nullifiers could identify a record that needed source hydration

if nullifers will never retrieve a record that needs source hydration, this is simpler. if all relevant tx on current block can be reliably identified by commitments from records on the current block, this is simpler.

@Valentine1898 can you explain the contents of compactBlock.sctUpdates.state_commitments and if/why it is different from the set of state commitments present in records on the block

should finish simplifying today

Valentine1898:
Yes, your reasoning is correct 
- new_notes and new_swaps cannot be spent in the same block as detected. Therefore `resolveNullifiers()` will never overwrite `new_notes` and `new_swaps` in `indexed-db`. 
- nullifier spent detection will never overwrite swap/note source. 
- all relevant tx on current block can certainly be identified by commitments AND SPENT NULLIFIERS from records on the current block  

Regarding `flush.sctUpdates.state_commitments` I don't know how SCT works, but from my observations `state_commitments` correspond to a set of state commitments present in records on the block

turbocrime (commented):
this seems done. a couple todos for release

https://github.com/penumbra-zone/web/blob/074b671a6aaca105652771e1b3ad5d4b0849d4bb/apps/extension/src/utils/download-proving-keys.ts#L6-L8

https://github.com/penumbra-zone/web/blob/074b671a6aaca105652771e1b3ad5d4b0849d4bb/packages/constants/config.ts#L9-L13


but those can't land until the core v0.65 goes live

packages/ui/components/ui/tx/view/memo-view.tsx
turbocrime, line 1: permit absent memoview, render nothing

packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -6,7 +6,6 @@ import { fromBaseUnitAmount } from '@penumbra-zone/types';
 
 export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
   if (!txv.bodyView) throw new Error('transaction view missing body view');
-  if (!txv.bodyView.memoView) throw new Error('body view missing memo view');
```
turbocrime, line 9: permit absent memoview

Valentine1898 (approved):
It was a lot more complicated than expected, great job @turbocrime

PR #360, opened 2024-01-06
use Transaction.toBinary instead of wasm encodeTx

turbocrime:
wasm just uses `pbt::Transaction::encode_to_vec`, which is byte-for-byte identical to bufbuild's `Transaction.toBinary`

https://github.com/penumbra-zone/penumbra/blob/4d4ae0197a1315fdb02bda6d40b044e35c838fe8/crates/wasm/src/tx.rs#L44-L52

https://github.com/penumbra-zone/penumbra//blob/1c957b5e1fc68cb2d26a63135b5392cd9357d2a9/transaction/src/transaction.rs#L306-L318

### DIFF
--- a/packages/query/src/block-processor/index.ts
+++ b/packages/query/src/block-processor/index.ts
@@ -15,8 +15,11 @@ import {
 import { RootQuerier } from '../root-querier';
 import { generateMetadata } from './metadata';
 import { Transactions } from './transactions';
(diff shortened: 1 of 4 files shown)


TalDerei (approved):

hdevalence:
we should file an issue upstream asking bufbuild to commit to producing ADR-27 compliant encodings https://docs.cosmos.network/main/build/architecture/adr-027-deterministic-protobuf-serialization

PR #362, opened 2024-01-08
Support fees in Penumbra web

### LINKED ISSUE #308: Support fees in Penumbra Web
Prior to mainnet, the web extension will need to support including transaction fees, paid in terms of the staking token (`penumbra`) and based on the current gas prices retrieved from a fullnode.

This is already implemented in the `Planner`. Examples can be seen in the `pcli` code, for example see https://github.com/penumbra-zone/penumbra/blob/37b1eab/crates/bin/pcli/src/command/tx.rs#L262-L271: 

```rust
        let gas_prices = app
            .view
            .as_mut()
            .context("view service must be initialized")?
            .gas_prices(GasPricesRequest {})
            .await?
            .into_inner()
            .gas_prices
            .expect("gas prices must be available")
            .try_into()?;
  
            // ...
            let mut planner = Planner::new(OsRng);
            planner.set_gas_prices(gas_prices);
```

When `Planner::plan` is called, the required gas will automatically be added to the transaction plan. After the plan has been finalized, the included fee can be retrieved with the `TransactionPlan`'s `fee` property.

After this initial implementation is complete, we should add support for fee _tiers_: https://github.com/penumbra-zone/penumbra/issues/3153 however this requires some small changes to the planner.

## Steps:

- [x] Fetch the current gas prices from the view service (`GasPricesRequest`)
- [x] Supply the current gas prices to the planner (`Planner::set_gas_prices`)
- [x] Display the transaction fees in the UI

Linked issue in main repo: https://github.com/penumbra-zone/penumbra/issues/3370

jessepinho:
## Background
Per #308, we want to support transaction fees in Penumbra Web. To do this, we need to make a `TransactionPlannerRequest` each time one of the values in the send form changes, then grab the gas fee off of the returned plan. 

## In this PR
- Refactor `planWitnessBuildBroadcast` in the `SendSlice` to extract a `getPlan` function that takes care of just the planning portion.
- Use that `getPlan` function to get the gas fee any time inputs change in the send form.
- Display the gas fee below the amount to send.
- Hardcode an undefined gas fee for the IBC form, which we're not yet using (per Henry).

Closes #308


apps/webapp/public/fuel.svg
jessepinho, line 1: Got this from [Lucide](https://lucide.dev/icons/fuel), which Zpoken told us is where we got some of our icons.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -30,9 +40,13 @@ export default function InputToken({
   setSelection,
   balances,
   tempPrice,
+  totalGasPrice,
   ...props
 }: InputTokenProps) {
   const vResult = validationResult(value, validations);
+  const currentBalance = selection?.asset
+    ? fromBaseUnitAmount(selection.asset.amount, selection.asset.denom.exponent).toFormat()
+    : '0';
```
jessepinho, line 49: Extracted to a `currentBalance` const to make it clear what this represents.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -29,11 +38,21 @@ export default function InputToken({
...
   balances,
-  tempPrice,
+  totalGasPrice,
   ...props
 }: InputTokenProps) {
   const vResult = validationResult(value, validations);
 
+  const currentBalance = selection?.asset
+    ? fromBaseUnitAmount(selection.asset.amount, selection.asset.denom.exponent).toFormat()
+    : '0';
```
jessepinho, line 48: Extracted to a `currentBalance` const to make it clear what this represents.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -29,11 +51,15 @@ export default function InputToken({
   inputClassName,
   setSelection,
   balances,
-  tempPrice,
+  totalGasPrice,
   ...props
 }: InputTokenProps) {
   const vResult = validationResult(value, validations);
 
+  const currentBalance = getCurrentBalance(selection?.asset);
```
jessepinho, line 59: Extracted to a helper to make it clear what this represents.

apps/webapp/public/fuel.svg
jessepinho, line 1: Got this from [Lucide](https://lucide.dev/icons/fuel), which Zpoken told us is where we got some of our icons.

It'd be great to eventually standardize an icon set, though — right now, we seem to be getting icons from Radix, Lucide, and our Figma mockups (and I don't know where those mockups' icons come from).

apps/webapp/src/components/send/send-form.tsx
```diff
@@ -1,16 +1,22 @@
...
+import { getGasPrices } from '../../fetchers/gas-prices.ts';
 
 export const SendAssetBalanceLoader: LoaderFunction = async (): Promise<AccountBalance[]> => {
-  const balancesByAccount = await getBalancesByAccount();
+  const [balancesByAccount, gasPrices] = await Promise.all([
+    getBalancesByAccount(),
+    getGasPrices(),
+  ]);
+
+  useStore.getState().send.setGasPrices(gasPrices);
```
jessepinho, line 19: I'm using the loader to load gas prices and set them in the state. 

Per [the docs](https://reactrouter.com/en/main/route/loader), loaders should make a request, then _return_ the result, thereby making that result available to the component via `useLoaderData()`. However, the existing code was using this loader to fetch data, potentially assign it to Zustand state, and return the data.

That seems perfectly acceptable to me, even if that wasn't `react-router-dom`'s original intent with loaders, so I'm doing the same thing as well. (An alternative would be to wrap this loading in a `useEffect`, but that would prevent SSR from making these calls before the component has rendered.) But please let me know if y'all think this is an OK pattern.

apps/webapp/src/components/send/send-form.tsx
```diff
@@ -82,7 +89,7 @@ export const SendForm = () => {
           },
         ]}
         balances={accountBalances}
-        tempPrice={1}
```
jessepinho, line 85: See my note below about removing the temp price and its corresponding UI

packages/router/src/grpc/view-protocol-server/gas-prices.ts
```diff
@@ -0,0 +1,13 @@
+import type { Impl } from '.';
+
+export const gasPrices: Impl['gasPrices'] = () => {
+  /** @todo Replace this stub with real gas prices. */
```
jessepinho, line 4: @Valentine1898 is currently working on this as part of #203

apps/webapp/src/state/send.ts
```diff
@@ -152,3 +161,17 @@ export const sendValidationErrors = (
...
+
+export const getTotalGasPriceFromGasPrices = (gasPrices: GasPrices | undefined) => {
+  if (!gasPrices) return undefined;
+
+  return (
+    gasPrices.blockSpacePrice +
+    gasPrices.compactBlockSpacePrice +
+    gasPrices.verificationPrice +
+    gasPrices.executionPrice
+  );
```
jessepinho, line 173: I'm summing up the different dimensions of the gas price to get the total. I'm not 100% certain that this logic is correct — this is where I'm bumping against the limits of my crypto knowledge — so reviewers, please let me know if I've got this right.

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
```diff
@@ -24,6 +26,11 @@ export const transactionPlanner: Impl['transactionPlanner'] = async (req, ctx) =
     fmdParams,
   });
 
+  const fetchedGasPrices = await gasPrices(req, ctx);
+  if (fetchedGasPrices.gasPrices) {
+    planner.setGasPrices(new GasPrices(fetchedGasPrices.gasPrices));
+  }
```
jessepinho, line 32: At the moment, these lines cause the following error when I press "Send":
image

Because of that, we have to wait until @Valentine1898 's work on #203 is finished before we can merge this PR.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -67,24 +93,17 @@ export default function InputToken({
...
-            value && 'text-muted-foreground',
+        <div className='flex items-start gap-2'>
+          {totalGasPriceAsString && (
+            <>
+              <img src='/fuel.svg' alt='Gas price' className='h-5 w-5' />
+              <p className='font-bold text-muted-foreground'>{totalGasPriceAsString}</p>
+            </>
           )}
-        >
-          ${displayAmount(Number(value) * tempPrice)}
```
jessepinho, line 76: As far as I can tell, this appeared to be a stub of the gas price. (I'm not sure of that, though — I'm not actually sure what `tempPrice` was for.) So I removed it and displayed the gas price in its place. If I'm missing something about the `tempPrice`, let me know.

apps/webapp/src/state/send.ts
turbocrime: it looks like the final step of calculating a total might be the location where the denominator is expected to be considered

https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/fee/src/gas.rs#L65-L74

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
turbocrime: would like to see this resolved first, yes

turbocrime (commented):
i think the way i would handle this, is instead of having a state selector that provides the total, provide a state selector that provides all of the price components. that way it could be displayed differently, or with more detail.

otherwise looks like good progress, pending the wasm impl

packages/router/src/grpc/view-protocol-server/gas-prices.ts
Valentine1898: It seems that gas prices do not actually change on the network and are always equal to these values
```
blockSpacePrice: 0n
compactBlockSpacePrice: 0n
executionPrice: 0n
verificationPrice: 0n
```

jessepinho: @Valentine1898 but that could change in the future, right? i.e., I should still be requesting them (using your new implementation) rather rather than depending on all-0 values?

Valentine1898: Yeah, I was just clarifying that we're currently getting all the 0 values from the network

jessepinho: Gotcha, thanks

turbocrime:
how to calculate gas price preview on send form

gas prices can be calculated from transaction plan information. the planner performs this calculation here

https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/planner.rs#L157-L174

the transactionplan's impl simply sums the costs of the action plans

https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/gas.rs#L225-L229

the send form only ever generates 'spend' and 'output' actions. cost for these plans are constant, based on their constant dimensions

https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/gas.rs#L23-L65

the final cost can be estimated using price information from GasPriceRequest https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/fee/src/gas.rs#L65-L74

hdevalence:
> how to calculate gas price preview on send form

It'd be strongly preferable not to replicate gas cost estimation logic in typescript. We will already be doing exact gas costing as part of the planner. Can we instead just read the fee out of the returned transaction plan?

turbocrime:
the ticket requested displaying a preview in the webapp, by our interpretation. displaying in the popup is certainly much easier :)

turbocrime:
or, should we have an extra stage between "plan came back" and "submit plan for auth"

hdevalence:
@turbocrime added an extra suggestion to my comment in an edit, i think a preview is certainly nice to have, wdyt of that strategy for having it?

the point is that the view service methods don't have user in the loop, the page can make as many planner requests as it wants and throw away the resulting plans

turbocrime:
> Instead, what if the form submitted planner requests to the extension as the values were input, so it can display a preview, but have the logic for computing the preview be the actual logic that will be used for tx planning?

current design would require sending a new plan request on every change, creating a new planner and re-planning the whole tx each time, instead of incrementally planning a single tx

turbocrime:
very doable

hdevalence:
my assumption is that planning should be pretty fast, fast enough that we could just spam requests on form change, it seems at least worth trying

Valentine1898:
> the existing webapp combines plan/witness/build/broadcast into a single fn and i think we were conceptually trapped lol

Maybe it makes sense to leave it that way and send separate planner requests for gas cost preview

jessepinho:
Thanks all — making planner requests for the gas costs sounds great. I'll go with that approach!

apps/webapp/src/components/send/send-form/use-refresh-fee.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react';
+import { sendSelector } from '../../../state/send';
+import { useStore } from '../../../state';
+
+/**
+ * Refreshes the fee in the state when the amount, recipient, selection, or memo
```
jessepinho, line 6: Note that we're refreshing even when just the memo changes. Am I correct in understanding that a larger memo could result in the fee changing, since a larger memo takes up more space? Or does the fee remain the same regardless of memo?

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -1,10 +1,23 @@
...
-import { displayAmount, fromBaseUnitAmount } from '@penumbra-zone/types';
+import { fromBaseUnitAmount, joinLoHiAmount } from '@penumbra-zone/types';
 import SelectTokenModal from './select-token-modal';
 import { Validation, validationResult } from './validation-result';
-import { AccountBalance } from '../../fetchers/balances';
+import { AccountBalance, AssetBalance } from '../../fetchers/balances';
 import { Selection } from '../../state/types';
+import { Fee } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1alpha1/fee_pb';
+
+const PENUMBRA_FEE_DENOMINATOR = 1000;
```
jessepinho, line 10: Note that we're currently hardcoding the penumbra fee denominator. @Valentine1898 pointed out that I could call [`getAllAssets`](https://github.com/penumbra-zone/web/blob/85fe38e4814b2600214f959a843f932e8004e217/apps/webapp/src/fetchers/assets.ts#L5) to get the denominator of the token referenced via the fee's `assetId` property; but @hdevalence, will there ever be a case where the fee will be paid in something other than the staking token (penumbra)?

apps/webapp/src/state/send.ts
```diff
@@ -84,9 +105,11 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
   };
 };
 
-const planWitnessBuildBroadcast = async ({ amount, recipient, selection, memo }: SendSlice) => {
-  if (typeof selection?.accountIndex === 'undefined') throw new Error('no selected account');
-  if (!selection.asset) throw new Error('no selected asset');
+type TransactionProps = Pick<SendSlice, 'amount' | 'recipient' | 'selection' | 'memo'>;
+
+const getPlan = async ({ amount, recipient, selection, memo }: TransactionProps) => {
```
jessepinho, line 110: Extracted a `getPlan` function, which we call to refresh the fee whenever a form input changes.

apps/webapp/src/state/send.ts
```diff
@@ -107,6 +130,19 @@ const planWitnessBuildBroadcast = async ({ amount, recipient, selection, memo }:
 
   const { plan } = await viewClient.transactionPlanner(req);
 
+  return plan;
+};
+
+const planWitnessBuildBroadcast = async ({
```
jessepinho, line 136: This works exactly as before, even throwing the same errors as before. I've only extracted a `getPlan` function above.

apps/webapp/src/components/shared/input-token.tsx
turbocrime: i believe currently tx fees are always PEN https://protocol.penumbra.zone/main/stake/validator-rewards.html

apps/webapp/src/components/send/send-form/use-refresh-fee.ts
turbocrime: i believe an encrypted memo is constant size, even if it contains no memo

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
turbocrime: has this resolved?

jessepinho: Yep!

I just tested it again though, and noticed that I'm still not seeing fees in the approval window:

image

This _may_ be because the amounts are all `0n`, but I'd still expect to see something in the JSON view... @Valentine1898 , since you just worked on it, do you happen to know if the fee amount is expected to be empty when each dimension of the fee is `0n`?

apps/webapp/src/components/send/send-form/use-refresh-fee.ts
jessepinho: OK, thanks — I've removed `memo` from the effect dependencies.

apps/webapp/src/state/send.ts
```diff
@@ -56,6 +60,23 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
...
+        return;
+      }
+
+      const plan = await getPlan({ amount, recipient, selection, memo });
+      if (!plan?.fee?.amount) return;
+
+      set(state => {
+        state.send.fee = plan.fee;
+      });
+    },
```
turbocrime, line 79: planning should be quick, but it does spin up a wasm runtime and make idb queries. input events can trigger very rapidly, so the request should be debounced to avoid hammering the endpoint. 

give it a largeish subsecond delay. if it's easy, apply CallOptions with an abort signal - we don't respect the signal in the service yet, but that should happen eventually.

https://github.com/connectrpc/connect-es/blob/main/packages/connect/src/call-options.ts#L38

turbocrime: you could debounce either here or in the ui component

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
Valentine1898: https://github.com/penumbra-zone/penumbra/blob/28ee5f71ba620e77b9ddf6ebcd740ed6a5be359c/crates/core/component/fee/src/gas.rs#L67
Based on this formula, if all gas prices are `0n`, then we will get a `0n` gas cost transaction. 
But it is strange that we get the display 
`"amount": {}`
I would expect `"amount": {"lo": 0n, "hi: 0n} `
but maybe it's just a serialization case.

apps/webapp/src/state/send.ts
jessepinho: @turbocrime good call — I actually ended up debouncing it [in the `useRefreshFee` hook](https://github.com/penumbra-zone/web/pull/362/files#diff-7676de151de670030b837855d561ce6f0a44fb45c3c31c086c033503c77d59ef).

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
jessepinho: @Valentine1898 that is indeed strange. I logged [the `tx` object](https://github.com/penumbra-zone/web/blob/jessepinho/web-308-gas-fees-in-web/apps/extension/src/routes/popup/approval/transaction.tsx#L7-L8) that the `TransactionApproval` component passes to the JSON Viewer, and `amount` is indeed an empty object. I'm looking into this more, but if you have any insight on why that could be even though [I called `setGasPrices`](https://github.com/penumbra-zone/web/blob/jessepinho/web-308-gas-fees-in-web/packages/router/src/grpc/view-protocol-server/transaction-planner.ts#L29-L32) from inside `transactionPlanner`, it'd be great to hear your thoughts!

image

apps/webapp/src/state/send.ts
jessepinho: Aaand I forgot to use the abort signal — will fix.

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
Valentine1898: Yes, I changed the gas prices values in indexed-db and it didn't affect anything. 
The issue seems to be in this logic 
https://github.com/penumbra-zone/penumbra/blob/28ee5f71ba620e77b9ddf6ebcd740ed6a5be359c/crates/wasm/src/planner.rs#L504

Valentine1898: The issue seems to be that we're just not passing the fee to planner

Valentine1898: No, I was wrong, I'm sorry. 
I only changed the blockSpacePrice in indexed-db so it had no effect. After I changed the executionPrice I got the correct display 
image

Valentine1898: I still think the mapping of `"amount": {} `is a serialization case, but not a serialization of `ts object -> json`, but possibly a serialization of `wasm -> ts object`

apps/webapp/src/components/shared/input-token.tsx
Valentine1898: https://discord.com/channels/824484045370818580/979785365437165658/1194805904311328808 
We can ignore the `assetId` for now

Valentine1898 (approved):

apps/webapp/src/state/send.ts
jessepinho: Per discussion in today's web weekly meeting, we're going to skip using the abort signal for now, especially since planning is so fast.

packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -2,11 +2,13 @@ import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbr
...
 import { ActionViewComponent } from './action-view';
 import { ViewBox, ViewSection } from './viewbox';
-import { fromBaseUnitAmount } from '@penumbra-zone/types';
+import { joinLoHiAmount } from '@penumbra-zone/types';
 
 export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
   if (!txv.bodyView) throw new Error('transaction view missing body view');
 
+  const fee = joinLoHiAmount(txv.bodyView.transactionParameters!.fee!.amount!).toString();
+
```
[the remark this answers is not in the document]
jessepinho: Good call — fixed

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -1,10 +1,23 @@
...
+
+const getFeeAsString = (fee: Fee | undefined) => {
+  if (!fee?.amount) return '';
+  return `${(Number(joinLoHiAmount(fee.amount)) / PENUMBRA_FEE_DENOMINATOR).toString()} penumbra`;
+};
+
+const getCurrentBalance = (assetBalance: AssetBalance | undefined) =>
+  assetBalance
+    ? fromBaseUnitAmount(assetBalance.amount, assetBalance.denom.exponent).toFormat()
+    : '0';
```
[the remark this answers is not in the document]
jessepinho: Discussed in Discord — this will be part of a follow-up PR (in progress here https://github.com/penumbra-zone/web/pull/440)

Issue #312, opened 2023-12-16
Stream large idb queries, avoid querying entire tables


turbocrime:
our `IndexedDb` wrapper class provide methods for things like getting "all transactions", "all swaps", etc

https://github.com/penumbra-zone/web/blob/83b95344315146c11a61be2aed46ff29bff37a58/packages/storage/src/indexed-db/index.ts#L159-L161

these query wrappers are used by service methods and elsewhere, which may await the return to load the entire table into memory, perform their own synchronous in-memory filtering, and return a faux "streaming" response

https://github.com/penumbra-zone/web/blob/83b95344315146c11a61be2aed46ff29bff37a58/packages/router/src/grpc/view-protocol-server/tx-info.ts#L13-L24

tables could eventually be large, so these should be real streaming queries. we are using idbp which [provides very nice cursor and asynciterator features](https://github.com/jakearchibald/idb#idbcursor-enhancements) for this purpose, so that change is minimal.

but more broadly, queriers should make specific queries. we should possibly remove "getAll" methods from `IndexedDb`.

grod220:
Oh for sure we should do this. Planning wise, normally we should keep things in the backlog until our planning meeting where we can prioritize the next sprint's Todos. But given we are in a kind of free form holiday season, think it's fine to drag in if you want.

grod220:
Applies to this filtering as well: https://github.com/penumbra-zone/web/pull/332

PR #363, opened 2024-01-09
filtered asset req

Vadim121197:


packages/router/src/grpc/view-protocol-server/assets.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const req = new AssetsRequest({
+      filtered: false,
+    });
+    for await (const res of assets(req, mockCtx)) {
+      responses.push(new AssetsResponse(res));
+    }
+    expect(responses.length).toBe(8);
+  });
+
+  test('req with filtered as true return empty list', async () => {
```
jessepinho, line 59: Could you rewrite this so that there's at least one asset that shows up, and modify the test name to indicate how the filtering works?

Right now, this test makes it seem like, any time `filtered: true` is passed, it will return an empty list. But in fact, there should only be an empty list if nothing matches the filters. So, this test would be improved by reflecting that, and thus improving the likelihood that this test will catch any future bugs in the filtering logic.

```suggestion
  test('returns only matching denominations when `filtered` is `true`', async () => {
```

(^ if I'm understanding the logic correctly)

jessepinho (approved):
Just one change to the tests, then LGTM!

turbocrime:
these filters should be implemented as a db query, instead of using allMetadata and then filtering in js, as per #312 

we are already doing it the wrong way everywhere, so i'm not going to block. but maybe this is an opportunity to start doing it the right way.

Vadim121197:
@turbocrime streaming db query is a lot of work and it is better to do it as a separate PR

Issue #291, opened 2023-12-11
Chain id does not show in extension


grod220:
<img width="370" alt="Screenshot 2023-12-11 at 3 51 47 PM" src="https://github.com/penumbra-zone/web/assets/16624263/b4882b6d-3f0d-4634-afe2-40adab23587c">

Normally it is in the top right

grod220:
<img width="383" alt="Screenshot 2023-12-11 at 3 56 33 PM" src="https://github.com/penumbra-zone/web/assets/16624263/118e25bc-168c-460b-8c69-179e3a6f1e8e">

It also does not show in the advanced menu

Issue #289, opened 2023-12-11
Update deps to 0.64.0


grod220:
New testnet released, pull the latest types, npm packages, & wasm.

PR #288, opened 2023-12-11
More descriptive errors in background connection manager

turbocrime:
Fixes #287

PR #364, opened 2024-01-09
view service:  impl gas prices

Valentine1898:


packages/constants/config.ts
```diff
@@ -8,6 +8,6 @@ export interface Config {
 
 export const testnetConstants: Config = {
   grpcEndpoint: 'https://grpc.testnet.penumbra.zone',
-  indexedDbVersion: 17,
+  indexedDbVersion: 18,
```
Valentine1898, line 11: We must increment the indexed-db version every time we change indexed-db's structure

packages/query/src/block-processor/index.ts
```diff
@@ -159,6 +159,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       abortSignal: this.abortController.signal,
     })) {
       if (block.fmdParameters) await this.indexedDb.saveFmdParams(block.fmdParameters);
+      if (block.gasPrices) await this.indexedDb.saveGasPrices(block.gasPrices);
```
Valentine1898, line 162: We get `GasPrices` from `CompactBlock` during the scanning process

jessepinho (approved):
Looks good to me! 

Just want to make sure I understand something correctly, though: it looks to me like we're setting gas prices in IndexedDB, detached from any one specific transaction. That is, any time we set the gas prices, they'll be set more or less "globally" in the `GAS_PRICES` object store.

Somehow this seems strange to me, although I'm having trouble identifying why... it's as if we're storing a transient value — that is, the current gas prices — in a non-transient store (IndexedDB). Then, each time we start a new transaction, we'll re-set the gas prices in IndexedDB. Wouldn't it be more appropriate to pass it around in memory while creating a new transaction, rather than storing it in IndexedDB? Or is that not possible, given the way our frontend architecture is structured (which I'm still learning about)? 

If there's a good reason for this, though, go ahead and merge! 🟢

TalDerei (approved):

Valentine1898:
> Looks good to me!
> 
> Just want to make sure I understand something correctly, though: it looks to me like we're setting gas prices in IndexedDB, detached from any one specific transaction. That is, any time we set the gas prices, they'll be set more or less "globally" in the `GAS_PRICES` object store.
> 
> Somehow this seems strange to me, although I'm having trouble identifying why... it's as if we're storing a transient value — that is, the current gas prices — in a non-transient store (IndexedDB). Then, each time we start a new transaction, we'll re-set the gas prices in IndexedDB. Wouldn't it be more appropriate to pass it around in memory while creating a new transaction, rather than storing it in IndexedDB? Or is that not possible, given the way our frontend architecture is structured (which I'm still learning about)?
> 
> If there's a good reason for this, though, go ahead and merge! 🟢

Whenever Gas prices change they are published in CompactBlock. 
We do not know in which block Gas prices have changed, and we have no way to get the current Gas prices from penumbra full node. 
In practice it looks like this 
#0 - gas prices published in block 
#1 
...
#288
#289 - gas prices published in block 
#290
#291
...
#382 
#383 - user creates transaction and needs gas prices 
In block #383 we do not know in which block the actual value of gas prices is written. 
For this we need to cache the last known value of gas prices in indexed-db while we scan the blocks.

jessepinho:
@Valentine1898 that explanation is suuuper helpful — thanks! ship it 🚢

grod220:
That is super helpful documentation @Valentine1898. Low priority, but can you add that context to say here:

packages/router/src/grpc/view-protocol-server/gas-prices.ts

Valentine1898:
> That is super helpful documentation @Valentine1898. Low priority, but can you add that context to say here:
> 
> packages/router/src/grpc/view-protocol-server/gas-prices.ts

Yeah, sure, I'll add

PR #365, opened 2024-01-10
Extract several shared components related to rendering addresses

### LINKED ISSUE #217: Standardize address component
All displays of the address should use the `AddressViewComponent`. We may need to strip down the styles inside of that component and accept a classname so consumers can customize it.

Screenshot 2023-11-03 at 10 35 16 AM

jessepinho:
Well, this turned out to be surprisingly annoying.

In short: the idea was to consolidate the display of addresses throughout the UIs and use the `AddressViewComponent`.

However, that didn't work in practice, since that component is designed to render [an `AddressView` object](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/keys/src/address/view.rs) which, once translated into JavaScript, contains properties that need to be [corralled into human-readable values](https://github.com/penumbra-zone/web/blob/jessepinho/web-217-standardize-address-component/packages/ui/components/ui/tx/view/address-view.tsx#L16-L17). The problem is, we render addresses in places where we don't have an `AddressView` object, but rather just an address string. Further complicating matters is the fact that they aren't all designed the same way: in some places, we show "Account #N" and a copy button; in others, we just show the address itself.

So I went through the various places where we render addresses, and extracted out components that they share:
- `<AddressIcon />` renders a consistently styled icon for a given address, so that we don't need to pass the same classes and props to `<Identicon />` everywhere that we render an address.
- `<Address />` renders an address, shortened to the prefix + 24 characters and styled consistently throughout the app.
- `<CopyToClipboardIconButton />` wraps `<CopyToClipboard />` and passes it the copy icon.*

I then used these new components in the various places where there was previously duplicative code. We should use these going forward whenever we need to display these different items. That said, the original goal of #217 was probably not really possible, since the designs of the various places where we render addresses are so different.

\* Weirdly, the linter complained about the `text-muted-foreground` class we were using for that icon — but only when it's inside this component! I don't get why it's only breaking things in this component and not elsewhere that it's used. But it seemed like a minor design thing anyway, so I removed that class to get the linter to pass. If anyone has ideas on how to fix this, I'm all 👂🏻s.

Closes #217

### DIFF
--- a/apps/extension/src/state/wallets.ts
+++ b/apps/extension/src/state/wallets.ts
@@ -4,7 +4,6 @@ import {
   getAddressByIndex,
   getEphemeralByIndex,
   getFullViewingKey,
-  getShortAddressByIndex,
   getWalletId,
 } from '@penumbra-zone/wasm-ts';
 import { Key } from '@penumbra-zone/crypto-web';
@@ -86,9 +85,6 @@ export const accountAddrSelector = (state: AllSlices) => (index: number, ephemer
 
   return {
     address: bech32Addr,
-    preview: ephemeral
-      ? bech32Addr.slice(0, 33) + '…'
-      : getShortAddressByIndex(active.fullViewingKey, index),
     index,
   };
 };
--- a/packages/ui/components/ui/copy-to-clipboard.tsx
+++ b/packages/ui/components/ui/copy-to-clipboard.tsx
@@ -18,7 +18,11 @@ const CopyToClipboard = React.forwardRef<HTMLButtonElement, CopyToClipboardProps
 
     return (
       <Button
-        className={cn(copied && 'cursor-default text-teal-500 hover:no-underline', className)}
+        className={cn(
+          copied && 'cursor-default text-teal-500 hover:no-underline',
+          'block',
+          className,
+        )}
         variant='link'
         ref={ref}
         size='sm'
--- a/packages/ui/components/ui/copy-to-clipboard.tsx
+++ b/packages/ui/components/ui/copy-to-clipboard.tsx
@@ -10,10 +10,11 @@ export interface CopyToClipboardProps extends React.ButtonHTMLAttributes<HTMLBut
   text: string;
   labelType?: 'text' | 'icon';
   label: React.ReactElement;
+  isSuccessCopyText?: boolean;
 }
 
 const CopyToClipboard = React.forwardRef<HTMLButtonElement, CopyToClipboardProps>(
-  ({ text, className, label, ...props }, ref) => {
+  ({ text, className, label, isSuccessCopyText, ...props }, ref) => {
     const [copied, setCopied] = useState(false);
 
     return (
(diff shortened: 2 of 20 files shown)


apps/extension/src/state/wallets.ts
```diff
@@ -86,9 +85,6 @@ export const accountAddrSelector = (state: AllSlices) => (index: number, ephemer
 
   return {
     address: bech32Addr,
-    preview: ephemeral
```
jessepinho, line 89: `preview` is no longer needed, as addresses are now shortened inside the `<Address />` component

packages/ui/components/ui/copy-to-clipboard.tsx
```diff
@@ -19,7 +19,11 @@ const CopyToClipboard = React.forwardRef<HTMLButtonElement, CopyToClipboardProps
 
     return (
       <Button
-        className={cn(copied && 'cursor-default text-teal-500 hover:no-underline', className)}
+        className={cn(
+          copied && 'cursor-default text-teal-500 hover:no-underline',
+          'block',
```
jessepinho, line 24: Added the `block` class to remove the weird spacing beneath it that sometimes happens when it's displayed inline. This helps with when the button is displayed in a flex row.

packages/ui/components/ui/tx/view/address-view.tsx
```diff
@@ -1,26 +1,20 @@
...
-import { CopyIcon } from '@radix-ui/react-icons';
-import { CopyToClipboard } from '../../copy-to-clipboard';
-import { Identicon } from '../../identicon';
+import { bech32Address } from '@penumbra-zone/types';
+import { CopyToClipboardIconButton } from '../../copy-to-clipboard-icon-button';
 
 interface AddressViewProps {
   view: AddressView | undefined;
   copyable?: boolean;
-  short_form?: boolean;
```
jessepinho, line 10: Not being used anywhere, so I removed it

packages/ui/tests-setup.ts
```diff
@@ -1,6 +1,8 @@
 import { afterEach } from 'vitest';
 import { cleanup } from '@testing-library/react';
 
+import '@testing-library/jest-dom/vitest';
```
jessepinho, line 4: Adds matchers like `expect(...).toHaveTextContent()` or `expect(...).toHaveStyle()`

packages/ui/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "tsconfig/react-library.json",
-  "include": ["."],
+  "include": [".", "./tests-setup.ts"],
```
jessepinho, line 3: Required so that TypeScript knows about the matchers I mentioned in `tests-setup.ts`

turbocrime: is there a reason this couldn't go under the existing paths?

packages/ui/package.json
```diff
@@ -22,6 +22,7 @@
     "@radix-ui/react-toast": "^1.1.5",
     "@radix-ui/react-toggle": "^1.0.3",
     "@radix-ui/react-tooltip": "^1.0.7",
+    "@testing-library/jest-dom": "^6.2.0",
```
turbocrime, line 25: cool w me but i avoid adding deps without asking gabe

packages/ui/vitest.config.ts
```diff
@@ -3,6 +3,6 @@ import { defineConfig } from 'vite';
 export default defineConfig({
   test: {
     environment: 'jsdom',
-    setupFiles: ['./tests-setup.js'],
+    setupFiles: ['./tests-setup.ts'],
```
turbocrime, line 6: ah, i guess you're just maintaining current location

turbocrime (approved):
my only holdup is added dep but i guess you had some testing-library stuff already merged by gabe so approve

PR #366, opened 2024-01-12
Show more meaningful approval dialogue when creating transactions

### LINKED ISSUE #325: Upgrade JSON view in approval dialog to semantically meaningful `TransactionPlanView`
The approval dialog should show the user the effects of the transaction plan, rather than just the raw JSON. (It should _also_ show the raw JSON, if requested, but this should be collapsed by default).

To do this, there should be a new `TransactionPlanComponent` component. This code will be structurally similar to, but different than, the `TransactionViewComponent`. We will need sub-components for each of the action types with `*Plan` variants, but we can re-use the view components for the actions which are their own plans.

Unlike the transaction _view_ component, which is processing a trusted, pre-interpreted view, the transaction _plan_ component MUST be doing its OWN interpretation, it cannot trust anyone else to interpret the plan for it. As a result, the transaction plan component will need to take an FVK, to do things like show whether the user controls an address, etc.

This code is important and security-critical, since it's bridging between the user's comprehension and the actual cryptographic effects of the system.

Blocked on https://github.com/penumbra-zone/penumbra/issues/3541 and will need to be coordinated with the next testnet reset since the plan format is changing.

jessepinho:
image

image

## Background
When a user initiates a transaction, thus triggering the extension approval popup to open, we want to show users an interpreted view of the transaction rather than just a JSON view.

## In this PR
🔧  Create helper functions inside the `types` package to convert a `TransactionPlan` to a `TransactionView`
🔀  Use them to render a converted `TransactionView` via the `TransactionViewComponent` inside the approval popup dialog
🍱  Rework the typings of the extension's Zustand state to use the beautiful new `Jsonified` type, so that we get accurate typings of the output of Protobufs' `.toJson()` methods.
✅  Write tests for everything

## Reviewer checklist
- [ ] Please review the "before merge" section right below this
- [ ] Please review the screenshots above, and make sure that the values in the approval dialogue look correct for the values being input into the form. (I took two screenshots to scroll through the entire popup dialog.) Note that there's a `1 test_usd` spend action — I'm not sure where that's coming from, or why it's being spent?
- [ ] Please review [my comment](https://github.com/penumbra-zone/web/pull/366#discussion_r1451209146) about `destAddress`. It seems strange that there are top-level `destAddress` and `value` keys in the output action, but then they get put into the output view's `note` property. 
- [ ] When reviewing, please keep in mind that this dialog presents, to the user, the authoritative truth about a transaction that they're about to approve or deny. So please use a fine-toothed comb when looking over the logic of translating a transaction plan to a transaction view, since, if I've done something wrong there, it could present incorrect information about a transaction and cause a user to e.g., spend more money than they meant to.

## Before merge (why this build is failing)
For some reason, the `apps/extension/src/state/tx-approval.test.ts` test suite [fails](https://github.com/penumbra-zone/web/actions/runs/7545267165/job/20540486861?pr=366) with this error message: `  TypeError: Unknown file extension ".wasm" for /home/runner/work/web/web/node_modules/.pnpm/@penumbra-zone+wasm-bundler@0.64.0/node_modules/@penumbra-zone/wasm-bundler/index_bg.wasm`.

I've been googling and poking around to figure out why this is happening for quite a while, and can't figure it out. It doesn't happen in any other test suites that include WASM-generated code, so I don't know why this particular test suite is complaining. If reviewers have any suggestions, I'd love to hear them!

UPDATE: One thing I've noticed is that it appears that the failing test suite is the only one in the `apps/extension` package that imports from `@buf/penumbra-zone_penumbra.bufbuild_es`. Looking into that more now...

## Punting for now
- A toggle to show/hide the JSON view. Once we're happy with this and merge it, I can go back in and make that toggle. The scope of this PR keeps expanding, so I'm trying to just get this merged, but then can go in and refine the UI afterward.
- A toggle to show the public/personal views of the transaction in the approval dialog. Again, trying to limit the scope of this PR.

Closes #325

### DIFF
--- a/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
+++ b/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
@@ -9,7 +9,6 @@ export const isTxApprovalReq = (req: PopupRequest): req is TxApproval => {
 
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
-    // @ts-expect-error Typescript doesn't like JsonValue could possibly be very deep
     state.txApproval.authorizeRequest = jsonReq.authorizeRequest;
     state.txApproval.denomMetadataByAssetId = jsonReq.denomMetadataByAssetId;
     state.txApproval.responder = responder;
--- a/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
+++ b/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
@@ -10,7 +10,7 @@ export const isTxApprovalReq = (req: PopupRequest): req is TxApproval => {
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
     // @ts-expect-error Typescript doesn't like JsonValue could possibly be very deep
-    state.txApproval.tx = jsonReq;
+    state.txApproval.authorizeRequest = jsonReq;
     state.txApproval.responder = responder;
   });
 };
--- a/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
+++ b/apps/extension/src/routes/popup/approval/tx-msg-handler.ts
@@ -10,7 +10,8 @@ export const isTxApprovalReq = (req: PopupRequest): req is TxApproval => {
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
     // @ts-expect-error Typescript doesn't like JsonValue could possibly be very deep
-    state.txApproval.authorizeRequest = jsonReq;
+    state.txApproval.authorizeRequest = jsonReq.authorizeRequest;
+    state.txApproval.denomMetadataByAssetId = jsonReq.denomMetadataByAssetId;
     state.txApproval.responder = responder;
   });
 };
--- a/packages/router/src/ctx/approver.ts
+++ b/packages/router/src/ctx/approver.ts
@@ -1,8 +1,14 @@
 import { sendPopupRequest, spawnDetachedPopup } from '@penumbra-zone/types/src/internal-msg/popup';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1alpha1/custody_pb';
+import { JsonValue } from '@bufbuild/protobuf';
 
-export const getTxApproval = async (req: AuthorizeRequest): Promise<void> => {
+type DenomMetadataAsJson = JsonValue;
+
+export const getTxApproval = async (
+  req: AuthorizeRequest,
+  denomMetadataByAssetId: Record<string, DenomMetadataAsJson>,
+): Promise<void> => {
   await spawnDetachedPopup('popup.html#/approval/tx');
 
   /**
@@ -11,7 +17,10 @@ export const getTxApproval = async (req: AuthorizeRequest): Promise<void> => {
    */
   const res = await sendPopupRequest<TxApproval>({
     type: 'TX-APPROVAL',
-    request: req.toJson(),
+    request: {
+      authorizeRequest: req.toJson(),
+      denomMetadataByAssetId,
+    },
   });
   if ('error' in res) throw res.error;
   if (!res.data) throw new Error('Transaction was not approved');
--- a/packages/router/src/ctx/approver.ts
+++ b/packages/router/src/ctx/approver.ts
@@ -1,13 +1,12 @@
 import { sendPopupRequest, spawnDetachedPopup } from '@penumbra-zone/types/src/internal-msg/popup';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1alpha1/custody_pb';
-import { JsonValue } from '@bufbuild/protobuf';
-
-type DenomMetadataAsJson = JsonValue;
+import { Jsonified } from '@penumbra-zone/types';
+import { DenomMetadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
 
 export const getTxApproval = async (
   req: AuthorizeRequest,
-  denomMetadataByAssetId: Record<string, DenomMetadataAsJson>,
+  denomMetadataByAssetId: Record<string, Jsonified<DenomMetadata>>,
 ): Promise<void> => {
   await spawnDetachedPopup('popup.html#/approval/tx');
 
@@ -18,7 +17,7 @@ export const getTxApproval = async (
   const res = await sendPopupRequest<TxApproval>({
     type: 'TX-APPROVAL',
     request: {
-      authorizeRequest: req.toJson(),
+      authorizeRequest: req.toJson() as Jsonified<AuthorizeRequest>,
       denomMetadataByAssetId,
     },
   });
--- a/packages/router/src/ctx/approver.ts
+++ b/packages/router/src/ctx/approver.ts
@@ -5,6 +5,10 @@ import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumb
 export const getTxApproval = async (req: AuthorizeRequest): Promise<void> => {
   await spawnDetachedPopup('popup.html#/approval/tx');
 
+  /**
+   * @todo: Should this include a request ID so as not to cross wires with other
+   * requests?
+   */
   const res = await sendPopupRequest<TxApproval>({
     type: 'TX-APPROVAL',
     request: req.toJson(),
--- a/packages/router/src/grpc/custody/authorize.ts
+++ b/packages/router/src/grpc/custody/authorize.ts
@@ -4,7 +4,7 @@ import { approverCtx, extLocalCtx, extSessionCtx, servicesCtx } from '../../ctx'
 import { generateSpendKey, authorizePlan } from '@penumbra-zone/wasm-ts';
 
 import { Key } from '@penumbra-zone/crypto-web';
-import { Box, Jsonified, uint8ArrayToBase64 } from '@penumbra-zone/types';
+import { Box, Jsonified, bech32AssetId } from '@penumbra-zone/types';
 
 import { ConnectError, Code, HandlerContext } from '@connectrpc/connect';
 import { DenomMetadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
@@ -15,8 +15,7 @@ const getDenomMetadataByAssetId = async (ctx: HandlerContext) => {
   const assetsMetadata = await walletServices.indexedDb.getAllAssetsMetadata();
   return assetsMetadata.reduce<Record<string, Jsonified<DenomMetadata>>>((prev, curr) => {
     if (curr.penumbraAssetId) {
-      prev[uint8ArrayToBase64(curr.penumbraAssetId.inner)] =
-        curr.toJson() as Jsonified<DenomMetadata>;
+      prev[bech32AssetId(curr.penumbraAssetId)] = curr.toJson() as Jsonified<DenomMetadata>;
     }
     return prev;
   }, {});
--- a/packages/router/src/grpc/custody/authorize.ts
+++ b/packages/router/src/grpc/custody/authorize.ts
@@ -1,12 +1,15 @@
 import type { Impl } from '.';
-import { approverCtx, extLocalCtx, extSessionCtx } from '../../ctx';
+import { approverCtx, extLocalCtx, extSessionCtx, servicesCtx } from '../../ctx';
 
 import { generateSpendKey, authorizePlan } from '@penumbra-zone/wasm-ts';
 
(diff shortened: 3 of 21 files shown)


packages/types/src/transaction/get-stub-action-view-from-plan.ts
```diff
@@ -0,0 +1,122 @@
...
+      case: 'visible',
+
+      value: {
+        note: {
+          value: getValueView(outputPlan.value, denomMetadataByAssetId),
+          address: {
+            addressView: {
+              case: 'opaque',
+              value: {
+                address: outputPlan.destAddress,
```
jessepinho, line 91: @hdevalence is this the correct mapping of the output plan's `destAddress` to the `outputView`? The `outputPlan` has a _top-level_ `destAddress` property next to a top-level `value` property, and I'm putting them both into the `note` of this `OutputView` instance. That feels wrong to me somehow, but I don't know enough about these object structures yet to know if that's incorrect. So I want to make sure I'm not misrepresenting the action to the user.

(This PR in general isn't ready for review yet, but I wanted to make this comment before I forget.)

packages/types/src/transaction/get-stub-action-view-from-plan.ts
```diff
@@ -0,0 +1,144 @@
...
+        /**
+         * Special case -- the `withdrawal` case in the action plan maps to the
+         * `ics20Withdrawal` case in the action view.
+         */
+        return new ActionView({
+          actionView: {
+            case: 'ics20Withdrawal',
+            value: {},
+          },
+        });
```
jessepinho, line 129: @hdevalence or @turbocrime would you mind checking my logic here?

It appears from the Protobufs that in the context of an action _plan_, an ICS20 Withdrawal has a case of `withdrawal` ([code](https://github.com/penumbra-zone/penumbra/blob/main/proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto#L255)). But in the context of an [_action_](https://github.com/penumbra-zone/penumbra/blob/main/proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto#L98) or an [_action view_](https://github.com/penumbra-zone/penumbra/blob/main/proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto#L188), it has a case of `ics20_withdrawal` (which is then camel-cased in JavaScript). So out of all of the different action cases, this is the only one that needs a special translation when going from action plan -> action view.

Let me know if I'm missing something, or if this is correct.

packages/router/src/ctx/approver.ts
```diff
@@ -1,13 +1,26 @@
 import { sendPopupRequest, spawnDetachedPopup } from '@penumbra-zone/types/src/internal-msg/popup';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1alpha1/custody_pb';
+import { JsonValue } from '@bufbuild/protobuf';
 
-export const getTxApproval = async (req: AuthorizeRequest): Promise<void> => {
+type DenomMetadataAsJson = JsonValue;
```
turbocrime, line 6: just use jsonvalue

packages/ui/components/ui/tx/view/action-view.tsx
```diff
@@ -3,6 +3,38 @@ import { SpendViewComponent } from './spend';
 import { OutputViewComponent } from './output';
```
jessepinho, line 3: The changes in this file are totally optional, and aren't strictly related to what the PR is about. But I found myself referring back to this component a lot, and realized it was a bit easier to understand if rewritten this way. That said, that could just be me, and I'm happy to revert this change if it's undesirable.

packages/router/src/grpc/custody/authorize.ts
```diff
@@ -15,6 +18,19 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   const sess = ctx.values.get(extSessionCtx);
   const local = ctx.values.get(extLocalCtx);
 
+  const services = ctx.values.get(servicesCtx);
+  const walletServices = await services.getWalletServices();
+  const assetsMetadata = await walletServices.indexedDb.getAllAssetsMetadata();
+  const denomMetadataByAssetId = assetsMetadata.reduce<Record<string, DenomMetadataAsJson>>(
+    (prev, curr) => {
+      if (curr.penumbraAssetId) {
+        prev[uint8ArrayToBase64(curr.penumbraAssetId.inner)] = curr.toJson();
```
turbocrime, line 27: i think it might be better to use bech32 like in discussion, bech32 is the canonical string representation of an assetid

turbocrime (commented):
many of the places where you are constructing a pb type, you don't actually need the features the constructor provides (conversion methods, etc) and it may be preferable to target the PlainMessage<Whatever> type instead of using a constructor.

the constructor accepts a partialmessage, in which all fields are optional, and there is no type enforcement. but a plainmessage type requires all fields and enforces field type.

https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#plainmessage

packages/router/src/grpc/custody/authorize.ts
```diff
@@ -15,6 +18,19 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   const sess = ctx.values.get(extSessionCtx);
   const local = ctx.values.get(extLocalCtx);
 
+  const services = ctx.values.get(servicesCtx);
+  const walletServices = await services.getWalletServices();
+  const assetsMetadata = await walletServices.indexedDb.getAllAssetsMetadata();
+  const denomMetadataByAssetId = assetsMetadata.reduce<Record<string, DenomMetadataAsJson>>(
```
turbocrime, line 24: request the metadata after the throws that are close to the approveReq call, we might bail out before we need any of it

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
```diff
@@ -10,7 +10,8 @@ export const isTxApprovalReq = (req: PopupRequest): req is TxApproval => {
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
     // @ts-expect-error Typescript doesn't like JsonValue could possibly be very deep
-    state.txApproval.tx = jsonReq;
+    state.txApproval.authorizeRequest = jsonReq.authorizeRequest;
+    state.txApproval.denomMetadataByAssetId = jsonReq.denomMetadataByAssetId;
```
turbocrime, line 14: it might be reasonable to do something like AuthorizeRequest.fromJson() here to 'deserialize'

turbocrime:
nice progress

packages/types/src/transaction/get-stub-action-view-from-plan.ts
hdevalence: I think this is downstream of the field naming in the .proto files and is the kind of minor inconsistency that would be good to fix asap before we stabilize formats and are stuck with it forever

jessepinho: Filed an issue [here](https://github.com/penumbra-zone/penumbra/issues/3614).

packages/router/src/grpc/custody/authorize.ts
jessepinho: done

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
jessepinho: We actually never ended up needing the JSON value of `AuthorizeRequest` — we just display it in a JSONViewer.

packages/router/src/grpc/custody/authorize.ts
jessepinho: Ah OK, done!

packages/types/src/transaction/get-stub-action-view-from-plan.ts
jessepinho: also CC @turbocrime

packages/types/src/transaction/view-transaction-plan/view-action-plan.test.ts
```diff
@@ -0,0 +1,286 @@
...
+      },
+    });
+
+    test('includes the destAddress', () => {
+      const actionView = viewActionPlan(metadataByAssetId)(validOutputActionPlan);
+      const outputView = actionView.actionView.value as OutputView;
+      const outputViewVisible = outputView.outputView.value as OutputView_Visible;
+
+      expect(outputViewVisible.note?.address?.addressView.value?.address).toEqual(destAddress);
+    });
```
jessepinho, line 152: See [this comment](https://github.com/penumbra-zone/web/pull/366#discussion_r1451209146). Not sure if the `destAddress` should be put into the note... there's kind of a weird structure to the `output` action, where it has top-level `destAddress` and `value` keys, which I _think_ (but am not sure) should go in the `OutputView`'s `note` property... but I'm not sure if that's an accurate representation of the output action...?

packages/ui/components/ui/tx/view/output.tsx
```diff
@@ -10,7 +10,7 @@ export const OutputViewComponent = ({ value }: { value: OutputView }) => {
       <ViewBox
         label='Output'
         visibleContent={
-          <div className='flex justify-between md:flex-col lg:flex-row'>
+          <div className='flex flex-col justify-between gap-2 sm:flex-row sm:gap-0'>
```
jessepinho, line 13: This (and the similar change below) makes the address move to below the amount on small screens

packages/types/src/transaction/view-transaction-plan/view-action-plan.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+};
+
+const getOutputView = (
+  outputPlan: OutputPlan,
+  denomMetadataByAssetId: Record<string, Jsonified<DenomMetadata>>,
+): OutputView => {
+  if (!outputPlan.destAddress) throw new Error('No destAddress in output plan');
+
+  return new OutputView({
```
jessepinho, line 84: Is there any add'l data we need in this `OutputView`? Didn't look like it to me, but want to make sure I'm not missing something.

hdevalence:
> Note that there's a 1 test_usd spend action — I'm not sure where that's coming from, or why it's being spent?

@jessepinho I believe this as an artifact of the current note selection logic. The transaction planner needs to construct a set of notes that will be sufficient to cover the requested amount, and I believe it currently does this in a naive way, picking notes one at a time. So my guess is that it selected the smaller note first, then spent another note to get enough balance.

packages/types/src/transaction/get-stub-action-view-from-plan.ts
hdevalence: Yes, this is actually correct. The `OutputPlan` contains all the data that will be used to generate the note.

The `Output` action consists of:

- a note commitment to the newly created note
- an encryption of the note itself, decryptable by both the sender and receiver
- a zk proof that the note commitment was correctly formed

So the reason that the output plan has the data separately (value, address) is that that data is used for each of those parts. 

Here where we're making only a "skeletonized" version, this is the correct translation.

hdevalence (approved):
This looks great! 

I think there are some further improvements to transaction comprehension that we should add, but not in this PR, we can add them in later changes

packages/router/src/ctx/approver.ts
```diff
@@ -1,13 +1,25 @@
...
+export const getTxApproval = async (
+  req: AuthorizeRequest,
+  denomMetadataByAssetId: Record<string, Jsonified<DenomMetadata>>,
+): Promise<void> => {
   await spawnDetachedPopup('popup.html#/approval/tx');
 
+  /**
+   * @todo: Should this include a request ID so as not to cross wires with other
+   * requests?
+   */
```
turbocrime, line 16: yes - i think we should implement this in this PR. just use crypto.getrandomuuid to make one at the moment the request is sent, and check that the response contains the same uuid.

turbocrime: in the zustand slice, we should
 
1) never return the uuid as part of a selector
2) throw if we are assigning data but a uuid is already present
3) clear the uuid when resolving the approval

packages/types/src/transaction/view-transaction-plan/view-action-plan.test.ts
turbocrime: did you find an answer on this?

packages/types/src/transaction/view-transaction-plan/view-action-plan.ts
```diff
@@ -0,0 +1,154 @@
...
+import { Jsonified } from '../../jsonified';
+import { bech32AssetId } from '../../asset';
+
+const getValueView = (
+  value: Value | undefined,
+  denomMetadataByAssetId: Record<string, Jsonified<DenomMetadata>>,
+): ValueView => {
+  if (!value) throw new Error('No value to view');
+  if (!value.assetId) throw new Error('No asset ID in value');
+  if (!value.amount) throw new Error('No amount in value');
```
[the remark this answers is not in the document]
turbocrime: you could potentially set an error.cause, attaching the entire Value, to convey the specific structure

turbocrime (approved):
one request on the uuid but that can be deferred to another issue, merge if you're satisfied

packages/types/src/transaction/view-transaction-plan/view-action-plan.test.ts
jessepinho: Yep, Henry answered it above.

jessepinho:
@turbocrime since the lack of request IDs is an existing issue, and because I don't fully understand yet what circumstances would require request IDs, I'm going to merge this as-is. I created #377 to address your request.

PR #367, opened 2024-01-15
210 ownedpositionidsrequest

Valentine1898:


Valentine1898:
I need to add the `POSITIONS` table to indexed-db and add saving the position data to` block-processor`, but it makes sense to do it after the refactoring in #357 is complete

grod220:
`0.65.0-pre.1` published to npm with the new `compute_position_id()` available to use

Valentine1898:
image
It's weird, broken tests that aren't related to this PR

packages/query/src/block-processor.ts
```diff
@@ -29,6 +29,7 @@ import {
   TransactionInfo,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 import { StateCommitment } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/crypto/tct/v1alpha1/tct_pb';
+import { computePositionId } from '@penumbra-zone/wasm-ts/src/dex';
```
grod220, line 32: Let's add `export * from './src/dex';` to `packages/wasm/index.ts` and we can shorten the import like the others

packages/storage/src/indexed-db/index.ts
```diff
@@ -340,4 +347,46 @@ export class IndexedDb implements IndexedDbInterface {
...
+      const positionId = PositionId.fromJson(positionCursor.value.id);
+      const position = Position.fromJson(positionCursor.value.position);
+
+      if (positionState && !positionState.equals(position.state)) {
+        continue;
+      }
+      if (tradingPair && !tradingPair.equals(position.phi?.pair)) {
+        continue;
+      }
+      ownedPositionIds.push(positionId);
```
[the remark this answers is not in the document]
Valentine1898: Good idea, will also have to revise tests

Valentine1898:
a lot of files changed just because of lint fix from #388

packages/router/src/grpc/view-protocol-server/index.ts
```diff
@@ -27,7 +28,7 @@ import { walletId } from './wallet-id';
 import { witness } from './witness';
 import { witnessAndBuild } from './witness-and-build';
 
-export const viewImpl: Omit<Impl, 'ownedPositionIds'> = {
+export const viewImpl: Impl = {
```
grod220, line 31: giphy

packages/storage/src/indexed-db/index.ts
```diff
@@ -349,4 +356,48 @@ export class IndexedDb implements IndexedDbInterface {
...
+
+      if (positionState && !positionState.equals(position.state)) {
+        continue;
+      }
+      if (tradingPair && !tradingPair.equals(position.phi?.pair)) {
+        continue;
+      }
+      yield positionId;
+    }
+  }
```
grod220, line 376: Fields should be required here:


```typescript
  async *getOwnedPositionIds(
    positionState: PositionState,
    tradingPair: TradingPair,
  ): AsyncGenerator<PositionId, void> {
    for await (const positionCursor of this.db.transaction('POSITIONS').store) {
      const position = Position.fromJson(positionCursor.value.position);

      if (!positionState.equals(position.state) || !tradingPair.equals(position.phi?.pair)) {
        continue;
      }

      yield PositionId.fromJson(positionCursor.value.id);
    }
  }
```

packages/router/src/grpc/view-protocol-server/ownned-position-ids.ts
```diff
@@ -0,0 +1,13 @@
+import type { Impl } from '.';
+import { servicesCtx } from '../../ctx';
+
+export const ownedPositionIds: Impl['ownedPositionIds'] = async function* (req, ctx) {
+  const services = ctx.values.get(servicesCtx);
+
+  const { indexedDb } = await services.getWalletServices();
+
+  const positionIds = indexedDb.getOwnedPositionIds(req.positionState, req.tradingPair);
```
grod220, line 9: We should do parameter validation here:

```typescript
  if (!req.positionState) throw new Error('Missing position state in request');
  if (!req.tradingPair) throw new Error('Missing trading pair in request');

  const positionIds = indexedDb.getOwnedPositionIds(req.positionState, req.tradingPair);
  
```

Valentine1898: But `positionState` and `tradingPair` should not be required. 
If they are required, the client will have to brute force all combinations of trading pairs and state to get all positions

Valentine1898: https://github.com/penumbra-zone/penumbra/blob/a483d8e377ea6728f865d844f9adc4e7fe8f8827/crates/view/src/storage.rs#L1655 
In rust, they are also not required

grod220: Ah, I misread the protos, thanks for correcting!

packages/storage/src/indexed-db/index.ts
```diff
@@ -349,4 +356,48 @@ export class IndexedDb implements IndexedDbInterface {
     }
     return Promise.resolve(notesForVoting);
   }
+
+  async *getOwnedPositionIds(
+    positionState: PositionState | undefined,
+    tradingPair: TradingPair | undefined,
+  ): AsyncGenerator<PositionId, void> {
+    for await (const positionCursor of this.db.transaction('POSITIONS').store) {
+      const positionId = PositionId.fromJson(positionCursor.value.id);
```
grod220, line 365: Think we can still drop this below so it doesn't deserialize unnecessarily:

```typescript
yield PositionId.fromJson(positionCursor.value.id);
```

grod220 (approved):
Nice job! Just one additional change I'd say.

packages/router/src/grpc/view-protocol-server/index.ts
TalDerei: omit = null

PR #368, opened 2024-01-15
Jsonified

turbocrime:
instead of aliasing jsonvalue, actually derive json-compatible type information

use it like:

```ts
const x = new AnyMessage({});
const xJson = x.toJson() as Jsonified<AnyMessage>;
```

### DIFF
--- a/apps/extension/src/offscreen.ts
+++ b/apps/extension/src/offscreen.ts
@@ -1,4 +1,3 @@
-import { Action } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import type { JsonObject, JsonValue } from '@bufbuild/protobuf';
 import {
   OffscreenRequest,
@@ -6,7 +5,6 @@ import {
   WasmBuildActionInput,
   isActionBuildRequest,
 } from '@penumbra-zone/types/src/internal-msg/offscreen';
-import { Jsonified } from '@penumbra-zone/types/src/internal-msg/shared';
 
 export const isOffscreenRequest = (req: unknown): req is OffscreenRequest =>
   req != null &&
@@ -49,16 +47,16 @@ export const buildActionHandler = (request: ActionBuildRequest) => {
 };
 
 const spawnWorker = (
-  transactionPlan: JsonObject & { actions: Jsonified<Action>[] },
+  transactionPlan: JsonObject & { actions: JsonValue[] },
   witness: JsonObject,
   fullViewingKey: string,
   actionPlanIndex: number,
-): Promise<Jsonified<Action>> => {
+): Promise<JsonValue> => {
   return new Promise((resolve, reject) => {
     const worker = new Worker(new URL('./wasm-build-action.ts', import.meta.url));
 
     const onWorkerMessage = (e: MessageEvent) => {
-      resolve(e.data as Jsonified<Action>);
+      resolve(e.data as JsonValue);
       worker.removeEventListener('error', onWorkerError);
       worker.terminate();
     };
--- a/apps/extension/src/wasm-build-action.ts
+++ b/apps/extension/src/wasm-build-action.ts
@@ -1,10 +1,9 @@
 import {
-  Action,
   TransactionPlan,
   WitnessData,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import type { WasmBuildActionInput } from '@penumbra-zone/types/src/internal-msg/offscreen';
-import { Jsonified } from '@penumbra-zone/types/src/internal-msg/shared';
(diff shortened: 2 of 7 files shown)


packages/types/src/jsonified.ts
turbocrime, line 1: jsonified.ts is the important part

currently, putting bufbuild message types in here requires wrapping in `PlainMessage<T>` first, maybe we could add support for doing that automatically

turbocrime:
minimized for @jessepinho use, but more wip available

packages/types/src/jsonified.ts
turbocrime: ok, added that

packages/types/src/jsonified.ts
```diff
@@ -0,0 +1,16 @@
...
+      [P in keyof T as T[P] extends Function ? never : P]: JsonField<T[P]>;
+    };
+
+// prettier-ignore
+type JsonField<F> = F extends (JsonValue) ? F
+                  : F extends (Date| Uint8Array | bigint) ? string
+                  : F extends (infer U)[] ? JsonField<U>[]
+                  : F extends Message<infer U> ? Jsonified<PlainMessage<U>>
+                  : F extends object ? Jsonified<F>
+                  : F;
```
jessepinho: It's... it's beautiful 🥺

packages/types/src/indexed-db.ts
```diff
@@ -93,27 +91,27 @@ export interface PenumbraDb extends DBSchema {
   // ======= Json serialized values =======
   // Allows wasm crate to directly deserialize
   ASSETS: {
-    key: Base64Str; // Jsonified<DenomMetadata['penumbraAssetId']['inner']>
-    value: Jsonified<DenomMetadata>;
+    key: Base64Str; // DenomMetadata['penumbraAssetId']['inner']
+    value: JsonValue;
```
jessepinho, line 95: Why change these to `JsonValue`? Isn't e.g., `Jsonified<DenomMetadata>` more useful now that we have a proper `Jsonified` type?

apps/extension/src/wasm-build-action.ts
```diff
@@ -43,7 +42,7 @@ async function executeWorker(
   witness: WitnessData,
   fullViewingKey: string,
   actionPlanIndex: number,
-): Promise<Jsonified<Action>> {
+): Promise<JsonValue> {
```
jessepinho, line 45: Same Q here — don't we want `Jsonified<Action>`?

apps/extension/src/offscreen.ts
```diff
@@ -49,16 +47,16 @@ export const buildActionHandler = (request: ActionBuildRequest) => {
...
   fullViewingKey: string,
   actionPlanIndex: number,
-): Promise<Jsonified<Action>> => {
+): Promise<JsonValue> => {
   return new Promise((resolve, reject) => {
     const worker = new Worker(new URL('./wasm-build-action.ts', import.meta.url));
 
     const onWorkerMessage = (e: MessageEvent) => {
-      resolve(e.data as Jsonified<Action>);
+      resolve(e.data as JsonValue);
```
jessepinho, line 59: Same Q here

jessepinho (commented):
Not sure I understand the changes to using `JsonValue` instead of `Jsonified<X>` — could you please explain this change before I approve? Otherwise, the `Jsonified` type looks 🔥

packages/types/src/indexed-db.ts
turbocrime: the upgrade has complex downstream effects :) for now it's just stubbed out

apps/extension/src/offscreen.ts
turbocrime: events arriving here are technically 'unknown' and must be inspected with a type guard

apps/extension/src/wasm-build-action.ts
turbocrime: we could probably assert that, but at the return statement we're calling a function that really does return `JsonValue` explicitly

jessepinho (approved):
OK, LGTM! Looks like there are some build/lint issues though — once those are green, ship it 🚢

grod220 (commented):
I'm a bit on the fence with this PR. Few things:

- We really shouldn't be interacting with json-serialized-object fields directly. We almost should pretend they are bytes. It's really the deserialized classes that are the first class citizens.
- The new `Jsonified` type may prove to be difficult to maintain.
- It doesn't appear that JsonValue === Jsonified<T> anymore. This is evident by the fact we can't use it to type the indexeddb values anymore.

Given this, I don't see a big benefit with expanding the type. Am I missing something?

jessepinho:
Not sure what you mean by "expanding the type," but the idea here is that `Jsonified<X>` is more useful than `JsonValue` since we know what shape it takes. That said, if we shouldn't be interacting with JSON-serialized objects directly, I suppose it's not that useful except as documentation of what object type it's serialized from.

PR #371, opened 2024-01-18
Notes for voting request

Valentine1898:


Valentine1898:
This PR does not require modification of `indexed-db`, all data for notes for voting is already stored

Valentine1898:
The main difficulty is that we need to read data from several tables by linking them by foreign keys 
https://github.com/penumbra-zone/penumbra/blob/eccf899efc07f2644d415dc70b1339a571d25c5a/crates/view/src/storage.rs#L1016
```
let spendable_note_records: Vec<SpendableNoteRecord> = dbtx
                .prepare(&format!(
                    "SELECT notes.note_commitment,
                        spendable_notes.height_created,
                        notes.address,
                        notes.amount,
                        notes.asset_id,
                        notes.rseed,
                        spendable_notes.address_index,
                        spendable_notes.source,
                        spendable_notes.height_spent,
                        spendable_notes.nullifier,
                        spendable_notes.position
                    FROM
                        notes JOIN spendable_notes ON notes.note_commitment = spendable_notes.note_commitment
                    WHERE
                        spendable_notes.address_index IS {address_clause}
                        AND notes.asset_id IN (
                            SELECT asset_id FROM assets WHERE denom LIKE '_delegation\\_%' ESCAPE '\\'
                        )
                        AND ((spendable_notes.height_spent IS NULL) OR (spendable_notes.height_spent > {votable_at_height}))
                        AND (spendable_notes.height_created < {votable_at_height})
                    ",
                ))?
                .query_and_then((), |row| row.try_into())?
                .collect::<anyhow::Result<Vec<_>>>()?;
``` 
Rust uses sqllite
Sqllite is a relational database and it looks easy, but indexed-db is not relational

Valentine1898:
The simple solution is that we can just read all the data from the `NOTES` and `ASSETS` tables and do in-memory selections and filtering 
In fact, up to this point, we've always done in-memory filtering for indexed-db data

Valentine1898:
But really, we're better off avoiding in-memory filtering. See #312

Valentine1898:
cc @jessepinho 
We didn't get a chance to work in pairs, so I'll try to reveal more details on working with this issue

Valentine1898:
The important point is that besides `SpendableNotes` we also need to get `IdentityKey` 
In rust it looks like this:
```rust
let identity_key = DelegationToken::from_str(&denom)
                    .context("invalid delegation token denom")?
                    .validator();
``` 
And good news, we can do this without wasm.  `udelegation_penumbravalid13m3vq62xwtjmesxyngayanp3tnllm2rta7ylc8pkz2vwqsag0v9snujwrz` is an example delegation asset, and it looks like the substring `valid13m3vq62xwtjmesxyngayanp3tnllm2rta7ylc8pkz2vwqsag0v9snujwrz` is a validator identity key in `bech32m` encoding

packages/storage/src/indexed-db/index.ts
```diff
@@ -254,4 +255,54 @@ export class IndexedDb implements IndexedDbInterface {
...
+      if (
+        relevantAssets.includes(note.note?.value?.assetId) &&
+        (note.heightSpent === 0n || note.heightSpent > votable_at_height) &&
+        note.heightCreated < votable_at_height
+      ) {
+        result.push(note);
+      }
+
+      noteCursor = await noteCursor.continue();
+    }
```
Valentine1898, line 304: wip: filtering with idb cursor is ready, but we still need to get IdentityKey from Asset

packages/storage/src/indexed-db/index.ts
```diff
@@ -254,4 +255,54 @@ export class IndexedDb implements IndexedDbInterface {
...
+    votable_at_height: bigint,
+  ): Promise<[]> {
+    const relevantAssets = [];
+
+    // Fetch all assets and filter out the relevant asset IDs
+    let assetCursor = await this.db.transaction('ASSETS').store.openCursor();
+
+    while (assetCursor) {
+      // Test denom with a regular expression
+      const regex = new RegExp('_delegation_.*');
```
jessepinho, line 270: I know this is still a draft PR, but I would rename this to indicate what the regex is looking for. For example:
```suggestion
      const isDelegationAssetType = new RegExp('_delegation_.*');
```

packages/storage/src/indexed-db/index.ts
```diff
@@ -254,4 +255,54 @@ export class IndexedDb implements IndexedDbInterface {
...
+        if (!note.addressIndex?.equals(addressIndex)) {
+          noteCursor = await noteCursor.continue();
+          continue;
+        }
+      }
+      // Check if asset_id exists in relevantAssets, and if height is < votable_at_height
+      if (
+        relevantAssets.includes(note.note?.value?.assetId) &&
+        (note.heightSpent === 0n || note.heightSpent > votable_at_height) &&
+        note.heightCreated < votable_at_height
```
jessepinho, line 298: Consider naming these conditions (especially the second one) with something human-friendly. I don't know enough about the context here to know what the _meaning_ of these conditions are, so I don't know if I can suggest something. But the idea is something like:

```TS
  const noteIsVotable = (note.heightSpent === 0n || note.heightSpent > votable_at_height);

  if (
        relevantAssets.includes(note.note?.value?.assetId) &&
        noteIsVotable &&
        note.heightCreated < votable_at_height
  ) { ... }
```

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
...
+        continue;
+      }
+
+      const isRelevantAsset = relevantAssets.has(uint8ArrayToHex(note.note.value.assetId.inner));
+      const noteIsVotable = note.heightSpent === 0n || note.heightSpent > votable_at_height;
+
+      if (isRelevantAsset && noteIsVotable && note.heightCreated < votable_at_height) {
+        const asset = relevantAssets.get(uint8ArrayToHex(note.note.value.assetId.inner));
+
+        const bech32idk = asset?.base.replace('udelegation_', '');
```
grod220, line 318: what's idk stand for?

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
...
+
+      const denomMetadata = DenomMetadata.fromJson(assetCursor.value);
+
+      if (isDelegationAssetType.test(denomMetadata.base) && denomMetadata.penumbraAssetId) {
+        relevantAssets.set(uint8ArrayToHex(denomMetadata.penumbraAssetId.inner), denomMetadata);
+      }
+      assetCursor = await assetCursor.continue();
+    }
+
+    const notesForVoting = [];
```
grod220, line 294: please annotate with type

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
...
+      }
+
+      const isRelevantAsset = relevantAssets.has(uint8ArrayToHex(note.note.value.assetId.inner));
+      const noteIsVotable = note.heightSpent === 0n || note.heightSpent > votable_at_height;
+
+      if (isRelevantAsset && noteIsVotable && note.heightCreated < votable_at_height) {
+        const asset = relevantAssets.get(uint8ArrayToHex(note.note.value.assetId.inner));
+
+        const bech32idk = asset?.base.replace('udelegation_', '');
+        if (bech32idk) {
```
grod220, line 319: Probably should throw here instead:

```typescript
if (!bech32idk) throw new Error("expected delegation token identity key not present")
```

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
...
     await this.u.update({ table: 'GAS_PRICES', value, key: 'gas_prices' });
   }
+
+  async getNotesForVoting(
+    addressIndex: AddressIndex | undefined,
+    votable_at_height: bigint,
+  ): Promise<NotesForVotingResponse[]> {
+    const relevantAssets = new Map<string, DenomMetadata>();
+
+    let assetCursor = await this.db.transaction('ASSETS').store.openCursor();
```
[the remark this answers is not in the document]
grod220: And then consume it with a `for await of...`

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
   async saveGasPrices(value: GasPrices): Promise<void> {
     await this.u.update({ table: 'GAS_PRICES', value, key: 'gas_prices' });
   }
+
+  async getNotesForVoting(
+    addressIndex: AddressIndex | undefined,
+    votable_at_height: bigint,
```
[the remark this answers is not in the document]
Valentine1898: Oh, how did I miss that

packages/storage/src/indexed-db/index.ts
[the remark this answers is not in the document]
Valentine1898: that makes sense

packages/storage/src/indexed-db/index.ts
Valentine1898: In this case one broken note will break the query, but I think this is correct, we shouldn't return data if some of it is broken

packages/storage/src/indexed-db/index.ts
```diff
@@ -264,4 +271,63 @@ export class IndexedDb implements IndexedDbInterface {
...
+    }
+
+    const notesForVoting = [];
+
+    let noteCursor = await this.db.transaction('SPENDABLE_NOTES').store.openCursor();
+    while (noteCursor) {
+      const note = SpendableNoteRecord.fromJson(noteCursor.value);
+
+      if (addressIndex) {
+        if (!note.addressIndex?.equals(addressIndex)) {
```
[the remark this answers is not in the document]
Valentine1898: I think it makes sense to add indexes for those fields for which we most often make selections. I think it would be better to do it in #312 for all tables

packages/storage/src/indexed-db/index.ts
```diff
@@ -87,6 +95,18 @@ export class IndexedDb implements IndexedDbInterface {
...
+  ) {
+    const tx = this.db.transaction(table);
+    let cursor = await tx.store.openCursor();
+
+    while (cursor) {
+      yield cursor.value;
+      cursor = await cursor.continue();
+    }
+  }
+
```
grod220, line 109: Ah, this kinda relates to https://github.com/penumbra-zone/web/issues/312 right? From this issue, there is a link to: https://github.com/jakearchibald/idb#idbcursor-enhancements. It looks like we can just do `for await` directly on the store?

We may not need this helper then?

```typescript
const tx = db.transaction(storeName);

for await (const cursor of tx.store) {
  // …
}
```

Valentine1898: Yeah, that turned out to be easier, we can just iterate over stores

grod220 (approved):
Nice job!

PR #372, opened 2024-01-18
Use a hash router

jessepinho:
This allows us to ship the web app as a static site that can be hosted by anyone anywhere, without any special server config required to support URL rewriting.

## Reviewers
- Do we want to also remove Firebase config from the `apps/webapp`, since that's specific to Penumbra Labs' own setup and consumers of this repo shouldn't have to worry about Firebase? (This would require reconfiguring our CI to apply its own config to Firebase, since `firebase.json` would be gone.) Or, are we OK with leaving it, since consumers of this repo can simply run `pnpm build` and get a statically-hostable version of the web app, which they can host wherever they want?

### DIFF
--- a/apps/webapp/src/components/root-router.tsx
+++ b/apps/webapp/src/components/root-router.tsx
@@ -1,4 +1,4 @@
-import { createBrowserRouter, redirect } from 'react-router-dom';
+import { createHashRouter, redirect } from 'react-router-dom';
 import { PagePath } from './metadata/paths.ts';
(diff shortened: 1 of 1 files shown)


turbocrime:
lmao

turbocrime:
delete firebase config y/n

turbocrime:
we do have some dynamic imports that i think need vite.config update

turbocrime (approved):

PR #378, opened 2024-01-18
Extend transaction approval dialog to show address visibility

### LINKED ISSUE #375: Extend transaction approval dialog to show address visibility
The initial approval dialog added in #366 doesn't allow users to clearly identify what funds are going where. The reason is that it shows all addresses as the "opaque" variant of the address view, meaning that it doesn't allow users to easily identify which funds are going back to their own wallet (as change) and which funds are going out to their counterparty.

To fix this, we should change all of the address views used in the skeleton `TransactionView` used to render the `TransactionPlan` to reflect the user's level of visibility into that address.

Recall that every Penumbra address contains encrypted data, and a Penumbra FVK (full viewing key) can be used to inspect an address and determine whether its structure is visible with that viewing key.

One idea of how to implement this would be:

- define a function in `view-action-plan.ts` that takes an address and an FVK and returns an `AddressView`
- that method can use the [`is_controlled_address`](https://rustdoc.penumbra.zone/main/penumbra_wasm/keys/fn.is_controlled_address.html) wasm method to populate the `AddressView` structure (either opaque, with just the address, or visible, with both address and address index, we can ignore the wallet id in this context)
- wrap all of the addresses used to construct the skeleton view structures with the viewing function

To make this work, we'll need to thread an FVK into the top-level function that turns a `TransactionPlan` into the skeleton `TransactionView`.

jessepinho:
## Background
In #366, the extension's transaction approval dialog was modified to render a `TransactionView` generated from the `TransactionPlan` it was passed.

One shortcoming of that PR was that addresses in the transaction view were all rendered as "opaque" — that is, as though the current user had no access to or knowledge about them. That meant that the approval dialog wouldn't be as intuitive to a user: they wouldn't readily see which addresses were their own vs. someone else's, so it wouldn't be as clear when, for example, a given output was in fact change being sent back to their own address. (See #375 for more context on this.)

The only way to get richer information about an address (such as which account index it is) is to use the user's full viewing key, then pass it to `isControlledAddress` (from `@penumbra-zone/wasm-ts`) to check whether the current user controls this address.

## In this PR
- Modify the message we send from the `authorize` impl to the popup: now, it contains the `ApproveRequest` and the `TransactionView` (which is generated from the approve request's `plan` property), rather than the `ApproveRequest` and the map of asset IDs to `DenomMetadata`s.
- Rework `viewTransactionPlan` to get an address index when possible via `isControlledAddress`.
- Move the `viewTransactionPlan` function and its related files to the `router` package. This is because this function now uses the `isControlledAddress` import from `@penumbra-zone/wasm-ts`. If I'd left `viewTransactionPlan` inside the `types` package, it would have created a cyclical dependency between `@penumbra-zone/types` and `@penumbra-zone/wasm-ts`, which breaks the CI build.
  - Note that this also has the effect of moving the translation of `TransactionPlan` -> `TransactionView` from the popup to the `authorize` impl. We'd discussed on Discord whether it was safe for the full viewing key to be sent to the popup for the purpose of that translation; Henry ultimately decided that it was indeed safe. However, that ended up being a moot point, because we needed to move the `viewTransactionPlan` function into the `router` package anyway due to the cyclical dependency described above.
- Get rid of the tests for `transactionViewSelector`, which no longer exists.

## Before (note the addresses in the popup)
image

## After (note the addresses in the popup)
image

Closes #375


apps/extension/src/state/tx-approval.test.ts
```diff
@@ -23,6 +23,7 @@ describe('TX Approval Slice', () => {
         'penumbra147mfall0zr6am5r45qkwht7xqqrdsp50czde7empv7yq2nk3z8yyfh9k9520ddgswkmzar22vhz9dwtuem7uxw0qytfpv7lk3q9dp8ccaw2fn5c838rfackazmgf3ahh09cxmz';
       const chainId = 'testnet';
       const expiryHeight = 100n;
+      const fullViewingKey = 'fvk';
```
hdevalence, line 26: https://github.com/penumbra-zone/penumbra/pull/3630 might be helpful here

hdevalence: The FVK corresponding to that return address is
```
penumbrafullviewingkey1vzfytwlvq067g2kz095vn7sgcft47hga40atrg5zu2crskm6tyyjysm28qg5nth2fqmdf5n0q530jreumjlsrcxjwtfv6zdmfpe5kqsa5lg09
```

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
```diff
@@ -10,7 +10,7 @@ export const isTxApprovalReq = (req: PopupRequest): req is TxApproval => {
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
     state.txApproval.authorizeRequest = jsonReq.authorizeRequest;
-    state.txApproval.denomMetadataByAssetId = jsonReq.denomMetadataByAssetId;
+    state.txApproval.transactionViewFromPlan = jsonReq.transactionViewFromPlan;
```
jessepinho, line 13: We're now translating the transaction plan into a transaction view inside the `authorize` impl [here](https://github.com/penumbra-zone/web/pull/378/files#diff-b2755ed7d656eafb2fa77b3ca380564a22ab97247124767e75f22fe233a191d7), so we have the `TransactionView` already once we're inside the popup. (Thus, we also no longer need `denomMetadataByAssetId` here, since it was previously being used to generate that `TransactionView` in the popup, but isn't any longer.)

apps/extension/src/state/tx-approval.ts
```diff
@@ -1,29 +1,17 @@
...
+  transactionViewFromPlan?: Jsonified<TransactionView>;
   // Holding the message responder function. Service worker will be "awaiting" the call of this.
   responder?: MessageResponder<TxApproval>;
 }
 
 export const createTxApprovalSlice: SliceCreator<TxApprovalSlice> = () => ({});
 
 export const txApprovalSelector = (state: AllSlices) => state.txApproval;
-
-export const transactionViewSelector = (state: AllSlices) => {
```
jessepinho, line 21: Got rid of this since we're no longer translating the plan -> view inside the popup

packages/router/src/grpc/custody/authorize.ts
```diff
@@ -32,14 +33,27 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
...
 
   const wallets = await local.get('wallets');
-  const { encryptedSeedPhrase } = wallets[0]!.custody;
+  const {
+    custody: { encryptedSeedPhrase },
+    fullViewingKey,
+  } = wallets[0]!;
+
+  if (!fullViewingKey)
+    throw new ConnectError('Unable to get full viewing key', Code.Unauthenticated);
```
jessepinho, line 42: Let me know if `Code.Unauthenticated` is the correct code to be using here. It seemed appropriate to me.

hdevalence:
We should merge this to keep iterating but longer term I don't think its a good design to move functionality into the router.

Interpreting the TransactionPlan is **the core responsibility** of the authorization module. Moving that out of the auth module into the router violates encapsulation and gives the router spend authority over the user's funds.

This should not block merging — getting this code in and implementing visibility previews is more important for the moment – but we should not make the router a trusted component for authorization.

packages/router/src/grpc/custody/view-transaction-plan/index.ts
```diff
@@ -28,7 +29,7 @@ export const viewTransactionPlan = (
 
   return new TransactionView({
     bodyView: {
-      actionViews: txPlan.actions.map(viewActionPlan(denomMetadataByAssetId)),
+      actionViews: txPlan.actions.map(viewActionPlan(denomMetadataByAssetId, fullViewingKey)),
       fee,
       memoView: {
         memoView: {
```
hdevalence, line 35: We'll also want to use the FVK to view the return address in the memo plaintext, in the screenshot it shows up as an opaque address but the user should see that it's their address and which account it's for

packages/ui/components/ui/tx/view/memo-view.tsx
```diff
@@ -1,20 +1,12 @@
...
 import { AddressViewComponent } from './address-view';
 import { ViewBox, ViewSection } from './viewbox';
-import { AddressView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1alpha1/keys_pb';
 
 export const MemoViewComponent = ({ memo }: { memo: MemoView }) => {
   if (memo.memoView.case === 'visible') {
     const mv = memo.memoView.value;
     const text = mv.plaintext?.text;
 
-    const av = new AddressView({
```
jessepinho, line 11: For some reason, `MemoViewComponent` was hard-coding the sender return address as opaque, rather than simply passing the memo's `returnAddress` property to `AddressViewComponent` and allowing `AddressViewComponent` to determine whether it was visible or opaque.

So I changed that logic here. As a result, the sender return address won't be opaque in the txn approval popup _or_ in the `TransactionViewComponent` for already-existing transactions:
image

packages/router/src/grpc/custody/view-transaction-plan/index.ts
jessepinho: Ah, good catch. I've addressed this in [this commit](https://github.com/penumbra-zone/web/pull/378/commits/39cc3ea2b2dac36d2cd7baba2b1e286c2acb8d36). (I also took the oppt'y to write tests for `getAddressView` in that commit.)

Screenshot:
image

jessepinho:
@hdevalence:
> Interpreting the TransactionPlan is the core responsibility of the authorization module

When you say "the authorization module," do you mean the popup? Because, while the `viewTransactionPlan` code is indeed moved into the router now, it's still inside the `authorize` impl. That is, interpreting the `TransactionPlan` is still only being done inside authorization-logic, even if not inside the popup code itself.

That said, if locating that code inside the `router` package isn't appropriate, I can definitely move it into the popup. I'll just have to figure out the cyclical import issue, but that should be relatively straightforward. (I'll do it in a separate PR, per your comment about getting this merged.)

hdevalence:
> @hdevalence:
> 
> > Interpreting the TransactionPlan is the core responsibility of the authorization module
> 
> When you say "the authorization module," do you mean the popup? Because, while the `viewTransactionPlan` code is indeed moved into the router now, it's still inside the `authorize` impl. That is, interpreting the `TransactionPlan` is still only being done inside authorization-related logic, even if not inside the popup code itself.

OK, this sounds better. I don't have a really detailed understanding of the extension's code structure, I was concerned because it sounded like it was in a part of the code that routes between services. I just wanted to hammer home the point that it's important to conceptualize the transaction comprehension part as being _equivalent_ to knowledge of the signing key itself. But it sounds like we're actually all on the same page on that.

> That said, if locating that code inside the `router` package isn't appropriate, I can definitely move it into the popup. I'll just have to figure out the cyclical import issue, but that should be relatively straightforward. (I'll do it in a separate PR, per your comment about getting this merged.)

No, let's not worry about it for now, I'm not equipped to make the judgements required to demand that it be a certain way, and it's much less important than getting the user-visible features in place.

Sorry for the noise, this is great work and let's keep rolling.

jessepinho:
OK, thanks @hdevalence — I'll merge as soon as I have your or @turbocrime 's approval. Note that I made the requested change to the sender's return address — it's prob easiest to review that bit via its [specific commit](https://github.com/penumbra-zone/web/pull/378/commits/39cc3ea2b2dac36d2cd7baba2b1e286c2acb8d36).

hdevalence (approved):

PR #383, opened 2024-01-23
Warn on unknown tendermint log

grod220:
Closes https://github.com/penumbra-zone/web/issues/352

It's probably too aggressive to throw an error on _any_ Tendermint log


jessepinho (approved):
I might rename `tendermintErrors` to `knownTendermintErrors` just to make it super clear. But up to you. 🚢

Issue #222, opened 2023-11-06
Extension should support GRPC proxying


hdevalence:
Penumbra users should be able to configure their own RPC endpoints. This is important for two reasons:

- **Decentralization**: If users cannot configure their own RPC endpoints, they're locked into using the RPC hardcoded into the frontend. This adds a single point of failure.
- **Privacy**: Users should be able to run their own nodes, or send requests only to trusted nodes. Longer-term, we should solve network privacy for FE requests, but this is a harder problem.

Both the extension and the FE website require RPC access. Users should only have to make one configuration setting, and have that RPC endpoint be used for all of their requests.  To accomplish this, we should have the extension proxy GRPC requests to the RPC endpoint it uses for scanning, similar to the architecture for `pclientd`:

```
   ┌────────┐  ┌─────────────────┐  ┌────────┐
   │  Client│  │         pclientd│  │Penumbra│
   │Software│◀─┼─┐             ┌─┼─▶│Fullnode│
   └────────┘  │ │             │ │  └────────┘
            ╭  │ │ ┌───────┐   │ │
     public │  │ │ │grpc   │   │ │
 chain data │  │ ├▶│proxy  │◀──┤ │
            ╰  │ │ └───────┘   │ │
               │ │             │ │
            ╭  │ │ ┌───────┐   │ │
    private │  │ │ │view   │   │ │
  user data │  │ ├▶│service│◀──┘ │
            ╰  │ │ └───────┘     │
               │ │               │
            ╭  │ │ ┌ ─ ─ ─ ┐     │
   spending │  │ │  custody      │
 capability │  │ └▶│service│     │
 (optional) ╰  │    ─ ─ ─ ─      │
               └─────────────────┘
```

grod220:
What would that look like design wise? Would it just be a checkbox in the current RPC settings page in the extension for `[x] proxy all traffic`?

Also, can you clarify how this is different than our existing RPC settings page where a user can choose another RPC provider url?

hdevalence:
> What would that look like design wise? Would it just be a checkbox in the current RPC settings page in the extension for `[x] proxy all traffic`?

Assuming by design you mean UX design, there would be no explicit design. The extension already has a setting for the RPC endpoint, and we'd leave that unchanged.

[...]

PR #386, opened 2024-01-24
recognize proxy services

turbocrime:
#222 

tried a few clients, and it seems to work. todo comprehensive testing

### DIFF
--- a/apps/extension/src/injected-connection-manager.ts
+++ b/apps/extension/src/injected-connection-manager.ts
@@ -3,10 +3,10 @@
  * client-side connection manager.
  */
 
-import type { InitPenumbraService } from './types/content-script-init';
-import type { serviceTypeNames } from '@penumbra-zone/types/src/registry';
-import { ChannelClientLabel } from '@penumbra-zone/transport/src/types';
 import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-runtime/client-connection-manager';
+import { ChannelClientLabel } from '@penumbra-zone/transport/src/types';
+import type { serviceTypeNames } from '@penumbra-zone/types/src/registry';
+import type { InitPenumbraService } from './types/content-script-init';
 
 /**
  * This could just be imported, but declaring it locally helps the bundler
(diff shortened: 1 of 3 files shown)


apps/extension/src/injected-connection-manager.ts
```diff
@@ -17,11 +17,21 @@ import { ClientConnectionManager } from '@penumbra-zone/transport/src/chrome-run
...
+  'penumbra.core.component.compact_block.v1alpha1.QueryService',
+  'penumbra.core.component.dex.v1alpha1.QueryService',
   'penumbra.core.component.dex.v1alpha1.SimulationService',
+  'penumbra.core.component.governance.v1alpha1.QueryService',
+  'penumbra.core.component.sct.v1alpha1.QueryService',
+  'penumbra.core.component.shielded_pool.v1alpha1.QueryService',
+  'penumbra.core.component.stake.v1alpha1.QueryService',
   'penumbra.util.tendermint_proxy.v1alpha1.TendermintProxyService',
-  'penumbra.view.v1alpha1.ViewProtocolService',
 ];
```
grod220, line 35: It's kinda a bummer we can't import these. Is the bundle _that_ much bigger if we do?

packages/types/src/registry.ts
```diff
@@ -20,17 +31,37 @@ import { ViewProtocolService } from '@buf/penumbra-zone_penumbra.connectrpc_es/p
...
+  ChainService.typeName,
+  CompactBlockService.typeName,
+  DexService.typeName,
+  DexSimulationService.typeName,
+  GovernanceService.typeName,
+  SctService.typeName,
+  ShieldedPoolService.typeName,
+  StakeService.typeName,
+  TendermintProxyService.typeName,
 ] as const;
```
grod220, line 67: It looks like we can have a single array with  `[CustodyProtocolService, ViewProtocolService...]` and spread them into `createRegistry` and do `serviceTypeNames = arr.map(s => s.typeName)`.

grod220: That single list could also be imported and used here too: https://github.com/penumbra-zone/web/pull/386/files#diff-b9be68cac5d37d21289e20d8d7fcb88d993de52b16e69a6500255f3fde092ae0R85-R98

apps/extension/src/injected-connection-manager.ts
grod220: Maybe you can document in code that this list needs to be in sync with the ones in the `types` package?

turbocrime: using the type import and a literal list reduces the size of this bundle to 3.9kb, which i think is important because we inject and execute this script on every single page that the user visits. with some transport refactor like i mentioned this morning, we could soon remove this list entirely.

when importing the actual array from registry.ts, even as 'const', the injected script is over 30x larger and includes every complete service and method definition.

importing the type from registry.ts enforces sync with the type system.

packages/types/src/registry.ts
turbocrime: last time i tried generating this list instead of writing a literal const, i was unable to satisfy ts and keep the injected bundle small.

turbocrime: that other list is a list of proxies - they are distinct, and different things happen to them. i will look into using one or the other for deduplication.

turbocrime:
will be part of #475

PR #387, opened 2024-01-24
Allow users to preview transaction visibility as the public or as the receiver

### LINKED ISSUE #376: Extend transaction approval dialog to highlight visibility.
The approval dialog added in #366 doesn't allow a user to understand which parts of their transaction will be public and which parts will be private -- and to whom. To fix this, we should add a visibility toggle to the approval dialog that allows a user to preview the transaction as seen by different parties.

This toggle should be at the top and should have either two or three settings, depending:

- "Your View"
- "Their View" (optional, see below)
- "Public View"

"Your View" should show the skeleton transaction view we generate from the transaction plan, as in the current code. The other views should show "attenuated" versions of that view. To do this, we need:

- [x] The existing code (post #375), which generates the user's view of the transaction they're about to create. This is the most complete view.
- [ ] A method `isExternal` that takes as input a transaction view, and returns `true` if any of the address views in any of the spend/output actions are opaque, signifying that there is an external counterparty for the transaction, rather than being a purely internal one.  (Similar logic exists in the transaction classifier: https://github.com/penumbra-zone/web/blob/13440a3eaaf8f471da9c66b7bed6c26c00f0d9cd/packages/types/src/transaction/classify.ts#L36 )
- [ ] A method `attenuateReceiver` that applies the following transformations to a transaction view:
  - All spend views become opaque
  - All swap views become opaque
  - All swap claim views become opaque
  - All delegator vote views become opaque
  - Output views become opaque if they are addressed to a _visible_ address (meaning the user's wallet) but remain visible if they are addressed to an _opaque_ address (meaning not the user's wallet)
  - The memo view remains visible, **but** the `return_address` `AddressView` in the `MemoPlaintextView` becomes opaque (the receiver will only see our address, not our account information).
- [ ] A method `attenuatePublic` that applies the following transformations to a transaction view:
  - All spend views become opaque
  - All output views become opaque
  - All swap views become opaque
  - All swap claim views become opaque
  - All delegator vote views become opaque
  - The memo view becomes opaque
- [ ] Code that calls `attenuateReceiver` and `attenuatePublic` on the full view to generate the "their view" and "public view" content
- [ ] A top-level toggle that has two settings (if `isExternal === false`) or three settings (if `isExternal === true`), and that switches the user's view of the transaction they're going to create.

There will be some code duplication between `attenuateReceiver` and `attenuatePublic` (note: could also have different names, just suggestions) but I think that this is better than trying to share code -- it makes the visibility logic easier to understand and audit if we just have to consider a pure functional transformation of the view for each separate case.

jessepinho:
🚨 Roughly 900 line changes are from tests

## Background
In the transaction approval popup, we want to show the user what the transaction will look like to A) the recipient, and B) the public. That way, a user will know — _before_ sending the transaction — what others will see once the transaction is submitted.

This PR accomplishes that by using a new "translator" pattern for translating different types of objects into each other — for example, translating a `TransactionView` from a sender's perspective into a `TransactionView` from a recipient's perspective. 

The result:

https://github.com/penumbra-zone/web/assets/1121544/a7a35a3a-f108-42a3-9647-4ed2cc1ed89f

## In this PR
- Create a new `ViewTabs` component underneath the `TransactionApproval` component that allows the user to switch between "Your View"/"Receiver's View" (if applicable)/"Public View".
  - Because of this, and because of the new hooks/etc. I created, I also moved the `TransactionApproval` component into its own directory.
- Create a new `translators` directory in the `types` package to translate e.g., the sender's view of a transaction to the public's view. **This is the bulk of the work in this PR.**
  - At the moment, this only contains `*View` -> `*View` translators, but it can have any kind of translators we want in the future.
  - If reviewers have design feedback on my approach here, I'd love to hear it. This was what took the most experimenting and back-and-forth. I tried imperative helpers, [lenses](https://www.bekk.christmas/post/2019/6/the-lens-pattern-in-typescript), and translator classes/interfaces, but this approach feels the most natural and readable.
- Refactor `classifyTransaction()` to return a machine-readable `TransactionClassification`, rather than a human-readable label.
  - Create a `getTransactionClassificationLabel()` function for the original functionality.

Closes #376 

## Remaining todos
- [x] Rewrite translators to optionally take a `ctx` object, rather than using HOFs for curried values.
- [x] Fix scroll bar issue.
- [ ] Use a data loader for translated transaction views, rather than the `useTransctionViews()` hook.


apps/webapp/src/fetchers/transactions.ts
```diff
@@ -1,6 +1,6 @@
 import { viewClient } from '../clients/grpc';
 import { streamToPromise } from './stream';
-import { classifyTransaction, uint8ArrayToHex } from '@penumbra-zone/types';
+import { getTransactionClassificationLabel, uint8ArrayToHex } from '@penumbra-zone/types';
```
jessepinho, line 3: This is actually a new function that does the exact same thing as the original. But I repurposed the original to return a machine-readable value rather than a human-readable value.

packages/ui/components/ui/tabs.tsx
```diff
@@ -28,7 +28,7 @@ const TabsTrigger = React.forwardRef<
   <TabsPrimitive.Trigger
     ref={ref}
     className={cn(
-      'inline-flex h-9 font-bold text-muted-foreground items-center justify-center whitespace-nowrap rounded-lg ring-offset-background transition-all focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 data-[state=active]:bg-teal data-[state=active]:text-muted data-[state=active]:shadow-sm',
+      'inline-flex h-9 font-bold text-muted-foreground items-center justify-center leading-4 rounded-lg ring-offset-background transition-all focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 data-[state=active]:bg-teal data-[state=active]:text-muted data-[state=active]:shadow-sm',
```
jessepinho, line 31: Needed to allow wrapping and tighten up the leading for this view:
image

packages/types/src/translators/action-view.test.ts
```diff
@@ -0,0 +1,326 @@
...
+      expect(
+        asPublicActionView(new ActionView({ actionView: { case: undefined } })).equals(
+          new ActionView(),
+        ),
+      ).toBe(true);
+    });
+  });
+
+  describe('when passed a spend action view', () => {
+    const actionView = new ActionView({
```
jessepinho, line 23: These view definitions get massive in these test files. In the future, I'd like to integrate with a library like [`factory.ts`](https://github.com/willryan/factory.ts) to more easily build objects for various use cases, but this PR is already bloated enough without that addition :D

packages/types/src/translators/README.md
```diff
@@ -0,0 +1,25 @@
...
+import { asOpaqueAddressView, asOpaqueMemoView } from '@penumbra-zone/types';
+```
+
+## Delegation
+
+Translators should be as simple as possible, delegating to other translators for nested object types. For example, the `asPublicTransactionView` translator doesn't itself modify a transaction view's nested memo view; rather, it delegates that work to `asOpaqueMemoView`. This way, each translator is easy to understand, and easy to test.
+
+## Higher-order functions
+
+Some "translators" are actually higher-order functions (HOFs) that _return_ translators. This is a useful pattern for when a translator needs access to additional variables besides the translatable object.
```
jessepinho, line 23: Curious what reviewers think of this design decision. The more I think about it, the more I'm not crazy about it — i.e., perhaps instead of HOFs, I should just pass a second `ctx` argument to translators which can hold whatever values they need...?

packages/types/src/translators/transaction-view.ts
```diff
@@ -0,0 +1,40 @@
...
+) => Translator<TransactionView, Promise<TransactionView>> =
+  isControlledAddress => async transactionView => {
+    if (!transactionView?.bodyView) return new TransactionView();
+
+    return new TransactionView({
+      bodyView: {
+        memoView: asReceiverMemoView(transactionView.bodyView.memoView),
+        actionViews: await Promise.all(
+          transactionView.bodyView.actionViews.map(asReceiverActionView(isControlledAddress)),
+        ),
```
jessepinho, line 33: There are a number of places like this where I've left out some properties like `detectionData`, since they don't get rendered to the screen in the `TransactionViewComponent` anyway. 

Are we OK with that? The risk that comes to mind is that, if we start rendering stuff like `detectionData` via `TransactionViewComponent` in the future, we'll likely forget to update that here, and data might not get rendered that should be.

hdevalence: yeah, i think this is fine, because the "translators" here are scoped to the confirmation dialog -- we don't have a problem of forgetting to add data in the confirmation dialog, because the confirmation dialog is the top-level part that _defines_ what data should be shown

apps/extension/src/routes/popup/approval/transaction/use-transaction-views.ts
```diff
@@ -0,0 +1,54 @@
...
+        .catch(() => {
+          /** no-op */
+        });
+    }
+  }, [transactionView]);
+
+  return asReceiver;
+};
+
+export const useTransactionViews = () => {
```
grod220, line 30: If it weren't for the async call, I'd say to create a synchronous selector out of this whole hook.  But given the call, I think we should use react-query here as the single hook (aka no useEffect) and a synchronous function to handle all the logic.

apps/extension/src/routes/popup/approval/transaction/is-controlled-address.ts
```diff
@@ -0,0 +1,29 @@
...
+    await grpcClient.indexByAddress({ address });
+    return true;
+  } catch (e) {
+    if (e instanceof ConnectError && e.code === Code.Unauthenticated) {
+      return false;
+    } else {
+      throw e;
+    }
+  }
+};
```
grod220, line 29: Think this workaround works for now, but I can see a good reason for a new rpc method here

apps/webapp/src/fetchers/transactions.ts
grod220: what's the difference between machine readable and human readable?

packages/types/src/translators/README.md
grod220: I found them to be a bit harder to mentally grasp. Perhaps it could be the double parenthesis thing. If we find that we only use it like so `asReceiverTransactionView(isControlledAddress)(transactionView)`, that probably means we don't need the HOF right? Otherwise we'd be using just `asReceiverTransactionView(isControlledAddress)` more often right?

packages/types/src/translators/action-view.test.ts
grod220: Ah yes. A factory or some other way to generate this data more sanely would be nice.

packages/ui/components/ui/tabs.tsx
grod220: Can we get the scroll bar flush to the right?

grod220 (commented):
Nice job with testing!

apps/extension/src/routes/popup/approval/transaction/use-transaction-views.ts
grod220: Other thoughts:

I really try to keep async stuff out of hooks. I've found this is the number one most challenging aspect of React and quite difficult to reason about. React-Query has helped, but I think `Data Loaders` are just better.

How about we make the selection a query parameter (or route), load the data in a react-router data loader, and then the component can expect that is available. This will allow us to avoid stuffing async logic in our UI code.

jessepinho: Ah, great idea. I haven't used the data loader pattern much yet, but it's a good one. Will rework this to use that.

packages/types/src/translators/README.md
jessepinho: The idea is that I'd like to have a standalone function that can be passed to e.g. `.map()`, like here: https://github.com/penumbra-zone/web/pull/387/files/c92b75e6b1b200de7ee7ffc52db1113625f07a34#diff-911daf37f54f94c28a4b25b6b2e923e5b45a3bf651ed7befef1da45f1d581ef7R32

That's only possible using HOFs . Otherwise, `.map()` needs to have an inline function defined.

That said, I agree that the HOF approach is harder to grasp, and that almost certainly outweighs the "need" for avoiding inline functions. I'll rewrite this with a `ctx` object that holds `isControlledAddress`.

apps/webapp/src/fetchers/transactions.ts
jessepinho: The machine-readable classifications are defined by [the `TransactionClassification` type](https://github.com/penumbra-zone/web/pull/387/files/0e4575541dc4bfdffc5978872139ac35d3bb3eee#diff-fdf698291d45d9c5719ccb4b9ebcdd6e0e15d4d71e73c8615e6fe1da1bf068fa), and can be used with e.g., `switch` or `if` statements to determine what kind of UI to render. Currently, [it's used inside `ViewTabs`](https://github.com/penumbra-zone/web/pull/387/files/0e4575541dc4bfdffc5978872139ac35d3bb3eee#diff-cbbb8cbe51aa9dff111ee1ae2c9e8036e43a3874cbd5c6c7cee6d08222f4e26aR16-R17) to determine which tabs to render, based on whether a transaction is a `send` vs. some other type.

jessepinho: Oops, I posted the comment before finishing.

The human-readable value is a label that shows up in the UI. So, while machine-readable values are defined in [the `TransactionClassification` type](https://github.com/penumbra-zone/web/pull/387/files/0e4575541dc4bfdffc5978872139ac35d3bb3eee#diff-fdf698291d45d9c5719ccb4b9ebcdd6e0e15d4d71e73c8615e6fe1da1bf068fa), the human-readable values are defined by [the function](https://github.com/penumbra-zone/web/pull/387/files/0e4575541dc4bfdffc5978872139ac35d3bb3eee#diff-4dc78f0802f529efed129c435d57ea2f67cab9853483548ba25752882fdff965R70-R79) that returns the label.

packages/ui/components/ui/tabs.tsx
jessepinho: image

Fixed. And I also fixed an issue where the bottom buttons overlapped the JSON view.

packages/types/src/translators/README.md
jessepinho: OK, I rewrote the translators to take a `ctx` object instead of using HOFs.

jessepinho:
@grod220 OK, ready for re-review.

- [This commit](https://github.com/penumbra-zone/web/pull/387/commits/9def0cb220d316c1b78bc1c5711d357d678a0cf3) fixes the scrollbar in the popup. It's now flush to the right, and the approve/deny buttons no longer overlap the JSON view.
- [This commit](https://github.com/penumbra-zone/web/pull/387/commits/d3f0eac4defc05dbbf9f185219d09c5c3866c6a5) replaces HOFs in the translators with the optional use of a second `ctx` argument.
- [This commit](https://github.com/penumbra-zone/web/pull/387/commits/3e521212bb2f21cf92da542f91e83f2fcabdd7e5) addresses your comment about using a loader, rather than a hook, for asynchronously compiling the sender/receiver/public transaction views. Note that I ended up not being able to use a loader: loaders can't use hooks since they're not part of the component lifecycle, so I couldn't access the transaction view in state via `useStore`. And calling `useStore.getState()` directly didn't work either, because the Zustand state wasn't yet populated when the loader was called. Instead, I moved the compilation of the three views into the `tx-msg-handler.ts` file, which is where there's a function for handling a transaction approval message from the extension to the popup. Since, as of #393, the transaction views are stored as JSON-stringified strings, I still had to keep the `useTransactionViews` hook (which I renamed to `useTransactionViewSwitcher`) to deserialize those strings to objects and memoize the results. But it's a lot easier to reason about now, and is totally synchronous.

apps/extension/src/state/tx-approval.ts
```diff
@@ -1,14 +1,23 @@
...
 import { MessageResponder } from '@penumbra-zone/types/src/internal-msg/shared';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
+import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 
 export interface TxApprovalSlice {
-  // zustand doesn't like JsonValue, because the type is infinitely deep. so store json strings
+  /**
+   * Zustand doesn't like JsonValue, because the type is infinitely deep. And we
+   * can't store instances of custom classes (like `TransactionView`s) in the
+   * store, because we're using Immer middleware for Zustand, which requires
```
grod220, line 12: I kinda feel that Immer+Zustand = Mobx. If so, why don't we just use mobx lol.

packages/types/src/translators/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```TS
+import { asOpaqueAddressView, asOpaqueMemoView } from '@penumbra-zone/types';
+```
+
+## Delegation
+
+Translators should be as simple as possible, delegating to other translators for nested object types. For example, the `asPublicTransactionView` translator doesn't itself modify a transaction view's nested memo view; rather, it delegates that work to `asOpaqueMemoView`. This way, each translator is easy to understand, and easy to test.
+
+## Contexts
```
grod220, line 21: Can we use an alternative name to `Context`? I feel like this may be confused with our repo's use of context from [connectrpc](https://connectrpc.com/docs/node/implementing-services/#context).

grod220: Maybe props, params, or env?

packages/ui/components/ui/tabs.tsx
grod220: - Can we get bottom margin on the title?
- Why do you think "Reciever's View" does not get proper padding? Can we adjust this a bit?

Screenshot 2024-01-26 at 12 25 37 PM

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
```diff
@@ -4,19 +4,38 @@ import { PopupRequest } from '@penumbra-zone/types/src/internal-msg/popup';
...
-    state.txApproval.responder = responder;
-  });
+
+    useStore.setState(state => {
+      state.txApproval.authorizeRequest = authorizeRequestAsString;
+      state.txApproval.asSender = asSender;
+      state.txApproval.asReceiver = asReceiver;
+      state.txApproval.asPublic = asPublic;
+      state.txApproval.transactionClassification = classifyTransaction(transactionView);
+      state.txApproval.responder = responder;
```
grod220, line 38: Hear me out. What if we kill nearly everything in `TxApproval` except for setting the `responder` in zustand:

https://github.com/penumbra-zone/web/blob/7d718d785dbc6fd0c63a5111d7f1b3c94706f648/apps/extension/src/routes/popup/approval/tx-msg-handler.ts#L20

Instead, pass all tx data as query params in the request URL:

https://github.com/penumbra-zone/web/blob/7d718d785dbc6fd0c63a5111d7f1b3c94706f648/packages/router/src/ctx/approver.ts#L11

We then kill all react hooks, handle data fetching in the react-router loader for that route, and the components have the data they need to render when they load. No more hooks+async required.

hdevalence:
Can we make sure this lands in the next published extension version?

grod220 (approved):
Discussed offline. Will merge now and follow up with a PR to explore loaders pattern vs async hooks.

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
grod220: In short, yes. Is it really _that_ big? If so, I suppose that is a hard blocker. If not, I think it's worth it. Url-as-state-management is basically the philosophy that gave rise to remixjs. It allows you to exit async react-hook land and prevents re-request waterfalls of re-render->request->re-render->request (I don't think this is the case for us here though).

grod220: Just asked chatgpt for an example json object of 2000 characters. It's smaller than I expected lol. Looks like it's too close to call 😢

grod220: Hmm: https://chromium.googlesource.com/chromium/src/+/master/docs/security/url_display_guidelines/url_display_guidelines.md#:~:text=URL%20Length,-In%20general%2C%20the&text=Chrome%20limits%20URLs%20to%20a,is%20used%20on%20VR%20platforms.

Not sure if this applies to chrome extension runtime though 🤔

jessepinho: OK! Interesting. I mean, we could test it. Doesn't it just _feel_ kinda wrong, though, to use the URL? Not that that's a great reason not to do it. It just feels like there must be a better way...

hdevalence:
Amazing!!

PR #391, opened 2024-01-24
Update Vite/Vitest

### LINKED ISSUE #271: Vite updates
We should update to the latest vite, vitest, & @vitest/browser versions. Looks like there are breaking changes.

Steps:
1. Do `pnpm up -r --latest`
2. Run pnpm test
3. See errors and make config changes to support new versions

Looks like there is a new warning that we need to add `type: "module"` to some of our packages. There are also changes to our vitest browser that need to be made.

### LINKED ISSUE #379: CI builds are failing intermittently for `packages/crypto` and `packages/storage`
If you navigate to [this test run](https://github.com/penumbra-zone/web/actions/runs/7590871408/job/20678594468), then toggle open the "Run all tests" section, then toggle open the `@penumbra-zone/crypto-web:test` section, you'll see the following error:

```
 ⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯
  
  Vitest caught 1 unhandled error during the test run.
  This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.
  
  ⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
  TypeError: Failed to fetch dynamically imported module: http://localhost:5173/@fs/home/runner/work/web/web/node_modules/.pnpm/@vitest+utils@0.34.6/node_modules/@vitest/utils/dist/index.js?v=981845ef
  This error originated in "src/encryption.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
  
   Test Files  no tests
        Tests  no tests
       Errors  1 error
     Start at  01:07:31
     Duration  6.50s (transform 0ms, setup 0ms, collect 0ms, tests 0ms, environment 0ms, prepare 0ms)
  
   ELIFECYCLE  Test failed. See above for more details.
  Error:  command finished with error: command (/home/runner/work/web/web/packages/crypto) /home/runner/setup-pnpm/node_modules/.bin/pnpm run test exited (1)
```

We don't actually use dynamic imports at the moment, so this error likely points at something inside Vite itself. It's worth noting that the `encryption.test.ts` test suite is run using [a special config](https://github.com/penumbra-zone/web/blob/main/packages/crypto/browser.config.ts#L6) for running in a browser environment, which may be causing this issue...?

UPDATE: We're now seeing these [test failures](https://github.com/penumbra-zone/web/actions/runs/7636547510/job/20804167822?pr=387) for `packages/storage` as well.

### Switch to JSDom
I tried switching to JSDom, in case [the chromium/playwright combination we're using](https://github.com/penumbra-zone/web/blob/34be34b8234d83b70ad33b51670faaf138f9733a/packages/crypto/browser.config.ts#L8-L9) is the source of this issue. `window.crypto` [is not implemented in JSDom](https://github.com/jsdom/jsdom/issues/1612#issuecomment-249441408), though. So I tried polyfilling it using the `webcrypto` import from `node:crypto`, per [this suggestion](https://github.com/jsdom/jsdom/issues/1612#issuecomment-1659591074). However, that resulted in a different [test failure](https://github.com/penumbra-zone/web/actions/runs/7591393707/job/20679391388):

```
     ❯ src/encryption.test.ts > encryption > Key > unseal > returns null when the box cannot be decrypted (OperationError)
       → The operation failed for an operation-specific reason
  ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
  
  
   FAIL  src/encryption.test.ts > encryption > Key > unseal > returns null when the box cannot be decrypted (OperationError)
  OperationError: The operation failed for an operation-specific reason
  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
  Serialized Error: { code: +0, INDEX_SIZE_ERR: 1, DOMSTRING_SIZE_ERR: 2, HIERARCHY_REQUEST_ERR: 3, WRONG_DOCUMENT_ERR: 4, INVALID_CHARACTER_ERR: 5, NO_DATA_ALLOWED_ERR: 6, NO_MODIFICATION_ALLOWED_ERR: 7, NOT_FOUND_ERR: 8, NOT_SUPPORTED_ERR: 9, INUSE_ATTRIBUTE_ERR: 10, INVALID_STATE_ERR: 11, SYNTAX_ERR: 12, INVALID_MODIFICATION_ERR: 13, NAMESPACE_ERR: 14, INVALID_ACCESS_ERR: 15, VALIDATION_ERR: 16, TYPE_MISMATCH_ERR: 17, SECURITY_ERR: 18, NETWORK_ERR: 19, ABORT_ERR: 20, URL_MISMATCH_ERR: 21, QUOTA_EXCEEDED_ERR: 22, TIMEOUT_ERR: 23, INVALID_NODE_TYPE_ERR: 24, DATA_CLONE_ERR: 25 }
  Caused by: Error: Cipher job failed
  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
  
   Test Files  1 failed | 1 passed (2)
        Tests  1 failed | 20 passed (21)
     Start at  02:00:44
     Duration  4.93s (transform 350ms, setup 85ms, collect 429ms, tests 2.04s, environment 2.33s, prepare 930ms)
  
   ELIFECYCLE  Test failed. See above for more details.
  Error:  command finished with error: command (/home/runner/work/web/web/packages/crypto) /home/runner/setup-pnpm/node_modules/.bin/pnpm run test exited (1)
```

jessepinho:
🚨 Most line changes are due to `pnpm-lock.yaml`

## Background
We wanted to update to the latest version of Vite and especially Vitest, since the version of Vitest we were using was still in beta.

## In this PR
- Update to the latest versions of `vite` and `vitest`.
- Update `@bufbuild/protobuf`, which otherwise broke things when updating to the latest `vite`/`vitest`.
- Update to the latest `@vitest/browser` in `packages/crypto`, and install it in `packages/storage`. This is required by `vitest` v1.0.0+.
- Re-enable the `packages/crypto` test suite that I'd previously disabled due to #379.
- Handle weird issue with the `buffer` library.
  - When Vadim was working on this (#272), he came across the same issue I did: a weird error saying that the `buffer` library `does not provide an export named 'Buffer'`. @grod220 [found](https://github.com/penumbra-zone/web/pull/272#issuecomment-1841412562) that changing the import to add a trailing slash would fix the issue, and indeed [`buffer`'s own docs support this](https://www.npmjs.com/package/buffer#usage). For some reason, though, that only fixed the test failures locally, not in CI.
  
    I couldn't figure out a way to resolve the issue with the `buffer` library, so I took a different approach: uninstalling the library completely and finding native JS ways to accomplish the tasks we were using `buffer` for. We were only using it to encode/decode data between different formats like base64, Uint8 arrays, etc., so I found native JS implementations of those functions and used them instead. Thankfully, the functions were well-tested, so I didn't touch the tests — I only updated the code and saw that the tests still passed, leaving comments for where I got the implementation.

    ~As part of this work, I also removed the `Buffer` global that was provided via Webpack in the extension. We're not using buffers in the extension anywhere anyway, so I'm guessing it's a leftover from when we were. Reviewers, please correct me if I'm wrong.~ Just kidding — looks like that `Buffer` global was being used by `bip39` under the hood.

## Manual testing
I imported the test account wallet locally and sent a transaction. Everything worked fine. If there's other stuff I should do to test that nothing's broken, let me know!

## Punting (until followup PRs)
- We're still getting this warning message during test runs: `The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.` Since it's not blocking anything, I'm going to handle that in a follow-up PR that will close #271, so that this PR doesn't get too big.
- I haven't updated _all_ packages yet via `pnpm up -r --latest` — I was just focusing on Vite/Vitest for this PR.

Relates to #271 
Closes #379

### DIFF
--- a/apps/extension/webpack/webpack.common.js
+++ b/apps/extension/webpack/webpack.common.js
@@ -69,9 +69,6 @@ module.exports = {
     },
   },
   plugins: [
-    new webpack.ProvidePlugin({
-      Buffer: ['buffer', 'Buffer'],
-    }),
     new CopyPlugin({
       patterns: [{ from: '.', to: '../', context: 'public' }],
       options: {},
--- a/apps/extension/webpack/webpack.common.js
+++ b/apps/extension/webpack/webpack.common.js
@@ -69,6 +69,9 @@ module.exports = {
     },
   },
   plugins: [
+    new webpack.ProvidePlugin({
+      Buffer: ['buffer', 'Buffer'],
+    }),
     new CopyPlugin({
       patterns: [{ from: '.', to: '../', context: 'public' }],
       options: {},
--- a/apps/extension/webpack/webpack.common.js
+++ b/apps/extension/webpack/webpack.common.js
@@ -70,6 +70,7 @@ module.exports = {
   },
   plugins: [
     new webpack.ProvidePlugin({
+      // Required by the `bip39` library
       Buffer: ['buffer', 'Buffer'],
     }),
     new CopyPlugin({
--- a/packages/types/src/base64.ts
+++ b/packages/types/src/base64.ts
@@ -19,7 +19,7 @@ export const InnerBase64Schema = z.object({ inner: Base64StringSchema });
 export const base64ToUint8Array = (base64: string): Uint8Array => {
   const validated = validateSchema(Base64StringSchema, base64);
   const binString = atob(validated);
-  return Uint8Array.from(binString, m => m.codePointAt(0)!);
+  return Uint8Array.from(binString, byte => byte.codePointAt(0)!);
 };
 
 export const uint8ArrayToBase64 = (byteArray: Uint8Array): string => {
--- a/packages/types/src/base64.ts
+++ b/packages/types/src/base64.ts
@@ -1,5 +1,4 @@
-// Re: the trailing slash, see https://www.npmjs.com/package/buffer#usage
-import { Buffer } from 'buffer/';
+import { Buffer } from 'buffer';
 import { z } from 'zod';
 import { validateSchema } from './validation';
 
--- a/packages/types/src/base64.ts
+++ b/packages/types/src/base64.ts
@@ -1,5 +1,6 @@
+// Re: the trailing slash, see https://www.npmjs.com/package/buffer#usage
+import { Buffer } from 'buffer/';
 import { z } from 'zod';
-import { Buffer } from 'buffer';
 import { validateSchema } from './validation';
 
 export const Base64StringSchema = z.string().refine(
--- a/packages/types/src/base64.ts
+++ b/packages/types/src/base64.ts
@@ -1,4 +1,4 @@
-import { Buffer } from 'buffer';
+import { Buffer } from 'buffer/';
 import { z } from 'zod';
 import { validateSchema } from './validation';
 
--- a/packages/types/src/base64.ts
+++ b/packages/types/src/base64.ts
@@ -1,6 +1,5 @@
-import { Buffer } from 'buffer/';
-import { z } from 'zod';
 import { validateSchema } from './validation';
+import { z } from 'zod';
 
 export const Base64StringSchema = z.string().refine(
   str => {
@@ -19,10 +18,11 @@ export const InnerBase64Schema = z.object({ inner: Base64StringSchema });
 
 export const base64ToUint8Array = (base64: string): Uint8Array => {
   const validated = validateSchema(Base64StringSchema, base64);
-  const buffer = Buffer.from(validated, 'base64');
-  return new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
+  const binString = atob(validated);
+  return Uint8Array.from(binString, m => m.codePointAt(0)!);
 };
 
 export const uint8ArrayToBase64 = (byteArray: Uint8Array): string => {
-  return Buffer.from(byteArray).toString('base64');
+  const binString = String.fromCodePoint(...byteArray);
+  return btoa(binString);
 };
--- a/packages/types/src/hex.ts
+++ b/packages/types/src/hex.ts
@@ -1,16 +1,36 @@
-import { Buffer } from 'buffer/';
-import { validateSchema } from './validation';
 import { Base64StringSchema } from './base64';
+import { validateSchema } from './validation';
 
+/**
+ * @see https://stackoverflow.com/a/39460727/974981
+ */
 export const base64ToHex = (base64: string): string => {
   const validated = validateSchema(Base64StringSchema, base64);
-  const buffer = Buffer.from(validated, 'base64');
-  return buffer.toString('hex');
+  const bytes = atob(validated);
+
+  let result = '';
+  for (let i = 0; i < bytes.length; i++) {
+    const hex = bytes.charCodeAt(i).toString(16);
+    result += hex.length === 2 ? hex : '0' + hex;
+  }
+
+  return result;
 };
 
+/**
+ * @see https://stackoverflow.com/a/41797377/974981
+ */
 export const hexToBase64 = (hex: string): string => {
-  const buffer = Buffer.from(hex, 'hex');
-  return buffer.toString('base64');
+  if (!hex) return '';
+
+  return btoa(
+    hex
+      .match(/\w{2}/g)!
+      .map(function (a) {
+        return String.fromCharCode(parseInt(a, 16));
+      })
+      .join(''),
+  );
 };
 
 /**
--- a/packages/types/src/hex.ts
+++ b/packages/types/src/hex.ts
@@ -1,5 +1,4 @@
-// Re: the trailing slash, see https://www.npmjs.com/package/buffer#usage
-import { Buffer } from 'buffer/';
+import { Buffer } from 'buffer';
 import { validateSchema } from './validation';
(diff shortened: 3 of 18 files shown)


apps/extension/webpack/webpack.common.js
```diff
@@ -70,6 +70,7 @@ module.exports = {
   },
   plugins: [
     new webpack.ProvidePlugin({
+      // Required by the `bip39` library
```
jessepinho, line 73: Added this note since I originally removed `buffer` from the extension, thinking we weren't using it anywhere.

packages/types/src/base64.ts
```diff
@@ -19,10 +18,11 @@ export const InnerBase64Schema = z.object({ inner: Base64StringSchema });
 
 export const base64ToUint8Array = (base64: string): Uint8Array => {
   const validated = validateSchema(Base64StringSchema, base64);
-  const buffer = Buffer.from(validated, 'base64');
-  return new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
+  const binString = atob(validated);
+  return Uint8Array.from(binString, m => m.codePointAt(0)!);
```
grod220, line 22: Can we confirm this will never throw? -> `m.codePointAt(0)!`

If not, think it's better to handle:

```typescript

  return Uint8Array.from(binString, m => {
    const codePoint = m.codePointAt(0);
    if (!codePoint) throw new Error('Invalid base64 string');
    return codePoint;
  });
```

packages/types/src/hex.ts
```diff
@@ -1,28 +1,55 @@
...
   // Check if the input string is a valid hexadecimal string
   if (!/^([0-9A-Fa-f]{2})*$/.test(hexString)) {
     throw new Error(`Invalid hexadecimal string: ${hexString}`);
   }
 
-  const buffer = Buffer.from(hexString, 'hex');
-  return new Uint8Array(buffer);
+  if (!hexString) return new Uint8Array();
+
+  return new Uint8Array(hexString.match(/.{1,2}/g)!.map(byte => parseInt(byte, 16)));
```
grod220, line 54: Think a code comment could be helpful:

```typescript
  // Split the string into pairs of characters
  const hexPairs = hexString.match(/.{1,2}/g)!;

  // Map each hexadecimal pair to the corresponding integer and create a Uint8Array from it
  return new Uint8Array(hexPairs.map(byte => parseInt(byte, 16)));
```

packages/types/src/hex.ts
```diff
@@ -1,28 +1,55 @@
...
+ * @see https://stackoverflow.com/a/41797377/974981
+ */
 export const hexToBase64 = (hex: string): string => {
-  const buffer = Buffer.from(hex, 'hex');
-  return buffer.toString('base64');
+  if (!hex) return '';
+
+  return btoa(
+    hex
+      .match(/\w{2}/g)!
```
grod220, line 28: Think an err assertion could be helpful

```typescript
 const hexPairs = hex.match(/[0-9A-Fa-f]{2}/g);
  if (!hexPairs) {
    throw new Error('Invalid hexadecimal input');
  }

  const binaryString = hexPairs
    .map(a => String.fromCharCode(parseInt(a, 16)))
    .join('');

   return btoa(binaryString)
```

grod220 (approved):
Great job! This has been a long running issue and a thorn in our side. Just had a few comments if they help.

packages/types/src/base64.ts
TalDerei: `atob` will only throws [InvalidCharacterError](https://developer.mozilla.org/en-US/docs/Web/API/atob#invalidcharactererror) for an invalid base64 encoding, which will be caught by `validateSchema`. 

We're asserting that `m.codePointAt(0)!` is non-null, but if `binString` is an empty string, then it'll be undefined and throw a runtime error. Can you have base64 encoded empty strings?

I think we should handle for errors regardless.

TalDerei (approved):

packages/types/src/hex.ts
jessepinho: good call — done

packages/types/src/hex.ts
jessepinho: Done!

packages/types/src/base64.ts
jessepinho: Hm... I don't think it's possible that this could throw. As @TalDerei pointed out, the string is already validated as Base64. Then, `Uint8Array.from()` is iterating over each character and calling `.codePointAt(0)` on it. Since it's a 1-character string, `.codePointAt(0)` would only return `undefined` if the character was somehow a byte that doesn't map to a Unicode code point. But since it's Base64, we know it does.

PR #393, opened 2024-01-25
don't attempt to type message json with Jsonified

turbocrime:
`Jsonified<AnyMessage>` becomes a black-box `JsonObject` to avoid any kind of structural inaccuracy, because protojson has rules that were not satisfied.

other types are converted more informatively.

jsonified is used in idb schema to identify contents and key types, which resolve to either `JsonObject` or primitives.

also, added type guards (and fixed some bugs) in offscreen.ts because it was a major internal user of jsonified.

use `Jsonified<T>` in your schemas/parameter/return types to convey the presence of unparsed json, and suggest what the json should deserialize to.

### DIFF
--- a/apps/extension/src/offscreen.ts
+++ b/apps/extension/src/offscreen.ts
@@ -1,9 +1,13 @@
-import type { JsonObject, JsonValue } from '@bufbuild/protobuf';
+import type { JsonValue } from '@bufbuild/protobuf';
 import {
   OffscreenRequest,
+  OffscreenResponse,
   ActionBuildRequest,
+  ActionBuildResponse,
   WasmBuildActionInput,
+  WasmBuildActionOutput,
   isActionBuildRequest,
+  isWasmBuildActionOutput,
 } from '@penumbra-zone/types/src/internal-msg/offscreen';
 
 export const isOffscreenRequest = (req: unknown): req is OffscreenRequest =>
@@ -19,20 +23,20 @@ export const offscreenMessageHandler = (
   sendResponse: (x: JsonValue) => void,
 ) => {
   if (!isOffscreenRequest(req)) return;
-  if (isActionBuildRequest(req.request)) {
-    const { type, request } = req;
+  const { type, request } = req;
+  if (isActionBuildRequest(request)) {
     void (async () => {
       const response = Promise.all(buildActionHandler(request));
       const res = await response
-        .then(data => ({ type, data }))
+        .then((data: ActionBuildResponse) => ({ type, data }))
         .catch((e: Error) => ({
           type,
           error: `Offscreen: ${e.message}`,
         }));
-      sendResponse(res);
+      sendResponse(res satisfies OffscreenResponse);
     })();
-  }
-  return true;
+    return true;
+  } else throw new Error('Unknown offscreen request');
 };
 
 chrome.runtime.onMessage.addListener(offscreenMessageHandler);
@@ -41,30 +45,31 @@ export const buildActionHandler = (request: ActionBuildRequest) => {
   // Destructure the data object to get individual fields
   const { transactionPlan, witness, fullViewingKey } = request;
 
-  return transactionPlan.actions.map((_, i) =>
-    spawnWorker(transactionPlan, witness, fullViewingKey, i),
+  return transactionPlan.actions.map((_, actionPlanIndex) =>
+    spawnWorker({ transactionPlan, witness, fullViewingKey, actionPlanIndex }),
   );
 };
 
-const spawnWorker = (
-  transactionPlan: JsonObject & { actions: JsonValue[] },
-  witness: JsonObject,
-  fullViewingKey: string,
-  actionPlanIndex: number,
-): Promise<JsonValue> => {
+const spawnWorker = ({
+  transactionPlan,
+  witness,
+  fullViewingKey,
+  actionPlanIndex,
+}: WasmBuildActionInput): Promise<WasmBuildActionOutput> => {
   return new Promise((resolve, reject) => {
     const worker = new Worker(new URL('./wasm-build-action.ts', import.meta.url));
 
-    const onWorkerMessage = (e: MessageEvent) => {
-      resolve(e.data as JsonValue);
+    const onWorkerMessage = (e: MessageEvent<unknown>) => {
       worker.removeEventListener('error', onWorkerError);
       worker.terminate();
+      if (isWasmBuildActionOutput(e.data)) resolve(e.data);
+      else reject(new Error(`No Action data from worker ${actionPlanIndex}`));
     };
 
     const onWorkerError = (error: ErrorEvent) => {
-      reject(new Error(`${error.filename}:${error.lineno} ${error.message}`));
       worker.removeEventListener('message', onWorkerMessage);
       worker.terminate();
+      reject(new Error(`${error.filename}:${error.lineno} ${error.message}`));
     };
 
     worker.addEventListener('message', onWorkerMessage, { once: true });
--- a/apps/extension/src/offscreen.ts
+++ b/apps/extension/src/offscreen.ts
@@ -66,12 +66,12 @@ const spawnWorker = ({
       else reject(new Error(`No Action data from worker ${actionPlanIndex}`));
     };
 
-    const onWorkerError = ({ filename, lineno, colno, message, error }: ErrorEvent) => {
+    const onWorkerError = (ev: ErrorEvent) => {
+      const { filename, lineno, colno, message } = ev;
       worker.removeEventListener('message', onWorkerMessage);
       worker.terminate();
-      const location = `Worker ${filename}:${lineno}:${colno}`;
-      console.error(location, message, error);
-      reject(error ?? new Error(`${location} ${message}`));
+      console.error('Worker ErrorEvent', ev);
+      reject(ev.error ?? new Error(`Worker ErrorEvent ${filename}:${lineno}:${colno} ${message}`));
     };
 
     worker.addEventListener('message', onWorkerMessage, { once: true });
--- a/apps/extension/src/offscreen.ts
+++ b/apps/extension/src/offscreen.ts
@@ -1,4 +1,5 @@
 import type { JsonValue } from '@bufbuild/protobuf';
+import { isJsonObject } from '@penumbra-zone/types';
 import {
   OffscreenRequest,
   OffscreenResponse,
@@ -7,7 +8,6 @@ import {
   WasmBuildActionInput,
   WasmBuildActionOutput,
   isActionBuildRequest,
-  isWasmBuildActionOutput,
 } from '@penumbra-zone/types/src/internal-msg/offscreen';
 
 export const isOffscreenRequest = (req: unknown): req is OffscreenRequest =>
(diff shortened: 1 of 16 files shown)


apps/extension/src/state/tx-approval.ts
```diff
@@ -1,13 +1,12 @@
...
-import { Jsonified } from '@penumbra-zone/types';
 import { MessageResponder } from '@penumbra-zone/types/src/internal-msg/shared';
 import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 
 export interface TxApprovalSlice {
-  authorizeRequest?: Jsonified<AuthorizeRequest>;
-  transactionViewFromPlan?: Jsonified<TransactionView>;
+  authorizeRequest?: AuthorizeRequest;
+  transactionViewFromPlan?: TransactionView;
```
turbocrime, line 9: not sure if zustand will tolerate this. other option is serialize to string because zustand doesn't tolerate `JsonValue`

packages/types/src/indexed-db.ts
turbocrime, line 1: these changes are entirely equivalent

apps/extension/src/state/tx-approval.ts
turbocrime: it doesn't like it, storing strings now

apps/extension/src/routes/popup/approval/transaction.tsx
```diff
@@ -16,10 +26,10 @@ export const TransactionApproval = () => {
           Confirm transaction
         </p>
 
-        <TransactionViewComponent txv={TransactionView.fromJson(transactionViewFromPlan)} />
+        <TransactionViewComponent txv={transactionViewFromPlan} />
 
         <div className='mt-8'>
-          <JsonViewer jsonObj={authorizeRequest} />
+          <JsonViewer jsonObj={authorizeRequest.toJson() as Jsonified<AuthorizeRequest>} />
```
grod220, line 32: I wonder if JsonViewer should take `JsonValue` (?)

apps/extension/src/state/tx-approval.ts
```diff
@@ -1,13 +1,10 @@
...
-import { Jsonified } from '@penumbra-zone/types';
 import { MessageResponder } from '@penumbra-zone/types/src/internal-msg/shared';
-import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 
 export interface TxApprovalSlice {
-  authorizeRequest?: Jsonified<AuthorizeRequest>;
-  transactionViewFromPlan?: Jsonified<TransactionView>;
+  authorizeRequest?: string;
+  transactionViewFromPlan?: string;
```
grod220, line 7: Why are json strings better than Json objects? Something like `Jsonified<AuthorizeRequest>` feels much more expressive here.

packages/types/src/internal-msg/offscreen.ts
```diff
@@ -10,29 +14,36 @@ export type ActionBuildMessage = InternalMessage<
...
   req.witness != null &&
+  typeof req.witness === 'object' &&
   'fullViewingKey' in req &&
   typeof req.fullViewingKey === 'string';
+
+export const isWasmBuildActionInput = (req: unknown): req is WasmBuildActionInput =>
+  isActionBuildRequest(req) && 'actionPlanIndex' in req && typeof req.actionPlanIndex === 'number';
+
+export const isWasmBuildActionOutput = (res: unknown): res is WasmBuildActionOutput =>
+  res != null && typeof res === 'object';
```
grod220, line 49: Is this too generic? This basically means any object could be a WasmBuildActionOutput?

packages/types/src/jsonified.ts
```diff
@@ -1,10 +1,10 @@
-import type { JsonValue, Message, PlainMessage } from '@bufbuild/protobuf';
+import type { JsonValue, JsonObject, AnyMessage } from '@bufbuild/protobuf';
 
 // prettier-ignore
 export type Jsonified<T> = T extends JsonValue ? T
```
grod220, line 4: Think this type needs some documentation about what all these extends do

apps/extension/src/routes/popup/approval/transaction.tsx
turbocrime: considered making that change but chose to limit scope

apps/extension/src/state/tx-approval.ts
turbocrime: it's an alias to jsonobject and zustand refuses to store jsonobject

turbocrime: wrapped with a `Stringified<J>` type that aliases to string

packages/types/src/internal-msg/offscreen.ts
turbocrime: replaced with a JsonObject type guard. we can't guard more than this without inspecting protojson, and we just parse it with fromJson immediately anyway, so it's fine.

packages/types/src/jsonified.ts
turbocrime: doc'd

apps/extension/src/routes/popup/approval/transaction.tsx
turbocrime: ok, now JsonViewer wants JsonObject (we will probably never just hand it a bare primitive)

apps/extension/src/offscreen.ts
```diff
@@ -41,30 +45,31 @@ export const buildActionHandler = (request: ActionBuildRequest) => {
...
       worker.terminate();
+      if (isJsonObject(e.data)) resolve(e.data);
+      else reject(new Error(`No Action data from worker ${actionPlanIndex}`));
     };
 
     const onWorkerError = (error: ErrorEvent) => {
-      reject(new Error(`${error.filename}:${error.lineno} ${error.message}`));
       worker.removeEventListener('message', onWorkerMessage);
       worker.terminate();
+      reject(new Error(`${error.filename}:${error.lineno} ${error.message}`));
```
grod220, line 72: I suppose we can't `reject(error)`?

apps/extension/src/state/tx-approval.ts
```diff
@@ -1,13 +1,13 @@
...
-import { Jsonified } from '@penumbra-zone/types';
 import { MessageResponder } from '@penumbra-zone/types/src/internal-msg/shared';
-import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
 import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
 
 export interface TxApprovalSlice {
-  authorizeRequest?: Jsonified<AuthorizeRequest>;
-  transactionViewFromPlan?: Jsonified<TransactionView>;
+  authorizeRequest?: Stringified<AuthorizeRequest>;
+  transactionViewFromPlan?: Stringified<TransactionView>;
```
grod220, line 10: Is this because of some zustand/immer type thing why we can't use JsonValue? Regardless, can you add a code comment above talking about why we specifically stringify?

grod220 (approved):

apps/extension/src/offscreen.ts
turbocrime: the 'error' is actually an `ErrorEvent`, not an `Error`. i can update the name to make that clear

jessepinho (commented):
OK, mostly LGTM except for one main hangup: I don't understand why we're switching to use a stringified version of the objects in the popup's Zustand state. Why not keep them as objects?

packages/types/src/internal-msg/offscreen.ts
```diff
@@ -10,29 +14,33 @@ export type ActionBuildMessage = InternalMessage<
...
+  'actions' in req.transactionPlan &&
+  Array.isArray(req.transactionPlan.actions) &&
   'witness' in req &&
-  typeof req.witness === 'object' &&
   req.witness != null &&
+  typeof req.witness === 'object' &&
   'fullViewingKey' in req &&
   typeof req.fullViewingKey === 'string';
+
+export const isWasmBuildActionInput = (req: unknown): req is WasmBuildActionInput =>
```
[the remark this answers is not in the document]
turbocrime: i'm not sure what you're asking here? this is a type guard

apps/extension/src/offscreen.ts
```diff
@@ -41,30 +45,31 @@ export const buildActionHandler = (request: ActionBuildRequest) => {
...
   return new Promise((resolve, reject) => {
     const worker = new Worker(new URL('./wasm-build-action.ts', import.meta.url));
 
-    const onWorkerMessage = (e: MessageEvent) => {
-      resolve(e.data as JsonValue);
+    const onWorkerMessage = (e: MessageEvent<unknown>) => {
       worker.removeEventListener('error', onWorkerError);
       worker.terminate();
+      if (isJsonObject(e.data)) resolve(e.data);
+      else reject(new Error(`No Action data from worker ${actionPlanIndex}`));
```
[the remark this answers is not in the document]
turbocrime: the worker could theoretically send us anything, but if it doesn't satisfy `JsonObject` something is very wrong.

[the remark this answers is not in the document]
turbocrime: ultimately the check is unnecessary, as immediately after return we attempt to deserialize it, and that would also fail if it isn't jsonobject. but, our typescript and linter config gets mad if i return any/unknown. i could just deserialize here but was going for a minimal set of changes.

apps/extension/src/routes/popup/approval/tx-msg-handler.ts
```diff
@@ -2,15 +2,21 @@ import { TxApproval } from '@penumbra-zone/types/src/internal-msg/tx-approval';
...
 export const handleTxApproval: InternalMessageHandler<TxApproval> = (jsonReq, responder) => {
   useStore.setState(state => {
-    state.txApproval.authorizeRequest = jsonReq.authorizeRequest;
-    state.txApproval.transactionViewFromPlan = jsonReq.transactionViewFromPlan;
+    state.txApproval.authorizeRequest = AuthorizeRequest.fromJson(
+      jsonReq.authorizeRequest,
+    ).toJsonString();
+    state.txApproval.transactionViewFromPlan = TransactionView.fromJson(
+      jsonReq.transactionViewFromPlan,
+    ).toJsonString();
```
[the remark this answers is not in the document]
turbocrime: zustand refuses to store jsonobject. the correct solution is storing a plainmessage. this is the least of our ser/de problems.

this is the minimal viable solution for removing jsonified pb types.

packages/types/src/internal-msg/offscreen.ts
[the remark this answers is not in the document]
turbocrime: it's not a very useful type guard, but it's not a very useful type :)

turbocrime:
> OK, mostly LGTM except for one main hangup: I don't understand why we're switching to use a stringified version of the objects in the popup's Zustand state. Why not keep them as objects?

Zustand refuses to store an object of potentially infinite depth, which `JsonObject` does represent. The minimal solution is to stringify.

PR #394, opened 2024-01-25
Fix Vite CJS warning

### LINKED ISSUE #271: Vite updates

jessepinho:
During test runs, we were getting the warning `The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.`. This is because some of our config files were using CJS modules (e.g., `const foo = require('foo')` instead of `import foo from 'foo'`, and `module.exports = {}` instead of `export default {}`).

## In this PR
- Modify Tailwind/PostCSS/Webpack config files to use ES modules instead of CJS modules (which are a bit outdated).
- Add `"type": "module"` to the packages that Vite was complaining about. I also added it to every other package as well, for consistency's sake.
- Rename ESLint config files from `.eslintrc.js` to `.eslintrc.cjs` (note the file extension). Unfortunately, ESLint doesn't support ES modules, so this is [required](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats) for packages that have `"type": "module"` in their `package.json`.

Closes #271


apps/extension/postcss.config.js
```diff
@@ -1 +1 @@
-module.exports = require('@penumbra-zone/ui/postcss.config');
+export { default } from '@penumbra-zone/ui/postcss.config.js';
```
jessepinho, line 1: Since `@penumbra-zone/ui` (and every other package in this monorepo) is now a `"type": "module"`, the Node resolver treats an import without an extension differently, expecting it to be listed in the module's `package.json` `files` setting (or something?). Adding `.js` makes it explicit that we are importing that file specifically.

grod220 (approved):
LGTM

TalDerei (approved):

TalDerei:
LGTM!

turbocrime (approved):

PR #395, opened 2024-01-25
Update all the things

jessepinho:
Note: reading the description below before reviewing will prob help you breeze through this review :D 

## In this PR
- Run `pnpm up -r --latest` (except for the `@buf/*` packages, since those require logic changes that I don't have the expertise to finish, and anyway I think @TalDerei is working on that).
- Address tons of ESLint/Prettier issues. The highlights:
  - The `plugin:tailwindcss/recommended` ESLint config prefers the `size-N` shorthand instead of `h-N w-N`, when width and height are the same. So **the vast majority of linting fixes are for this**.
  - Prettier complained about lack of trailing slashes in our `tsconfig.json` files, which is a little weird to me, since JSON doesn't technically support trailing slashes. But I updated them nonetheless.
  - ESLint now expects promises to reject with an instance of `Error`, rather than with any old value. So I had to make a few changes to make ESLint happy. However, they're in places that I don't really know how to test, so if reviewers have suggestions, that'd be helpful.


apps/extension/src/offscreen.ts
```diff
@@ -71,7 +71,11 @@ const spawnWorker = ({
       worker.removeEventListener('message', onWorkerMessage);
       worker.terminate();
       console.error('Worker ErrorEvent', ev);
-      reject(ev.error ?? new Error(`Worker ErrorEvent ${filename}:${lineno}:${colno} ${message}`));
+      reject(
+        ev.error instanceof Error
+          ? ev.error
+          : new Error(`Worker ErrorEvent ${filename}:${lineno}:${colno} ${message}`),
+      );
```
jessepinho, line 78: Reviewers: [the v6.19.0 release](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.19.0) of the TypeScript ESLint plugin introduced a [rule](https://eslint.org/docs/latest/rules/prefer-promise-reject-errors) that Promise rejections should contain an `Error` instance. I assume we're preferring to stick to the default ESLint config in general, so I updated this code to adhere to the rule. However, I have no idea how to test it manually :)

apps/extension/webpack/webpack.dev.js
```diff
@@ -1,4 +1,4 @@
-import merge from 'webpack-merge';
+import { merge } from 'webpack-merge';
```
jessepinho, line 1: This file was breaking without this change, which is weird, since it appears that `merge` is _also_ the default export of `webpack-merge`. But nonetheless, this change fixed it, so 🤷🏻‍♂️

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -20,7 +20,6 @@ interface ContextsRequest {
 let active = 0;
 
 const activateOffscreen = async () => {
-  // @ts-expect-error: no types available yet
```
jessepinho, line 23: Apparently, this doesn't trigger a TS error (at least, not anymore?), so TS complained:
image

packages/transport/src/create.ts
```diff
@@ -185,7 +185,8 @@ export const createChannelTransport = ({
         return streamImpl as MethodImpl<typeof method>;
       }
       default: {
-        const noImpl: unknown = () => Promise.reject(`${MethodKind[method.kind]} unimplemented`);
+        const noImpl: unknown = () =>
+          Promise.reject(new Error(`${MethodKind[method.kind]} unimplemented`));
```
jessepinho, line 189: This is due to that same linting issue where `reject()` should take an `Error` object. Again, I'm not sure how to test this.

packages/types/src/internal-msg/shared.ts
```diff
@@ -18,9 +18,10 @@ export type InternalMessageHandler<M extends InternalMessage<string, unknown, un
     : (request: M['request'], response: MessageResponder<M>) => void;
 
 // The awaited response sent back to requestor
-export type AwaitedInternalResponse<T> = T extends InternalMessage<infer Type, unknown, infer Res>
-  ? { type: Type; data: Awaited<Res> }
-  : never;
+export type AwaitedInternalResponse<T> =
```
jessepinho, line 21: Prettier formatting change — the type itself is the same.

packages/types/src/std-route.ts
```diff
@@ -8,11 +8,12 @@ export interface ServiceWorkerRequest<T extends SwRequestMessage> {
 }
 
 // The core message sent via content-script/popup
-export type IncomingRequest<T> = T extends InternalMessage<infer Type, infer Req, unknown>
-  ? Req extends undefined
-    ? { type: Type }
-    : { type: Type; arg: Req }
-  : never;
+export type IncomingRequest<T> =
```
jessepinho, line 11: Prettier formatting change — the type itself is the same.

packages/types/src/transaction/perspective.ts
```diff
@@ -69,7 +69,7 @@ export const viewFromEmptyPerspective = (tx: Transaction): TransactionView => {
       }),
       actionViews: tx.body.actions.map(action => {
         return viewActionFromEmptyPerspective(action)!;
-      })!,
+      }),
```
jessepinho, line 72: Apparently makes no difference, so ESLint said to get rid of it.

apps/webapp/src/fetchers/stream.ts
```diff
@@ -65,7 +65,9 @@ export const streamToPromise = <T>(query: AsyncIterable<T>): Promise<T[]> => {
         }
         resolve(result);
       } catch (e) {
-        reject(e);
+        if (e instanceof Error) reject(e);
+        else if (typeof e === 'string') reject(new Error(e));
+        else reject(new Error('Unknown error in `streamToPromise`'));
```
jessepinho, line 70: Due to [the aforementioned ESLint rule](https://eslint.org/docs/latest/rules/prefer-promise-reject-errors) about only rejecting with an instance of `Error`, I had to rework this `catch` clause a bit to ensure that it's actually rejecting an `Error` instance. I figured that the `Error`/`string` cases should catch most errors; anything else, the developer encountering this error will have to dig a bit deeper.

I don't know when this would even get triggered, since this is related to some view server stuff that I'm less familiar with. But if devs have suggestions, I'm all 👂🏻 👂🏻

apps/webapp/src/fetchers/stream.test.ts
```diff
@@ -0,0 +1,47 @@
+import { beforeEach, describe, expect, test } from 'vitest';
+import { streamToPromise } from './stream';
+
+describe('streamToPromise()', () => {
```
jessepinho, line 4: See my description inside the `stream.ts` file below — I added these tests due to having to rework the `catch` logic

grod220 (approved):
The new json trailing comma has me puzzled. I feel like that could be a eslint bug (?). May be worth checking the release notes for the update... or maybe not! Up to you.

grod220:
Oh dear, drama: https://github.com/prettier/prettier/issues/15956. I guess we let the community duke this one out.

jessepinho:
Ha, good find @grod220. Maybe we'll have to switch back our TSConfigs if they revert that decision lol

PR #418, opened 2024-01-27
wip smoke tests

turbocrime:
first attempt at playwright tests

1. it seems like playwright doesnt provide access to the anchor segment of the page uri, which we use heavily
2. none of our elements have roles/labels/ids/aria so it's very difficult to write selectors
3. playwright tests should exec after page settles, but it seems like they are executing too early
4. passphrase input doesn't behave properly with the automated input

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 
 # testing
 coverage
+e2e/playwright-user-data
 
 # build output
 .next/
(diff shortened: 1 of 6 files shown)


turbocrime:
nothing in this (incomplete, broken) set of tests actually used a feature of playwright that fails to expose through vitest, so this could be rewritten in vitest.

Issue #390, opened 2024-01-24
Swap UI [part 1]


grod220:
- Re-enable frontend tab
- New denom-out UI component (non-editable)
- `planWitnessBuildBroadcast` should be a shared function between send/ibc/swap
- Implement proper Denom Registry. Related: https://github.com/penumbra-zone/web/issues/140
- Add swap/swap commitment to view-protocol-server/transaction-planner.ts

turbocrime:
lpnft asset display names look pretty ugly in the ui right now

grod220:
Can you share a screenshot here? @turbocrime

PR #419, opened 2024-01-29
Swap [ui part 1]

grod220:
Closes: https://github.com/penumbra-zone/web/issues/390

The first part of the multi-part PR series in doing swaps. Does a few things:

- Implements new swap asset selector components
- Refactor swap slice
- Implement transaction wasm planning for swaps

### DIFF
--- a/apps/webapp/src/components/header/navbar.tsx
+++ b/apps/webapp/src/components/header/navbar.tsx
@@ -15,8 +15,7 @@ export const Navbar = () => {
             to={link.href}
             className={cn(
               'font-bold py-[10px] select-none',
-              link.subLinks &&
-                link.subLinks.includes(pathname) &&
+              (link.href === pathname || link.subLinks?.includes(pathname)) &&
                 'bg-button-gradient-secondary px-[30px] rounded-lg',
             )}
           >
--- a/apps/webapp/src/components/swap/asset-out-box.tsx
+++ b/apps/webapp/src/components/swap/asset-out-box.tsx
@@ -1,0 +1,71 @@
+import { useStore } from '../../state';
+import { swapSelector } from '../../state/swap';
+import { AccountBalance, AssetBalance, groupByAsset } from '../../fetchers/balances';
+import { Input } from '@penumbra-zone/ui';
+import { AssetOutSelector } from './asset-out-selector';
+import { uint8ArrayToBase64 } from '@penumbra-zone/types';
+import {
+  DenomMetadata,
+  ValueView,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
+import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
+
+interface AssetOutBoxProps {
+  balances: AccountBalance[];
+}
+
+export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
+  const { assetOut, setAssetOut } = useStore(swapSelector);
+
+  const aggregatedBalances = balances
+    .flatMap(b => b.balances)
+    .reduce<AssetBalance[]>(groupByAsset, []);
+
+  // TODO: with https://github.com/penumbra-zone/web/issues/392 convert to use `getValueViewByAccount`
+  const balanceOfDenom = aggregatedBalances.find(
+    b => uint8ArrayToBase64(b.assetId.inner) === assetOut?.penumbraAssetId.inner,
+  );
+  const valueView = balanceOfDenom
+    ? new ValueView({
+        valueView: {
+          case: 'knownDenom',
+          value: {
+            amount: balanceOfDenom.amount,
+            denom: new DenomMetadata({
+              display: balanceOfDenom.denom.display,
+              denomUnits: [balanceOfDenom.denom],
+            }),
+          },
+        },
+      })
+    : new ValueView();
+
+  return (
+    <div className='flex flex-col rounded-lg border bg-background px-4 pb-5 pt-3'>
+      <div className='mb-2 flex items-center justify-between gap-1 md:gap-2'>
+        <p className='text-sm font-bold md:text-base'>Swap into</p>
+      </div>
+      <div className='flex items-center justify-between gap-4'>
+        <Input
+          variant='transparent'
+          type='number'
+          className='font-bold leading-10 md:h-8 md:w-[calc(100%-80px)] md:text-xl  xl:h-10 xl:w-[calc(100%-160px)] xl:text-3xl'
+          // TODO: estimate actual swap out amount button: https://github.com/penumbra-zone/web/issues/421
+          value=''
+        />
+        <AssetOutSelector
+          balances={aggregatedBalances}
+          assetOut={assetOut}
+          setAssetOut={setAssetOut}
+        />
+      </div>
+      <div className='mt-[6px] flex items-start justify-between'>
+        <div />
+        <div className='flex items-start gap-1'>
+          <img src='./wallet.svg' alt='Wallet' className='size-5' />
+          <ValueViewComponent view={valueView} />
+        </div>
+      </div>
+    </div>
+  );
+};
--- a/apps/webapp/src/components/swap/asset-out-box.tsx
+++ b/apps/webapp/src/components/swap/asset-out-box.tsx
@@ -3,7 +3,6 @@ import { swapSelector } from '../../state/swap';
 import { AccountBalance, AssetBalance, groupByAsset } from '../../fetchers/balances';
 import { Input } from '@penumbra-zone/ui';
 import { AssetOutSelector } from './asset-out-selector';
-import { uint8ArrayToBase64 } from '@penumbra-zone/types';
 import {
   DenomMetadata,
   ValueView,
@@ -22,9 +21,7 @@ export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
     .reduce<AssetBalance[]>(groupByAsset, []);
 
   // TODO: with https://github.com/penumbra-zone/web/issues/392 convert to use `getValueViewByAccount`
-  const balanceOfDenom = aggregatedBalances.find(
-    b => uint8ArrayToBase64(b.assetId.inner) === assetOut?.penumbraAssetId.inner,
-  );
+  const balanceOfDenom = aggregatedBalances.find(b => b.assetId.equals(assetOut?.penumbraAssetId));
   const valueView = balanceOfDenom
     ? new ValueView({
         valueView: {
--- a/apps/webapp/src/components/swap/asset-out-selector.tsx
+++ b/apps/webapp/src/components/swap/asset-out-selector.tsx
@@ -1,0 +1,49 @@
+import { AssetBalance } from '../../fetchers/balances';
+import { Dialog, DialogClose, DialogContent, DialogHeader, DialogTrigger } from '@penumbra-zone/ui';
+import { AssetIcon } from '../shared/asset-icon';
+import { assets } from '@penumbra-zone/constants/assets';
+import { Asset } from '@penumbra-zone/types';
+
+interface AssetOutSelectorProps {
+  balances: AssetBalance[];
+  assetOut: Asset | undefined;
+  setAssetOut: (asset: Asset) => void;
+}
+
+export const AssetOutSelector = ({ balances, setAssetOut, assetOut }: AssetOutSelectorProps) => {
+  return (
+    <Dialog>
+      <DialogTrigger disabled={!balances.length}>
+        <div className='flex h-9 min-w-[100px] items-center justify-center gap-2 rounded-lg bg-light-brown px-2'>
+          {assetOut?.display && <AssetIcon name={assetOut.display} />}
+          <p className='font-bold text-light-grey md:text-sm xl:text-base'>{assetOut?.display}</p>
+        </div>
+      </DialogTrigger>
+      <DialogContent className='max-w-[312px] bg-charcoal-secondary md:max-w-[400px]'>
+        <div className='relative z-10 flex flex-col gap-4 pb-5'>
+          <DialogHeader className='border-b'>Select asset</DialogHeader>
+          <div className='flex flex-col gap-4 px-[30px]'>
+            <div className='flex flex-col gap-2'>
+              {assets.map(a => (
+                <div key={a.display} className='flex flex-col'>
+                  <DialogClose>
+                    <div
+                      className='grid cursor-pointer grid-cols-3 py-[10px] font-bold text-muted-foreground hover:-mx-4 hover:bg-light-brown hover:px-4'
+                      onClick={() => setAssetOut(a)}
+                    >
+                      <div className='flex justify-start gap-[6px]'>
+                        <AssetIcon name={a.display} />
+                        <p>{a.display}</p>
+                      </div>
+                    </div>
+                  </DialogClose>
+                </div>
+              ))}
+            </div>
+          </div>
+        </div>
+        <div className='absolute inset-0 z-0 bg-card-radial opacity-20' />
+      </DialogContent>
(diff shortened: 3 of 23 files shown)


apps/webapp/src/components/header/navbar.tsx
```diff
@@ -15,8 +15,7 @@ export const Navbar = () => {
             to={link.href}
             className={cn(
               'font-bold py-[10px] select-none',
-              link.subLinks &&
-                link.subLinks.includes(pathname) &&
+              (link.href === pathname || link.subLinks?.includes(pathname)) &&
```
grod220, line 18: Felt a bit annoying that a path had to add itself to its own set of sublinks

apps/webapp/src/components/swap/asset-out-box.tsx
grod220, line 1: Screenshot 2024-01-29 at 4 01 16 PM

apps/webapp/src/components/swap/asset-out-selector.tsx
grod220, line 1: Screenshot 2024-01-29 at 4 01 21 PM

packages/storage/src/indexed-db/index.ts
```diff
@@ -223,46 +223,6 @@ export class IndexedDb implements IndexedDbInterface {
     }
   }
 
-  private addSctUpdates(txs: IbdUpdates, sctUpdates: ScanBlockResult['sctUpdates']): void {
```
grod220, line 226: A lint in webstorm re-orders these fyi (alphabetically?)

apps/webapp/src/components/shared/select-token-modal.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import { fromBaseUnitAmount } from '@penumbra-zone/types';
 import { cn } from '@penumbra-zone/ui/lib/utils';
 import { AccountBalance } from '../../fetchers/balances';
-import { AssetIcon } from './asset-icon';
+import { AssetIcon } from './asset-icon.tsx';
```
jessepinho, line 14: Is this intentional?

I've noticed that, throughout the web monorepo, we're importing from `some-file.tsx` in some places, and just `some-file` in others. Any idea why that is? Feels like a mistake.

packages/constants/assets.ts
```diff
@@ -1,165 +1,148 @@
...
 };
 
-// TODO: This should only have things like icon or other fields not available in assets proto
-export const assets: Asset[] = [
+export interface Asset {
+  metadata: DenomMetadata;
+  icon?: string;
+}
+
+export const AssetRegistry: Asset[] = [
```
jessepinho, line 25: Not at all a blocker, but what do you think of making this an object, keyed by e.g., display name? Then you could just look up metadata on the object by name, rather than iterating over an array and memoizing the result like you're doing [here](https://github.com/penumbra-zone/web/pull/419/files#diff-4ec70347e05c7bcd69a141c28ebc60dfbf558278548fe0b51ce29b002ff07188R7).

apps/webapp/src/state/swap.ts
```diff
@@ -1,103 +1,105 @@
...
-  pay: SwapAssetInfo;
-  receive: SwapAssetInfo;
-  validationErrors: SwapValidationFields;
-  setAmount: (type: SwapInputs) => (amount: string) => void;
-  setAsset: (type: SwapInputs) => (asset: AssetId) => void;
-  replaceAsset: () => void;
-  setAssetBalance: (amount: number) => void;
+  assetIn: Selection | undefined;
+  setAssetIn: (selection: Selection) => void;
+  amount: string;
```
jessepinho, line 20: Shouldn't this be a `number`? (and get converted in the `InputToken`'s [`onChange` handler](https://github.com/penumbra-zone/web/pull/419/files#diff-0a9c5fc7496ea1a47f7fef2e351c40f80cdc243873cbf333c5412ddac5ee1f81R30-R33))?

apps/webapp/src/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <InputToken
+        label='Amount to swap'
+        placeholder='Enter an amount'
+        className='mb-1'
+        selection={assetIn}
+        setSelection={setAssetIn}
+        value={amount}
+        onChange={e => {
+          if (Number(e.target.value) < 0) return;
+          setAmount(e.target.value);
```
jessepinho, line 32: See my note in the Zustand state — shouldn't this be converted to a number first?

apps/webapp/src/components/swap/asset-out-box.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
+  const { assetOut, setAssetOut } = useStore(swapSelector);
+
+  const aggregatedBalances = balances
+    .flatMap(b => b.balances)
+    .reduce<AssetBalance[]>(groupByAsset, []);
+
+  const balanceOfDenom = aggregatedBalances.find(b =>
+    b.assetId.equals(assetOut?.metadata.penumbraAssetId),
+  );
```
jessepinho, line 21: I'm personally a fan of doing as little logic in the template as possible, so I'd prefer adding another const below this one, something like this:
```suggestion
  );
  const formattedBalanceOfDenom = balanceOfDenom
              ? fromBaseUnitAmount(balanceOfDenom.amount, balanceOfDenom.denom.exponent).toFormat()
              : '0';
```

Then, you can just render `{formattedBalanceOfDenom}` in the markup below.

(Not a blocker though — just a nitpick.)

apps/webapp/src/state/send.ts
```diff
@@ -97,7 +97,7 @@ const assembleRequest = async ({ amount, recipient, selection, memo }: SendSlice
     ],
     source: new AddressIndex({ account: selection.accountIndex }),
     memo: new MemoPlaintext({
-      returnAddress: await getAddressByIndex(selection.accountIndex),
+      returnAddress: await getEphemeralAddress(selection.accountIndex),
```
jessepinho, line 100: Wait, why this change? We always want sends to be from an ephemeral address?

apps/webapp/src/state/swap.ts
```diff
@@ -1,103 +1,105 @@
...
+          state.send.txInProgress = false;
+        });
+      } catch (e) {
+        set(state => {
+          state.send.txInProgress = false;
+        });
+        dismiss();
+        toastFn(errorTxToast(e));
+        throw e;
+      }
```
jessepinho, line 74: Up to you, but some of this could be pulled into a `finally`:

```suggestion
        toastFn(successTxToast(txHash));

        // Reset form
        set(state => {
          state.send.amount = '';
        });
      } catch (e) {
        toastFn(errorTxToast(e));
        throw e;
      } finally {
        set(state => {
          state.send.txInProgress = false;
        });
        dismiss();
      }
```

apps/webapp/src/state/swap.ts
```diff
@@ -1,103 +1,105 @@
...
+
+  return new TransactionPlannerRequest({
+    swaps: [
+      {
+        targetAsset: assetOut.metadata.penumbraAssetId,
+        value: {
+          amount: toBaseUnit(BigNumber(amount), assetIn.asset.denom.exponent),
+          assetId: assetIn.asset.assetId,
+        },
+        claimAddress: await getEphemeralAddress(assetIn.accountIndex),
```
jessepinho, line 91: Why an ephemeral address? Is that just a default we want to use? (Seems unnecessary in a swap though, since no one else can read anything in the transaction, right?)

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
```diff
@@ -28,10 +48,27 @@ export const transactionPlanner: Impl['transactionPlanner'] = async (req, ctx) =
...
+    if (!value || !targetAsset || !fee || !claimAddress)
+      throw new Error('no value, targetAsset, fee or claimAddress in swap');
+
+    const intoDenomMetadata = await getDenomMetadata(indexedDb, targetAsset, querier);
+    planner.swap(value, intoDenomMetadata, fee, claimAddress);
+  }
+
+  for (const { swapCommitment } of req.swapClaims) {
+    if (!swapCommitment) throw new Error('no swapCommitment in swapClaim');
+    await planner.swapClaim(swapCommitment);
```
jessepinho, line 66: Interesting... so, we can plan the swap and the swap claim at the same time, even though they'll be in separate transactions/blocks?

apps/webapp/src/components/swap/swap-form.tsx
```diff
@@ -0,0 +1,44 @@
+import { Button } from '@penumbra-zone/ui';
+import { useToast } from '@penumbra-zone/ui/components/ui/use-toast';
+import InputToken from '../shared/input-token.tsx';
+import { useLoaderData } from 'react-router-dom';
+import { AccountBalance } from '../../fetchers/balances.ts';
+import { useStore } from '../../state';
+import { swapSelector } from '../../state/swap.ts';
+import { AssetOutBox } from './asset-out-box.tsx';
+
+export const SwapForm = () => {
```
jessepinho, line 10: I can't actually navigate to this UI in my local git repo to test it, since I'm getting the "missing trailer" issue we're discussing in Discord. But the code looks fine!

jessepinho (commented):
LGTM, with some suggestions/nitpicks that I'll leave up to you. I also am still unclear on how swaps + swap claims will be handled in the UI, since they're separate transactions/blocks.

jessepinho (approved):
whoops, forgot to approve

apps/webapp/src/components/swap/asset-out-box.tsx
hdevalence: why is this component manually formatting the balances, instead of having that logic only in one place, in a component rendering a `ValueView` (the same way we only format addresses once in a component rendering an `AddressView`)?

apps/webapp/src/state/send.ts
hdevalence: i'm guessing this was a rebase or something, but we never want to use ephemeral addresses here

apps/webapp/src/state/swap.ts
hdevalence: we want to avoid using one-time addresses anywhere except ibc deposits, where they're absolutely necessary, this should indeed be using the default address for the account

packages/storage/src/indexed-db/index.ts
turbocrime: can we avoid lint/format changes that aren't defined in ci by eslint

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
hdevalence: i don't know about the surrounding context, but in general, planning the swap claim requires knowledge of the `BatchSwapOutputData` for the block the swap was submitted in, which arrives in the `CompactBlock` we use to detect when the swap was committed on-chain. so the swap claim can't be planned at the same time as the swap but it can be planned as soon as the swap is detected

turbocrime: this would only attempt to plan a swap and swapClaim at the same time if the caller requests that a swap and swapClaim be planned at the same time by including both in the TransactionPlannerRequest.

the caller probably does not have the information to make such a request, unless they can independently plan and build transactions, and write to idb. but it's possible by pb spec.

packages/constants/assets.ts
turbocrime: note there are about a thousand assets that will eventually be in this table

apps/webapp/src/components/swap/asset-out-box.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <p className='text-sm font-bold md:text-base'>Swap into</p>
+      </div>
+      <div className='flex items-center justify-between gap-4'>
+        <Input
+          variant='transparent'
+          type='number'
+          className='font-bold leading-10 md:h-8 md:w-[calc(100%-80px)] md:text-xl  xl:h-10 xl:w-[calc(100%-160px)] xl:text-3xl'
+          // TODO: estimate actual swap out amount
+          value=''
+        />
```
jessepinho, line 35: Ah, just noticed this — so, this will be done in a follow-up PR?

packages/constants/assets.ts
```diff
@@ -1,165 +1,148 @@
...
+  unbondingToken: new RegExp('^unbonding_'),
+  votingReceipt: new RegExp('^voted_on_'),
 };
 
-// TODO: This should only have things like icon or other fields not available in assets proto
-export const assets: Asset[] = [
+export interface Asset {
+  metadata: DenomMetadata;
+  icon?: string;
+}
```
turbocrime, line 23: this should be renamed so it is never confused with the pb Asset type that is used in many places

apps/webapp/src/components/swap/layout.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <div className='relative mx-auto grid gap-6 md:grid-cols-2 md:gap-4 xl:max-w-[1276px] xl:grid-cols-3 xl:gap-5'>
+      <div className='hidden xl:order-1 xl:block' />
+      <Card gradient className='order-2 row-span-2 flex-1 p-5 md:order-1 md:p-4 xl:p-5'>
+        <SwapForm />
+      </Card>
+      <EduInfoCard
+        className='order-1 md:order-2'
+        src='./receive-gradient.svg'
+        label='Swap me'
+        content={EduPanel.TEMP_FILLER}
```
jessepinho, line 42: https://film.jessepinho.com/lorem-ipsum

😏

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
turbocrime: i had a comment here about why it isn't possible to make such a request but then accidentally deleted it just now, lol. but

1. it's valid pb to make such a request
2. caller doesn't have the information to do so unless they can plan/build independently
3. such a request would likely fail in wasm, which queries idb for that swap commitment, which should not be present yet

packages/constants/assets.ts
hdevalence: I don't have context here but why do we want this at all? shouldn't the icon be part of the `DenomMetadata`?

hdevalence:
I think it could be good to minimize the amount of further changes we do to asset info in this PR and work on an asset registry as a separate effort, since it's something of a project in its own right.

One thing I noticed by looking at this PR is that the existing web code has a lot of ad hoc types for asset information. I think it would be good to just use `DenomMetadata` everywhere, and ensure it's fit for that purpose. We shouldn't be ending up in a place where the web code has to define a type that's "`DenomMetadata` plus an icon", the `DenomMetadata` should just have that field. It's a type we made up, after all.

packages/types/src/transaction/perspective.ts
```diff
@@ -43,6 +47,20 @@ export const viewActionFromEmptyPerspective = (action: Action): ActionView | und
...
+          value: new SwapView({
+            swapView: {
+              case: 'opaque',
+              value: new SwapView_Opaque({
+                swap: action.action.value,
+              }),
+            },
+          }),
+        },
+      });
```
turbocrime, line 63: i guess this matches other sections of this file, but i'd like to point out you don't have to nest constructors like this - it's fine to just use the top-level type

```suggestion
    case 'swap':
      return new ActionView({
        actionView: {
          case: 'swap',
          value: {
            swapView: {
              case: 'opaque',
              value: {
                swap: action.action.value,
              },
            },
          },
        },
      });
```

turbocrime (commented):
this needs some rebase, notably i went ahead and killed those weird asset types, and the new proto specs change methods to streaming, which all definitely has downstream effects on you. but otherwise nothing stands out and im sleepy

PR #426, opened 2024-01-30
more v0.65 updates

turbocrime:


turbocrime:
currently it seems like fmdParameters is no longer arriving in the compact block cc @erwanor 

it looks like most places were covered by [penumbra-zone/penumbra/04208aa](https://github.com/penumbra-zone/penumbra/commit/04208aa1a0b974c153ba3add78afb7b0cfade223) but maybe fmd needs to be set deliberately

is https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/compact-block/src/component/rpc.rs involved?

hdevalence:
Looks like they're still there in the proto, maybe they're just not being populated at the start?

IMO not essential to worry about since we never change them (we need to fix that pre mainnet, but it would look like just noticing that there are parameters set in the CB and updating)

apps/webapp/src/components/shared/asset-icon.tsx
turbocrime, line 1: casualty of types/src/asset.ts

apps/webapp/src/state/swap.test.ts
turbocrime, line 1: casualty of types/src/asset.ts

package.json
turbocrime, line 1: pulled typescript, connectrpc, bufbuild, tsconfig deps up to top level

packages/constants/assets.ts
turbocrime, line 1: casualty of types/src/asset.ts

packages/constants/package.json
turbocrime, line 1: all of these deps are top-level now

packages/crypto/package.json
turbocrime, line 1: etc

packages/types/src/asset.ts
turbocrime, line 1: these were pointless and will also be refactored soon, don't worry about it

turbocrime:
tx works. icons now broken

turbocrime:
those weird custom asset types broke with spec change, so i just ripped them out. all it broke was icons, afaict. would like to defer icons to #428

apps/webapp/src/components/shared/asset-icon.tsx
jessepinho: Won't this component still work as-is? Why not reset it to how it was in `main`, and let @grod220 's PR #419 take care of changing this?

package.json
jessepinho: I don't have any strong opinion about this, but it'd be nice to have a best practice for which deps go in the root `package.json` vs. which belong in individual packages' `package.json`s.

package.json
```diff
@@ -11,23 +11,29 @@
     "clean": "turbo clean",
     "format": "prettier --write .",
     "format-check": "prettier --check .",
-    "all-check": "pnpm install && pnpm format && pnpm format-check && pnpm lint && pnpm test && pnpm build"
```
jessepinho, line 14: Was this an intentional removal? My understanding is that `pnpm format` will only auto-fix the things it can, but it might not be able to auto-fix everything — in which case, `pnpm format-check` is still useful to keep here.

packages/constants/package.json
jessepinho: 👏🏻

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
-  if (!transaction) throw new Error('no transaction in response');
+  if (!transactionPlan) throw new Error('no plan in response');
 
-  const { id } = await viewClient.broadcastTransaction({ transaction, awaitDetection: true });
-  if (!id) throw new Error('no id in broadcast response');
+  let transaction;
+  for await (const { status } of viewClient.authorizeAndBuild({ transactionPlan })) {
+    switch (status.case) {
+      case 'buildProgress':
+        console.log('buildProgress', status.value.progress);
```
jessepinho, line 19:
```suggestion
        // no-op
```

(I assume we don't want actual logs in production, right?)

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
 
-  const { transaction } = await viewClient.authorizeAndBuild({
-    transactionPlan: plan,
-  });
-  if (!transaction) throw new Error('no transaction in response');
+  if (!transactionPlan) throw new Error('no plan in response');
 
-  const { id } = await viewClient.broadcastTransaction({ transaction, awaitDetection: true });
-  if (!id) throw new Error('no id in broadcast response');
+  let transaction;
```
jessepinho, line 15:
```suggestion
  let transaction: Transaction | undefined;
```

hdevalence (approved):
At a high level, this seems good. I just merged https://github.com/penumbra-zone/penumbra/pull/3702 , so it should be published now, and it could be used to add back the images that are removed here.

See https://github.com/penumbra-zone/penumbra/blob/main/crates/core/asset/src/asset/denom_metadata.rs#L527-L561 for an example JSON format. Also, at least the Rust implementations of JSON deserialization are now capable of reading the entries in the cosmos asset registry, so I would expect the Buf generated code is too (possibly with configuration to ignore unknown fields).

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
+  const broadcastStream = viewClient.broadcastTransaction({ transaction, awaitDetection });
+
+  const expectId = new TransactionId({ inner: await sha256Hash(transaction.toBinary()) });
+
+  let detectionHeight;
+
+  for await (const { status } of broadcastStream) {
+    switch (status.case) {
+      case 'broadcastSuccess':
+        console.log('broadcastSuccess', status.value);
```
jessepinho, line 39:
```suggestion
```

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
+  for await (const { status } of viewClient.authorizeAndBuild({ transactionPlan })) {
+    switch (status.case) {
+      case 'buildProgress':
+        console.log('buildProgress', status.value.progress);
+        break;
+      case 'complete':
+        transaction = status.value.transaction;
+        break;
+      default:
+        throw new Error(`unknown authorizeAndBuild status: ${status.case}`);
```
jessepinho, line 25: Nice thoroughness 👏🏻

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
+    }
+  }
+  if (!transaction) throw new Error('did not build transaction');
 
-  return uint8ArrayToHex(id.inner);
+  const broadcastStream = viewClient.broadcastTransaction({ transaction, awaitDetection });
+
+  const expectId = new TransactionId({ inner: await sha256Hash(transaction.toBinary()) });
+
+  let detectionHeight;
```
jessepinho, line 34:
```suggestion
  let detectionHeight: bigint | undefined;
```

apps/webapp/src/state/helpers.ts
```diff
@@ -1,18 +1,54 @@
...
 import { uint8ArrayToHex } from '@penumbra-zone/types';
+import { TransactionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/txhash/v1alpha1/txhash_pb';
+import { sha256Hash } from '@penumbra-zone/crypto-web';
 
-export const planWitnessBuildBroadcast = async (req: TransactionPlannerRequest) => {
-  const { plan } = await viewClient.transactionPlanner(req);
-  if (!plan) throw new Error('no plan in response');
+export const planWitnessBuildBroadcast = async (
+  req: TransactionPlannerRequest,
+  awaitDetection = true,
```
jessepinho, line 9:
```suggestion
```
We're not actually passing this argument in anywhere — OK if we just hard-code it as `true` in the call to `viewClient.broadcastTransaction` on line 30 below?

packages/router/src/grpc/view-protocol-server/broadcast-transaction.ts
```diff
@@ -15,15 +19,38 @@ export const broadcastTransaction: Impl['broadcastTransaction'] = async (req, ct
   // start subscription early to prevent race condition
   const subscription = indexedDb.subscribe('TRANSACTION_INFO');
 
-  const id = await tendermint.broadcastTx(req.transaction);
+  const id = new TransactionId({ inner: await sha256Hash(req.transaction.toBinary()) });
+
+  const broadcastId = await tendermint.broadcastTx(req.transaction);
+  if (!id.equals(broadcastId)) {
+    console.error('broadcast transaction id disagrees', id, broadcastId);
```
jessepinho, line 26:
```suggestion
```

packages/router/src/grpc/view-protocol-server/broadcast-transaction.ts
```diff
@@ -15,15 +19,38 @@ export const broadcastTransaction: Impl['broadcastTransaction'] = async (req, ct
   // start subscription early to prevent race condition
   const subscription = indexedDb.subscribe('TRANSACTION_INFO');
 
-  const id = await tendermint.broadcastTx(req.transaction);
+  const id = new TransactionId({ inner: await sha256Hash(req.transaction.toBinary()) });
```
jessepinho, line 22: I like how you used the name `expectId` in `webapp/src/state/helpers`, and that could be nice here to make the purpose of this `id` clear. But up to you.

jessepinho (approved):
LGTM! I made a number of small suggestions throughout, but nothing that I think is blocking. Please just review my comment on your `pnpm all-check` change, and remove the `console.log`s.

Also, it'd be great to have robust tests for the things being changed here — especially e.g., `planWitnessBuildBroadcast` and `broadcastTransaction`. Totally get if you want to just merge this first, but I'd be happy to pair on (or just do it myself) writing some tests for those two functions in a follow-up PR.

apps/webapp/src/components/shared/asset-icon.tsx
turbocrime: it's updated now to fit the spec that just added this feature

apps/webapp/src/state/helpers.ts
turbocrime: this is a helper method now and not local to any specific slice, let's allow it to remain versatile

apps/webapp/src/state/helpers.ts
turbocrime: linter and type system are already happy with it

apps/webapp/src/state/helpers.ts
turbocrime: linter and type system are already happy with it

packages/router/src/grpc/view-protocol-server/broadcast-transaction.ts
turbocrime: no. this is important because if for some reason conversion to string fails, it's still logged

apps/webapp/src/state/helpers.ts
jessepinho: I meant more for human reasons than linting reasons: a type annotation here makes it immediately clear to readers what this variables can be.

apps/webapp/src/state/helpers.ts
grod220: This feels like a linter issue 🤔 . `let transaction;` would qualify as an `any` no? We should investigate why this was allowed through.

grod220: Ah, this actually _is_ allowed with our current config. Going to open a new PR adding a lint rule to catch these.

PR #430, opened 2024-01-30
Fix client import

jessepinho:
This fixes an import issue.

Relatedly: I've messaged @grod220 about requiring PRs to be built against latest `main` to avoid broke import issues like this in the future.

### DIFF
--- a/packages/types/src/translators/address-view.test.ts
+++ b/packages/types/src/translators/address-view.test.ts
@@ -36,7 +36,7 @@ describe('asOpaqueAddressView()', () => {
         },
       });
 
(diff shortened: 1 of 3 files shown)


packages/types/src/translators/address-view.test.ts
```diff
@@ -36,7 +36,7 @@ describe('asOpaqueAddressView()', () => {
         },
       });
 
-      expect(asOpaqueAddressView(addressView)!.equals(expected)).toBe(true);
+      expect(asOpaqueAddressView(addressView).equals(expected)).toBe(true);
```
jessepinho, line 39: Caught by the linter as unnecessary, and was breaking the build

turbocrime (approved):

PR #431, opened 2024-01-30
Provide default FmdParams in the transaction planner when not present in the request

jessepinho:
Temporary workaround for missing FmdParameters from testnet-preview. Should be reverted after core is fixed, although it'd be nice to keep the test setup.


turbocrime:
fmd params are now incoming on the block again

Issue #425, opened 2024-01-29
Don't strictly validate recovery phrase


turbocrime:
currently, we enforce recovery phrase validation

https://github.com/penumbra-zone/web/blob/1f7390c77b07bfab7bd4f286d357d3e54b68c248/apps/extension/src/routes/page/onboarding/import.tsx#L34-L45

this is a misuse of the bip39 library, which prominently recommends that phrase validity should be displayed, but should not be enforced

https://github.com/bitcoinjs/bip39/blob/master/README.md#reminder-for-developers

Issue #319, opened 2023-12-22
bundle offscreen js into offscreen.html


turbocrime:
in anticipation of offscreen pr merging

offscreen.html fetches javascript internally, but it could bundle the initial js containing event handlers. webpack should be configured to do this.

https://github.com/penumbra-zone/web/blob/0743e85482cb3f298e0ec5b3b6d7696ca1175059/apps/extension/webpack/webpack.common.js#L14

https://github.com/penumbra-zone/web/blob/bdf5fd486d675a7c0f08b78871c6fa12589f7e96/apps/extension/public/offscreen.html#L10

i believe this should improve reliability of `await chrome.offscreen.createDocument`, because the js becomes part of the initial document load, instead of a subsequent fetch.

turbocrime:
can't inline js due to chrome disallows script-src CSP hashes in the manifest

PR #434, opened 2024-01-31
Bundler config

turbocrime:
- uses CopyWebpackPlugin for keys instead of downloading straight to dist
- uses HtmlWebpackPlugin for html output instead of three identical templates with hard-coded paths
- more deliberate chunking
- smaller bundled output for both webapp and extension
- some deps moved to root workspace

eventually, for the extension, something like https://github.com/crxjs/chrome-extension-tools/ which uses rollup would be nice, so it's more similar to our webapp config.

might have some more commits for better extension chunking, but don't want to go too far with this.


turbocrime:
would like to bundle js #319

apps/extension/src/state/seed-phrase/generate.ts
```diff
@@ -4,8 +4,17 @@ import {
...
+
+export const generateValidationFields = (
+  seedPhrase: string[],
+  amount: number,
+): ValidationField[] => {
+  const allWords: ValidationField[] = seedPhrase.map((word, index) => ({ word, index }));
+  const shuffleWords = allWords.sort(() => 0.5 - Math.random());
+  const pickWords = shuffleWords.slice(0, amount);
+  return pickWords.sort((a, b) => a.index - b.index);
+};
```
grod220, line 17: Why do we need to refactor the old code? Also, we seem to have lost the tests for this. Being co-located with the mnemonic stuff in the crypto package feels like a better spot.

apps/extension/webpack/webpack.common.js
```diff
@@ -70,13 +74,40 @@ export default {
     },
   },
   plugins: [
+    new webpack.CleanPlugin(),
```
grod220, line 77: What does this do? can you add a code comment?

apps/extension/webpack/webpack.common.js
```diff
@@ -70,13 +74,40 @@ export default {
...
     new webpack.ProvidePlugin({
       // Required by the `bip39` library
       Buffer: ['buffer', 'Buffer'],
     }),
+    new webpack.IgnorePlugin({
+      // Not required by the `bip39` library, but very nice
+      checkResource(resource) {
+        return /.*\/wordlists\/(?!english).*\.json/.test(resource);
+      },
+    }),
```
grod220, line 87: Does this ignore any bip39 wordlist that isn't english? Won't this make our validator non-comprehensive and prevent users from adding valid phrases in other languages?

apps/extension/webpack/webpack.common.js
```diff
@@ -70,13 +74,40 @@ export default {
...
     new CopyPlugin({
-      patterns: [{ from: '.', to: '../', context: 'public' }],
-      options: {},
+      patterns: ['public', { from: 'bin', to: 'bin', transform: { cache: true } }],
```
grod220, line 89: What's the thoughts on downloading to `/bin` + copying to `/dist` versus just downloading to `/dist`? Does it help us?

apps/webapp/vite.config.ts
```diff
@@ -1,8 +1,19 @@
...
   base: './',
-  plugins: [react()],
+  plugins: [react(), splitVendorChunkPlugin()],
+  build: {
+    rollupOptions: {
+      output: {
+        manualChunks: (id: string) => {
+          if (id.includes('@buf')) return 'protobuf';
+          else if (id.includes('node_modules')) return 'vendor';
+          return;
```
grod220, line 14: What does this do? Can you add a code comment?

apps/extension/react-root.html
```diff
@@ -0,0 +1,9 @@
+<!doctype html>
+<html lang="en" translate="no">
+  <head>
+    <meta charset="UTF-8" />
+  </head>
+  <body>
+    <div id="<%= htmlWebpackPlugin.options.rootId || 'root' %>"></div>
```
grod220, line 7: Ah, yes this is better

apps/extension/src/state/seed-phrase/generate.ts
turbocrime: it was the only use of the `lodash` package in the entire monorepo, for a single random subset function. the entire lodash package was being bundled into every script on both the frontend and backend, so it was a good target to reduce js bundle size.

'validationFields' is poorly named, but its purpose is to confirm that the user has backed up their mnemonic phrase during onboarding, by satisfying a form validation prop.

it makes sense to me that we should put form validation props with the forms that they are used for. going for minimal changes, i just dropped the refactored function directly at the location of its former import. but, it should probably either stay here in the slice and be explained, or move to confirm-backup.tsx

https://github.com/penumbra-zone/web/blob/24be9588a0cf4ee5b1a29447acfed80bf9507bab/apps/extension/src/routes/page/onboarding/confirm-backup.tsx#L18-L49

the corresponding tests in the crypto package were not particularly useful, and the slice's comprehensive behavior including the functionality of this code is covered by existing tests in generate.test.ts immediately adjacent.

apps/extension/webpack/webpack.common.js
turbocrime: previously, webpack left all existing content in dist while building. mostly this worked as new assets clobbered old assets, but this kind of thing commonly leads to issues with old content from old builds persisting in the new builds.

apps/extension/webpack/webpack.common.js
turbocrime: yes, this ignores any bip39 wordlist that isn't in english.

previously, every bip39 wordlist in every language was bundled into every script in both webapp and extension. they are not insignificant so it was a good target to reduce js bundle size.

we don't currently generate any phrases that aren't english, so we won't ever give an extension user a phrase we can't validate.

we shouldn't enforce validation on phrases that the user inputs, as instructed by the library, and as issued in #425

the bip39 library we are using specifically and prominently recommends this approach and this webpack configuration, at the top of their readme, which is linked in that issue

if at some point we support additional languages, we should chunk languages separately, and not load bip39 unless we are performing phrase validation

apps/extension/webpack/webpack.common.js
turbocrime: makes it easy to delete dist without redownloading all keys. in general it would be nice to consider dist to be territory of the bundler, and avoid manual steps

apps/webapp/vite.config.ts
turbocrime: this is the chunking config for vite. protobuf is a nice conceptual unit, so i put it in its own chunk. vendor chunk is just anything from a package.

i guess line 13 is unnecessary, it is achieved by the vendor chunk plugin above.

https://vitejs.dev/guide/build.html#chunking-strategy

apps/extension/src/state/seed-phrase/generate.ts
grod220: > the entire lodash package was being bundled

I'm not exactly married to lodash, but this doesn't sound right to me. Webpack does tree shaking. We should only be importing the specific code for that function. Everything else should not be being pulled in.

> into every script on both the frontend and backend

By frontend, do you mean web app and backend, is the extension code? The web app does not use this code, only the extension. And given the whole bundle is pre-loaded on disk when you download it up front, it has no noticeable affect on the user's performance.

> it makes sense to me that we should put form validation props with the forms that they are used for

In theory, this would also apply to the whole of the crypto package as the password slice is the only one using most of that. The crypto package feels like a good logic unit and domain specific grouping. And if so, mnemonic validation feels properly placed in `mnemonic.ts`

apps/webapp/vite.config.ts
grod220: I think this is fine for now. But likely the best chunking possible is by page. This is something meta-frameworks do automatically for us and another reason to explore them.

apps/extension/webpack/webpack.common.js
turbocrime: it looks like we eventually use the seed phrase in wasm, which uses SeedPhrase::from_str, which uses SeedPhrase::validate_checksum, which checks for presence in a list which is exactly equivalent to our english list from the bip39 package.

https://github.com/penumbra-zone/penumbra/blob/66b13e98643b8af8285238c62657ded81e8824f1/crates/core/keys/src/keys/seed_phrase/words.rs 

i think enforcing validation on a bip39 phrase input is wrong, but if we are going to validate, we had better only validate english words.

the current state of this pr satisfies that, so i'm keeping it as-is

apps/extension/src/state/seed-phrase/generate.ts
turbocrime: >I'm not exactly married to lodash, but this doesn't sound right to me. Webpack does tree shaking. We should only be importing the specific code for that function. Everything else should not be being pulled in. [...] By frontend, do you mean web app and backend, is the extension code? The web app does not use this code, only the extension.

haha yeah :( our current bundle config is bad. getting rid of lodash in crypto changes the bundle size of the frontend

>And given the whole bundle is pre-loaded on disk when you download it up front, it has no noticeable affect on the user's performance.

when it's in the webapp, every node will serve it.

in the extension, it was getting bundled into every script using the crypto package, which used to be most things. improving extension bundling is important because we do things like launch popups and workers that should be as fast as possible.

> In theory, this would also apply to the whole of the crypto package as the password slice is the only one using most of that.

i like this line of thought, but i'm not sure that's entirely true. let's not get ahead of ourselves

apps/extension/webpack/webpack.common.js
grod220: Oh! If our systems currently only support english, yep, we should just validate for that then. I suppose it's a bigger task for later to support more languages then.

apps/extension/src/state/seed-phrase/generate.ts
grod220: > getting rid of lodash in crypto changes the bundle size of the frontend

Here's the reason: https://github.com/penumbra-zone/web/blob/b3a1b463bb5aec45559f4189079ac646e7c88143/apps/webapp/src/state/helpers.ts#L5

Looks like the web app is making use of one of the functions in that package. But... there's still a tree shaking issue. Nothing in that function requires lodash. I suspect it has to do with our "barrel exports" https://github.com/penumbra-zone/web/blob/main/packages/crypto/index.ts. I'm fine with getting rid of lodash, but we should investigate this when we have more time.

> improving extension bundling is important because we do things like launch popups and workers that should be as fast as possible.

For the extension, this is a function of how fast javascript can run on the machine, not network loading latency (which is usually the performance consideration of increasing the size of a bundle). The difference in the extension case is not noticeable. In fact, the most noticeable difference would be in the algorithm difference (if for some reason one was faster than the other). In any case, we are discussing the difference in single digit milliseconds.

grod220 (commented):
If you re-locate your updated mnemonic validator back to the crypto package + re-enable the tests, I'd be happy to approve

grod220 (approved):

PR #439, opened 2024-01-31
AddressByIndex tests

Valentine1898:


packages/router/src/grpc/view-protocol-server/address-by-index.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      protocolName: 'mock',
+      requestMethod: 'MOCK',
+      contextValues: createContextValues().set(servicesCtx, mockServices),
+    });
+  });
+
+  test('should successfully get AddressByIndex with default index', async () => {
+    const addressByIndexResponse = new AddressByIndexResponse(
+      await addressByIndex(new AddressByIndexRequest(), mockCtx),
+    );
```
grod220, line 42: The response object seems to be wrapped by another response object? You can just do:

```typescript
    const response = await addressByIndex(new AddressByIndexRequest(), mockCtx);
    expect(response.address).toBeInstanceOf(Address);
```

Valentine1898: In this case, yes. In other cases wrap is necessary because `addressByIndex()` returns `AddressByIndexResponse | PartialMessage<AddressByIndexResponse>.`

grod220 (approved):

Issue #427, opened 2024-01-30
Refactor balances.ts to make use of `DenomMetadata` and `ValueView`


grod220:
- [ ] The extension and frontend code should be making use of the `DenomMetadata` at all times. For instance, the assets list makes use of an `AssetBalance`, which has an ad-hoc subset of different parts of the `DenomMetadata`, plus an amount. But we already have a way to represent a typed amount (`Value`) and a way to attach metadata to that typed amount (`ValueView`).
- [ ] `Value`s should be rendered using standardized components that take a `ValueView`, rather than having custom rendering in different places (similarly to how we have a standardized `AddressView` component)
- [ ] Values should use the ticker symbols defined in the `DenomMetadata`.
- [ ] Do a search for `fromBaseUnitAmount`. Should find a view examples of where ValueView should replace.
- [ ] https://github.com/penumbra-zone/web/issues/429

Will affect things downstream (swap/send/ibc). Ensure those are not formatting their own values but using `ValueView`.

PR #440, opened 2024-01-31
Refactor selections/balances to use `DenomMetadata`, and to render a ValueViewComponent when possible

jessepinho:
This started as a change to how we render a user's balance on the Send form. @grod220 [pointed out](https://github.com/penumbra-zone/web/pull/362#discussion_r1472411058) that we should be using a `ValueViewComponent`, as that is the standard for how we render `ValueView`s throughout the app.

As I worked on this, though, I realized this change wasn't possible without also changing the form of data that we pass around in `apps/webapp/src/fetchers/balances.ts`. So I refactored the helpers in that file so that we could work with full `DenomMetadata` objects, rather than just a few properties pulled off of those objects. 

When I did this, I also realized there were a few places where we were making incorrect assumptions and calculations about how to render financial values! I'll comment on those in the code so you can check my work, but I believe I corrected them.

## In this PR
- Modify the `AssetBalance` type to convert the `denom` property to a full `denomMetadata` property containing all metadata about the asset.
- Create a couple new helpers for converting between amounts and display values (based on the denom metadata).
- Fix a couple bugs in `ValueViewComponent` where the default exponent was assumed to be `1`, when it should be `0`. (Reviewers, please double-check this logic — I'll leave a comment where I'm referring to).
- Write tests for `ValueViewComponent` to ensure we're rendering financial figures correctly.

### DIFF
--- a/apps/webapp/src/fetchers/balances.ts
+++ b/apps/webapp/src/fetchers/balances.ts
@@ -5,7 +5,7 @@ import {
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 import {
   AssetId,
-  DenomUnit,
+  DenomMetadata,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
 import { getAddresses, IndexAddrRecord } from './address';
 import { getAllAssets } from './assets';
@@ -16,10 +16,7 @@ import { viewClient } from '../clients/grpc';
 import { streamToPromise } from './stream';
 
 export interface AssetBalance {
-  denom: {
-    display: DenomUnit['denom'];
-    exponent: DenomUnit['exponent'];
-  };
+  denomMetadata: DenomMetadata;
   assetId: AssetId;
   amount: Amount;
   usdcValue: number;
@@ -35,29 +32,19 @@ type NormalizedBalance = AssetBalance & {
   account: { index: number; address: string };
 };
 
-// Given an asset has many denom units, the amount should be formatted using
-// the exponent of the display denom (e.g. 1,954,000,000 upenumbra = 1,954 penumbra)
-export const displayDenom = (res?: AssetsResponse): { display: string; exponent: number } => {
-  const display = res?.denomMetadata?.display;
-  if (!display) return { display: 'unknown', exponent: 0 };
-
-  const match = res.denomMetadata?.denomUnits.find(d => d.denom === display);
-  if (!match) return { display, exponent: 0 };
-
-  return { display, exponent: match.exponent };
-};
-
-const getDenomAmount = (res: BalancesResponse, metadata: AssetsResponse[]) => {
+const getDenomAmount = (
+  res: BalancesResponse,
+  metadata: AssetsResponse[],
+): { amount: Amount; denomMetadata: DenomMetadata } => {
   const assetId = uint8ArrayToBase64(res.balance!.assetId!.inner);
   const match = metadata.find(m => {
     if (!m.denomMetadata?.penumbraAssetId?.inner) return false;
     return assetId === uint8ArrayToBase64(m.denomMetadata.penumbraAssetId.inner);
   });
 
-  const { display, exponent } = displayDenom(match);
   const amount = res.balance?.amount ?? new Amount();
 
-  return { display, exponent, amount };
+  return { amount, denomMetadata: match?.denomMetadata ?? new DenomMetadata() };
 };
 
 const normalize =
@@ -66,10 +53,10 @@ const normalize =
     const index = res.account?.account ?? 0;
     const address = indexAddrRecord?.[index] ?? '';
 
-    const { display, exponent, amount } = getDenomAmount(res, metadata);
+    const { denomMetadata, amount } = getDenomAmount(res, metadata);
 
     return {
-      denom: { display, exponent },
+      denomMetadata,
       assetId: res.balance!.assetId!,
       amount,
       //usdcValue: amount * 0.93245, // TODO: Temporary until pricing implemented
@@ -81,7 +68,7 @@ const normalize =
 const groupByAccount = (balances: AccountBalance[], curr: NormalizedBalance): AccountBalance[] => {
   const match = balances.find(b => b.index === curr.account.index);
   const newBalance = {
-    denom: curr.denom,
+    denomMetadata: curr.denomMetadata,
     amount: curr.amount,
     usdcValue: curr.usdcValue,
     assetId: curr.assetId,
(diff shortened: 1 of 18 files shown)


apps/webapp/src/fetchers/balances.ts
```diff
@@ -35,29 +32,19 @@ type NormalizedBalance = AssetBalance & {
...
-// the exponent of the display denom (e.g. 1,954,000,000 upenumbra = 1,954 penumbra)
-export const displayDenom = (res?: AssetsResponse): { display: string; exponent: number } => {
-  const display = res?.denomMetadata?.display;
-  if (!display) return { display: 'unknown', exponent: 0 };
-
-  const match = res.denomMetadata?.denomUnits.find(d => d.denom === display);
-  if (!match) return { display, exponent: 0 };
-
-  return { display, exponent: match.exponent };
-};
```
jessepinho, line 48: Eliminated this helper function entirely, as it just takes an `AssetsResponse` (a thin wrapper around `DenomMetadata`) and returns just the display denom's exponent.

Going forward, we want to use the `ValueViewComponent` to display balances, which requires an entire `DenomMetadata` object. So we don't want to just strip a few properties out of that object here.

packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -17,7 +18,7 @@ export const ValueViewComponent = ({ view }: ValueViewPrpos) => {
     const encodedAssetId = bech32AssetId(value.assetId!);
     return (
       <div className='flex font-mono'>
-        <p className='text-[15px] leading-[22px]'>{fromBaseUnitAmount(amount, 1).toFormat()}</p>
+        <p className='text-[15px] leading-[22px]'>{fromBaseUnitAmount(amount, 0).toFormat()}</p>
```
jessepinho, line 21: I'm like... 90% sure that using `1` for the default exponent for an unknown denomination is a mistake. An exponent of `1` means we're multiplying the `amount` by "10 to the power of 1" — i.e., we're multiplying it by 10. 

We _should_ (right?!) be multiplying it by `1`  (AKA, "10 to the power of 0") when we don't know the denomination.

Reviewers, please check my logic!

packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -36,8 +37,8 @@ export const ValueViewComponent = ({ view }: ValueViewPrpos) => {
     const value = view.valueView.value;
     const amount = value.amount ?? new Amount();
     const display_denom = value.denom?.display ?? '';
-    // The first denom unit in the list is the display denom, according to cosmos practice
-    const exponent = value.denom?.denomUnits[0]?.exponent ?? 1;
+    const exponent = value.denom ? getDisplayDenomExponent(value.denom) : 0;
```
jessepinho, line 40: Same issue here, changing the exponent from 1 to 0.

Also, the comment which I deleted says:
> The first denom unit in the list is the display denom, according to cosmos practice

However, I'm not sure this is correct. Cosmos's [own ADR docs](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-024-coin-metadata.md) contain an example asset where the first denom is _not_ the display denom. So, to be safe, I'm calling `getDisplayDenomExponent` to make sure we're getting the right exponent.

packages/types/src/amount.ts
```diff
@@ -1,15 +1,24 @@
...
 import BigNumber from 'bignumber.js';
+import { DenomMetadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1alpha1/asset_pb';
+import { getDisplayDenomExponent } from './denom-metadata';
 
 export const joinLoHiAmount = (amount: Amount): bigint => {
   return joinLoHi(amount.lo, amount.hi);
 };
 
-export const fromBaseUnitAmount = (amount: Amount, exponent: number): BigNumber => {
+export const fromBaseUnitAmount = (amount: Amount, exponent = 0): BigNumber => {
```
jessepinho, line 11: Since we're occasionally passing in an empty `DenomMetadata` upstream of this call, I'll leave the default exponent as `0`, so that we're just multiplying by `1`.

packages/types/src/lo-hi.ts
```diff
@@ -76,7 +76,7 @@ export const fromBaseUnit = (lo = 0n, hi = 0n, exponent: number): BigNumber => {
  * @param {number} exponent - The exponent to be applied.
  * @returns {LoHi} An object with properties `lo` and `hi`, representing the low and high 64 bits of the multiplied value.
  */
-export const toBaseUnit = (value: BigNumber, exponent: number): LoHi => {
+export const toBaseUnit = (value: BigNumber, exponent = 0): LoHi => {
```
jessepinho, line 79: Since we're occasionally passing in an empty `DenomMetadata` upstream of this call, I'll leave the default exponent as `0`, so that we're just multiplying by `1`.

packages/ui/components/ui/tx/view/value.tsx
turbocrime: i believe this is correct, but i would confirm by viewing a value through this and then comparing it with the same value viewed with pcli

packages/ui/components/ui/tx/view/value.tsx
turbocrime: i agree it's preferable to certainly identify it rather than assume, especially since it doesn't seem to be an expensive search

turbocrime (approved):

packages/ui/components/ui/tx/view/value.tsx
grod220: Ah good catch!

grod220 (approved):
Nice job! Left some comments about taking this further in https://github.com/penumbra-zone/web/issues/427. I'd love to integrate this work and build off of it. I'm going to do something I never do approve+merge 😱 (maybe it _should_ be the standard? haha) and rebase this against my other branch and continue it from there.

packages/types/src/amount.ts
jessepinho: @grod220 hmm yeah, possibly. I think the case where we'd have missing denomMetadata is when we have a transaction but we don't have denom metadata for its asset type — which, if I understand correctly how Penumbra works, should never happen, right? In which case you're right that we should throw (somewhere in `fetchers/balances.ts`).

packages/types/src/denom-metadata.ts
```diff
@@ -0,0 +1,17 @@
...
+ * each denomination in relation to the base unit. For example, upenumbra is
+ * penumbra's base unit -- the unit which can not be further divided into
+ * decimals. 1 penumbra is equal to 1,000,000 (AKA, 10 to the 6th) upenumbra, so
+ * penumbra's display exponent -- the exponent used to multiply the base unit
+ * when displaying a penumbra value to a user -- is 6. (For a non-crypto
+ * example, think of US dollars. The dollar is the display unit; the cent is the
+ * base unit; the display exponent is 2 (10 to the 2nd).)
+ */
+export const getDisplayDenomExponent = (denomMetadata: DenomMetadata): number | undefined =>
+  denomMetadata.denomUnits.find(denomUnit => denomUnit.denom === denomMetadata.display)?.exponent;
```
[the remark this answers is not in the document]
jessepinho: Good call. If it's OK, I'll leave that up to you in your PR.

PR #441, opened 2024-01-31
Support fee tiers on web

### LINKED ISSUE #309: Support tiered fees in web
After #308 and https://github.com/penumbra-zone/penumbra/issues/3369 are complete, the extension should be modified to support tiered fees.

## Steps:

- [ ] Add a "low/medium/high" selector to the web extension transaction plan UI
- [ ] Provide the selected fee tier to the planner

jessepinho:
🚨 6248 line insertions and 1576 line deletions came from `pnpm-lock.yaml` (due to Storybook installation)

image

## In this PR
- Using the new support for `feeMode` in `TransactionPlannerRequest` (as of https://github.com/penumbra-zone/penumbra/pull/3722), set the fee tier based on the user's preference and pass it to the planner.
- Install and set up [Storybook](https://storybook.js.org/) for viewing and testing components. (More detail below.)
- Create a `SegmentedPicker` component:
    image
- Create a `GasFee` component that uses `SegmentedPicker` to let the user pick a gas fee tier: 
image
- Simplify `InputBlock` to basically be a standardized field wrapper, and use it with the gas fee field.
  - Also, refactor `InputToken` to use `InputBlock` rather than duplicating a lot of its styling + markup.

## Storybook
[Storybook](https://storybook.js.org/) gives us a nice out-of-the-box display case for our UI components, allowing us to view our library of components, interact with them, and change their props via a control panel — all in isolation from the app itself. 

I'd love to see us add more components to Storybook over time, especially as we get updated component designs in the near future.

image

Closes #309


apps/webapp/src/components/shared/input-block.tsx
```diff
@@ -1,25 +1,17 @@
...
 
-interface InputBlockProps extends InputProps {
+interface InputBlockProps {
   label: string;
-  className?: string;
-  inputClassName?: string;
-  validations?: Validation[];
-  value: string;
+  className?: string | undefined;
+  validations?: Validation[] | undefined;
```
jessepinho, line 8: FWIW, needing to add `| undefined` to a prop that's already optional is another example of the downside of enforcing `exactOptionalPropertyTypes` in our tsconfigs. Obviously not a big problem, but just mentioning it since it's been a topic.

apps/webapp/src/components/shared/input-block.tsx
```diff
@@ -30,17 +22,12 @@ export const InputBlock = ({
...
         <p className='text-base font-bold'>{label}</p>
         {vResult ? <div className={cn('italic', 'text-red-400')}>{vResult.issue}</div> : null}
       </div>
-      <Input
-        variant='transparent'
-        className={inputClassName}
-        placeholder={placeholder}
-        value={value}
-        {...props}
-      />
```
jessepinho, line 43: This now has to be passed in as `children`

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -57,31 +48,12 @@ export default function InputToken({
...
-      )}
-    >
-      <div className='mb-2 flex items-center justify-between gap-1 md:gap-2'>
-        <p className='text-sm font-bold md:text-base'>{label}</p>
-        {vResult ? (
-          <div className={cn('italic text-[12px] md:text-[15px]', 'text-red-400')}>
-            {vResult.issue}
-          </div>
-        ) : null}
-      </div>
```
jessepinho, line 84: All of this was duplicating markup from `InputBlock`, albeit with slightly different font sizes etc. So I just refactored it to use `InputBlock` directly.

apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -98,19 +70,11 @@ export default function InputToken({
...
 
       <div className='mt-[6px] flex items-center justify-between gap-2'>
-        <div className='flex items-start gap-2'>
-          {feeAsString && (
-            <>
-              <img src='/fuel.svg' alt='Gas fee' className='size-5' />
-              <p className='font-bold text-muted-foreground'>{feeAsString}</p>
-            </>
-          )}
-        </div>
```
jessepinho, line 108: We're no longer rendering the fee inside `InputToken`, but rather as its own first-class field.

apps/webapp/src/components/send/send-form/use-refresh-fee.ts
```diff
@@ -24,5 +24,5 @@ export const useRefreshFee = () => {
     }, DEBOUNCE_MS);
   }, [refreshFee]);
 
-  useEffect(debouncedRefreshFee, [amount, recipient, selection, debouncedRefreshFee]);
+  useEffect(debouncedRefreshFee, [amount, feeTier, recipient, selection, debouncedRefreshFee]);
```
jessepinho, line 27: `refreshFee()` is now called every time `feeTier` is changed, too.

apps/webapp/src/components/send/ibc/ibc-form.tsx
```diff
@@ -72,21 +72,21 @@ export default function IbcForm() {
...
         validations={[penumbraAddrValidation()]}
-      />
+      >
+        <Input
+          variant='transparent'
+          placeholder='Enter the address'
+          value={destinationChainAddress}
+          onChange={e => setDestinationChainAddress(e.target.value)}
+        />
+      </InputBlock>
```
jessepinho, line 89: `InputBlock` no longer renders its own `Input`, so we have to render it ourselves as a child.

package.json
```diff
@@ -16,9 +16,9 @@
   "dependencies": {
     "@buf/cosmos_ibc.bufbuild_es": "1.7.2-20240123184419-30c7be3837b0.1",
     "@buf/cosmos_ibc.connectrpc_es": "1.3.0-20240123184419-30c7be3837b0.1",
-    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.7.2-20240201164939-04d44cf9ba58.1",
-    "@buf/penumbra-zone_penumbra.connectrpc_es": "1.3.0-20240201164939-04d44cf9ba58.1",
-    "@bufbuild/protobuf": "^1.7.1",
+    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.7.2-20240202021444-2081ef4dca12.1",
+    "@buf/penumbra-zone_penumbra.connectrpc_es": "1.3.0-20240202021444-2081ef4dca12.1",
```
jessepinho, line 20: So that we get the latest version of `TransactionPlannerRequest`, which includes the `feeMode` property.

packages/ui/.storybook/preview.js
```diff
@@ -0,0 +1,16 @@
+import '../styles/globals.css';
```
jessepinho, line 1: Storybook boilerplate file

packages/ui/.storybook/main.js
```diff
@@ -0,0 +1,29 @@
+import { join, dirname } from 'path';
```
jessepinho, line 1: Storybook boilerplate file

packages/ui/components/ui/segmented-picker.tsx
```diff
@@ -0,0 +1,62 @@
...
+  value: ValueType;
+  onChange: (value: ValueType) => void;
+  options: SegmentedPickerOption<ValueType>[];
+}) => {
+  return (
+    <div className='flex flex-row gap-0.5'>
+      {options.map((option, index) => (
+        <div
+          key={option.value.toString()}
+          role='button'
```
turbocrime, line 46:
```suggestion
    <div role='radiogroup' className='flex flex-row gap-0.5'>
      {options.map((option, index) => (
        <div
          key={option.value.toString()}
          role='radio'
          aria-checked={value === option.value}
```

turbocrime (approved):
nice

packages/ui/components/ui/segmented-picker.tsx
jessepinho: good catch! fixing...

apps/webapp/src/components/shared/input-block.tsx
grod220: It sounded like the majority of folks are interested in disabling `exactOptionalPropertyTypes`. Think we should feel free to and revisit it if it ends up not catching a bug in the future.

apps/webapp/src/state/send.ts
```diff
@@ -127,6 +138,17 @@ const assembleRequest = async ({ amount, recipient, selection, memo }: SendSlice
       },
     ],
     source: new AddressIndex({ account: selection.accountIndex }),
+
+    // Note: we currently don't provide a UI for setting the fee manually. Thus,
+    // a `feeMode` of `manualFee` is not supported here.
```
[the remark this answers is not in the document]
jessepinho: I just don't know if it's prioritized or important at the moment, but happy to take that on next!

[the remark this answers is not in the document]
grod220: Could you create a github issue for this feature? 🙏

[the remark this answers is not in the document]
jessepinho: @grod220 https://github.com/penumbra-zone/web/issues/461

Issue #384, opened 2024-01-23
New streaming tx build rpcs


grod220:
`witnessAndBuild` and `broadcastTransaction` now are streaming rpcs (not unary anymore). Update to new bufbuild types to see breaking changes. Our UI should support the API with new states:

- Build progress bar
- Broadcast success
- Tx confirmed

grod220:
@TalDerei realizing this may conflict with the work you are in progress with. This new RPC change likely requires us to rethink a lot with our build implementation. `build_parallel` and `offscreenClient` do not give in-progress updates. This means we need to change them into streaming methods as well. Another round of internal messaging re-work...

TalDerei:
Can I fold these changes and new broadcast toast (#327) into #369?

grod220:
Given the new RPC definition, I think you will have to. However, it does increase the scope of the task quite a bit fyi.

hdevalence:
Wait, what's the requirement for the change?

grod220:
Frontend requirements:

Before 
- unary request/response
- Toast on build-in-progress, toast on broadcast, toast on confirmation

After
- streaming request/response
- Probably should not be using toasts anymore and have a table for:
  - build progress bar
  - broadcasting loading
  - tx confirmed


Extension requirements:

Before:
- Unary request flow:
 - frontend —> service worker —> offscreen window —> web worker

After:
- Convert rpc handler to async generator
- Offscreen window needs message streaming capabilities
- Spawn and handle web workers not with Promise.All() but with streaming or generators
- Determine rules on communicating build progress. % of actions built?

hdevalence:
I see, thanks for mapping it out. To start I think it's fine to just change the interfaces and not report progress info (so the bar goes from 0 to 100% directly). Then we have the plumbing in place and can change the reporting in a follow up chunk of work.

TalDerei:
@grod220 The new streaming RPCs don't actually conflict with #369. To limit the scope, let's merge that in. I already started implementing some of the new interface changes separately.

turbocrime:
still needs ui

turbocrime:
ui provided by #327

PR #443, opened 2024-01-31
even more v0.65 updates 

turbocrime:
wip. having difficulty with some kind of proto spec mismatch

edit: ready to go

### DIFF
--- a/apps/extension/src/hooks/last-block-synced.ts
+++ b/apps/extension/src/hooks/last-block-synced.ts
@@ -4,7 +4,6 @@ import { TendermintQuerier } from '@penumbra-zone/query';
 import { PopupLoaderData } from '../routes/popup/home';
 import { useStore } from '../state';
 import { networkSelector } from '../state/network';
-import { internalSwClient } from '@penumbra-zone/router';
 
 // There is a slight delay with Zustand loading up the last block synced.
 // To prevent the screen flicker, we use a loader to read it from chrome.storage.local.
@@ -17,15 +16,6 @@ const useLastBlockSynced = (): number => {
 export const useSyncProgress = () => {
   const lastBlockSynced = useLastBlockSynced();
   const { grpcEndpoint } = useStore(networkSelector);
-
-  // Wake up service worker so block syncing can resume
-  useQuery({
-    queryKey: ['sync-blocks', grpcEndpoint],
-    queryFn: async () => {
-      await internalSwClient.syncBlocks();
-      return true; // react-query cannot return undefined
-    },
-  });
 
   const { data } = useQuery({
     queryKey: ['lastBlockHeight'],
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -1,8 +1,6 @@
 import { RootQuerier } from './root-querier';
 
-import { Code, ConnectError } from '@connectrpc/connect';
 import { bech32 } from 'bech32';
-import { backOff } from 'exponential-backoff';
 
 import { sha256Hash } from '@penumbra-zone/crypto-web';
 import {
@@ -47,7 +45,7 @@ export class BlockProcessor implements BlockProcessorInterface {
   private readonly indexedDb: IndexedDbInterface;
   private readonly viewServer: ViewServerInterface;
   private readonly abortController: AbortController = new AbortController();
-  private blockSyncPromise: Promise<void> | undefined;
+  private syncPromise: Promise<void> | undefined;
 
   constructor({ indexedDb, viewServer, querier, fullViewingKey }: QueryClientProps) {
     this.fullViewingKey = fullViewingKey;
@@ -56,22 +54,24 @@ export class BlockProcessor implements BlockProcessorInterface {
     this.querier = querier;
   }
 
-  // If syncBlocks() is called multiple times concurrently,
-  // they'll all wait for the same promise rather than each starting their own sync process.
-  async syncBlocks(): Promise<void> {
-    if (!this.blockSyncPromise) {
-      this.blockSyncPromise = this.autoRetrySync();
-    }
-    return this.blockSyncPromise;
+  // If syncBlocks() is called multiple times concurrently, they'll all wait for
+  // the same promise rather than each starting their own sync process.
+  public async sync(): Promise<void> {
+    this.syncPromise ??= this.syncAndStore().catch(e => {
+      // prefer abort error
+      this.abortController.signal.throwIfAborted();
+      throw e;
+    });
+    return this.syncPromise;
   }
 
-  stopSync() {
-    this.abortController.abort();
+  public stop(r?: string) {
+    this.abortController.abort(`Sync abort ${r}`);
   }
 
   // TODO: should this be here? it's only ever used as a fallback by view protocol server
   // identify failures?
-  async getTransactionInfo(id: TransactionId): Promise<TransactionInfo> {
+  public async getTransactionInfo(id: TransactionId): Promise<TransactionInfo> {
     const { transaction, height } = await this.querier.tendermint.getTransaction(id);
     const { txp: perspective, txv: view } = await transactionInfo(
       this.fullViewingKey,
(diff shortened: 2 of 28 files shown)


turbocrime:
Screenshot 2024-01-31 at 15 15 31

appParameters deserialization is broken

pr is using wasm v0.65.0-alpha.1b which corresponds to https://github.com/penumbra-zone/penumbra/commit/f0ecf2210ad7431435a7dade8b2786993273df02 but using buf registry corresponding to https://github.com/penumbra-zone/penumbra/commit/d99153001adc54179c010104aa31b74e300cf635

for some reason there is no newer buf registry package available - main might be moving too fast

error still present, when using a local build of wasm corresponding to https://github.com/penumbra-zone/penumbra/commit/d99153001adc54179c010104aa31b74e300cf635

packages/router/src/grpc/view-protocol-server/app-parameters.test.ts
```diff
@@ -44,8 +44,8 @@ describe('AppParameters request handler', () => {
 
 // TODO chainParams will be removed, chain id will be a top-level field https://github.com/penumbra-zone/penumbra/pull/3703
 const testData = new AppParameters({
-  chainParams: {
-    chainId: 'penumbra-testnet-titan',
+  chainId: 'penumbra-testnet-titan',
+  sctParams: {
```
erwanor, line 48: Will also need to add `shieldedPoolParams` containing `fmdParameters`

packages/storage/src/indexed-db/index.ts
```diff
@@ -212,6 +214,22 @@ export class IndexedDb implements IndexedDbInterface {
...
+    if (!json) return undefined;
+    return AppParameters.fromJson(json);
+  }
+
+  async saveAppParams(app: AppParameters): Promise<void> {
+    // this always fails, because the chainId is corrupted
+    //if (app.chainId != this.c.name) throw new Error(`New chainId mismatch: stored ${this.c.name}, new ${app.chainId}`);
+    await this.u.update({
+      table: 'APP_PARAMETERS',
+      value: app.toJson() as Jsonified<AppParameters>,
```
erwanor, line 228: I suppose this is where we're noticing the missing shielded pool field, right?

turbocrime: we aren't using the shieldedPool field for anything yet. but anywhere we make an appParameters request to grpc, the deserialized response is corrupt. right now it matters for the chainId which we get from that endpoint.

turbocrime: we do use sctParams from this endpoint, when asking wasm to plan a tx. it seems that 'epochDuration' is deserialized as `1n`, which doesn't seem correct

turbocrime:
Screenshot 2024-01-31 at 15 46 56

pcli main is broken

turbocrime:
deployment is fixed and now this works 😎

packages/query/src/block-processor.ts
```diff
@@ -81,27 +81,6 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      });
-    } catch {
-      // Was signaled to abort, can swallow this error
-    }
-  }
-
-  /**
-   * Populates IndexedDB with data from the blockchain. Most of what's in
-   * IndexedDB comes from this method.
-   */
```
turbocrime, line 104: now that app parameters can be updated, it's possible for the chainId to change mid-sync. our idb name corresponds to chainId, so if a chainId changes we need to reconnect to another database.

thus, it's important to update the sync error handling in order to react appropriately. i moved retry logic to where it is possible to re-init services when necessary, with the proper idb connection.

formerly, this retry logic called `viewServer.resetTreeToStored`, which cannot complete successfully if the database is closed. the `viewServer` is clobbered if services re-init - it seems like unflushed information might be lost this way, but that can't be fixed without refactoring wasm.

packages/router/src/std/routes/sync.ts
turbocrime, line 1: this is not actually used anywhere.

packages/services/src/index.ts
```diff
@@ -21,30 +22,40 @@ export class Services implements ServicesInterface {
...
+
+          // backoff will swallow the error, so log it
+          console.warn('Sync failure', e);
+          // next caller will re-init services
+          this.walletServicesPromise = undefined;
+          console.log('Sync retry', attemptNumber);
+          return true;
+        },
+      },
+    );
```
turbocrime, line 53: this is the relocated retry logic. anything that isn't a deliberate abort will completely re-init services. it is possible that some in-flight request may be holding a reference to the old services, and unfortunately it may still be possible to use them

packages/storage/src/indexed-db/index.ts
```diff
@@ -212,6 +215,25 @@ export class IndexedDb implements IndexedDbInterface {
...
+    if (!json) return undefined;
+    return AppParameters.fromJson(json);
+  }
+
+  async saveAppParams(app: AppParameters): Promise<void> {
+    // if chainId changes, we should use a different database.
+    if (app.chainId !== this.chainId) {
+      this.db.close();
+      throw new Error(`Updated chainId: idb ${this.chainId} != new ${app.chainId}`);
+    }
```
turbocrime, line 229: db is closed to prevent writes from any remaining handles floating around

apps/extension/src/hooks/last-block-synced.ts
turbocrime, line 1: i think this file should probably be using a channel client, and not independently constructing queriers

packages/router/src/std/routes/clear-cache.ts
turbocrime, line 1: interestingly, the handler here was using the wrong type. it worked the whole time.

packages/query/src/block-processor.ts
hdevalence: I am not sure it's possible for the chain ID to change — I know there are some parameters the governance module doesn't allow changing, maybe the chain ID is one of them. I think it's not essential to handle that possibility at the moment.

packages/query/src/queriers/app.ts
```diff
@@ -14,19 +12,20 @@ export class AppQuerier implements AppQuerierInterface {
...
-  }
-
-  async txsByHeight(height: bigint) {
-    const req = new TransactionsByHeightRequest({ blockHeight: height });
-    return await this.client.transactionsByHeight(req);
+  async txsByHeight(blockHeight: bigint): Promise<Transaction[]> {
+    const { blockHeight: responseHeight, transactions } = await this.client.transactionsByHeight({
+      blockHeight,
+    });
+    if (responseHeight !== blockHeight)
```
grod220, line 25: oh dear, is this a real possibility? We get a different block than we requested?

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
```diff
@@ -35,16 +35,15 @@ export const transactionPlanner: Impl['transactionPlanner'] = async (req, ctx) =
     viewServer: { fullViewingKey },
     querier,
   } = await services.getWalletServices();
-  const chainParams = await services.querier.app.chainParams();
   const fmdParams = await indexedDb.getFmdParams();
-  if (!fmdParams) throw new ConnectError('Fmd Params not in indexeddb', Code.FailedPrecondition);
+  if (!fmdParams) throw new ConnectError('FmdParameters not available', Code.FailedPrecondition);
```
grod220, line 39: Are there benefits for throwing a `ConnectError` versus a normal `Error`? If so, should this be the standard in all our rpc methods?

packages/services/src/index.ts
```diff
@@ -21,30 +22,40 @@ export class Services implements ServicesInterface {
...
+    this._querier = new RootQuerier({ grpcEndpoint: this.config.grpcEndpoint });
 
-      await this.tryToSync();
-    } catch (e) {
-      // Logging here as service worker does not appear to bubble the errors up
-      console.error(e);
-      throw e;
-    }
+    // await startup for any errors
+    await this.getWalletServices();
```
grod220, line 29: Can we add a bit more to the comment? I think something like this?

```
// initialize walletServices separately without exponential backoff to bubble up errors immediately
```

packages/services/src/index.ts
grod220: Hopefully that would only be our rpc methods which can be refetched and the services get re-requested

grod220 (approved):

apps/extension/src/hooks/last-block-synced.ts
```diff
@@ -18,15 +17,6 @@ export const useSyncProgress = () => {
...
 
-  // Wake up service worker so block syncing can resume
-  useQuery({
-    queryKey: ['sync-blocks', grpcEndpoint],
-    queryFn: async () => {
-      await internalSwClient.syncBlocks();
-      return true; // react-query cannot return undefined
-    },
-  });
-
```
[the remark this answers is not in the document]
turbocrime: service worker sync is robust, and always running if the worker is running. it is "woken up" by any message emit, if you're on a client page it's awake

packages/query/src/queriers/app.ts
turbocrime: idk but it's certainly something that would ruin all of our downstream work

TalDerei:
Consumes #384

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
turbocrime: - we can specify a meaningful error code
- we can deliberately attach information or context that serializes
- there are some wrappers to catch and rethrow but generally yes

packages/query/src/block-processor.ts
turbocrime: ok, moved logic back into block-processor where it's simpler. leaving the throw on chainId mismatch because that would indicate some kind of failure

PR #445, opened 2024-02-01
400 add tests for broadcasttransactionrequest

Valentine1898:


Valentine1898:
The tests work locally, but here I'm getting 
`expected [Function] to throw error including 'broadcast failed' but got 'crypto is not defined'`

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
```diff
@@ -0,0 +1,483 @@
...
+  });
+
+  test('should throw error if broadcast transaction id disagrees', async () => {
+    mockTendermint.broadcastTx.mockResolvedValue(new TransactionId());
+    await expect(
+      (async () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-empty
+        for await (const _ of broadcastTransaction(broadcastTransactionRequest, mockCtx)); // eslint-disable-line   no-empty
+      })(),
+    ).rejects.toThrow('broadcast transaction id disagrees');
```
Valentine1898, line 125: I haven't found a better way to verify that async generator throw an error

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
```diff
@@ -0,0 +1,483 @@
...
+import { servicesCtx } from '../../ctx';
+import { Transaction } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1alpha1/transaction_pb';
+import { broadcastTransaction } from './broadcast-transaction';
+import type { Services } from '@penumbra-zone/services';
+import { TransactionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/txhash/v1alpha1/txhash_pb';
+
+const mockSha256 = vi.hoisted(() => vi.fn());
+vi.mock('@penumbra-zone/crypto-web', () => ({
+  sha256Hash: mockSha256,
+}));
```
Valentine1898, line 18: I'm not sure if it's necessary, but the CI-CD pipeline was throwing a 'crypto is not defined' error, although locally everything worked without this mock.

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
```diff
@@ -0,0 +1,483 @@
...
+}
+interface TendermintMock {
+  broadcastTx: Mock;
+}
+interface MockServices {
+  getWalletServices: Mock<[], Promise<{ indexedDb: IndexedDbMock }>>;
+  querier: {
+    tendermint: TendermintMock;
+  };
+}
```
grod220, line 31: These feel like they should live in a shared place. Maybe a `test-utils.ts` file?

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
grod220: Another way is to do this:

```typescript
  test('shsould throw error if broadcast transaction id disagrees', async () => {
    mockTendermint.broadcastTx.mockResolvedValue(new TransactionId());

    try {
      for await (const _ of broadcastTransaction(broadcastTransactionRequest, mockCtx)) {}
      throw new Error('broadcastTransaction did not throw an error as expected');
    } catch (error) {
      expect((error as ConnectError).message).contains('broadcast transaction id disagrees');
    }
  });
```

But don't feel strongly

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
grod220: You can't just do?:

```typescript
vi.mock('@penumbra-zone/crypto-web', () => ({
  sha256Hash: vi.fn(),
}));
```

grod220 (approved):
Few comments, feel free to merge afterward

packages/router/src/grpc/view-protocol-server/broadcast-transaction.test.ts
Valentine1898: `sha256` must return the correct hash, otherwise we will always get the error 'broadcast transaction id disagrees'.

```typescript
const id = new TransactionId({ inner: await sha256Hash(req.transaction.toBinary()) });

  const broadcastId = await tendermint.broadcastTx(req.transaction);
  if (!id.equals(broadcastId)) {
    console.error('broadcast transaction id disagrees', id, broadcastId);
    throw new Error(
      `broadcast transaction id disagrees: expected ${uint8ArrayToHex(id.inner)} != tendermint ${uint8ArrayToHex(broadcastId.inner)}`,
    );
  }
```

Valentine1898:
#400

PR #449, opened 2024-02-01
Move web app deploy from continuous to manual

grod220:
Follow up from web weekly. Given the _unstable_ times we are in, moving the web app deployment to manual (syncing with extension publishes) in order to not break user experiences between testnet launches.

After this, will manually run this action on an old commit that works with our current extension. This should revive the current experience for users.


jessepinho (approved):
🥲

PR #450, opened 2024-02-01
v0.65 updates round 4

turbocrime:
- JsonReadOptions are now more permissive and consistent throughout repo
- getAddressByIndex now handles missing inputs
- fetchers for same slightly adjusted

### DIFF
--- a/apps/webapp/src/fetchers/address.ts
+++ b/apps/webapp/src/fetchers/address.ts
@@ -1,11 +1,4 @@
-import {
-  Address,
-  AddressIndex,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1alpha1/keys_pb';
-import {
-  AddressByIndexRequest,
-  EphemeralAddressRequest,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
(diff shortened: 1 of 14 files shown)


apps/webapp/src/fetchers/address.ts
```diff
@@ -31,19 +24,16 @@ export const getAddresses = async (accounts: (number | undefined)[]): Promise<In
     }, {});
 };
 
-export const getAddressByIndex = async (account: number | undefined): Promise<Address> => {
-  const req = new AddressByIndexRequest();
-  if (account) req.addressIndex = new AddressIndex({ account });
-  const res = await viewClient.addressByIndex(req);
-  if (!res.address) throw new Error('Address not in getAddressByIndex response');
-  return res.address;
+export const getAddressByIndex = async (account = 0): Promise<Address> => {
```
grod220, line 27: I go back and forth between whether we should have default account numbers for these functions. If someone is trying to get an address, they never should be passing nothing right?

grod220 (approved):
One comment, up to you

apps/webapp/src/fetchers/address.ts
turbocrime: well, *we* are sometimes passing nothing, from our own frontend. so

hdevalence: imo passing nothing should be interpreted as the default account (index 0)

PR #451, opened 2024-02-02
Fix bug in how we check for account param's existence

jessepinho:


jessepinho:
Nvm  — looks like @turbocrime already fixed this

PR #454, opened 2024-02-04
idb watching refactor

turbocrime:
made while working on broadcast but it's conceptually separate

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
+++ b/packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
@@ -1,33 +1,34 @@
 import type { Impl } from '.';
 import { servicesCtx } from '../../ctx';
+import { watchSubscription } from './util/watch-subscription';
 
-import {
-  NoteByCommitmentResponse,
-  SpendableNoteRecord,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
+import { SpendableNoteRecord } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 
 import { ConnectError, Code } from '@connectrpc/connect';
 
-export const noteByCommitment: Impl['noteByCommitment'] = async (req, ctx) => {
+export const noteByCommitment: Impl['noteByCommitment'] = async (
+  { noteCommitment: findNoteCommitment, awaitDetection },
+  ctx,
+) => {
   const services = ctx.values.get(servicesCtx);
   const { indexedDb } = await services.getWalletServices();
-  if (!req.noteCommitment)
+  if (!findNoteCommitment)
     throw new ConnectError('Missing note commitment in request', Code.InvalidArgument);
 
-  const noteByCommitment = await indexedDb.getSpendableNoteByCommitment(req.noteCommitment);
-  if (noteByCommitment) return { spendableNote: noteByCommitment };
-  if (!req.awaitDetection) throw new ConnectError('Note not found', Code.NotFound);
-
-  // Wait until our DB encounters a new note with this commitment
-  const response = new NoteByCommitmentResponse();
+  // start subscription early to avoid race condition
   const subscription = indexedDb.subscribe('SPENDABLE_NOTES');
 
-  for await (const update of subscription) {
-    const note = SpendableNoteRecord.fromJson(update.value);
-    if (note.noteCommitment?.equals(req.noteCommitment)) {
-      response.spendableNote = note;
-      break;
-    }
-  }
-  return response;
+  const spendableNote =
+    (await indexedDb.getSpendableNoteByCommitment(findNoteCommitment)) ??
+    (awaitDetection &&
+      SpendableNoteRecord.fromJson(
+        await watchSubscription(subscription, update => {
+          const scannedNote = SpendableNoteRecord.fromJson(update.value);
+          return findNoteCommitment.equals(scannedNote.noteCommitment);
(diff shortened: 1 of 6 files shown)


packages/router/src/grpc/view-protocol-server/swap-by-commitment.ts
```diff
@@ -1,26 +1,34 @@
...
 import { servicesCtx } from '../../ctx';
+import { watchSubscription } from './util/watch-subscription';
 
 import { SwapRecord } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 
 import { ConnectError, Code } from '@connectrpc/connect';
 
-export const swapByCommitment: Impl['swapByCommitment'] = async (req, ctx) => {
+export const swapByCommitment: Impl['swapByCommitment'] = async (
+  { swapCommitment: findSwapCommitment, awaitDetection },
```
jessepinho, line 10: OK if we leave this name as it was? It seems like an arbitrary rename, and the new name sounds more like a function than like a message object.

packages/router/src/grpc/view-protocol-server/swap-by-commitment.ts
```diff
@@ -1,26 +1,34 @@
...
+        }).catch(),
+      ));
 
-  if (req.awaitDetection) {
-    for await (const { value: swapJson } of indexedDb.subscribe('SWAPS')) {
-      const swap = SwapRecord.fromJson(swapJson);
-      if (swap.swapCommitment?.equals(swapCommitment)) return { swap };
-    }
-  }
+  if (swap) return { swap };
```
jessepinho, line 31: Hm... what's the reason for this change? To me, this feels a bit harder to read than how it was before.

packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
```diff
@@ -1,33 +1,34 @@
...
-  NoteByCommitmentResponse,
-  SpendableNoteRecord,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
+import { SpendableNoteRecord } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1alpha1/view_pb';
 
 import { ConnectError, Code } from '@connectrpc/connect';
 
-export const noteByCommitment: Impl['noteByCommitment'] = async (req, ctx) => {
+export const noteByCommitment: Impl['noteByCommitment'] = async (
+  { noteCommitment: findNoteCommitment, awaitDetection },
```
jessepinho, line 10: Same here as below — I feel like the earlier name is clearer. If you want to emphasize that we're looking for this note commitment, you could call it `noteCommitmentToFind`. But right now it sounds like a function name.

packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
```diff
@@ -1,33 +1,34 @@
...
+    (await indexedDb.getSpendableNoteByCommitment(findNoteCommitment)) ??
+    (awaitDetection &&
+      SpendableNoteRecord.fromJson(
+        await watchSubscription(subscription, update => {
+          const scannedNote = SpendableNoteRecord.fromJson(update.value);
+          return findNoteCommitment.equals(scannedNote.noteCommitment);
+        }).catch(),
+      ));
+
+  if (spendableNote) return { spendableNote };
```
jessepinho, line 31: Same note as below — this is a bit harder to read and understand. I get why you're refactoring to use `watchSubscription`, but doing it inside of a `&&` which itself is nested inside a `??` is a bit hard to grok.

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -1,23 +1,19 @@
...
-): Promise<U> => {
-  for await (const update of subscription) if (test(update)) return update;
-  throw new Error('Subscription ended');
-};
-
-export const nullifierStatus: Impl['nullifierStatus'] = async (req, ctx) => {
-  const { nullifier } = req;
-  if (!nullifier) throw new ConnectError('No nullifier passed', Code.InvalidArgument);
+export const nullifierStatus: Impl['nullifierStatus'] = async (
+  { awaitDetection, nullifier: findNullifier },
```
jessepinho, line 13: Same note here re: `findNullifier` name

packages/router/src/grpc/view-protocol-server/nullifier-status.ts
```diff
@@ -33,26 +29,28 @@ export const nullifierStatus: Impl['nullifierStatus'] = async (req, ctx) => {
...
+          watchSubscription(swapStream, ({ value: swapJson }) => {
+            const swap = SwapRecord.fromJson(swapJson);
+            return Boolean(swap.heightClaimed) && findNullifier.equals(swap.nullifier);
+          }),
+          watchSubscription(noteStream, ({ value: noteJson }) => {
+            const note = SpendableNoteRecord.fromJson(noteJson);
+            return Boolean(note.heightSpent) && findNullifier.equals(note.nullifier);
+          }),
+        ])),
+    );
```
jessepinho, line 53: A bit hard to follow. I'd pull out this bit into its own variable above `const spent`, and then refer to that variable here.

jessepinho (commented):
The `watchSubscription` refactor makes sense, but I find some of the changes unnecessarily hard to read and understand. I'd avoid doing too much work inside of a single expression, if things would be more clearly expressed by breaking up the work into multiple statements and variables.

PR #459, opened 2024-02-05
403 add tests for indexbyaddressrequest

Valentine1898:


packages/router/src/grpc/view-protocol-server/index-by-address.ts
```diff
@@ -17,11 +17,7 @@ export const indexByAddress: Impl['indexByAddress'] = async (req, ctx) => {
 
   const addressIndex = isControlledAddress(fullViewingKey, address);
 
-  if (!addressIndex)
-    throw new ConnectError(
-      'Address is not controlled by view service full viewing key',
-      Code.Unauthenticated,
-    );
+  if (!addressIndex) return {};
```
Valentine1898, line 20: We should not return an error if the address is not associated with our FVK, but instead return an empty response like rust does.
My guideline is that the `extension view service` and the `rust viewer service` should have the same behavior and be interchangeable.

apps/webapp/src/fetchers/index-by-address.ts
Valentine1898, line 1: Deleted it because it's not used, and it's not clear how it can be used given this line:
`  return res.addressIndex ?? new AddressIndex();`

packages/router/src/grpc/view-protocol-server/index-by-address.ts
jessepinho: I don't know if I'm the best person to make the call on whether this reasoning is OK, but it sounds fine to me :D

packages/router/src/grpc/view-protocol-server/index-by-address.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      method: ViewService.methods.indexByAddress,
+      protocolName: 'mock',
+      requestMethod: 'MOCK',
+      contextValues: createContextValues().set(servicesCtx, mockServices),
+    });
+
+    testAddress = getAddressByIndex(testFVK, 0);
+  });
+
+  test('should successfully get index for address with index 0', async () => {
```
jessepinho, line 41:
```suggestion
  test('should successfully get index for a given address', async () => {
```

The test name should reflect _what_ you're testing, not _how_ you're testing it. `index 0` is _how_ you're testing it — i.e., you're passing it an address that happens to have an index of 0, and then asserting that the response has an index of 0. But of course, `indexByAddress` can return any account index, not just the 0th one :D

jessepinho (approved):
LGTM! Very minor note on your test name, but otherwise looks great.

packages/router/src/grpc/view-protocol-server/index-by-address.ts
grod220: Normally, I'd disagree, but it appears the proto definitions have this comment:

```
  /**
   * Will be absent if given an address not viewable by this viewing service
   *
   * @generated from field: penumbra.core.keys.v1alpha1.AddressIndex address_index = 1;
   */
  addressIndex?: AddressIndex;
```
So I suppose this is the expected behavior

grod220 (approved):

PR #460, opened 2024-02-05
402 add tests for fmdparametersrequest

Valentine1898:


packages/router/src/grpc/view-protocol-server/app-parameters.test.ts
```diff
@@ -64,7 +57,7 @@ const testData = new AppParameters({
     epochDuration: 719n,
   },
   shieldedPoolParams: {
-    fmdParams: { asOfBlockHeight: 1n, precisionBits: 0 },
+    fixedFmdParams: { asOfBlockHeight: 1n, precisionBits: 0 },
```
Valentine1898, line 60: This did not cause an error

packages/router/src/grpc/view-protocol-server/ephemeral-address.ts
```diff
@@ -3,7 +3,6 @@ import { servicesCtx } from '../../ctx';
 
 import { getEphemeralByIndex } from '@penumbra-zone/wasm-ts';
 
-// TODO handle req.displayConfirm
```
Valentine1898, line 6: `displayConfirm` has been removed, this TODO is not relevant

packages/router/src/grpc/view-protocol-server/fmd-parameters.ts
```diff
@@ -7,6 +7,6 @@ export const fMDParameters: Impl['fMDParameters'] = async (_, ctx) => {
   const services = ctx.values.get(servicesCtx);
   const { indexedDb } = await services.getWalletServices();
   const parameters = await indexedDb.getFmdParams();
-  if (!parameters) throw new ConnectError('No FMD parameters', Code.FailedPrecondition);
+  if (!parameters) throw new ConnectError('No FMD parameters', Code.Unavailable);
```
Valentine1898, line 10: Error status changed to the same as rust view service. 
And in general it seems that `unavailable` is better in this case because the client will not know when FMD parameters will be available

grod220: This code is reserved for this:

```typescript
    /**
     * Unavailable, client should back off and retry.
     */
    Unavailable = 14,
```

Think we should revert to the previous error

grod220 (approved):
Just one comment 👍

Issue #467, opened 2024-02-06
Add swaps and swap claims to transaction classifications


jessepinho:
Right now, a swap or swap claim shows up like this in the transactions list:

![Image](https://github.com/penumbra-zone/web/assets/1121544/abde84d2-8b9b-46e6-b2a4-9750dd5e736f)

This ticket is to:
1. Add `swap` and `swapClaim` types to `packages/types/src/transaction/classification.ts`
2. Handle those types in `classifyTransaction` and `getTransactionClassificationLabel` in `packages/types/src/transaction/classify.ts`
3. Write tests for this behavior.

jessepinho:
Closed by #470

Issue #469, opened 2024-02-07
Don't swallow errors inside the `ViewProtocolServer`


jessepinho:
At the moment, any errors thrown inside a `ViewProtocolServer` impl, like this

![Image](https://github.com/penumbra-zone/web/assets/1121544/143e8759-b033-4046-951e-f610862b471a)

...get swallowed and converted into a useless, generic "internal error":

![Image](https://github.com/penumbra-zone/web/assets/1121544/627c377e-057b-40d7-b5a2-bb0be84d790d)

It looks like [`rethrowImplErrors`](https://github.com/penumbra-zone/web/blob/main/apps/extension/src/service-worker.ts#L81-L109) is supposed to handle that, but somehow errors are still getting swallowed. This ticket is to fix that.

jessepinho:
cc @turbocrime — is this a known issue with `rethrowImplErrors`?

turbocrime:
rethrowImplErrors didn't work on streams because streams have their own internal error handling, and now we're hitting that since we're using more streams

grod220:
Think this is a duplicate of: https://github.com/penumbra-zone/web/issues/437

jessepinho:
Good catch — closing this as a duplicate

PR #470, opened 2024-02-07
Support swap claims when doing a swap

### LINKED ISSUE #420: Second follow up swap claim action
Automatic claiming after a swap. Comparable to pcli code: https://github.com/penumbra-zone/penumbra/blob/7a4ccef16e940ce7e97e415c4d6b06bb9bcba389/crates/bin/pcli/src/command/tx.rs#L477-L532

jessepinho:
🚨 471 line additions are just test files
🚨 Note that the "Swap into" field is not yet used. This PR is just for adding the swap claim step after a swap.

https://github.com/penumbra-zone/web/assets/1121544/22da28c4-328d-4a03-927b-e5d3d2c63d5d

## In this PR
- Build and submit a swap claim transaction automatically after building and submitting the swap transaction.
- Fix bug where we were trying to load proving keys in `camelCase` when we should have been loading them in `snake_case`
- Update `planWitnessBuildBroadcast` to now return the entire built transaction; extract a `getTransactionHash` helper so we can use that where needed.
- Update `classifyTransaction` to also handle `swap`/`swapClaim` transactions.

## Manual testing
If reviewers want to manually test, here's how I tested:
1. Check out penumbra core on the v0.65.0 tag
2. `rm -rf ~/.penumbra/testnet_data`
3. `cargo run --release --bin pd -- testnet generate`
4. Start Penumbra + CometBFT
5. Configure pcli to run against the local devnet, not testnet/testnet-preview
6. `cargo run --bin pcli tx position order buy 1000penumbra@5test_usd` (This sets the exchange rate at 1penumbra = 5test_usd)
7. On the swap page, swap 1 penumbra for test_usd:
    image
8. When it's finished, go back to the homepage to see that you have 1 less penumbra, and 5 more test_usd.

Closes #420, #467

### DIFF
--- a/apps/extension/src/wasm-build-action.ts
+++ b/apps/extension/src/wasm-build-action.ts
@@ -4,6 +4,7 @@ import {
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 import type { WasmBuildActionInput } from '@penumbra-zone/types/src/internal-msg/offscreen';
 import type { JsonValue } from '@bufbuild/protobuf';
+import { camelToSnakeCase } from '@penumbra-zone/types';
 
 // necessary to propagate errors that occur in promises
 // see: https://stackoverflow.com/questions/39992417/how-to-bubble-a-web-worker-error-in-a-promise-via-worker-onerror
@@ -50,7 +51,7 @@ async function executeWorker(
   const actionPlanType = transactionPlan.actions[actionPlanIndex]?.action.case;
   if (!actionPlanType) throw new Error('No action key provided');
 
-  await penumbraWasmModule.loadProvingKey(actionPlanType);
+  await penumbraWasmModule.loadProvingKey(camelToSnakeCase(actionPlanType));
 
   // Build action according to specification in `TransactionPlan`
   return penumbraWasmModule
--- a/apps/webapp/src/state/swap.ts
+++ b/apps/webapp/src/state/swap.ts
@@ -58,9 +58,10 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
 
         const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
         const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq);
+        const swapClaimTxHash = await getTransactionHash(swapClaimTx);
 
         dismiss();
-        toastFn(successTxToast(await getTransactionHash(swapClaimTx)));
+        toastFn(successTxToast(swapClaimTxHash));
 
         // Reset form
         set(state => {
--- a/apps/webapp/src/state/swap.ts
+++ b/apps/webapp/src/state/swap.ts
@@ -57,7 +57,9 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
         const swapCommitment = getSwapCommitment(swapTx);
 
         const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
-        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq);
+        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq, {
+          rpcMethod: 'witnessAndBuild',
+        });
         const swapClaimTxHash = await getTransactionHash(swapClaimTx);
 
         dismiss();
--- a/apps/webapp/src/state/swap.ts
+++ b/apps/webapp/src/state/swap.ts
@@ -58,6 +58,13 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
 
         const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
         const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq, {
+          /**
+           * Swap claims are self-authenticating, so we'll use `witnessAndBuild`
+           * rather than `authorizeAndBuild` for the swap claim. That way, we
+           * won't trigger a second, unnecessary approval popup.
+           *
+           * @see https://protocol.penumbra.zone/main/zswap/swap.html#claiming-swap-outputs
+           */
           rpcMethod: 'witnessAndBuild',
         });
         const swapClaimTxHash = await getTransactionHash(swapClaimTx);
--- a/apps/webapp/src/state/swap.ts
+++ b/apps/webapp/src/state/swap.ts
@@ -2,12 +2,15 @@ import { AllSlices, SliceCreator } from './index';
 import { errorTxToast, loadingTxToast, successTxToast } from '../components/shared/toast-content';
 import { toast } from '@penumbra-zone/ui/components/ui/use-toast';
 import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
-import { planWitnessBuildBroadcast } from './helpers';
+import { getTransactionHash, planWitnessBuildBroadcast } from './helpers';
 import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { AssetBalance } from '../fetchers/balances';
 import { getDisplayDenomExponent, toBaseUnit } from '@penumbra-zone/types';
 import BigNumber from 'bignumber.js';
 import { getAddressByIndex } from '../fetchers/address';
+import { Swap } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/dex/v1/dex_pb';
+import { Transaction } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import { StateCommitment } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/crypto/tct/v1/tct_pb';
 
 export interface SwapSlice {
   assetIn: AssetBalance | undefined;
@@ -49,10 +52,15 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
       const { dismiss } = toastFn(loadingTxToast);
 
       try {
-        const req = await assembleRequest(get().swap);
-        const txHash = await planWitnessBuildBroadcast(req);
+        const swapReq = await assembleSwapRequest(get().swap);
+        const swapTx = await planWitnessBuildBroadcast(swapReq);
+        const swapCommitment = getSwapCommitment(swapTx);
+
+        const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
+        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq);
+
         dismiss();
-        toastFn(successTxToast(txHash));
+        toastFn(successTxToast(await getTransactionHash(swapClaimTx)));
 
         // Reset form
         set(state => {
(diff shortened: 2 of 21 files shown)


apps/extension/src/wasm-build-action.ts
```diff
@@ -50,7 +51,7 @@ async function executeWorker(
   const actionPlanType = transactionPlan.actions[actionPlanIndex]?.action.case;
   if (!actionPlanType) throw new Error('No action key provided');
 
-  await penumbraWasmModule.loadProvingKey(actionPlanType);
+  await penumbraWasmModule.loadProvingKey(camelToSnakeCase(actionPlanType));
```
jessepinho, line 54: This was the beast that took forever to solve 😩 Per #469, errors inside `ViewProtocolServer` impl's are being swallowed, so it took rebuilding the WASM crate, adding log lines everywhere, etc. to discover that we were trying to load the `swapClaim` proving key when we should have been trying to load the `swap_claim` key 🤦🏻‍♂️

apps/webapp/src/state/swap.ts
```diff
@@ -71,7 +79,7 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
   };
 };
 
-const assembleRequest = async ({ assetIn, amount, assetOut }: SwapSlice) => {
+const assembleSwapRequest = async ({ assetIn, amount, assetOut }: SwapSlice) => {
```
jessepinho, line 82: Renamed to differentiate from `assembleSwapClaimRequest` below

packages/router/src/grpc/custody/view-transaction-plan/view-action-plan.ts
```diff
@@ -94,6 +100,51 @@ const getOutputView = (
   });
 };
 
+const getSwapView = (swapPlan: SwapPlan): SwapView => {
```
jessepinho, line 103: This, and `getSwapClaimView` below it, are quite important functions, since they determine what we'll be showing the user when they decide whether to approve or deny a transaction. Please review these thoroughly to make sure they're accurately copying over from Plan -> View

packages/types/src/transaction/classify.ts
```diff
@@ -7,6 +7,9 @@ export const classifyTransaction = (txv?: TransactionView): TransactionClassific
     return 'unknown';
   }
 
+  if (txv.bodyView?.actionViews.some(a => a.actionView.case === 'swap')) return 'swap';
+  if (txv.bodyView?.actionViews.some(a => a.actionView.case === 'swapClaim')) return 'swapClaim';
```
jessepinho, line 11: (Is it ever possible for a single transaction to have multiple types of actions at once? e.g., where you send money to another person, delegate some tokens to a validator, and do a swap — all in the same transaction? if so, this `classifyTransaction` function is pretty useless, since it's assuming there's only one primary purpose per transaction.)

grod220:
Going to rebase your work on latest testnet-preview changes so I can play around with this locally (currently broken)

grod220 (commented):
Have you found it actually makes the swap? I was having two issues:

- Getting a `No return address found in transaction plan` error. We don't require return addresses for swaps, so I think we need to relax [this error](https://github.com/penumbra-zone/web/blob/1dbfefe899dd99b1a93618e3f7adf08f20d914cb/packages/router/src/grpc/custody/view-transaction-plan/index.ts#L25-L26).
- Have you gotten a swap+claim to work in practice? Swaps seem to take my funds, but the swap claims appear to just return them to my wallet. It's weird.
- Think a follow up PR to remove the second auth makes sense. We just need to drop the  authorizationData requirement for the `WitnessAndBuild` request (we are currently error guarding in the rpc method) and get support from core for making `effect_hash` optional. At the moment, that is blocking us.

grod220:
Update: 
- After [this](https://github.com/penumbra-zone/penumbra/pull/3763) gets merged, we can generate a new wasm crate version to use. 
- This will allow us to use `viewClient.witnessAndBuild()` instead of `viewClient.authorizeAndBuild()` for the swap claim  excluding the `authorizationData` field.
- We should delete this [error guard](https://github.com/penumbra-zone/web/blob/1dbfefe899dd99b1a93618e3f7adf08f20d914cb/packages/router/src/grpc/view-protocol-server/witness-and-build.ts#L12-L13)

packages/router/src/grpc/custody/view-transaction-plan/view-action-plan.ts
```diff
@@ -94,6 +100,73 @@ const getOutputView = (
...
+  fullViewingKey: string,
+): SwapClaimView => {
+  return new SwapClaimView({
+    swapClaimView: {
+      case: 'visible',
+      value: {
+        output1: {
+          address: swapClaimPlan.swapPlaintext?.claimAddress
+            ? getAddressView(swapClaimPlan.swapPlaintext.claimAddress, fullViewingKey)
+            : undefined,
```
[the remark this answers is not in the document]
erwanor: I think they should be both the claimed address. I'm intrigued that one of them is not recognized, happy to pair on this with either of you this afternoon.

packages/router/src/grpc/custody/view-transaction-plan/view-action-plan.ts
```diff
@@ -94,6 +100,73 @@ const getOutputView = (
...
+          value: swapClaimPlan.outputData?.lambda1
+            ? getValueView(
+                new Value({
+                  amount: swapClaimPlan.outputData.lambda1,
```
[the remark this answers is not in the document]
erwanor: You're right that `delta` = input and `lambda` = output, but the output data in the swap claim is for the entire batch of trades, so you need to compute the user's share of that batch (also referred to as "pro-rated share" in other parts of the codebase). Maybe this is done elsewhere, I haven't went through all of it yet.

I am writing from my phone but will be in the office in a sec, but basically:
1. Get the Swap inputs
2. Take the batch output data
3. Compute the user's share of the input

5. Multiply that by the total output of that assetrq

TalDerei:
> Update:
> 
> * After [this](https://github.com/penumbra-zone/penumbra/pull/3763) gets merged, we can generate a new wasm crate version to use.
> * This will allow us to use `viewClient.witnessAndBuild()` instead of `viewClient.authorizeAndBuild()` for the swap claim  excluding the `authorizationData` field.
> * We should delete this [error guard](https://github.com/penumbra-zone/web/blob/1dbfefe899dd99b1a93618e3f7adf08f20d914cb/packages/router/src/grpc/view-protocol-server/witness-and-build.ts#L12-L13)

We currently call `const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq)` which internally invokes the `viewClient.authorizeAndBuild` RPC. We should probably rename existing `planWitnessBuildBroadcast` calls to `planAuthorizeBuildBroadcast`, and separately use `planWitnessBuildBroadcast` that calls `witnessAndBuild` for swap claims. Or we can pass a flag to `planWitnessBuildBroadcast` to detects whether it's a swap claim, and switch on viewClient.witnessAndBuild or viewClient.authorizeAndBuild.

packages/router/src/grpc/custody/view-transaction-plan/view-action-plan.ts
TalDerei: swap claims and views are working properly for me

apps/extension/src/wasm-build-action.ts
TalDerei: @jessepinho We can actually simply remove `{ keyType: 'nullifierDerivation', file: 'nullifier_derivation_pk.bin' }` and change convert to `{ keyType: 'undelegateClaim', file: 'convert_pk.bin' }`.

jessepinho:
@TalDerei yeah, I'm planning that for a follow-up PR. I'm actually thinking of just adding an `rpcMethod` argument to (to be-renamed) `planWitnessBuildBroadcast` that allows you to choose between calling `witnessAndBuild` vs. `authorizeAndBuild`.

apps/extension/src/wasm-build-action.ts
jessepinho: @grod220 We're [calling out to WASM](https://github.com/penumbra-zone/web/blob/main/packages/wasm/src/utils.ts#L18) with the snake-cased key type; WASM then maps snake-cased keys to filenames [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/keys.rs#L26-L45). So I figured it'd be best to keep the names consistent between the two. Thoughts?

apps/webapp/src/state/swap.ts
```diff
@@ -49,10 +52,15 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
+        const swapReq = await assembleSwapRequest(get().swap);
+        const swapTx = await planWitnessBuildBroadcast(swapReq);
+        const swapCommitment = getSwapCommitment(swapTx);
+
+        const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
+        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq);
+
         dismiss();
-        toastFn(successTxToast(txHash));
+        toastFn(successTxToast(await getTransactionHash(swapClaimTx)));
```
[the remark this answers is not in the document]
jessepinho: fixed

jessepinho:
@grod220 the errors you're getting came after the rebase — working on those now

apps/webapp/src/state/swap.ts
```diff
@@ -49,10 +52,15 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
 
       try {
-        const req = await assembleRequest(get().swap);
-        const txHash = await planWitnessBuildBroadcast(req);
+        const swapReq = await assembleSwapRequest(get().swap);
+        const swapTx = await planWitnessBuildBroadcast(swapReq);
+        const swapCommitment = getSwapCommitment(swapTx);
+
+        const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
+        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq);
```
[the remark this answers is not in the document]
jessepinho: Discussed on Discord — this is now a separate ticket (#473)

packages/router/src/grpc/custody/view-transaction-plan/view-action-plan.ts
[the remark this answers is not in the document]
jessepinho: Note to self: this appears to just be a rendering issue. `index.account` is indeed set to `0` here — will investigate why it's not rendering though.

apps/extension/src/wasm-build-action.ts
grod220: Ah, I see. Hmm. Kinda on the fence. Ideally, the type system could assist us in ensuring we don't accidentally put the wrong strings there. Given `load_proving_key` is made for the web, it feels ok to convert those to camelCase so that we can use the bufbuild types.

jessepinho:
OK @grod220 this is officially ready for review. If you want to test, I added manual testing steps to the PR description.

apps/extension/src/wasm-build-action.ts
grod220: This change will be easier to make after migrating to web repo

apps/webapp/src/state/swap.ts
```diff
@@ -49,10 +52,18 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
-        const req = await assembleRequest(get().swap);
-        const txHash = await planWitnessBuildBroadcast(req);
+        const swapReq = await assembleSwapRequest(get().swap);
+        const swapTx = await planWitnessBuildBroadcast(swapReq);
+        const swapCommitment = getSwapCommitment(swapTx);
+
+        const swapClaimReq = assembleSwapClaimRequest(swapCommitment);
+        const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq, {
+          rpcMethod: 'witnessAndBuild',
+        });
```
grod220, line 62: Wonder if we should have a code comment about swap claims not requiring second auth

grod220 (approved):

apps/extension/src/wasm-build-action.ts
jessepinho: Ah, forgot about this comment. But OK, will leave it as-is then.

apps/webapp/src/state/swap.ts
jessepinho: Good call — I added a comment

PR #472, opened 2024-02-07
Fix imports from tests to use core v1

jessepinho:
Looks like a couple imports were missing in the recent upgrade to v1


Valentine1898 (approved):

Valentine1898:
Sorry, that's my lapse, didn't notice before merging #465.

jessepinho:
@Valentine1898 all good!

PR #474, opened 2024-02-07
WIP: Show multiple toasts for swap + swap claim transactions

jessepinho:


packages/ui/components/ui/toaster.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { Toaster } from './toaster';
+import { toast } from 'sonner';
+import { Button } from './button';
+
+const meta: Meta<typeof Toaster> = {
```
jessepinho, line 7: image

jessepinho:
Moving back to draft while I resolve rebase conflicts

jessepinho:
OK @turbocrime this is ready for review again. Note that this PR now also includes #493, which I had branched off of this branch to avoid merge conflicts, so you can safely ignore all those changes since you already previously reviewed them

apps/extension/src/routes/popup/popup-layout.tsx
```diff
@@ -26,7 +26,7 @@ export const PopupLayout = () => {
     <div
       className={cn(
         'relative flex flex-col min-h-full',
-        isDarkBg ? 'bg-chachoal' : 'bg-charcoal-secondary',
+        isDarkBg ? 'bg-charcoal' : 'bg-charcoal-secondary',
```
grod220, line 29: oh nice catch

apps/webapp/src/state/ibc.ts
```diff
@@ -57,32 +62,37 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
...
 
-      const { dismiss, update } = toastFn(buildingTxToast());
+      const txToastId = toast(...buildingTxToast());
 
       try {
         const transactionPlan = await plan(await getPlanRequest(get().ibc));
-        const transaction = await authWitnessBuild(update, { transactionPlan });
-        const { txHash, detectionHeight } = await broadcast(update, { transaction });
+        const transaction = await authWitnessBuild({ transactionPlan }, status =>
+          toast(...buildingTxToast(status, txToastId)),
```
grod220, line 75: Instead of the consumer being required to install the npm package + spread the results of `buildingTxToast()` into it, could we change the api of `buildingTxToast()` to do this for us?

```diff
- toast(...buildingTxToast(status, txToastId))
+ buildingTxToast(status, txToastId)
```

apps/webapp/src/state/ibc.ts
```diff
@@ -57,32 +62,37 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
...
+        );
+        const txHash = await getTxHash(transaction);
+        const { detectionHeight } = await broadcast({ transaction }, status =>
+          toast(...broadcastingTxToast(txHash, status, txToastId)),
+        );
 
-        update(successTxToast(txHash, detectionHeight));
+        toast(...successTxToast(txHash, detectionHeight));
```
grod220, line 82: _(referencing https://sonner.emilkowal.ski/)_

Could we make use of the different toast "types"? Looks like success/error/promise could be useful here?

Can we also use the `richColors` setting?

apps/webapp/src/components/send/send-form/index.tsx
```diff
@@ -56,7 +54,7 @@ export const SendForm = () => {
       className='flex flex-col gap-4 xl:gap-3'
       onSubmit={e => {
         e.preventDefault();
-        void sendTx(toast);
+        void sendTx();
```
grod220, line 57: Ah, that's nice we don't have to pass these anymore

grod220 (approved):
Some comments, not sure if they are possible/relevant though

apps/webapp/src/state/ibc.ts
turbocrime, line 82: `txHash` is available from `authWitnessBuild`, you don't have to calculate it independently

apps/webapp/src/state/swap.ts
```diff
@@ -50,56 +55,73 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
-        const { txHash: swapClaimHash, detectionHeight: swapClaimDetectionHeight } =
-          await broadcast(update, {
-            transaction: swapClaimTx,
-          });
-
-        update(successTxToast(swapClaimHash, swapClaimDetectionHeight));
+        const { detectionHeight: swapClaimDetectionHeight } = await broadcast(
+          { awaitDetection: true, transaction: swapClaimTx },
+          status => toast(...broadcastingTxToast(swapClaimTxHash, status, swapClaimTxToastId)),
+        );
```
turbocrime, line 103: this has broken the parallelism of the claim build during swap broadcast. now they happen sequentially

apps/webapp/src/state/ibc.ts
turbocrime: this happens in the other slices as well

apps/webapp/src/state/swap.ts
grod220: We can't build a swap claim without the swap record in the database fyi:
Removed this here: https://github.com/penumbra-zone/web/pull/498#discussion_r1489442684

apps/webapp/src/state/ibc.ts
jessepinho: @turbocrime it's available from `broadcast,` not `authWitnessBuild`, right? And since the `broadcastingTxToast` requires a `txHash`, I had to calculate the hash from the transaction, rather than getting it back from `broadcast`.

jessepinho: ah, yeah, I did that in a few places but missed this, apparently. I'll fix it (which will be easier when using your approach above re: calling `toast()` from the `toast-content` helpers).

jessepinho: image

done

PR #475, opened 2024-02-08
Connection refactor

turbocrime:
Yeah this is pretty large, sorry. Includes:

## packages
- Split transport into two packages, move some generic extension code into the transport packages.
- Provide a client package that exposes tools for connecting to Prax

## extension content scripts
- Provide visibility of the extension to the page
- Provide a way to request access to the extension
- Provide a way to acquire connection to the extension

## extension ui
- connection approval dialog
- no review of connections in settings, yet (that's next, trying to limit scope)

## extension background
- some internal reorganization
- removed a lot of the cross-package imports that were formerly necessary for messaging
- internal messaging is now simpler
  - the extension only handles one unprompted message from content-scripts: the connection request
  - internal handlers are located with their relevant behavior

## transport
- refactored init to support above
- refactored internals, now it's service-generic, given the proper registry
- now much better organized

### cut from scope
- init with reduced permissions. experimenting with this took a good chunk of time. i still think this is a good idea, but this would change the entire init process again.
- management of existing connected sites. really, it's a simple change now that the state and init exists


apps/webapp/src/clients.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Query as IbcClientService } from '@buf/cosmos_ibc.connectrpc_es/ibc/core/client/v1/query_connect';
+import { jsonOptions } from '@penumbra-zone/types/src/json-options';
+
+import { penumbra } from '@penumbra-zone/types/penumbra-global';
+
+export const getPenumbraPort = async () => {
+  if (!window[penumbra]) throw Error('No Penumbra global (no provider installed)');
+
+  const providers = Array.from(Object.values(window[penumbra]));
+  if (!providers[0] || typeof providers[0].request != 'function')
```
jessepinho, line 15: Since we're hard-coding `PRAX_ORIGIN` for now in the webapp anyway, why not also hard-code it here for now? Why use the first item in the array?

apps/webapp/src/clients.ts
```diff
@@ -0,0 +1,34 @@
...
+import { CustodyService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/custody/v1/custody_connect';
+import { Query as IbcClientService } from '@buf/cosmos_ibc.connectrpc_es/ibc/core/client/v1/query_connect';
+import { jsonOptions } from '@penumbra-zone/types/src/json-options';
+
+import { penumbra } from '@penumbra-zone/types/penumbra-global';
+
+export const getPenumbraPort = async () => {
+  if (!window[penumbra]) throw Error('No Penumbra global (no provider installed)');
+
+  const providers = Array.from(Object.values(window[penumbra]));
```
jessepinho, line 14:
```suggestion
  const providers = Object.values(window[penumbra]);
```
(Right? Do we need `Array.from()` here?)

packages/constants/config.ts
```diff
@@ -7,7 +7,7 @@ export interface Config {
 }
 
 export const testnetConstants: Config = {
-  grpcEndpoint: 'https://grpc.testnet.penumbra.zone',
+  grpcEndpoint: 'https://grpc.testnet.penumbra.zone/',
```
jessepinho, line 10: Is this an intentional/necessary change?

packages/types/promise-with-resolvers.ts
```diff
@@ -0,0 +1,17 @@
+// Promise.withResovlers is a convenience function already shipping in major
```
jessepinho, line 1:
```suggestion
// Promise.withResolvers is a convenience function already shipping in major
```

packages/types/promise-with-resolvers.ts
```diff
@@ -0,0 +1,17 @@
...
+  reject: (reason?: any) => void;
+}
+
+declare global {
+  interface PromiseConstructor {
+    withResolvers<T>(): PromiseWithResolvers<T>;
+  }
+}
+
+export {};
```
jessepinho, line 17: If you rename this file to `.d.ts`, is the export still necessary?

packages/types/src/lo-hi.ts
```diff
@@ -34,9 +34,9 @@ export const splitLoHi = (value: bigint): Required<LoHi> => {
  * returns a bigint that combines these two parts.
  * To achieve this, we shift `hi` 8 bytes to the left.
  *
- * @param {bigint} lo - The low order 64 bits of the number.
- * @param {bigint} hi - The high order 64 bits of the number.
```
jessepinho, line 38: Not in scope for this PR, but these `@param` declarations should be moved to `/** ... */`-style comments before the params themselves in the function definition. JSDoc style comments aren't needed in TypeScript land :)

packages/types/src/lo-hi.ts
```diff
@@ -34,9 +34,9 @@ export const splitLoHi = (value: bigint): Required<LoHi> => {
...
- * @returns {bigint} The combined 128-bit number represented as a single bigint.
+ * @param  lo - The low order 64 bits of the number.
+ * @param  hi - The high order 64 bits of the number.
+ * @returns  The combined 128-bit number represented as a single bigint.
  */
 export const joinLoHi = (lo = 0n, hi = 0n): bigint => {
```
jessepinho, line 41: Clarifying my comment above:
```suggestion
export const joinLoHi = (
  /** The low order 64 bits of the number. */
  lo = 0n,
  /** The high order 64 bits of the number. */
  hi = 0n,
): bigint => {
```

packages/types/src/registry.ts
```diff
@@ -21,18 +30,35 @@ import { SimulationService } from '@buf/penumbra-zone_penumbra.connectrpc_es/pen
  */
 
 export const typeRegistry = createRegistry(
```
jessepinho, line 32: Add comment here instructing devs to add more services to this list as they are created in core

apps/extension/src/routes/popup/approval/origin.tsx
turbocrime, line 1: copied mostly from settings page

apps/extension/src/routes/popup/approval/approve-deny.tsx
turbocrime, line 1: component reused with tx-approval

apps/extension/src/routes/popup/settings/settings-connected-sites.tsx
turbocrime, line 1: deferred to next pr

packages/router/src/ctx/approver.ts
turbocrime, line 1: approver was moved into the extension, where it's injected. this should be the pattern for all context items: local interface definition and handle, and injection from the adapter.

turbocrime, line 1: transaction approver was moved into the extension, and is now injected.

to achieve portability of the service implementations, this should be the pattern for all context items: an interface and key defined in the service, implemented and injected from the host adapter.

packages/router/src/grpc/custody/authorize.ts
turbocrime, line 1: since approver moved to external injection, it's possibly undefined

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -26,8 +26,6 @@ export const balances: Impl['balances'] = async function* (req, ctx) {
       accounts[accountNumber] = {};
     }
 
-    // not throwing ConnectError here, just a normal Error, because these
-    // failures are truly internal. they did not arise from request input.
```
turbocrime, line 30: this isn't relevant to PR but this comment is incorrect by now and that bothered me

packages/services/src/index.ts
```diff
@@ -12,7 +16,16 @@ export interface ServicesConfig {
...
+  constructor(private readonly config: ServicesConfig) {
+    chrome.runtime.onMessage.addListener((req, sender, respond) => {
+      if (req != ServicesMessage.ClearCache) return false;
+      if (sender.origin != origin) return false;
+
+      void this.clearCache().then(() => respond());
+
+      return true;
+    });
+  }
```
turbocrime, line 28: clear cache message handled here

grod220 (commented):
Great job!

apps/extension/src/approve-origin.ts
```diff
@@ -0,0 +1,45 @@
...
+        origin: urlOrigin,
+        favIconUrl: tab.favIconUrl,
+        title: tab.title,
+      },
+    });
+    if ('error' in res)
+      throw errorFromJson(res.error as JsonValue, undefined, ConnectError.from(res));
+
+    connectedSites[urlOrigin] = res.data.attitude;
+    void localExtStorage.set('connectedSites', connectedSites); // TODO: is there a race condition here?
```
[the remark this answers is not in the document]
turbocrime: we could, but there's no reason to. awaiting this would only block return of the handler.

apps/extension/src/approve-origin.ts
```diff
@@ -0,0 +1,45 @@
...
+}: chrome.runtime.MessageSender): Promise<boolean> => {
+  if (!senderOrigin?.startsWith('https://') || !tab?.id || frameId)
+    throw new Error('Unsupported sender');
+
+  // parses the origin and returns a consistent format
+  const urlOrigin = new URL(senderOrigin).origin;
+
+  const connectedSites = await localExtStorage.get('connectedSites');
+
+  if (connectedSites[urlOrigin] == null) {
```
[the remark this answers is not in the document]
turbocrime: now comparing with undefined

packages/services/src/index.ts
```diff
@@ -12,7 +16,16 @@ export interface ServicesConfig {
 export class Services implements ServicesInterface {
   private walletServicesPromise: Promise<WalletServices> | undefined;
 
-  constructor(private readonly config: ServicesConfig) {}
+  constructor(private readonly config: ServicesConfig) {
+    chrome.runtime.onMessage.addListener((req, sender, respond) => {
+      if (req != ServicesMessage.ClearCache) return false;
+      if (sender.origin != origin) return false;
```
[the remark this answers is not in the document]
turbocrime: added lint rule in another pr

apps/extension/src/approve-origin.ts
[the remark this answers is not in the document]
turbocrime: just kidding, now using typeof, which is what i actually wanted

packages/client/get-port.ts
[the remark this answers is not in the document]
turbocrime: yes oops

apps/extension/src/clients.ts
```diff
@@ -0,0 +1,14 @@
+import { ViewService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/view/v1/view_connect';
+import { createPromiseClient } from '@connectrpc/connect';
+import { transportOptions } from '@penumbra-zone/types/registry';
+import { createChannelTransport } from '@penumbra-zone/transport-dom/create';
+import { CRSessionClient } from '@penumbra-zone/transport-chrome/session-client';
+
+const port = CRSessionClient.init(PRAX);
```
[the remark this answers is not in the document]
turbocrime: Chrome Runtime

apps/extension/src/state/origin-approval.ts
```diff
@@ -0,0 +1,75 @@
+import { ConnectError } from '@connectrpc/connect';
+import { PopupType, OriginApproval } from '../message/popup';
+import { AllSlices, SliceCreator } from '.';
+import { InternalRequest, InternalResponse } from '@penumbra-zone/types/src/internal-msg/shared';
+import { errorToJson } from '@connectrpc/connect/protocol-connect';
+
+export interface OriginApprovalSlice {
+  responder?: (m: InternalResponse<OriginApproval>) => void;
+  favIconUrl?: string;
```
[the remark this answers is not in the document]
turbocrime: this is the name that chrome runtime uses on a sender object

https://github.com/penumbra-zone/web/blob/connection-refactor/apps/extension/src/approve-origin.ts#L30-L37

apps/extension/src/utils/auto-login.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * This is a convenience script to auto-login the user in development builds.
+ * You can use it as a side-effectful import in service-worker.ts
+ */
+
+import { localExtStorage, sessionExtStorage } from '@penumbra-zone/storage';
+import { Key, KeyPrint } from '@penumbra-zone/crypto-web';
+
+// in development builds, attempt auto-login with a blank password
+if (process.env['NODE_ENV'] === 'development')
```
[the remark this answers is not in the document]
turbocrime: webpack injects node env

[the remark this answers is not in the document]
turbocrime: it only runs if you include it. i'll remove

apps/extension/src/content-scripts/injected-penumbra-global.ts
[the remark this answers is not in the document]
turbocrime: commented

turbocrime:
https://github.com/penumbra-zone/web/assets/134443988/d090011c-0fde-47ea-b28a-e10f2bb0e0cc

apps/extension/src/content-scripts/message.ts
```diff
@@ -0,0 +1,33 @@
+import { Prax } from '../message/prax';
+
+// @ts-expect-error - ts can't understand the injected string
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type PraxMessage<T = unknown> = { [PRAX]: T };
```
grod220, line 5: What if we just do:

```typescript
export type PraxMessage<T> = Record<typeof PRAX, T>;
const isPraxMessageEventData = <T>(p: unknown): p is PraxMessage<T> =>
  typeof p === 'object' && p != null && PRAX in p;
```

I don't think `{ [PRAX]: T }` is giving us typesafety anyway.

apps/extension/src/popup.ts
```diff
@@ -0,0 +1,73 @@
...
+  req: PopupRequest<M>,
+): Promise<PopupResponse<M>> =>
+  spawnPopup(req.type)
+    .then(async () => {
+      // We have to wait for React to bootup, navigate to the page, and render the components
+      await new Promise(resolve => {
+        setTimeout(resolve, 1000);
+      });
+    })
+    .then(() => chrome.runtime.sendMessage<InternalRequest<M>, InternalResponse<M>>(req));
```
grod220, line 19: Let's not mix .then() and awaits here. Think this is more readable:

```typescript
export const popup = async <M extends PopupMessage>(
  req: PopupRequest<M>,
): Promise<PopupResponse<M>> => {
  await spawnPopup(req.type);
  await new Promise(resolve => setTimeout(resolve, 1000));
  return chrome.runtime.sendMessage<InternalRequest<M>, InternalResponse<M>>(req);
};
```

apps/extension/src/popup.ts
```diff
@@ -0,0 +1,73 @@
...
+      documentUrls: [chrome.runtime.getURL(path)],
+    })
+    .then(popupContexts => {
+      if (popupContexts.length) throw Error('Popup already open');
+    });
+
+const spawnPopup = async (pop: PopupType) => {
+  const popUrl = new URL(chrome.runtime.getURL('popup.html'));
+
+  if (!(await sessionExtStorage.get('passwordKey'))) {
```
grod220, line 57: let's pull the await out of the condition so it's easier to read

apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
```diff
@@ -31,10 +31,8 @@ export const SettingsClearCache = () => {
             size='lg'
             className='w-full'
             onClick={() => {
-              void (async () => {
-                await internalSwClient.clearCache();
-                navigate(PopupPath.INDEX);
-              })();
+              void chrome.runtime.sendMessage(ServicesMessage.ClearCache);
+              navigate(PopupPath.INDEX);
```
grod220, line 35: Nice, I think we do this in a few other place too

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -46,13 +46,14 @@ export const SettingsRPC = () => {
         setNewChainId(appParameters.chainId);
         await setGRPCEndpoint(rpcInput);
         // If the chain id has changed, our cache is invalid
-        if (appParameters.chainId !== currentChainId) void internalSwClient.clearCache();
+        if (appParameters.chainId !== currentChainId)
+          void chrome.runtime.sendMessage(ServicesMessage.ClearCache);
```
grod220, line 50: This is a good candidate for brackets

apps/extension/src/service-worker.ts
```diff
@@ -2,136 +2,72 @@
...
+ * - rpc services, router, and adapter
+ * - session manager for rpc entry
  */
 
-import { Services } from '@penumbra-zone/services';
-import { localExtStorage } from '@penumbra-zone/storage';
-
-import { jsonOptions } from '@penumbra-zone/types/src/json-options';
-import { custodyCtx, servicesCtx } from '@penumbra-zone/router/src/ctx';
+// import './auto-login';
```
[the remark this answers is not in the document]
grod220: Can we remove this? I don't think we need auto-login in dev when we've removed the password requirement

apps/webapp/src/components/layout.tsx
```diff
@@ -1,31 +1,40 @@
...
+      isConnected: true;
       chainId: string;
     };
 
 export const LayoutLoader: LoaderFunction = async (): Promise<LayoutLoaderResult> => {
-  const isInstalled = isExtensionInstalled();
-  if (!isInstalled) return { isInstalled };
+  const isInstalled = isPraxAvailable();
+  if (!isInstalled) return { isInstalled, isConnected: false };
+  const isConnected = isPraxConnected() || (await isPraxConnectedTimeout(1000));
```
grod220, line 27: 1 second felt a little long to me, could we get away with 500ms?

packages/client/prax.ts
```diff
@@ -0,0 +1,88 @@
...
+import { createChannelTransport } from '@penumbra-zone/transport-dom/create';
+import { jsonOptions } from '@penumbra-zone/types/registry';
+import { PenumbraSymbol } from './global';
+
+export class PraxNotAvailableError extends Error {}
+export class PraxNotConnectedError extends Error {}
+export class PraxManifestError extends Error {}
+
+export const getPraxPort = async () =>
+  window[PenumbraSymbol]?.['chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe']?.connect()!;
```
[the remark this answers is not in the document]
grod220: or better, it should be imported from `/packages/constants`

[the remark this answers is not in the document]
grod220: Also, does it need `!` at the end?

packages/client/prax.ts
```diff
@@ -0,0 +1,88 @@
...
+  Boolean(
+    window[PenumbraSymbol]?.['chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe']?.isConnected(),
+  );
+
+export const isPraxConnectedTimeout = (timeout: number) =>
+  new Promise<boolean>(resolve => {
+    if (
+      window[PenumbraSymbol]?.['chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe']?.isConnected()
+    )
+      resolve(true);
```
grod220, line 34: The conditional is so long, that without the brackets, I couldn't tell where it ended. Can we add them here? Or can you lift the statement out of the conditional and do:

```typescript
    const isConnected = window[PenumbraSymbol]?.['chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe']?.isConnected();
    if (isConnected) resolve(true);
```

packages/client/prax.ts
```diff
@@ -0,0 +1,88 @@
...
+        clearInterval(interval);
+        resolve(true);
+      }
+    }, 100);
+
+    setTimeout(() => {
+      clearInterval(interval);
+      resolve(false);
+    }, timeout);
+  });
```
grod220, line 51: What if we did a Promise race here?

```typescript
export const isPraxConnectedTimeout = (timeout: number): Promise<boolean> => {
  const connectPromise = new Promise<boolean>(resolve => {
    const interval = setInterval(() => {
      if (isPraxConnected()) {
        clearInterval(interval);
        resolve(true);
      }
    }, 100);
  });

  const timeoutPromise = new Promise<boolean>(resolve => setTimeout(resolve, timeout, false));

  return Promise.race([connectPromise, timeoutPromise]);
};
```

packages/client/prax.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const throwIfPraxNotConnectedTimeout = async (timeout: number = 500) => {
+  if (!(await isPraxConnectedTimeout(timeout)))
+    throw new PraxNotConnectedError('Prax not connected');
+};
+
+export const isPraxAvailable = () =>
+  Boolean(window[PenumbraSymbol]?.['chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe']);
+
+export const throwIfPraxNotAvailable = () => {
```
grod220, line 61: Is available the same as "installed"? Can we change it to that?

packages/client/prax.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (!isPraxAvailable()) throw new PraxNotAvailableError('Prax not available');
+};
+
+export const throwIfPraxNotConnected = () => {
+  if (!isPraxConnected()) throw new PraxNotConnectedError('Prax not connected');
+};
+
+export const getPraxManifest = async () => {
```
grod220, line 69: Looks like some of these are unused. But would they still be helpful for clients?

packages/client/get-port.ts
```diff
@@ -0,0 +1,9 @@
+import penumbra from '.';
+
+export const getAnyPenumbraPort = async (request?: boolean) => {
```
grod220, line 3: What's the word `Any` imply here? Could we have multiple at once and it doesn't matter? Also it seems unused.

packages/polyfills/Array.fromAsync.ts
grod220, line 1: Seems unused. Do we need this for this PR? If not, can we remove?

packages/transport-dom/direct.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const directGetPort =
+  (entry: ChannelHandlerFn, jsonOptions?: Partial<JsonReadOptions> & Partial<JsonWriteOptions>) =>
+  (): Promise<MessagePort> => {
+    const { port1: servicePort, port2: clientPort } = new MessageChannel();
+
+    // straight-pipe connection, no conditions or transformations
+    const directEntryHandler = async ({ requestId, message }: TransportMessage) => {
+      const transportResponse: TransportEvent = await entry(message)
+        .then(response =>
```
grod220, line 27: Can we choose .then() or awaits here?

apps/extension/src/routes/popup/approval/origin.tsx
```diff
@@ -0,0 +1,77 @@
...
+          Connect
+        </h1>
+        <div className='mx-auto size-20'>
+          <LinkGradientIcon />
+        </div>
+        <div className='flex flex-1 flex-col items-start justify-between px-[30px] pb-[30px]'>
+          <div className='flex w-full flex-col gap-4'>
+            <div className='flex flex-col gap-2'>
+              <div className='flex h-11 w-full items-center rounded-lg border bg-background px-3 py-2 text-muted-foreground'>
+                {favIconUrl ? <img src={favIconUrl} alt='icon' className='h-11' /> : null}
```
grod220, line 43: The favicon appears to exceed the size of the box. Can you make it the height (minus padding) of the box?

grod220 (approved):
Good job! Left comments. After those, think it's best to get this in given how many lines of code it is.

Also, I seem to still have the issue where the popup does not trigger 🤔 . I only get `postRequest` in the web app console. And this in the extension console:

```
popup.ts:16 Uncaught (in promise) TypeError: chrome.action.openPopup is not a function
```

turbocrime:
> `popup.ts:16 Uncaught (in promise) TypeError: chrome.action.openPopup is not a function`

https://developer.chrome.com/docs/extensions/reference/api/action#method-openPopup

since chrome 118. there have been several 0day since then btw

apps/extension/src/approve-origin.ts
```diff
@@ -0,0 +1,44 @@
...
+  tab,
+  frameId,
+}: chrome.runtime.MessageSender): Promise<boolean> => {
+  if (!senderOrigin?.startsWith('https://') || !tab?.id || frameId)
+    throw new Error('Unsupported sender');
+
+  // parses the origin and returns a consistent format
+  const urlOrigin = new URL(senderOrigin).origin;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const connectedSites = (await localExtStorage.get('connectedSites')) ?? {};
```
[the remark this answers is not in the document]
turbocrime: we do set a default, but that does not guarantee this will never fail. in the case where the storage object is somehow not present, failing to handle this creates an unrecoverable error. i'll change the storage type to account for this and enforce checks

apps/extension/src/approve-origin.ts
```diff
@@ -0,0 +1,44 @@
...
+      origin: urlOrigin,
+      favIconUrl: tab.favIconUrl,
+      title: tab.title,
+    },
+  });
+  if ('error' in res)
+    throw errorFromJson(res.error as JsonValue, undefined, ConnectError.from(res));
+
+  connectedSites[urlOrigin] = res.data.attitude;
+  void localExtStorage.set('connectedSites', connectedSites); // TODO: is there a race condition here?
```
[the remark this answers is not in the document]
turbocrime: i left a larger comment in the PR chained into this one

apps/extension/src/content-scripts/message.ts
turbocrime: it's useful to have a composeable type

apps/webapp/src/components/layout.tsx
turbocrime: no, we couldn't

packages/client/prax.ts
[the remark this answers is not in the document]
turbocrime: trying to make this fairly standalone. it does need the exclamation

packages/client/prax.ts
turbocrime: yes

packages/polyfills/Array.fromAsync.ts
turbocrime: it's not unused

turbocrime: there is a pr for it and jesse blocked it

turbocrime: https://github.com/penumbra-zone/web/pull/644 broke this out into its own pr with explanation/tests

packages/transport-dom/direct.ts
turbocrime: we actually can't, without getting significantly more complicated. i can collapse the 'catch' into a second 'then' parameter.

apps/extension/src/routes/popup/approval/origin.tsx
turbocrime: yes. this was a deliberate aesthetic choice

packages/client/prax.ts
turbocrime: no, the timeout needs to be able to clear the interval

apps/extension/src/approve-origin.ts
```diff
@@ -0,0 +1,44 @@
...
+    request: {
+      origin: urlOrigin,
+      favIconUrl: tab.favIconUrl,
+      title: tab.title,
+    },
+  });
+  if ('error' in res)
+    throw errorFromJson(res.error as JsonValue, undefined, ConnectError.from(res));
+
+  connectedSites[urlOrigin] = res.data.attitude;
```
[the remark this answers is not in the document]
grod220: This feedback doesn't look like it was addressed, please follow up with a PR on this one

PR #476, opened 2024-02-08
Fix a number of issues with how assets were displayed in the swap form

jessepinho:
This fixes a few UI issues I discovered when I added some exchange rates via `pcli tx position replicate...`

## Before
image
image

## After
image

image


apps/webapp/src/components/shared/input-token.tsx
```diff
@@ -48,7 +48,7 @@ export default function InputToken({
       </div>
 
       <div className='mt-[6px] flex items-center justify-between gap-2'>
-        <div className='flex items-start gap-1'>
+        <div className='flex items-start gap-1 truncate'>
```
jessepinho, line 51: `truncate` = `overflow-hidden text-ellipsis whitespace-nowrap`

packages/ui/components/ui/dialog.tsx
```diff
@@ -37,7 +37,35 @@ const DialogContent = React.forwardRef<
...
       className={cn(
-        'fixed left-[50%] top-[50%] z-50 grid w-full max-w-lg translate-x-[-50%] translate-y-[-50%] gap-4 rounded-lg bg-background shadow-lg duration-200 data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[state=closed]:slide-out-to-left-1/2 data-[state=closed]:slide-out-to-top-[48%] data-[state=open]:slide-in-from-left-1/2 data-[state=open]:slide-in-from-top-[48%] sm:rounded-lg md:w-full data-[state=closed]:fill-mode-forwards data-[state=open]:fill-mode-forwards',
+        'fixed',
+        'left-[50%]',
+        'top-[50%]',
+        'z-50',
+        'grid',
+        'w-full',
+        'max-w-lg',
+        'max-h-screen',
```
jessepinho, line 47: This ^ is the only class I added. The rest was just formatting the class list as a list of arguments, rather than a single string.

grod220 (approved):
Aha, this exposes another issue. We should not have lpnft's in that selector drop down box at all. We should think of the best way to filter these out 🤔

Issue #218, opened 2023-11-03
Wasm crate refactor


grod220:
We should split up the current Rust wasm crate into two:

A) penumbra-wasm
Should have only the `wasm-bindgen` functions that expose utility functions

B) penumbra-extension
Should have the penumbra-extension specific functionality here.

(Do these in later issues)
C̶o̶n̶s̶i̶d̶e̶r̶a̶t̶i̶o̶n̶s̶:̶
-̶ ̶I̶n̶d̶e̶x̶e̶d̶d̶b̶ ̶g̶l̶o̶b̶a̶l̶ ̶v̶s̶ ̶d̶e̶p̶e̶n̶d̶e̶n̶c̶y̶ ̶i̶n̶j̶e̶c̶t̶i̶o̶n̶
-̶ ̶N̶P̶M̶ ̶p̶u̶b̶l̶i̶s̶h̶i̶n̶g̶ ̶w̶o̶r̶k̶f̶l̶o̶w̶
-̶ ̶T̶e̶s̶t̶i̶n̶g̶ ̶s̶t̶r̶a̶t̶e̶g̶y̶

hdevalence:
I think the `penumbra-wasm` code should continue to live in the core protocol repo, and the `penumbra-extension` code should move into this repo.  This way, the parts of the extension that use typescript and that use Rust can be updated in lockstep.

grod220:
> the penumbra-extension code should move into this repo

We should think through this because there are significant trade offs.

### Current flow:

- Wasm crate lives in core repo
- Web code consumes crate npm package published on new tags. Mid-cycle npm publishing done via manual github action runs.
- Dev experience for changes require local compiling and pointing web monorepo npm dependencies to the build output on disk.
- Breaking changes in core require fixing wasm crate simultaneously. A kind of core-team maintenance commitment. Though not all things are caught (e.g. fee tiers implemented in pcli, but not wasm crate).

### Migrate to web monorepo

- Crate would be an additional package (would require to have a package.json that mapped build/lint commands from javascript to rust)
- Web code consumes compiled wasm not from crates.io, but from the local crate and local npm dependencies pointing to the build output.
- Penumbra core would be required to publish all crates to crates.io so it can be remotely consumed by relocated wasm crate. Published on tags.
- Core team members would not fix breaking changes in relocated wasm crate. Upon a new release, web team is responsible for pulling latest dependencies and making fixes (similar to spec changes now).

====

[...]

PR #478, opened 2024-02-08
refactor: native wasm integration 

TalDerei:
References https://github.com/penumbra-zone/web/issues/218


TalDerei:
@grod220 this represents preliminary work on porting the `penumbra-extension` to the web. Based on our discussion during the web sync, having local dependencies means we won't need to continue publishing NPM packages since local development would point to locally compiled dependencies. 

Regarding turborepo tooling considerations, I'm currently integrating the rust code into the build system and implementing sanity checking with sha256 hashing to verify we're using to the latest version.

grod220:
Instead of putting the crate inside `src/` next to the typescript code, think we can have two main folders alongside each other: `packages/wasm/web/` and `packages/wasm/crate`.

We should drag over some of those helpers in `penumbra/crates/wasm/publish/run.ts` to help us compile as well. Given taht compiling takes so long, we shouldn't need to do it on every `pnpm build`. We should have a hash of the files in `packages/wasm/crate` saved somewhere and if there is a diff, we should throw an error, requiring the dev to re-compile and generate a new hash.

turbocrime:
#492

turbocrime:
CI is broken. lint now depends on presence of wasm output (d.ts files) and i've configured that in turbo.json but something in CI is failing.

turbocrime:
also, since lint depends on wasm output, it would be cool if we could either

1) compile wasm once and forward that to other steps
2) have just one step that compiles wasm, then lints, then tests, then builds

TalDerei:
Currently, necessary rust toolchain configs are missing from our yml files which seem to be the source of the failing CI workflows.

TalDerei:
@turbocrime why are we nuking our existing CI? I was close to getting the CI to pass, and `.github/workflows/wasm.yml` was checking that the wasm crate builds cleanly and executed a full headless browser unit test which was removed.

TalDerei:
Making a note here. While pairing with L, we decided to move the CI workflows into a single yml file with multiple jobs since `turbo-build, turbo-lint, and turbo-test` depend on the cache contents from executing `turbo-compile`. This simplifies our workflow succinctly.

turbocrime:
:D

turbocrime (approved):

PR #481, opened 2024-02-08
Idb iterators

Valentine1898:


packages/router/src/grpc/view-protocol-server/notes.test.ts
```diff
@@ -15,39 +14,58 @@ import {
...
 
   test('empty request return no spend notes', async () => {
+
     const responses: NotesResponse[] = [];
     const req = new NotesRequest({});
     for await (const res of notes(req, mockCtx)) {
       responses.push(new NotesResponse(res));
     }
-    expect(responses.length).toBe(5);
+    expect(responses.length).toBe(6);
```
Valentine1898, line 68: records in testData array have always been 6, but this test was passed before, how?

packages/types/src/indexed-db.ts
```diff
@@ -54,10 +54,10 @@ export interface IndexedDbInterface {
...
   saveSpendableNote(note: SpendableNoteRecord): Promise<void>;
-  getAllSpendableNotes(): Promise<SpendableNoteRecord[]>;
+  iterateSpendableNotes(): AsyncGenerator<SpendableNoteRecord, void>;
   saveTransactionInfo(tx: TransactionInfo): Promise<void>;
   getTransactionInfo(txId: TransactionId): Promise<TransactionInfo | undefined>;
-  getAllTransactionInfo(): Promise<TransactionInfo[]>;
+  iterateTransactionInfo(): AsyncGenerator<TransactionInfo, void>;
   getAssetsMetadata(assetId: AssetId): Promise<Metadata | undefined>;
   saveAssetsMetadata(metadata: Metadata): Promise<void>;
   getAllAssetsMetadata(): Promise<Metadata[]>;
```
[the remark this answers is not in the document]
Valentine1898: Yes, I get it, this PR is still in progress, I decided to fix all the broken tests first, and that turned out to be not so easy

packages/router/src/grpc/view-protocol-server/notes.ts
```diff
@@ -8,25 +8,24 @@ import { Amount } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/nu
...
+    if (amountToSpend && assetId && !includeSpent) {
       const noteAmount = n.note?.value?.amount ?? new Amount();
       spent = addAmounts(spent, noteAmount);
-
-      return joinLoHiAmount(spent) <= joinLoHiAmount(amountToSpend);
-    })
-    .map(noteRecord => ({ noteRecord }));
+      if (joinLoHiAmount(spent) >= joinLoHiAmount(amountToSpend)) break;
+    }
+  }
```
Valentine1898, line 30: Fixed here

packages/router/src/grpc/custody/authorize.ts
```diff
@@ -18,7 +18,11 @@ import { viewTransactionPlan } from './view-transaction-plan';
 const getMetadataByAssetId = async (ctx: HandlerContext) => {
   const services = ctx.values.get(servicesCtx);
   const walletServices = await services.getWalletServices();
-  const assetsMetadata = await walletServices.indexedDb.getAllAssetsMetadata();
+
+  const assetsMetadata: Metadata[] = [];
+  for await (const metadata of walletServices.indexedDb.iterateAssetsMetadata()) {
+    assetsMetadata.push(metadata);
+  }
```
grod220, line 25: Not for this PR, but we should refactor `viewTransactionPlan()` to accept a parameter that allows for retrieving individual metadata async.

grod220 (approved):
Nice job!

packages/router/src/grpc/custody/authorize.ts
Valentine1898:
```typescript
/**
 * @todo As more asset types get used, the amount of asset metadata we store
 * will grow. Loading all the asset metadata into memory for the purpose of
 * compiling a transaction view may not be sustainable in the long term.
 * Eventually, we may want to scan through the transaction plan, extract all the
 * asset IDs in it, and then query just those from IndexedDB instead of grabbing
 * all of them.
 */
``` 
Looks like there's already a todo for that

Issue #428, opened 2024-01-30
Import Asset Registry from `Cosmos Asset List`


grod220:
The extension needs a way to import its own DenomMetadata definitions that refine those defined on-chain. In particular, the on-chain definitions won't have any denom units or other information about IBC assets. The extension needs to be able to import those from the [Cosmos Asset List](https://github.com/cosmos/chain-registry/tree/master#assetlists). This is a somewhat security-critical step, because it's how we present to users information about which assets are "real" and which are random other channels. This is also an opportunity, since we can, e.g., show source chain info in the icon or ticker.

See: https://github.com/penumbra-zone/web/issues/392 for discussion on this

hdevalence:
I'm not sure we need to import the entire cosmos asset list to start. None of those assets will correspond directly to assets on Penumbra, to start, since we'll have to define IBC mappings for all of them.

I think a good first go at this would be to:

- [ ] in JSON files, define "refined" `DenomMetadata` for assets on the testnet chain (we already have some JSON sitting around, from what I've seen in other PRs, it could move out into configuration files)
- [ ] when initializing the view server database, add a step that reads the configuration data and registers the contents as known assets

Then when syncing, when the view server sees an asset for the "first time", it will already have been registered with the configuration data's `DenomMetadata`.

The configuration data should be "refined" in that it should have fields filled in that aren't present in the on-chain metadata. For instance, the on-chain data never includes the "symbol", but the extension should be rendering assets via symbol, so the extension should be specifying `DenomMetadata` with the symbol set.

hdevalence:
As of https://github.com/penumbra-zone/penumbra/pull/3702, this exists in the protos, and at least the Rust implementation of `DenomMetadata` can parse the JSON from the cosmos asset list without problems. I would expect the same thing is true for the Buf generated code, though it might be necessary to configure the deserialization to ignore unknown fields.

PR #482, opened 2024-02-08
Pre-load db with local assets

grod220:
Closes https://github.com/penumbra-zone/web/issues/428

Our hard-coded assets list is a source of truth we want to pre-load our `ASSETS` table with. This PR does this at database instantiation.


packages/storage/src/indexed-db/indexed-db.test-data.ts
grod220, line 1: Test data overlaps with actual data in our `localAssets` list. Renamed to avoid this conflict.

hdevalence:
Can we have the known asset data stored in a distinct `.json` file that gets loaded by the extension on init? That way we have an easy source of truth for the extension's assets and we can write other tooling to periodically regenerate that JSON file.

grod220:
Currently that file is here: [packages/constants/assets.ts](https://github.com/penumbra-zone/web/blob/main/packages/constants/assets.ts), but it's saved as a typescript file. I figured we could change these by hand as we want to add new assets. However, if we anticipate future tooling to drop these assets in here as JSON, we should change this up a bit.

hdevalence:
Yes, we'll want to eventually have tooling that

1. reads data about chains and channels from the cosmos chain registry
2. transports token definitions along our preferred channels to create the ibc token definitions that will be on penumbra
3. writes out those modified definitions into our data format

For this reason I think it would be good to have the asset metadata be a JSON file that's accessed at runtime. This also means we'll be able to update it later, not just on view database init.

jessepinho (approved):
LGTM! Just confirming — incrementing the indexedDB version results in every web client's DB being blown away and recreated, right? (i.e., that's how we do data migrations?)

grod220:
> incrementing the indexedDB version results in every web client's DB being blown away and recreated, right? 

Yes, that's right. It's not really necessary for us to do migrations as our db can just re-sync. I incremented in this case so users get the pre-populated metadata in their dbs.

PR #484, opened 2024-02-08
Implement getters, per the RFC

### LINKED ISSUE #458: RFC: Determine how to handle type assertions at both compile- and run-time
Per our web check-in on Feb 5, 2024, we want a clean and expressive way to assert that particular properties exist on objects that come over the wire and are deserialized from Protobufs.

## The problem
We are currently using the classes generated by `bufbuild` to represent data types, even though those classes are designed for _in-transit_ data. This means that, even if we know that some data _should_ have certain properties in certain places, we're still treating all properties as optional, since that's how `bufbuild` generates classes.

Thus, throughout the codebase, we have to use ugly type guards everywhere ([example](https://github.com/penumbra-zone/web/pull/453#discussion_r1476332514)). For example, let's say we have an `<AddressViewComponent />` that renders an `AddressView`. `AddressView` may or may not have a `case` of `visible`; if it does, it may or may not have an `index.account` property, an `address` property, and/or a `walletId` property. Inside `<AddressViewComponent />`, we'll need to assert that `addressView.case === 'visible' && !!addressView.value.address && !!addressView.value.index.account` before we can e.g., render the text `Account #1` here:
image

This problem gets even worse if you have multiple levels of nesting of Protobuf types. A `TransactionView`, for instance, contains underneath it `ActionView`s (which themselves contain `SpendView`s, `OutputView`s, etc.), `MemoView`s, etc. There is a huge number of possible branches that one has to deal with based on whether certain properties are present vs. missing at each level of the object structure.

And the worst part is that, in many cases, we actually know via our business logic that certain properties must already exist — i.e., that it isn't actually possible for certain properties to be missing. For example, when the approval dialog opens for a "Send" transaction, we know that the current user is the one creating the transaction, and thus that the `memoView` must have a `case` of `visible`. Any other possibility wouldn't make sense. And yet we still have to act as though all properties are optional, like we don't know what the `case` is, etc.

## Requirements for solving this problem
- A **minimum of repeated code**. We don't want to have to repeat imperative type guards at various levels of the component tree.
- A **minimum of imperative code**. An expressive, declarative option would be preferable.
- A **marriage of compile- and run-time type safety.** We don't want to have to e.g., use `!` to tell the compiler, "Trust us, we know this property exists." Rather, we should have a way of run-time checking our objects to make sure they're in the shape we expect; and the run-time checking should result in compile-time type assertions so that we can fully trust our types.

### Utilize `PartialMessage` + type predicate functions
`PartialMessage` is a type provided by `bufbuild` that represents an all-optional version of a Protobuf message. That is, all fields are optional, recursively to the deepest-nested object.

We could create TypeScript utility interfaces that assert that some deeply nested properties _do_ exist, and then create type predicate functions with those, like so:

```TS

interface WithAccountIndex {
  addressView: {
    case: 'decoded';
    value: {
      index: {
        account: number;
      };
    };
  };
}

const hasAccountIndex = (
  addressView: PartialMessage<AddressView>,
): addressView is PartialMessage<AddressView> & WithAccountIndex => {
  return (
    addressView.addressView?.case === 'decoded' &&
    typeof addressView.addressView.value.index?.account !== 'undefined'
  );
};
```

Then, in our components, instead of needing to check the `case` and the existence of the account index, we'd just:

```TSX
<div>
  {hasAccountIndex(addressView) && <span>{addressView.addressView.value.index.account /* <-- no need for `?`s or `!`s */}</span>}
</div>
```

(I'd suggest we neatly organize where all these utility interfaces and type predicate functions live, of course.)

### Use Zod for compile- and run-time type validation
This is a pretty similar solution to the one above, but using Zod to more expressively combine the run-time and compile-time checking:

```TS
export const hasAccountIndex = isType(
  z.object({
    addressView: z.object({
      case: z.literal('decoded'),
      value: z.object({
        index: z.object({
          account: z.number(),
        }),
      }),
    }),
  }),
);
```

(`isType` is defined [here](https://github.com/penumbra-zone/web/pull/453/files#diff-6055854e07829e5ebccce89e07268f49e452eba337899c35199d7ce726f0eb4bR38-R69).)

Same as above, in our components, we'd use this helper like so:

```TSX
<div>
  {hasAccountIndex(addressView) && <span>{addressView.addressView.value.index.account /* <-- no need for `?`s or `!`s */}</span>}
</div>
```

### Create a parallel set of types or classes, with required properties
The shortcoming of both of the above solutions is that they require using type guards at all. One could argue that there should never be, e.g., an `AddressView` with a defined `index.account` but an undefined `address`. If that's true, then perhaps we should define classes for specific use cases, such as `CurrentUserAddressView` which _always_ has both an `index.account` and an `address`, since we know that we'll always have access to that for addresses belonging to the current user. 

On the other hand, that solution isn't really that much more elegant than the ones above, and could get extremely messy when it comes to deeply nested objects. A `TransactionView`, for example, might have some `CurrentUserAddressView`s nested underneath it, but also some `AddressView`s belonging to other users. Would we have to define a type of `TransactionView` that indicates that it contains a `CurrentUserAddressView` as well as some opaque `AddressView`s, and possibly other combinations of object types underneath it? What would the name of such a class be? `TransactionViewWithCurrentUserAddressInMemoButOtherUserInOutputAction` or something crazy like that? I don't think is really a feasible option, unless I'm thinking incorrectly about how such an object would be defined to enable access to nested properties without tons of type guards and safe navigation operators.

## Outcome
Create an `ADRs` directory in the `web` repo to store our ADRs for future reference, and include a summary of the outcome of this RFC.

## Commenters
Please feel free to flesh out the solutions I listed above if I'm scant on detail, or add anything else you think is relevant to this issue.

## Notes to self
- If we end up wanting to use named functions, [this](https://dev.to/tmikeschu/dynamically-assigning-a-function-name-in-javascript-2d70) should come in handy.

jessepinho:
This PR wraps up what we discussed in today's web weekly meeting. This introduces a new pattern that we'll have to build some muscle memory with, but I think it makes handling of deeply nested properties a bit more elegant.

## In this PR
- Create a `createGetter` function that creates a getter with some nifty features.
- Create a few getters for the `assembleSwapRequest` use case as a proof of concept.
- Write an ADR and a README for getters. There's some overlap between the two, but the ADR is more focused on _why_ we chose getters, while the README is focused on _how_ to use them.

## A note on syntax
It might feel a little funny to call a function that returns a function, as we do e.g. [here](https://github.com/penumbra-zone/web/pull/484/files#diff-b26f9cb2a251b58ff9dcfb22a7f2f375617e1e6742814729c92398f299cf4e43R110):
```
getMetadata.pipe(getDisplayDenomExponent).orThrow()(assetIn.value)
                                                 ^^^^^^^^^^^^^^^^^
```

The reason for this is chainability. `.orThrow()` shouldn't be a getter itself, but rather should _return_ a getter, so that it can be piped (e.g., `getMetadata.orThrow().pipe(...)`). It also makes it easier to assign a getter to a variable that can then be called like a "regular" function, like this:

```tsx
const getAddressIndexFromMemoView = getAddress
  .orThrow('The address is missing!')
  .pipe(getAddressIndex.orThrow('The address index is missing!'));

<div>
  <span>{getAddressIndexFromMemoView(memoView)</span>
</div>
```

Closes #458


apps/webapp/src/state/swap.ts
```diff
@@ -90,26 +96,20 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
-        targetAsset: assetOut.penumbraAssetId,
+        targetAsset: getAssetId.orThrow(assetOut),
         value: {
           amount: toBaseUnit(
             BigNumber(amount),
-            getDisplayDenomExponent(assetIn.value.valueView.value.metadata),
+            getMetadata.pipe(getDisplayDenomExponent).orThrow(assetIn.value),
           ),
-          assetId: assetIn.value.valueView.value.metadata.penumbraAssetId,
+          assetId: getMetadata.pipe(getAssetId).orThrow(assetIn.value),
```
jessepinho, line 110: Reviewers, if you don't like inlining the `pipe`ing here, I can also extract these above, e.g., 

```ts
const getAssetIdFromValueView = getMetadata.pipe(getAssetId).orThrow

return new TransactionPlannerRequest({
  swaps: [
    {
      value: {
        // ...
        assetId: getAssetIdFromValueView(assetIn.value)
      },
    }
})
```

apps/webapp/src/state/swap.ts
```diff
@@ -90,26 +96,22 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
   return new TransactionPlannerRequest({
     swaps: [
       {
-        targetAsset: assetOut.penumbraAssetId,
+        targetAsset: getAssetId.orThrow()(assetOut),
         value: {
           amount: toBaseUnit(
             BigNumber(amount),
-            getDisplayDenomExponent(assetIn.value.valueView.value.metadata),
+            getMetadata.pipe(getDisplayDenomExponent).orThrow()(assetIn.value),
```
grod220, line 110: I think we can move all the `pipe`'s to `types/src/getters/value-view.ts` and have a helper function there. If not, I suspect we'll need to use redundant pipes in different places.

packages/types/src/getters/utils/create-getter.ts
```diff
@@ -0,0 +1,23 @@
...
+  ) => AllowsUndefined extends true ? TargetType | undefined : TargetType,
+): Getter<SourceType, TargetType, AllowsUndefined> => {
+  const getter: Getter<SourceType, TargetType, AllowsUndefined> = value => getterFunction(value);
+
+  getter.orThrow = errorMessage => {
+    const getOrThrow = createGetter<SourceType, TargetType, false>(value => {
+      const result = getterFunction(value);
+      if (typeof result === 'undefined') throw new Error(errorMessage);
+      return result;
+    });
```
grod220, line 15: Instead of writing our own errors, think we have it write its own:

```typescript
  getter.orThrow = () => {
    return createGetter<SourceType, TargetType, false>(value => {
      const result = getterFunction(value);
      if (typeof result === 'undefined') {
        const errorMessage = `${getterFunction.name} failed to extract from ${JSON.stringify(value)}`;
        throw new Error(errorMessage);
      }
      return result;
    });
  };
```

grod220 (commented):
Interesting functional patterns here 👀

apps/webapp/src/state/swap.ts
jessepinho: Good call — [done](https://github.com/penumbra-zone/web/pull/484/commits/fa259d239a8e592e1d2c24e17dcd084cc5fa1d6f).

apps/webapp/src/state/swap.ts
```diff
@@ -90,26 +96,22 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
-  if (assetIn?.value.valueView.case !== 'knownAssetId') throw new Error('unknown denom selected');
-  if (!assetIn.value.valueView.value.metadata?.penumbraAssetId)
-    throw new Error('missing metadata for assetIn');
-  if (assetIn.address.addressView.case !== 'decoded')
-    throw new Error('address in view is not decoded');
-  if (!assetIn.address.addressView.value.index) throw new Error('No index for assetIn address');
-  if (assetOut?.penumbraAssetId === undefined) throw new Error('assetOut has no asset id');
+  if (!assetIn) throw new Error('`assetIn` was undefined');
+
+  const addressIndex = getAddressIndex.orThrow('No index for assetIn address')(assetIn.address);
```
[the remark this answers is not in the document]
jessepinho: Definitely an improvement. Done in [this commit](https://github.com/penumbra-zone/web/pull/484/commits/16c294a79be22d442a310f4b8962d82ae1431f39).

packages/types/src/getters/utils/create-getter.ts
jessepinho: Good call. I added the second half of that error message [here](https://github.com/penumbra-zone/web/pull/484/commits/16c294a79be22d442a310f4b8962d82ae1431f39#diff-0aafb5759e68a96372f36cbab5c7a0347e7a39227ef6b3f3905796047f42a206R12). Note that the getter function has no name, because `createGetter` takes an anonymous function:
```ts
const getFirstName = createGetter((employee: Employee) => employee.firstName);
                               // ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                               // anonymous function
```

We could require that people define function names:
```ts
const getFirstName = createGetter(function getFirstName(employee: Employee) { return employee.firstName) });
```

But that doesn't feel as ergonomic. Thoughts?

packages/types/src/getters/utils/create-getter.ts
turbocrime, line 1: i like this

packages/types/src/getters/utils/create-getter.ts
```diff
@@ -0,0 +1,35 @@
...
+    const result = getterFunction(value);
+    if (typeof result === 'undefined' && !optional) {
+      const errorMessage = `Failed to extract from ${JSON.stringify(value)}`;
+      throw new GetterMissingValueError(errorMessage);
+    }
+    return result as Optional extends true ? TargetType | undefined : TargetType;
+  };
+
+  getter.optional = () =>
+    createGetter<SourceType, TargetType, true>(value => getterFunction(value), true);
```
turbocrime, line 18: it makes sense to me to write Getter as a class or a more detailed object literal, so we can createGetter optional on access, instead of requiring an empty call. this would turn the `.optional()(etc)` syntax into just `.optional(etc)`

```ts
get optional() {
    return createGetter<SourceType, TargetType, true>(value => getterFunction(value), true);
}
```

packages/types/src/getters/utils/getter.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Pipes the output of this getter to another getter or getter function.
+   *
+   * @example
+   * ```ts
+   * // This will throw if any step along the way returns `undefined`.
+   * const assetId1 = getMetadata.pipe(getAssetId).pipe(getInner).orThrow()(valueView);
+   * // This will throw only if the asset ID specifically is missing.
+   * const assetId2 = getMetadata.pipe(getAssetId.orThrow()).pipe(getInner)(valueView);
```
turbocrime, line 41: will we ever actually need a mid-pipe throw like this?

turbocrime:
instead of the `.pipe(etc)` syntax we should return a getter for each field on the return of any getter

packages/types/src/getters/utils/create-getter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    state: 'California',
+  },
+};
+
+const getFirstName = createGetter((employee?: Employee) => employee?.firstName);
+const getLastName = createGetter((employee?: Employee) => employee?.lastName);
+const getAddress = createGetter((employee?: Employee) => employee?.address);
+const getCity = createGetter((address?: Address) => address?.city);
+const getCountry = createGetter((address?: Address) => address?.country);
+const getFirstLetter = createGetter((value?: string) => value?.[0]);
```
turbocrime, line 29: simple getters like this could be automatically created and attached to the return of a getter, eliminating some of the need for pipe syntax

packages/types/src/getters/utils/create-getter.ts
```diff
@@ -0,0 +1,35 @@
+import { Getter } from './getter';
+import { GetterMissingValueError } from './getter-missing-value-error';
+
+export const createGetter = <SourceType, TargetType, Optional extends boolean = false>(
+  getterFunction: (value: SourceType | undefined) => TargetType | undefined,
+  optional?: Optional,
+): Getter<SourceType, TargetType, Optional> => {
+  const getter: Getter<SourceType, TargetType, Optional> = value => {
+    const result = getterFunction(value);
+    if (typeof result === 'undefined' && !optional) {
```
grod220, line 10: This is the equivalent right?

```diff
- typeof result === 'undefined'
+ result === undefined

File: packages/types/src/getters/utils/create-getter.ts [hunk 9]
```diff
@@ -0,0 +1,35 @@
...
+  getter.optional = () =>
+    createGetter<SourceType, TargetType, true>(value => getterFunction(value), true);
+
+  getter.pipe = <PipedTargetType = unknown>(
+    next: Getter<TargetType, PipedTargetType, Optional>,
+  ) => {
+    return createGetter<SourceType, PipedTargetType, Optional>(value => {
+      try {
+        const result = next(getterFunction(value));
+        return result;
```
grod220, line 26: Can inline:

```typescript
return next(getterFunction(value));
```

packages/types/src/getters/utils/create-getter.ts
grod220: Ah, I see, it's a getter syntax trick

packages/types/src/getters/utils/getter.ts
grod220: Think the `orThrow()` syntax is from previous commits and needs updating

grod220 (approved):
Nice job overall! We should open source this to the other typscript protobuf folks :)

packages/types/src/getters/utils/getter.ts
jessepinho: Ah, good catch — fixed.

PR #490, opened 2024-02-10
webapp dev server ssl

### LINKED ISSUE #284: Local dev server SSL
would like to restrict extension connections to TLS-only, dev server blocks it

needs vite config

turbocrime:
vite plugin provides a simple self-signed cert for localhost ssl. this is nice because

- some browser api are only available on ssl pages
- simplifies dev when restricting transport connections to ssl pages

i don't think this interferes with any CI or tests

fixes #284

### DIFF
--- a/apps/webapp/package.json
+++ b/apps/webapp/package.json
@@ -34,6 +34,7 @@
     "@types/react": "^18.2.55",
     "@types/react-dom": "^18.2.19",
     "@types/react-helmet": "^6.1.11",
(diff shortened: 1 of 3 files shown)


turbocrime:
aside: note the big pnpm lock diff. someone doesn't have buf registry configured and keeps stripping the registry information from the lockfile

grod220 (approved):
I suppose we'll have to keep an eye out for subsequent PRs to see who's doesn't have the registry installed

PR #492, opened 2024-02-12
wasm refactor suggestions

turbocrime:
package structure, turbo, vitest without mocking wasm


TalDerei:
Is there a reason for not differentiating between `@penumbra-zone/wasm-bundler` and `@penumbra-zone/wasm-nodejs` for dev and prod builds? Instead of `../crate/pkg`, can we alias that to something cleaner like `@penumbra-zone/wasm` instead of using a relative path?

turbocrime:
differentiating between bundler and nodejs is unnecessary. i added some vitest config that understands how to load the wasm, and soon they'll support it natively without the plugin. had to fix a bunch of tests now that we're actually executing wasm instead of mocking it

TalDerei:
all unit tests are passing on my end

Screenshot 2024-02-12 at 1 08 54 PM

TalDerei (approved):

turbocrime:
it seems like vitest has some longstanding and unclear issues with 'browser mode' and its utils package, somehow interacting with pnpm, that are causing my error. not sure what to do about it

PR #493, opened 2024-02-12
Tweaks to ESLint + Tailwind integration

jessepinho:
Currently, VSCode complains when using any of our [custom theme colors](https://github.com/penumbra-zone/web/blob/ee9e02b/packages/ui/tailwind.config.js#L49-L52) in class names:

image

This is because of [an issue with the VSCode ESLint extension](https://github.com/microsoft/vscode-eslint/issues/1706#issuecomment-1916389417) when used in a monorepo.

To fix the issue, I extracted Tailwind config into its own package in the monorepo, and pointed our ESLint config at it (per the suggestion in the link above). I also configured ESLint to lint `cn()` calls, and fixed a couple places where we had invalid class names.

## In this PR
- Extract Tailwind config into `@penumbra-zone/tailwind-config`
- Tell our ESLint config about our Tailwind config, for better VSCode/etc. integration
- Add `cn()` calls to the list of function calls that `tailwindcss/no-custom-classname` should lint.
- Fix a few places where we had invalid classnames.


packages/tailwind-config/index.js
```diff
@@ -0,0 +1,149 @@
+import tailwindCssAnimatePlugin from 'tailwindcss-animate';
```
jessepinho, line 1: Nothing new here — this file is just moved over to this new package from `packages/ui`.

packages/ui/tailwind.config.js
```diff
@@ -1,149 +1 @@
-import tailwindCssAnimatePlugin from 'tailwindcss-animate';
```
jessepinho, line 1: Moved to the new `packages/tailwind-config` package

packages/ui/components/ui/select.tsx
```diff
@@ -26,10 +26,7 @@ const SelectTrigger = React.forwardRef<
     {children}
     <SelectPrimitive.Icon asChild>
       <ChevronDownIcon
-        className={cn(
-          'h-6 w-6 text-muted transition-all duration-500 ease-bounce',
```
jessepinho, line 30: ESLint was complaining that `ease-bounce` isn't a valid class — should it be [`animate-bounce`](https://github.com/jamiebuilds/tailwindcss-animate?tab=readme-ov-file#basic-usage)?

packages/ui/components/ui/navigation-menu.tsx
```diff
@@ -77,7 +77,7 @@ const NavigationMenuViewport = React.forwardRef<
   <div className={cn('absolute left-0 top-full flex justify-center')}>
     <NavigationMenuPrimitive.Viewport
       className={cn(
-        'origin-top-center relative mt-1.5  h-[var(--radix-navigation-menu-viewport-height)] w-full overflow-hidden rounded-md border bg-background text-popover-foreground shadow-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-90 md:w-[var(--radix-navigation-menu-viewport-width)]',
+        'origin-top relative mt-1.5  h-[var(--radix-navigation-menu-viewport-height)] w-full overflow-hidden rounded-md border bg-background text-popover-foreground shadow-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-90 md:w-[var(--radix-navigation-menu-viewport-width)]',
```
jessepinho, line 80: [`origin-top-center` doesn't exist](https://tailwindcss.com/docs/transform-origin). I'm guessing the intent was to use `origin-top`.

packages/eslint-config-custom/index.js
```diff
@@ -30,6 +30,9 @@ module.exports = {
     'tailwindcss/no-custom-classname': [
       'error',
       {
+        // All of these callees are the Tailwind defaults, except `cn`, which is
+        // our own custom helper.
+        callees: ['classnames', 'clsx', 'cn', 'ctl', 'cva', 'tv'],
```
jessepinho, line 35: This is one of the main changes to this PR: adding `cn` to the list of `callees` ([docs](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-custom-classname.md#callees-default-classnames-clsx-ctl-cva-tv)). This ensures that ESLint will still check our class names when they're wrapped in a `cn()` call.

packages/eslint-config-custom/index.js
```diff
@@ -50,5 +53,11 @@ module.exports = {
   ignorePatterns: ['dist/*'],
   settings: {
     react: { version: 'detect' },
+    tailwindcss: {
+      // Ensures that ESLint knows about our Tailwind config, such as color
+      // names, etc. Without this, ESLint integrations (such as with the VSCode
+      // extension) break.
+      config: require.resolve('@penumbra-zone/tailwind-config'),
```
jessepinho, line 60: And this is one of the other main changes: ensuring ESLint knows about our Tailwind config, which makes for better integration with VSCode.

packages/ui/components.json
```diff
@@ -3,7 +3,7 @@
   "style": "default",
   "rsc": true,
   "tailwind": {
-    "config": "tailwind.config.js",
+    "config": "@penumbra-zone/tailwind-config",
```
jessepinho, line 6: I actually don't know how this setting works or what it does, so I'm not sure how to test that this change worked...

turbocrime (approved):
typescript-eslint parser releases a new version tomorrow, and would enable us to switch to new eslint flat config

edit: a few hours ago, actually

go ahead with this but i'd like a larger reconfig towards that, because it supports large projects with a single top-level eslint file

jessepinho:
@turbocrime good find. Mind making a ticket for that?

Issue #423, opened 2024-01-29
Dedicated unclaimed swaps box


grod220:

hdevalence:
I feel somewhat strongly that there shouldn't be UI for this if at all possible. Can we do the right thing automatically?

grod220:
I should have filled out this with more detail (instead of using the tracking issue create shortcut). 99% of the cases should be covered here: https://github.com/penumbra-zone/web/issues/420, where it's automatic.

The intent of this issue was for the user that (for whatever reason) has unclaimed swaps present. If we detect the case, we'd show a box giving the user the ability to continue where they left off. I'm afraid if we don't have this, there will be that odd occasion in discord where a user is requesting to be rescued.

Or would you say this case is so improbable, we should disregard it?

PR #498, opened 2024-02-13
Add recovery box for unclaimed swaps

grod220:
Closes: https://github.com/penumbra-zone/web/issues/423

On page load, if an unclaimed swap is detected, the page will show an `Unclaimed Swaps` box where the user can resume the second step of their swap. It's quite unlikely users will get into this state, but it is possible (e.g. network errs on swap claim). This box helps them recover from that state.

https://github.com/penumbra-zone/web/assets/16624263/1c145107-32fc-4e2a-aa9e-462448752d74


apps/webapp/src/state/swap.ts
```diff
@@ -127,24 +115,25 @@ const assembleSwapRequest = async ({ assetIn, amount, assetOut }: SwapSlice) =>
...
-  return new TransactionPlannerRequest({
+export const issueSwapClaim = async (swapCommitment: StateCommitment) => {
+  const swapClaimReq = new TransactionPlannerRequest({
     swapClaims: [
       {
         swapCommitment,
       },
     ],
   });
+  const swapClaimTx = await planWitnessBuildBroadcast(swapClaimReq, {
```
jessepinho, line 126: When rebasing, this will need to use the new helpers @turbocrime split up in the helpers file.

apps/webapp/src/components/swap/unclaimed-swaps.tsx
```diff
@@ -0,0 +1,75 @@
...
+              <span>↔</span>
+              <AssetIcon metadata={asset2} />
+            </div>
+          </div>
+          <Button
+            className='w-20'
+            onClick={() => void claimSwap(swap, toast, setInProgress, revalidate)}
+            disabled={inProgress}
+          >
+            Claim
```
jessepinho, line 69: What do you think of:
```suggestion
            {inProgress ? 'Claiming' : 'Claim'}
```

I tried this locally, where I had two unclaimed swaps. Both buttons got greyed out and when I pressed "Claim," so I was wondering if it was claiming both of them simultaneously — although it became clear once it finished that only one of them was being claimed.

For that matter, do _all_ Claim buttons need to be disabled while any _one_ swap is being claimed? Can't we just disable the button next to the swap that's currently in progress, so that the user can click a bunch and have a bunch of them go in parallel? (We could keep track of which ones are in progress using e.g., the base 64-encoded swap commitment bytes as their IDs?)

apps/webapp/src/components/swap/unclaimed-swaps.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <p className='bg-text-linear bg-clip-text font-headline text-xl font-semibold leading-[30px] text-transparent md:text-2xl md:font-bold md:leading-9'>
+        Unclaimed Swaps
+      </p>
+      <p className='text-gray-400'>
+        Swaps on Penumbra are a two step process. The first transaction issues the request and the
+        second claims the result of the swap action. For some reason, these second transactions were
+        not completed. Claim them!
+      </p>
+      {unclaimedSwaps.map(({ swap, asset1, asset2 }, i) => (
+        <div key={i} className='mt-4 flex justify-between'>
```
jessepinho, line 55: It's generally a no-no to use an array index as the React key, as this can lead to rendering bugs. Could you try encoding the swap commitment to base64 and using that?

packages/types/src/getters/transaction.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Swap } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/dex/v1/dex_pb';
+
+const getSwap = createGetter(
+  (transaction?: Transaction) =>
+    transaction?.body?.actions.find(action => action.action.case === 'swap')?.action.value as
+      | Swap
+      | undefined,
+);
+
+const getCommitment = createGetter((swap?: Swap) => swap?.body?.payload?.commitment);
```
jessepinho, line 12: I meant to put this in the README for getters, but like with the translators, I was trying to organize the files by the type that you pass to the getter. So, all getters in this file should accept a `Transaction` as their argument. Since `getCommitment` accepts a `Swap`, it should be in a file called `swap.ts`. (Same goes for the various types in the `swaps.ts` file above — `getSwapRecordCommitment()` should go in a file called `swap-record.ts` since it accepts a `SwapRecord`, etc.)

Not a blocker — I'd leave this up to you. I just would personally keep it organized that way because we'll probably end up with dozens of getters before long, and we'll always know where to look for them if they're organized by the type they accept as their argument.

jessepinho (approved):
Looks great. Turns out I had some unclaimed swaps locally that this let me clean up :D 

Couple notes, but no blockers.

apps/webapp/src/components/swap/unclaimed-swaps.tsx
grod220: Yeah, this is probably a better user experience. Will create a new slice for this logic.

apps/webapp/src/state/swap.ts
```diff
@@ -55,51 +54,32 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
-
-        // start to build swap claim
-        const swapCommitment = getSwapCommitment(swapTx);
-
-        const swapClaimPlan = await plan(assembleSwapClaimRequest(swapCommitment));
-        const swapClaimTx = await witnessBuild(update, { transactionPlan: swapClaimPlan });
+        await broadcast(update, { awaitDetection: true, transaction: swapTx });
 
-        // after swap broadcast success, emit the claim
-        await swapBroadcast;
```
grod220, line 83: In order to plan a swap claim, wasm needs access to the swap record in our db. So this will throw.

grod220:
Fyi, Toasts don't seem to be working as expected 🤔 . However, it looks like we have a few open PRs about refactoring it. So will revisit this later.

PR #500, opened 2024-02-13
Don't swallow errors thrown in streaming RPC methods

### LINKED ISSUE #437: Stream error propagation
At the moment, error messages do not propagate for streaming requests. To test:

- In `view-protocol-server/balances.ts` throw an error to the top of the function (e.g. `throw new Error('Error xyz');`)
- Likely will need to add `// @ts-expect-error xyz` to a few lines below
- Reload extension
- Load dashboard page in web app
- Error will read `ConnectError: [canceled] ConnectError: [internal] internal error` instead of error thrown in `balances.ts`

jessepinho:
Previously, `rethrowImplErrors` was only catching errors thrown by view service methods with a `kind` of `MethodKind.Unary`. This meant that, for `MethodKind.ServerStreaming` methods, any errors thrown would get swallowed by the ConnectRPC method, and turned into a generic, unhelpful "internal error". (See [this ticket](https://github.com/penumbra-zone/web/issues/437).)

This PR fixes that by extending `rethrowImplErrors` to handle server streaming methods.

## Manual testing
1. Update the `balances` impl to throw on its first line:
    image
2. Reload the extension.
3. Visit the web app.

### Before
image

### After
image

Closes #437.

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -84,8 +84,9 @@ const rpcImpls = [
   [
     CustodyService,
     /**
-     * @todo Remove `as` below once `custodyImpl` implemented the entire custody
-     * service. It's just there temporarily to make `rethrowImplErrors` happy.
+     * @todo Remove `as` below once `custodyImpl` has implemented the entire
+     * custody service. It's just there temporarily to make `rethrowImplErrors`
+     * happy.
      */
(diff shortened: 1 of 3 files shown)


turbocrime (approved):
rebased, i tested without 16e475810abec37d838edf7e67a1ba7e7ccfed2e and it does seem to throw those properly so i'm satisfied.

merge if you'd like i'm going to rebase my streaming fix on this

PR #503, opened 2024-02-13
407 add tests for swapbycommitmentrequest

### LINKED ISSUE #407: Add tests for SwapByCommitmentRequest

Valentine1898:
close #407

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
+++ b/packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
@@ -1,10 +1,7 @@
 import type { Impl } from '.';
 import { servicesCtx } from '../../ctx';
 
-import {
-  NoteByCommitmentResponse,
-  SpendableNoteRecord,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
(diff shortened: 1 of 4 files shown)


packages/router/src/grpc/view-protocol-server/note-by-commitment.ts
```diff
@@ -16,18 +13,13 @@ export const noteByCommitment: Impl['noteByCommitment'] = async (req, ctx) => {
...
-      response.spendableNote = note;
-      break;
+  if (req.awaitDetection) {
+    for await (const update of indexedDb.subscribe('SPENDABLE_NOTES')) {
+      const spendableNote = SpendableNoteRecord.fromJson(update.value);
+      if (spendableNote.noteCommitment?.equals(req.noteCommitment)) return { spendableNote };
     }
   }
-  return response;
+  throw new ConnectError('Note not found', Code.NotFound);
```
Valentine1898, line 24: I noticed that `swapByCommitment` and `noteByCommitment` differently handled the case when `awaitDetection` is true but no record was detected. 
I think we should throw error rather than an return empty response in such a case

grod220 (approved):

PR #504, opened 2024-02-13
Keep IDB queries awake

turbocrime:
~~This didn't actually solve the problem, because it turns out ConnectRPC will use await internally, and break the IDB transaction.~~

~~But, it involves more robust rethrow logic to bypass ConnectRPC's obscuring rethrow.~~

~~Removed the rethrow wrapper in service-worker.ts and just wrapped every impl with a try/catch.~~

~~May be able to address the IDB issue more fundamentally by rewriting IDB queries as streams.~~

# edit

This is is now a complete solution

rebased onto #500 which provides a better rethrow wrapper

now using a ReadableStream to iterate idb, which provides buffering and async iteration while allowing IDB to perceive its caller as remaining active, while still allowing free async behavior outside the stream

https://github.com/penumbra-zone/web/assets/134443988/2955a70a-5db6-4c84-8bf1-ae8db268ad81

### DIFF
--- a/packages/storage/src/indexed-db/stream.ts
+++ b/packages/storage/src/indexed-db/stream.ts
@@ -1,0 +1,24 @@
+import { AnyMessage, JsonValue, Message, MessageType } from '@bufbuild/protobuf';
+import { PenumbraDb, PenumbraStoreNames } from '@penumbra-zone/types';
+import { IDBPCursorWithValue } from 'idb';
+
+export class IdbCursorSource<N extends PenumbraStoreNames, T extends Message<T> = AnyMessage>
+  implements UnderlyingDefaultSource<T>
+{
+  constructor(
+    private cursor: Promise<IDBPCursorWithValue<PenumbraDb, [N], N> | null>,
+    private messageType: MessageType<T>,
+  ) {}
+
+  start(cont: ReadableStreamDefaultController<T>) {
+    // immediately stream as fast as possible, to prevent idb from closing the tx
+    void (async () => {
+      let cursor = await this.cursor;
+      while (cursor) {
+        cont.enqueue(this.messageType.fromJson(cursor.value as JsonValue));
+        cursor = await cursor.continue();
+      }
+      cont.close();
+    })();
+  }
+}
--- a/packages/router/src/grpc/view-protocol-server/assets.ts
+++ b/packages/router/src/grpc/view-protocol-server/assets.ts
@@ -1,7 +1,6 @@
 import type { Impl } from '.';
 import { servicesCtx } from '../../ctx';
 import { assetPatterns } from '@penumbra-zone/constants';
-import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 
 export const assets: Impl['assets'] = async function* (req, ctx) {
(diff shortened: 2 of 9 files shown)


jessepinho (approved):

grod220:
If this doesn't solve our IDB issue, I'm not sure why we'd want to do this. It feels like boilerplate that should be abstracted away at a different level automatically.

turbocrime:
> If this doesn't solve our IDB issue, I'm not sure why we'd want to do this. It feels like boilerplate that should be abstracted away at a different level automatically.

it's difficult to do automatically in a robust way. we've had a few iterations of that, and repeatedly discovered some shortcoming or another. the naive approach is robust, it's very few lines per method, and simple to understand.

jessepinho:
@grod220 I'm currently working on a more robust version of [`rethrowImplErrors`](https://github.com/penumbra-zone/web/pull/504/files#diff-b9be68cac5d37d21289e20d8d7fcb88d993de52b16e69a6500255f3fde092ae0L81-L109) that accounts for streaming methods too. Still in progress...

jessepinho:
PR for the error-swallowing issue here: https://github.com/penumbra-zone/web/pull/500

jessepinho:
@turbocrime if you don't rethrow errors with `ConnectError.from()`, does your fix still work? If so, can you remove those changes in favor of #500?

turbocrime:
okay, ReadableSteam fix works. i'll rebase onto #500

packages/storage/src/indexed-db/stream.ts
```diff
@@ -0,0 +1,24 @@
...
+    // immediately stream as fast as possible, to prevent idb from closing the tx
+    void (async () => {
+      let cursor = await this.cursor;
+      while (cursor) {
+        cont.enqueue(this.messageType.fromJson(cursor.value as JsonValue));
+        cursor = await cursor.continue();
+      }
+      cont.close();
+    })();
+  }
```
grod220, line 23: Ah, clever work around. But does this mean that all of the values are loaded up in the stream (aka in memory) all at once? In the end, is this the equivalent of a getAll()?

packages/storage/src/indexed-db/stream.ts
```diff
@@ -0,0 +1,24 @@
...
+export class IdbCursorSource<N extends PenumbraStoreNames, T extends Message<T> = AnyMessage>
+  implements UnderlyingDefaultSource<T>
+{
+  constructor(
+    private cursor: Promise<IDBPCursorWithValue<PenumbraDb, [N], N> | null>,
+    private messageType: MessageType<T>,
+  ) {}
+
+  start(cont: ReadableStreamDefaultController<T>) {
+    // immediately stream as fast as possible, to prevent idb from closing the tx
```
grod220, line 14: We should probably add a bit more documentation here and link to: https://github.com/jakearchibald/idb?tab=readme-ov-file#transaction-lifetime

packages/storage/src/indexed-db/stream.ts
Valentine1898: If this is really the equivalent of `getAll()`, then we are not optimally filtering. In this case we should filter inside the `while (cursor)` loop to avoid reading all data.

Issue #421, opened 2024-01-29
Estimate swap out amount


grod220:

hdevalence:
We shouldn't do this automatically, it should be an explicit user action (with a "simulate" button or similar).

Why? because

- estimating the output requires making a network request that leaks the user's trade intent to their rpc
- the rpc may or may not even support simulation

PR #511, opened 2024-02-14
Simulate swap output

grod220:
Closes: https://github.com/penumbra-zone/web/issues/421

https://github.com/penumbra-zone/web/assets/16624263/cf2955f1-dedc-4c77-b045-8e32942cea11


apps/webapp/src/components/shared/select-token-modal.tsx
```diff
@@ -65,9 +64,8 @@ export default function SelectTokenModal({
             </div>
             <div className='flex flex-col gap-2'>
               {balances.map((b, i) => {
-                const index =
-                  b.address.addressView.case === 'decoded' &&
-                  b.address.addressView.value.index?.account;
+                const index = getAddressIndex(b.address).account;
```
jessepinho, line 67: Note that this is a change in behavior — this component will now throw if the address index is undefined. Assuming that's intentional?

apps/webapp/src/components/shared/select-token-modal.tsx
```diff
@@ -65,9 +64,8 @@ export default function SelectTokenModal({
...
+                const metadata = getMetadata.optional()(b.value);
```
jessepinho, line 68: ah, looks like it was intentional, since you called `.optional()` here :D

apps/webapp/src/components/swap/asset-out-box.tsx
```diff
@@ -27,7 +28,8 @@ interface AssetOutBoxProps {
 }
 
 export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
-  const { assetOut, setAssetOut } = useStore(swapSelector);
+  const { assetOut, setAssetOut, simulateSwap, simulateOutResult } = useStore(swapSelector);
+  const { toast } = useToast();
```
jessepinho, line 32:
```suggestion
```

apps/webapp/src/components/swap/asset-out-box.tsx
```diff
@@ -37,13 +39,20 @@ export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
...
-        />
+        {simulateOutResult ? (
+          <ValueViewComponent view={simulateOutResult} showDenom={false} />
+        ) : (
+          <Button
+            variant='secondary'
+            className='w-32 md:h-9'
+            onClick={e => {
+              e.preventDefault();
+              void simulateSwap(toast);
```
jessepinho, line 50:
```suggestion
              void simulateSwap();
```

apps/webapp/src/state/swap.test.ts
```diff
@@ -70,4 +70,35 @@ describe('Swap Slice', () => {
...
+
+  test('changing parameters clears simulation', () => {
+    expect(useStore.getState().swap.simulateOutResult).toBeUndefined();
+    useStore.setState(state => {
+      state.swap.simulateOutResult = new ValueView();
+      return state;
+    });
+    expect(useStore.getState().swap.simulateOutResult).toBeDefined();
+
+    // Asset in change resets simulateOutResult
```
jessepinho, line 82: I would write each of these as a separate `test` so that we know exactly where the point of failure is when a test fails.

apps/webapp/src/state/swap.ts
```diff
@@ -26,29 +30,49 @@ export interface SwapSlice {
   setAssetOut: (metadata: Metadata) => void;
   initiateSwapTx: (toastFn: typeof toast) => Promise<void>;
   txInProgress: boolean;
+  simulateSwap: (toastFn: typeof toast) => Promise<void>;
```
jessepinho, line 33:
```suggestion
  simulateSwap: () => Promise<void>;
```

apps/webapp/src/state/swap.ts
```diff
@@ -26,29 +30,49 @@ export interface SwapSlice {
...
+        const assetIn = get().swap.assetIn;
+        const assetOut = get().swap.assetOut;
+        if (!assetIn || !assetOut) throw new Error('Both asset in and out need to be set');
+
+        const outputVal = await simulateSwapOutput(assetIn.value, assetOut);
+        set(({ swap }) => {
+          swap.simulateOutResult = outputVal;
+        });
+      } catch (e) {
+        toastFn(errorTxToast(e));
```
jessepinho, line 73:
```suggestion
        errorTxToast(e);
```

apps/webapp/src/state/swap.ts
```diff
@@ -26,29 +30,49 @@ export interface SwapSlice {
...
-      set(state => {
-        state.swap.amount = amount;
+      set(({ swap }) => {
+        swap.amount = amount;
+        swap.simulateOutResult = undefined;
       });
     },
     txInProgress: false,
+    simulateOutResult: undefined,
+    simulateSwap: async toastFn => {
```
jessepinho, line 62:
```suggestion
    simulateSwap: async () => {
```

jessepinho, line 62:
```suggestion
    simulateSwap: async () => {
```

packages/types/src/getters/value-view.ts
```diff
@@ -9,3 +10,26 @@ export const getMetadata = createGetter((valueView?: ValueView) =>
...
 
 export const getAssetIdFromValueView = getMetadata.pipe(getAssetId);
+
+export const getAmountFromValueView = createGetter((valueView?: ValueView) =>
+  valueView?.valueView.value ? valueView.valueView.value.amount : undefined,
+);
+
+export const getDisplayDenomFromView = createGetter((view?: ValueView) => {
+  if (view?.valueView.case === 'unknownAssetId') {
+    if (!view.valueView.value.assetId) throw new Error('no asset id for unknown denom');
```
jessepinho, line 20: Hmm... I think we should return `undefined` here, so that this getter still works with `.optional()`.

jessepinho (approved):
LGTM with a couple comments. Note that you'll need to make a couple changes after rebasing on `main` — I pointed them out in my comments.

Also, I always get a `0` when I estimate the swap — even if I've opened a position of say 1penumbra:5test_usd. Is that expected? Not sure I understand this enough to know if it should be working differently. My extension is pointed at testnet-preview.

jessepinho:
@grod220 one more thought — what do you think of adding a warning to the user that pressing "estimate swap" will expose them to the network? (Could be in a follow-up PR)

hdevalence:
> Also, I always get a 0 when I estimate the swap — even if I've opened a position of say 1penumbra:5test_usd. Is that expected? Not sure I understand this enough to know if it should be working differently. My extension is pointed at testnet-preview.

I'm not sure there's any infrastructure currently running on `testnet-preview` that would actually create liquidity positions. So this would be expected. If you aim at `testnet`, I've created some liquidity positions that should cause there to be price information.

jessepinho:
One other note: when I point my extension at my local devnet, then click the "estimate swap" button, I get the error `ConnectError: [unimplemented] HTTP 404`.

My local devnet is on the latest `main`, and I've `rm -rf`ed my test data and re-`generate`d everything. Any ideas why that's happening?

hdevalence:
The `SimulateTrade` RPC isn't enabled by default because it does computation (vs most of the other RPCs that just return data). Try `--enable-expensive-rpc`:
```
$ cargo run --release --bin pd -- start --help
...
    -e, --enable-expensive-rpc
            Enable expensive RPCs, such as the trade simulation service. The trade simulation
            service allows clients to simulate trades without submitting them. This is useful for
            approximating the cost of a trade before submitting it. But, it is a potential DoS
            vector, so it is disabled by default
```

jessepinho:
Awesome, thanks @hdevalence ! And I did indeed confirm that I got prices when pointing at testnet rather than testnet-preview.

jessepinho:
One other thing I'm not understanding, but not sure if it's a bug or a misunderstanding on my part:
1. I reset my local devnet.
2. `cargo run --bin pcli tx position order buy 1000penumbra@5test_usd` <-- this should set the exchange rate at 1penumbra:5test_usd
3. In the swap page, I enter 1 penumbra as the amount to swap, and test_usd as the asset type to swap into. 
4. I press "estimate swap."

Expected behavior: it should estimate 5 test_usd.
Actual behavior: it estimates 5,000 test_usd.

I figured this was an issue of an incorrect exponent somewhere, but couldn't find any incorrect handling of amounts. For that matter, if I changed the input ("amount to swap") to e.g. 1,000,000 penumbra, the estimate remains at 5,000.

hdevalence:
@erwanor any ideas on what's happening there?

erwanor:
@jessepinho The fact that you get the maximum amount of `test_usd` for `1penumbra` and `10000..penumbra` hints that something might be wrong with the denom unit amount. I can take a deeper look tomorrow, but my intuition is that we could be applying the scaling factor twice.

Could you paste the log of what `assetIn.value` is when you make the `1penumbra` query?

grod220:
A few things:
- @erwanor the logs revealed the truth! I was passing the _wallet_ balance of that asset to the swap simulation and not the user's swap input #fail. Fixed this. The results look good now. Thanks @jessepinho for prodding on this one!
- Added tooltip that gives a privacy note on this action possibly revealing swap intentions to node
- CI/CD is currently broken, but all passes locally. Think this is a low risk merge.

PR #522, opened 2024-02-15
Create UI for when extension is not installed

### LINKED ISSUE #520: Show a better UI when the Penumbra Chrome extension is not installed
We currently call [`throwIfExtNotInstalled()`](https://github.com/penumbra-zone/web/blob/main/apps/webapp/src/fetchers/is-connected.ts#L6-L8) in a number of components. This results in rather ugly UIs like this:

Image

and this:

Image

This ticket is to improve that UI. Ideally, we could handle this at a top-level component that detects if the extension is not installed, and then disables the _whole_ UI and shows a nicely designed prompt to install the extension.

jessepinho:
Tests seem to be failing due to an issue in `main`.

### Dashboard
image

### Swap page
image

## After (any page when the extension is not installed)
image

Closes #520

### DIFF
--- a/apps/extension/src/routes/page/onboarding/success.tsx
+++ b/apps/extension/src/routes/page/onboarding/success.tsx
@@ -1,43 +1,25 @@
-import {
-  Button,
-  Card,
-  CardContent,
-  CardHeader,
-  CardTitle,
-  CompressedVideoLogo,
-  FadeTransition,
-} from '@penumbra-zone/ui';
+import { Button, SplashPage } from '@penumbra-zone/ui';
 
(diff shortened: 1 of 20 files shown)


jessepinho:
@grod220 great idea. Updated the design to this, so it's less of an error and more of a splash screen:
image

The logo in the back fits a little weirdly, so I'll open a separate PR to fix that since it also touches a few other screens.

grod220 (approved):

grod220:
The background video seems frozen on my incognito browser. Maybe it's just my computer though. Can you double check before merge?

Issue #529, opened 2024-02-15
Assets table shows equivalent value of all assets


grod220:
Make use of pricing now available.

Design inspiration:

<img width="876" alt="Screenshot_2024-02-14_at_10 30 17_AM" src="https://github.com/penumbra-zone/web/assets/16624263/3b5bced3-86c0-4cea-a211-3d464c8d62c2">

Each asset type requires its own special rendering in `ValueViewComponent`. Will need to ditch table design.

hdevalence:
I think a good first step would be to change the `ValueView` component (even if we aren't currently set up with the plumbing required to track equivalent values).

PR #551, opened 2024-02-15
Update ValueViewComponent to be all-in-one

hdevalence:
Work towards #529.

Some of this will likely need further work, this just does a minimal set of changes. However the results are already better.

The assets table now has space for us to have action triggers:

Screenshot 2024-02-15 at 3 03 28 PM

Everywhere that was using a ValueView gets a formatting upgrade:

Screenshot 2024-02-15 at 3 05 15 PM
Screenshot 2024-02-15 at 3 06 42 PM
Screenshot 2024-02-15 at 3 07 07 PM


packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -1,21 +1,90 @@
...
+        <Identicon
+          uniqueIdentifier={metadata?.symbol ?? '?'}
+          size={24}
+          className='rounded-full'
+          type='solid'
+        />
+      )}
+    </>
+  );
+};
```
hdevalence, line 37: I didn't know how to deduplicate this with the other `AssetIcon` or move it between packages. Anyways, the existing one is wrongly defined now that we have images in the metadata, there's no need for a memoized search.

packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -1,21 +1,90 @@
...
+
+export const ValueViewComponent = ({ view, showDenom = true, showValue = true, showIcon = true, showEquivalent = true }: ValueViewProps) => {
   if (!view) return <></>;
 
+  if (view.valueView.case === 'knownAssetId' && view.valueView.value.metadata) {
+    const value = view.valueView.value;
+    const metadata = view.valueView.value.metadata;
+    const amount = value.amount ?? new Amount();
+    const exponent = getDisplayDenomExponent(metadata);
+    const formattedAmount = fromBaseUnitAmount(amount, exponent).toFormat();
```
hdevalence, line 48: Ideally this would round the number of decimal places to something reasonable (see the USD screenshot) but I wasn't sure how to do this, passing "6" to `.toFormat()` gave `.000000` even for round numbers.

packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -5,17 +5,71 @@ import {
...
+        {showEquivalent &&
+          value.equivalentValues.map((equivalentValue, index) => {
+            const exponent = getDisplayDenomExponent(equivalentValue.numeraire);
+            const formattedEquivalent = fromBaseUnitAmount(
+              equivalentValue.equivalentAmount ?? new Amount(),
+              exponent,
+            )
+              .toFormat(6)
+              .replace(/(\.\d*?[1-9])0+$|\.0*$/, '$1');
+            const symbol = equivalentValue.numeraire?.symbol || 'Unknown Asset';
```
hdevalence, line 60: the linter is complaining but i did this on purpose because it is the correct behavior here

turbocrime:
```suggestion
            // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
            const symbol = equivalentValue.numeraire?.symbol || 'Unknown Asset';
```

hdevalence: tyvm!

PR #553, opened 2024-02-16
Build out a Toast class for interacting more easily with toasts

### LINKED ISSUE #515: Use nicer interface for toast lifecycles
While working on https://github.com/penumbra-zone/web/pull/474, I realized that the process of managing toasts was a bit convoluted. Sonner, the library we use for managing toasts, [returns a toast ID](https://sonner.emilkowal.ski/toast#updating-toasts) that can be used for programmatically updating or dismissing a toast; but passing this around to various toast helper functions gets a bit cumbersome, and makes toasts a bit hard to reason about. 

It'd be nice to have e.g., a `Toast` class that represents the lifecycle of a toast. It would allow devs to instantiate a `Toast` instance without actually showing it yet; then update the toast to different contents based on the status of a transaction build, etc.

The benefit is that, instead of making `toast()` calls and passing a toast ID, you'd pass around a `Toast` instance. It'd be very clear which toast you're dealing with at any given point because each instance would be assigned to a variable of its own name.

Taking the swap + swap claim flow as an example, instead of this:

```ts
      const swapToastId = buildingTxToast(undefined, undefined, 'Building swap transaction');

      try {
        // build swap
        const swapReq = await assembleSwapRequest(get().swap);
        const swapPlan = await plan(swapReq);
        const swapTx = await authWitnessBuild({ transactionPlan: swapPlan }, status =>
          buildingTxToast(status, swapToastId, 'Building swap transaction'),
        );
        const swapTxHash = await getTxHash(swapTx);
        await broadcast({ awaitDetection: true, transaction: swapTx }, status =>
          broadcastingTxToast(swapTxHash, status, swapToastId),
        );
        successTxToast(swapTxHash, undefined, swapToastId, 'Swap transaction succeeded! 🎉');
    }
```

you could do this:

```ts
      const swapToast = new TransactionToast('swap')); // Note: `TransactionToast` would be a subclass of `Toast` for transaction-specific lifecycles, which would take a human-readable `transactionType` parameter to be used in building/success messages (e.g., `Building swap transaction`)

      try {
        // build swap
        const swapReq = await assembleSwapRequest(get().swap);
        const swapPlan = await plan(swapReq);
        swapToast.show();
        const swapTx = await authWitnessBuild({ transactionPlan: swapPlan }, status =>
          swapToast.building(status),
        );
        const swapTxHash = await getTxHash(swapTx);
        await broadcast({ awaitDetection: true, transaction: swapTx }, status =>
          swapToast.broadcasting(swapTxHash, status),
        );
        swapToast.succeeded(swapTxHash, undefined);
    }
```

jessepinho:
🚨 About half the line changes are from tests

Per #515, the way we currently manage the lifecycle of a toast is a bit unergonomic:
1. We need to keep track of a toast's ID to be able to update it, and pass it around to helper functions.
2. We have helper functions for using preset toasts, but those can't customized before showing them to the user, and they're conceptually separate from the toast ID you have to keep track of.

It would be better to have a single object that can both A) represent the toast instance (so you don't have to keep track of, and pass around, a toast ID), and B) update the toast instance as needed.

## In this PR
- Create a well-tested `Toast` class that represents an instance of a toast. It has methods to render and update a toast, keeping track of the toast ID internally.
- Create a `TransactionToast` class that manages the lifecycle of toasts for transactions specifically, through the stages of building, broadcasting, completion success or failure, and user authorization denial.
- Refactor the swap/send/ibc slices to use `TransactionToast`.
- Create an `errorToast` preset that uses the `Toast` interface under the hood.
- For funsies, show a progress bar on the transaction toast when a transaction is being build.

Closes #515


apps/webapp/src/state/helpers.ts
```diff
@@ -110,3 +111,17 @@ export const getTxId = (tx: Transaction | PartialMessage<Transaction>) =>
...
+ * bit. Right now, if we throw a `ConnectError` with `Code.PermissionDenied` (as
+ * we do in the approver), it gets swallowed by ConnectRPC's internals and
+ * rethrown via `ConnectError.from()`.  This means that the original code is
+ * lost, although the stringified error message still contains
+ * `[permission_denied]`. So we'll (somewhat hackily) check the stringified
+ * error message for now; but in the future, we need ot get the error flow
+ * working properly so that we can actually check `e.code ===
+ * Code.PermissionDenied`.
+ */
+export const userDeniedTransaction = (e: unknown): boolean =>
```
jessepinho, line 126: Realized this was only in the `send` slice, but should have also been in the `swap` and `ibc` slices, so I pulled it out into a helper.

apps/webapp/src/state/swap.ts
```diff
@@ -83,40 +86,21 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get) => {
...
         const swapCommitment = getSwapCommitmentFromTx(swapTx);
         await issueSwapClaim(swapCommitment);
-
-        // Reset form
         set(state => {
           state.swap.amount = '';
         });
-      } catch (e) {
-        errorTxToast(e, swapToastId);
-        throw e;
```
jessepinho, line 119: Moved almost all of this to `issueSwap` below

packages/ui/package.json
```diff
@@ -32,6 +32,7 @@
     "lucide-react": "^0.331.0",
     "react-dom": "^18.2.0",
     "react-json-view": "^1.21.3",
+    "react-router-dom": "^6.22.0",
```
jessepinho, line 35: Needed for rendering `<Link />` in the `TransactionToast` class

apps/webapp/tests-setup.ts
```diff
@@ -0,0 +1,10 @@
+import { afterEach } from 'vitest';
+import { cleanup } from '@testing-library/react';
+
+import '@testing-library/jest-dom/vitest';
```
jessepinho, line 4: See note above about handling DOM in the webapp

apps/webapp/vite.config.ts
```diff
@@ -17,4 +17,7 @@ export default defineConfig({
       },
     },
   },
+  test: {
+    environment: 'jsdom',
```
jessepinho, line 21: Needed this because `webapp` references `TransactionToast` in its tests. `TransactiionToast`, in turn, uses React elements (for the progress bar and the `<Link />` to the transaction details). As a result, webapp tests were breaking since they didn't know how to handle DOM elements.

turbocrime (approved):

PR #558, opened 2024-02-16
Experiment with symbols for delegation tokens

hdevalence:
This PR experiments with symbols for delegation tokens. Two possible approaches are illustrated:

1. The extension's asset registry can be augmented with the validator-specific token info, as I did for the two main testnet validators.
2. The extension could auto-create a symbol for the delegation token, but we must be very very careful about this, because the `symbol` field is shown to users and it is potentially attacker-controlled. For instance, if we just read the label off the chain, someone could spoof another validator's identity without consequence.

However, this seems like a good place to start.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -8,6 +8,7 @@ import {
   IndexedDbInterface,
   ViewServerInterface,
 } from '@penumbra-zone/types';
+import { assetPatterns } from '@penumbra-zone/constants';
 import { computePositionId, decodeSctRoot, transactionInfo } from '@penumbra-zone/wasm';
 
 import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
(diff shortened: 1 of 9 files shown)


packages/query/src/block-processor.ts
```diff
@@ -251,6 +252,26 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+        const id = metadata.display.substring(index + 1, index + 1 + 24);
+
+        metadata.symbol = 'Delegated UM (' + id + '...)';
+      }
+
+      // TODO: unbonding tokens?
+
+      // Note: the below code is incorrect, the asset ID is the hash of the denom,
+      // so this is actually generating metadata for a different asset. Not sure
+      // when/if this is used.
```
grod220, line 274: This block of code seems to be redundant to: https://github.com/penumbra-zone/web/blob/main/packages/types/src/asset.ts which we should fix too. However, it appears to be the same as https://github.com/penumbra-zone/penumbra/blob/6ab217514f49c03eafbca1399fe57e19e6a96925/crates/core/asset/src/asset/id.rs#L97-L105. Or no?

packages/query/src/block-processor.ts
```diff
@@ -251,6 +252,26 @@ export class BlockProcessor implements BlockProcessorInterface {
       let metadata: Metadata | undefined;
       metadata = await this.querier.shieldedPool.assetMetadata(assetId);
 
+      // If the metadata is for a delegation token, customize its symbol.
+      if (metadata && assetPatterns.delegationTokenPattern.test(metadata.display)) {
+        // We can't trust the validator's self-described name, so use their validator ID.
```
grod220, line 257: What field is their "self-described" name? Is `metadata.display` safe to use?

packages/query/src/block-processor.ts
```diff
@@ -251,6 +252,26 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+        // We know it's delegation_penumbravalid1... so use substrings:
+
+        // TODO: what's the best way to handle delegation tokens to unknown validators?
```
grod220, line 260: If we have made it within this if clause, this means they aren't in our registry. So this will always be unknown validators.

grod220 (commented):
I think it makes sense to do both of these methods. Either:

A) They are a known/trusted validator. We add them to our local registry list with their company logo in images array. This gets pre-loaded into our db.
B) They are an unknown validator. If a user has a delegation token from them, we don't find it in our db, so we assemble the metadata w/ the symbol in the block processor. If we can trust the id in `metadata.display` then we are safe.

You're taking off for holiday soon, happy to take this PR over on Monday if you'd like

packages/query/src/block-processor.ts
hdevalence: The difference is that the Display impl there isn't creating an entire Metadata, it's just converting to a string. It's correct to render the raw asset ID using bech32, the incorrect part is calling that string a "denom". But probably it doesn't cause problems if it's only ever used as a display data...

packages/query/src/block-processor.ts
hdevalence: I mean fields in the validator definition like the name, description, etc (like "Penumbra Labs CI 1"). The metadata.display field has the validator ID in it, but this is long and unwieldy.

PR #563, opened 2024-02-16
Overwrite metadata on init

grod220:
Discord context: https://discord.com/channels/824484045370818580/1044457038886998137/1208072520784420864

At the moment, we are saving new additions to the local assets registry to the database when the view server starts up. We'll want to overwrite even existing entries though. There are cases where a local asset is edited (say the image icon) and we want our database to pick this up.

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -189,14 +189,7 @@ export class IndexedDb implements IndexedDbInterface {
 
   // Save all hard-coded assets in config to database
   async saveLocalAssetsMetadata() {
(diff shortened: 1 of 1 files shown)


hdevalence (approved):
Makes sense, this is run on every init right?

grod220:
Yes, every init of the view service. So in the lifecycle of the extension, when it spins up (after being dormant or new install of extension).

grod220:
Will hang tight on merging until CI/CD stuff is settled

turbocrime:
cross-branch CI cache restore works 😄

Issue #377, opened 2024-01-18
Add a request ID for transaction approval requests


jessepinho:
Currently, when transactions are being sent to the approval popup, there's no ID attached to them to ensure that wires aren't crossed if multiple requests are being sent to the popup before any approvals. We should add a request ID when sending transactions to the popup.

(@turbocrime, can you please add more detail here on how wires could get crossed? Not sure I understand the vulnerability here.)

Per @turbocrime's [comment](https://github.com/penumbra-zone/web/pull/366#discussion_r1458048719):

> just use crypto.getrandomuuid to make one at the moment the request is sent, and check that the response contains the same uuid.
> in the zustand slice, we should
> 
> 1) never return the uuid as part of a selector
> 2) throw if we are setting request into slice but a uuid is already present
> 3) clear the uuid when resolving the approval

turbocrime:
closed #565

PR #565, opened 2024-02-16
improve tx approval slice

turbocrime:
this makes more sense in the context of connection-refactor, but it's split out here for scope concerns


turbocrime:
closes #377 by confirming request contents match

apps/extension/src/routes/popup/internal-msg-handler.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (isTxApprovalReq(req)) handleTxApproval(req.request, sendResponse);
+      else throw new Error('Unknown popup request');
+    } catch (e) {
+      const response = {
+        type: req.type,
+        error: String(e),
+      } as PopupResponse;
+      sendResponse(response);
+    }
+    return true;
```
jessepinho, line 27:
```suggestion
```
Per [the docs](https://developer.chrome.com/docs/extensions/reference/api/runtime#event-onMessage), looks like there's no meaning to the return value of the function you pass to `addListener`. So it's misleading to return a boolean.

apps/extension/src/routes/popup/internal-msg-handler.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  return false;
```
jessepinho, line 29:
```suggestion
```

apps/extension/src/routes/popup/internal-msg-handler.ts
```diff
@@ -0,0 +1,42 @@
...
+import { InternalMessageHandler } from '@penumbra-zone/types/src/internal-msg/shared';
+import { useStore } from '../../state';
+import { txApprovalSelector } from '../../state/tx-approval';
+
+export const popupMsgHandler = (
+  req: unknown,
+  _: chrome.runtime.MessageSender,
+  sendResponse: (x: unknown) => void,
+) => {
+  if (isPopupRequest(req)) {
```
jessepinho, line 16: In the interest of not nesting `if`s and `try`s too deeply, I'd just return early here if this is false, e.g., 
```suggestion
  if (!isPopupRequest(req)) return;
```

packages/types/src/internal-msg/popup.ts
```diff
@@ -1,11 +1,13 @@
-import { TxApproval } from './tx-approval';
-import { InternalRequest, InternalResponse, Ping } from './shared';
+import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1/custody_pb';
+import { Jsonified } from '../jsonified';
+import { InternalMessage, InternalRequest, InternalResponse } from './shared';
 
-type PopupSupportedMsg = TxApproval | Ping; // Add more as needed
```
jessepinho, line 4: Why are we removing the `Ping` type? Is that not being used anywhere anymore?

packages/router/src/std/router.ts
```diff
@@ -17,12 +17,11 @@ export const stdRouter = (
   services: ServicesInterface,
 ) => {
   (async function () {
-    const result = await typedMessageRouter(req.penumbraSwReq, services);
+    await typedMessageRouter(req.penumbraSwReq, services);
     sendResponse({
       sequence: req.sequence,
       penumbraSwRes: {
         type: req.penumbraSwReq.type,
-        data: result,
```
jessepinho, line 25: I don't understand enough about the router to know if this removal is OK. If it is fine, then LGTM. But in the future, for changes like this, could you please add inline comments explaining these changes to make it easier on reviewers?

apps/extension/src/state/tx-approval.ts
```diff
@@ -13,15 +23,114 @@ export interface TxApprovalSlice {
...
+
+  // these are generated by setTransactionView
   asSender?: Stringified<TransactionView>;
   asReceiver?: Stringified<TransactionView>;
   asPublic?: Stringified<TransactionView>;
   transactionClassification?: TransactionClassification;
-  // Holding the message responder function. Service worker will be "awaiting" the call of this.
-  responder?: MessageResponder<TxApproval>;
+
+  attitude?: boolean;
```
jessepinho, line 44: Is this a normal name for a boolean representing user consent? To me, it seems like it'd be clearer to use a name like `isApproved` or something. But if `attitude` is a normal name for this concept that I'm not familiar with, LGTM :)

jessepinho (approved):
LGTM! I added a few small notes about naming etc.

As I mentioned in one of the comments too, it'd be really helpful to write up A) PR descriptions about what to expect in this PR, so that reviewers like me who are less familiar with routing can understand what we're looking at, and B) inline comments where needed to explain changes whose purposes are not readily apparent.

packages/types/src/internal-msg/popup.ts
turbocrime: it isn't used anywhere. i guess that's an excessive change for this, but in particular this was pulled out of a larger refactor PR

apps/extension/src/routes/popup/internal-msg-handler.ts
turbocrime: the documentation is misleading. the truthiness of a handler return indicates to the runtime if this handler intends to asynchronously provide a response.

apps/extension/src/routes/popup/internal-msg-handler.ts
turbocrime: i constructed it this way to scope the `true` return, and make the fallback return a `false`, to improve the accuracy of intent to respond expressed to the runtime. we occasionally have subtle failures and this makes that failure space smaller.

packages/router/src/std/router.ts
turbocrime: i'll add PR comments in the future. i expected to pair on you with this. inline code comments feel out of place, these are vestigial remnants of the old and non-functional router

apps/extension/src/routes/popup/internal-msg-handler.ts
jessepinho: Yep, makes sense now — my comment here only applied if we were nixing the `true`/`false` return value

apps/extension/src/state/tx-approval.ts
```diff
@@ -13,15 +23,114 @@ export interface TxApprovalSlice {
...
+      });
+    }
+  },
+});
 
 export const txApprovalSelector = (state: AllSlices) => state.txApproval;
+
+/**
+ * Returns a boolean indicating whether the given address is controlled by the
+ * current user.
```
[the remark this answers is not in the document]
turbocrime: it looks like this is only ever used as a parameter for asRecieverTransactionView and its associated functions, and never anywhere else. so should probably be commented there and perhaps renamed to be less likely to confuse. to avoid confusion i've now tightened the scope and am no longer exporting it.

we might reconsider making view generation the responsibility of the component instead of the slice, which would allow us to use hooks/fetchers/queriers and permit this slice to become fully synchronous.

PR #567, opened 2024-02-17
Simplify the Card component to no longer need z-indexes

jessepinho:
I noticed that the dropdown menu on tablets was hiding behind the form on the Send page. This was due to some weird use of absolute positioning just to create a gradient background on our `Card` component. To fix this, I removed all the absolute positioning, and modified the gradient to include the semi-transparency and background color that was previously being achieved via markup.

## Before
image

## After
image

### DIFF
--- a/apps/webapp/src/components/send/layout.tsx
+++ b/apps/webapp/src/components/send/layout.tsx
@@ -13,7 +13,7 @@ export const SendLayout = () => {
     <div className='relative mx-auto grid gap-6 md:grid-cols-2 md:gap-4 xl:max-w-[1276px] xl:grid-cols-3 xl:gap-5'>
       <div className='hidden xl:order-1 xl:block' />
       <Card gradient className='order-2 row-span-2 flex-1 p-5 md:order-1 md:p-4 xl:p-5'>
-        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto w-[75%] lg:w-[372px]' />
+        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto flex w-[75%] lg:w-[372px]' />
         <Outlet />
       </Card>
       <EduInfoCard
--- a/packages/tailwind-config/index.js
+++ b/packages/tailwind-config/index.js
@@ -131,8 +131,16 @@ export default {
         },
       },
       backgroundImage: {
-        'card-radial':
-          'radial-gradient(33% 50% at 15% 44%, var(--rust), transparent),radial-gradient(33% 40% at 105% 42%, var(--teal), transparent),radial-gradient(33% 80% at 85% 124%, var(--teal), transparent)',
+        // The final `linear-gradient` is just to make a solid charcoal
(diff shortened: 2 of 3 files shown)


apps/webapp/src/components/send/layout.tsx
```diff
@@ -13,7 +13,7 @@ export const SendLayout = () => {
     <div className='relative mx-auto grid gap-6 md:grid-cols-2 md:gap-4 xl:max-w-[1276px] xl:grid-cols-3 xl:gap-5'>
       <div className='hidden xl:order-1 xl:block' />
       <Card gradient className='order-2 row-span-2 flex-1 p-5 md:order-1 md:p-4 xl:p-5'>
-        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto w-[75%] lg:w-[372px]' />
+        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto flex w-[75%] lg:w-[372px]' />
```
jessepinho, line 16: This was weird. Something about the previous version of the card had made the tabs centered. After my change, they were no longer centered, because they default to being `inline-flex`. This seems like some messy CSS that I eventually want to pull the thread on, but not today :D

packages/tailwind-config/index.js
```diff
@@ -131,8 +131,16 @@ export default {
         },
       },
       backgroundImage: {
-        'card-radial':
-          'radial-gradient(33% 50% at 15% 44%, var(--rust), transparent),radial-gradient(33% 40% at 105% 42%, var(--teal), transparent),radial-gradient(33% 80% at 85% 124%, var(--teal), transparent)',
+        // The final `linear-gradient` is just to make a solid charcoal
+        // background color for the radial gradients to sit on top of. If
+        // there's a way to make a solid background color without
+        // `linear-gradient`, feel free to update this.
+        'card-radial': `
```
jessepinho, line 138: Three changes here:
1. I formatted this string with line breaks to make it more readable.
2. I added `color-mix()` to set the radial gradients to 20% opacity. Previously, this was done by making the div on which the gradients appeared 20% opaque.
3. I added a `linear-gradient` at the bottom to form the background color.

packages/ui/components/ui/card.tsx
```diff
@@ -8,13 +8,12 @@ export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
...
-        <div className='absolute inset-0 z-0 rounded-lg bg-card-radial p-[30px] opacity-20 ' />
-      </div>
-    ) : (
-      <div ref={ref} className={cn(baseClasses, className)} {...props}>
+    return (
+      <div
+        ref={ref}
+        className={cn(baseClasses, gradient && 'bg-card-radial', className)}
+        {...props}
+      >
```
jessepinho, line 16: We no longer need all the absolute/relative positioning and z-indexes. We just need to apply `bg-card-radial` if the `gradient` prop is truthy.

grod220 (approved):

Issue #524, opened 2024-02-15
Add max height + scroll to asset selector


grod220:
<img width="884" alt="Screenshot_2024-02-14_at_3 40 18_PM" src="https://github.com/penumbra-zone/web/assets/16624263/d44c03b4-417e-4656-8eeb-61fabeeee64e">

jessepinho:
@grod220 are you sure you weren't seeing this on an old commit? I fixed this a while back. Here's what it looks like for me:



![Image](https://github.com/penumbra-zone/web/assets/1121544/bd51728a-58a7-4c7f-b1f9-26be87cebc6e)

jessepinho:
(moving to blocked for now until that question is resolved)

PR #568, opened 2024-02-17
Truncate overly long balances

jessepinho:
Found while investigating #524 

## Before (note the overflowing penumbra balance)
image

## After
image


grod220 (approved):
Can we also get a hover tooltip or something that displays the full amount if it's truncated and hovered for a few seconds? I can make another issue if it's out of scope for now.

hdevalence:
The dropdown should be patched to eliminate the "Token Name" column entirely, now that we have a component for rendering the balance and token symbol in one.

jessepinho:
superseded by v2 library

Issue #541, opened 2024-02-15
User's delegations


grod220:
On staking page, display:
- user's current delegations (unbonding, bonded)
- Amount able to delegate

jessepinho:
I'm going to close this since it's now split into several ticket: #586, #587, and #588

Issue #463, opened 2024-02-06
Tracking Issue: Staking


grod220:
Support in mini-front the ability to view validators and stake penumbra with them

- [ ] Add new `Staking` tab 
- [ ] Query: List of all validators in network (fields: voting power, share, commission, state, bonding state, validator info)
- [ ] View: user's current delegations (unbonded, bonded)
- [ ] Tx: 
  - [ ] Delegate to a validator
  - [ ] Undelegate from validator (show unbonding status)
  - [ ] Undelegate claim (after unbonding)

hdevalence:
xref https://github.com/penumbra-zone/web/pull/569#issuecomment-1957738474

hdevalence:
#540 and #541 should not be separate issues, since the staking information to display is token-based.

PR #569, opened 2024-02-17
Staking/validators page

### LINKED ISSUE #540: Validators page
Add new Staking tab + page

List of all validators in network (fields: voting power, share, commission, state, bonding state, validator info)

Some design inspiration: https://portal.dymension.xyz/dymension/staking

jessepinho:
image

## In this PR
- Create Staking page with initial columns. NOTE: There's no "delegate"/"undelegate" button yet; that will come as part of #541.
- Add `StakingService` to the service registry, create a proxy impl for it in `service-worker.ts`, and create a service client for it in `webapp/src/clients/grpc.ts`.
- Move the max-width that was shared by a number of pages to the `main` tag of the root `<Layout />` component.
- Add a bunch of getters for validator info.
- Set up UI testing libraries inside `webapp` (`@testing-library/react` and `@testing-library/jest-dom`) along with test setup files and Vitest config.

Closes #540


apps/webapp/src/components/shared/edu-panels/content.tsx
```diff
@@ -21,6 +22,8 @@ export const eduPanelContent: Record<EduPanel, string> = {
     'Every Penumbra account has a stable default address and many one-time IBC deposit addresses.  All addresses for the same account deposit to the same pool of funds.  Use a freshly generated IBC deposit address to preserve privacy when sending funds from a transparent chain to Penumbra.',
   [EduPanel.IBC_WITHDRAW]:
     'IBC to a connected chain. Note that if the chain is a transparent chain, the transaction will be visible to others.',
+  [EduPanel.STAKING]:
+    'Explore the available validator nodes and their associated rewards, performance metrics, and staking requirements. Select the validator you wish to delegate your tokens to, based on factors like uptime, reputation, and expected returns. Stay informed about validator performance updates, rewards distribution, and any network upgrades to ensure a seamless staking experience.',
```
jessepinho, line 26: Did ChatGPT write this? Hmm, who can really say...

apps/webapp/src/components/dashboard/layout.tsx
```diff
@@ -10,7 +10,7 @@ export const DashboardLayout = () => {
   const pathname = usePagePath<DashboardTab>();
 
   return (
-    <div className='relative mx-auto grid gap-6 md:gap-4 lg:grid-cols-3  xl:max-w-[1276px] xl:gap-5'>
+    <div className='grid gap-6 md:gap-4 lg:grid-cols-3 xl:gap-5'>
```
jessepinho, line 13: Moved these classes to `webapp/src/components/layout`

apps/webapp/src/components/header/constants.tsx
```diff
@@ -33,6 +33,12 @@ export const headerLinks: HeaderLink[] = [
     active: true,
     mobileIcon: <SwapIcon />,
   },
+  {
+    href: PagePath.STAKING,
+    label: 'Staking',
+    active: true,
```
jessepinho, line 39: Changed this to `true`, and reordered it to be earlier in the menu

apps/webapp/src/components/send/layout.tsx
```diff
@@ -10,10 +10,10 @@ export const SendLayout = () => {
   const pathname = usePagePath<SendTab>();
 
   return (
-    <div className='relative mx-auto grid gap-6 md:grid-cols-2 md:gap-4 xl:max-w-[1276px] xl:grid-cols-3 xl:gap-5'>
+    <div className='grid gap-6 md:grid-cols-2 md:gap-4 xl:grid-cols-3 xl:gap-5'>
       <div className='hidden xl:order-1 xl:block' />
       <Card gradient className='order-2 row-span-2 flex-1 p-5 md:order-1 md:p-4 xl:p-5'>
-        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto w-[75%] lg:w-[372px]' />
+        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto w-[75%] lg:min-w-[300px]' />
```
jessepinho, line 16: the `min-w` change fixes a layout bug where the tabs were previously too wide

apps/webapp/src/components/staking/validator-info-row.tsx
```diff
@@ -0,0 +1,54 @@
...
+  loading,
+  validatorInfo,
+  votingPowerByValidatorInfo,
+}: {
+  loading: boolean;
+  validatorInfo: ValidatorInfo;
+  votingPowerByValidatorInfo: Map<ValidatorInfo, number>;
+}) => {
+  const rewardRate = getValidatorRewardRateFromValidatorInfoOptional(validatorInfo);
+  const commission = rewardRate ? Number(joinLoHiAmount(rewardRate)) : '';
```
jessepinho, line 32: Not sure how to calculate commission. At the moment, I'm just `Number`ifying the reward rate. Reviewers, how is commission calculated?

packages/transport/src/proxy.ts
```diff
@@ -19,6 +19,15 @@ export const simpleContextHandler: ProxyContextHandler = (i, ctx) => {
...
 
+/**
+ * Creates an implementation of a service type that proxies all requests via a
+ * gRPC-web transport.
+ *
+ * To do this, it iterates over each method in the service type definition,
+ * creating a method on the new proxy object with the same name that calls out
+ * to the gRPC-web transport. Thus, the returned implementation is basically
+ * just a bunch of methods that call out to an RPC endpoint.
+ */
```
jessepinho, line 30: Just added this as a "breadcrumb" to keep document what this does, once I'd followed the trail of its code.

packages/ui/components/ui/table.tsx
```diff
@@ -3,9 +3,7 @@ import { cn } from '../../lib/utils';
 
 const Table = React.forwardRef<HTMLTableElement, React.HTMLAttributes<HTMLTableElement>>(
   ({ className, ...props }, ref) => (
-    <div className='relative w-full overflow-auto'>
```
jessepinho, line 6: Wrapped `<div>` isn't needed, as far as I can tell.

apps/webapp/src/components/send/layout.tsx
```diff
@@ -10,10 +10,14 @@ export const SendLayout = () => {
...
   return (
-    <div className='relative mx-auto grid gap-6 md:grid-cols-2 md:gap-4 xl:max-w-[1276px] xl:grid-cols-3 xl:gap-5'>
+    <div className='grid gap-6 md:grid-cols-2 md:gap-4 xl:grid-cols-3 xl:gap-5'>
       <div className='hidden xl:order-1 xl:block' />
       <Card gradient className='order-2 row-span-2 flex-1 p-5 md:order-1 md:p-4 xl:p-5'>
-        <Tabs tabs={sendTabs} activeTab={pathname} className='mx-auto flex w-[75%] lg:w-[372px]' />
+        <Tabs
+          tabs={sendTabs}
+          activeTab={pathname}
+          className='mx-auto flex w-[75%] lg:min-w-[300px]'
```
jessepinho, line 19: the `min-w` change fixes a layout bug where the tabs were previously too wide

jessepinho:
Converting back to draft while I fix the commission

apps/webapp/src/components/staking/helpers.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const toSum = (prev: number, curr: number) => prev + curr;
+
+/**
+ * Given a `ValidatorInfo`, returns the sum of all commission rates in terms of
+ * [basis points](https://en.wikipedia.org/wiki/Basis_point).
+ */
+export const calculateCommission = (validatorInfo: ValidatorInfo): number => {
+  const fundingStreams = getFundingStreamsFromValidatorInfo(validatorInfo);
+  return fundingStreams.map(getRateBpsFromFundingStream).reduce(toSum);
```
jessepinho, line 27: This is the total commission, per the logic @erwanor described [here](https://github.com/penumbra-zone/web/issues/540#issuecomment-1953271544) and per pcli's implementation [here](https://github.com/penumbra-zone/penumbra/blob/48e32b4/crates/bin/pcli/src/command/query/validator.rs#L109-L115).

jessepinho:
OK @grod220 / @turbocrime , this is ready for review again

apps/webapp/src/components/staking/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * To do this, we convert the rate from [basis
+ * points](https://en.wikipedia.org/wiki/Basis_point) (which are one hundredth
+ * of one percent).
+ */
+export const calculateCommissionAsPercentage = (validatorInfo: ValidatorInfo): number => {
+  const fundingStreams = getFundingStreamsFromValidatorInfo(validatorInfo);
+  const totalBps = fundingStreams.map(getRateBpsFromFundingStream).reduce(toSum);
+
+  return totalBps / 100;
```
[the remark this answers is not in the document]
erwanor: LGTM!

apps/webapp/src/components/staking/use-validator-infos.ts
```diff
@@ -0,0 +1,77 @@
...
+  loading: boolean;
+  error: unknown;
+} => {
+  /**
+   * Use a ref so that it doesn't constantly re-fetch when the component
+   * re-renders.
+   *
+   * @todo: How to trigger a refresh?
+   */
+  const query = useRef(getValidatorInfos());
```
grod220, line 41: I have psychological trauma about hooks calling hooks calling hooks (all those unmaintainable frontends I left behind 😅 ).

What do you think about moving all this to a react-router loader? (using async/await instead of hooks)

However, if we anticipate potentially slow or very long response, then it may not be best.

apps/webapp/src/components/staking/use-validator-infos.ts
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+const toTotalVotingPower = (prev: number, curr: ValidatorInfo) =>
+  prev + Number(joinLoHiAmount(getVotingPowerFromValidatorInfo(curr)));
+
+const getFormattedVotingPower = (validatorInfo: ValidatorInfo, totalVotingPower: number) =>
+  Math.round(
+    (Number(joinLoHiAmount(getVotingPowerFromValidatorInfo(validatorInfo))) / totalVotingPower) *
+      100,
+  );
```
grod220, line 24: looks like some more staking stuff that could move to a staking utils file in packages somewhere

grod220 (approved):

apps/webapp/src/components/staking/layout.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <div className='col-span-2'>
+        <Card gradient>
+          <CardContent>
+            <ValidatorsTable />
+          </CardContent>
+        </Card>
+      </div>
+
+      <div>
+        <EduInfoCard label='Staking' content={EduPanel.STAKING} src='./funds-gradient.svg' />
```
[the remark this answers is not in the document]
jessepinho: Done. Where'd you get this?

apps/webapp/src/components/staking/use-validator-infos.ts
jessepinho: Yeah, that was my original approach: I called `getValidatorInfos()` in the loader and then waited for the stream to finish. That took quite a while, though: testnet is a bit slow right now, and there are like 80 validators (300 if you include inactive ones, which I imagine we'll eventually do). So I figured we should stick with streaming the results.

apps/webapp/src/components/staking/layout.tsx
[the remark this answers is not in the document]
grod220: Got it from https://www.svgrepo.com/ and had ChatGPT add the gradient in

apps/webapp/src/components/staking/use-validator-infos.ts
hdevalence: I don't think we need to support inactive validators in the list, only supporting members of the consensus set (and any validators the user has delegation tokens for) is fine.

apps/webapp/src/components/staking/use-validator-infos.ts
jessepinho: Done

apps/webapp/src/components/staking/helpers.ts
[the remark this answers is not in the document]
jessepinho: Done

hdevalence:
As a general comment, this work is (or should be) entangled with #558: staking in Penumbra is token-based, and the staking UI should reflect that.

**This UI shouldn't be centered around a list of validators, it should be centered around a list of tokens, since a user's staking state is represented via tokens.** (That's why I opened up #558, to flag that as a problem to solve, but I could have been more explicit about the intent, sorry about that).

OK, but what does it mean for the UI to be centered around tokens?

First, we need to set out what tokens there are. There are three relevant classes of tokens:

1. The staking token UM itself, representing unbonded stake. This is the simplest, since it's just one token.
2. Delegation tokens, representing shares of a particular validator's delegation pool. These are not just one token but a family of tokens, parameterized by the validator identity. (The chain does staking rewards by adjusting the exchange rate between the staking token and each delegation token).
3. Unbonding tokens, representing stake that was delegated to a particular validator but is in the process of being undelegated. These have even more variety than delegation tokens: they're parameterized both by the validator identity and the epoch at which unbonding started. (In the happy path, unbonding tokens are 1:1 with the staking token, but the chain delays conversion to allow slashing if the validator were to misbehave during the unbonding period).

Rendering the user's staking state requires handling all of these tokens. However, because these tokens are _how_ that state is recorded, it does not require any _more_ than handling these three classes of tokens.

I'd suggest the following organization for the staking page. There should be several tabular blocks, showing each of these categories of tokens. Each row should have staking-specific rendering (ie additional contextual information) of the user's balance of a particular token.

The first block should be titled "Unstaked Tokens" and have one row, showing the user's balance of the UM token.

The second block should be titled "Staked Tokens" and have one row for each delegation token for which the user has a nonzero balance. If there are no rows, the block should be omitted. Each row should show the amount of delegation tokens, the equivalent amount of staking tokens and USD (this functionality being provided by the equivalent_values feature in the valueview, which the extension needs to but does not yet implement), information about the validator, and buttons for "Delegate" and "Undelegate".

The third block should be titled "Unbonding Tokens" and have one row for each unbonding token for which the user has a nonzero balance. If there are no rows, the block should be omitted. Each row should show the amount of unbonding tokens (initially we can ignore penalties and treat these as 1:1 for UM tokens), information about the validator, and a time-based estimate of when unbonding will finish. The row should have a button labeled "Claim" that does the UndelegateClaim, this should be disabled until unbonding is believed to be finished.

The fourth block should be labeled "Active Validators" and have one row for each active validator, _excepting_ validators the user already has delegations to. Each row should be identical to the rows used for the "Staked Tokens" block, the only difference being that the rows are rendering a 0-balance of that validator's delegation token and so should have a disabled "Undelegate" button.

This UI shows a complete picture of the user's staking state, which is recorded by tokens, shows them the most relevant information (their balances and delegations) first, and provides convenient hooks for all staking-related actions.

grod220:
I feel this comment is best placed in the very next task that is currently being worked on: https://github.com/penumbra-zone/web/issues/463

> This UI shouldn't be centered around a list of validators, it should be centered around a list of tokens

In general, this feels like you are saying something was built incorrectly and something is wrong. But, really, this page is simply not complete. We have follow up issues specifically for the user's positions and taking actions. We really can't expect a user to be able to take an action without being able to pull a list of validators they can delegate to.

apps/webapp/src/components/staking/helpers.ts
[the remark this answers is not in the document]
hdevalence: Yes, my point is that we shouldn't be splitting out helper functions until we know what helper functions are actually helpful, and since this code needs to be redesigned, we don't have that information yet.

jessepinho:
Thanks @grod220 / @hdevalence. It seems like I've built the "Active Validators" block except for its delegate buttons. My suggestion:

1. I'll add delegate buttons as part of #541.
2. I'll create additional tickets for each of the remaining blocks described above.

hdevalence:
Yep, that sounds great. In that refactoring it would be good to refactor into a component that takes BOTH:

- a ValueView of the delegation token
- additional validator data

and reconceptualize it as being a rendering of the delegation token amount. We should be planning to use the asset Metadata to render the delegation tokens for the reasons discussed in #558 — it's important that we have consistent rendering of delegation tokens.

Issue #310, opened 2023-12-15
Build output -> Html file based routing


grod220:
At the moment, our web app builds a single `index.html` entry point file. This requires the html server that is serving the site to do something like this: https://github.com/penumbra-zone/web/blob/main/apps/webapp/firebase.json#L6-L11. Wildcard the route so it broadly points to that entrypoint html file. At that point, React routing takes over.

This is problematic as it puts responsibility on the html server where some don't have this ability (aka in our case with pd). An example to test:

```
check out bundle-frontend-into-pd on monorepo
do pd start
go to localhost:8080/index.html
(it doesn't have a / -> /index.html redirect at the moment but this isn't an inherent limitation)
```

We need to change our build output to be html files per route. There are Vite plugins that can help us with these kinds of things. Should investigate:

- https://vitejs.dev/guide/ssr
- vite-plugin-pages
- vite-plugin-ssr/plugin

We also should update our firebase config when we've moved to the new method.

turbocrime:
this involves changing some routes. things like transaction view URIs are formatted like `baseurl/tx/7r4ns4c710n-h4sh` and a static site served without rewriting can't accept that.

such 'parameters' should be applied via the 'anchor' segment of the url, like `baseurl/tx#7r4ns4c710n-h4sh`. this has the benefit of remaining local, the static server will not be aware of the anchor segment.

turbocrime:
alternatively, a hash router could be used for all routes. for the webapp i think this is an okay choice.

grod220:
(from our conversation today @turbocrime @jessepinho)

It looks like we've made some progress on the support of our web app via the penumbra node. With the new hash router: https://github.com/penumbra-zone/web/pull/372, it appears to work. However, hash routing is not idiomatic web routing. It's really a bit of a hack of a tool meant for something else.

Our current Vite build tool leans into the single-page-app philosophy (that is feeling more and more dated). It outputs a single html file and points to the single javascript bundle. We could write code that makes vite supports multiple entry points (multiple html files that points to same bundle). Quite soon after, we'll likely want to use a plugin or write code to do code splitting. And then after, we may want to solve our multiple round trip issue (html -> javascript -> rendering -> more requests). So more custom code for pre-rendering...

[...]

PR #572, opened 2024-02-19
ADR 002: Client-side rendering

grod220:
Consequence of: https://github.com/penumbra-zone/web/issues/310

### DIFF
--- a/docs/adrs/002-framework.md
+++ b/docs/adrs/002-framework.md
@@ -35,7 +35,7 @@ to a "meta-framework" that would handle all of this for us. However, we discover
 It allows any network adversary to watch a user's activity as they navigate the app (since the requests are predictable and content-free, they may leak paths through the length of the encrypted data, which TLS does not protect).
 
 If we are not code-splitting, pre-rendering html by route also doesn't make sense.
-Unfortunately, the better performance characteristics is not compatible with our stated philosophical goals.
+Unfortunately, the better performance characteristics are not compatible with our stated philosophical goals.
 The momentum in the industry of web stack tech (putting more responsibility on the server) does not serve us well.
 
 ### Settled design
--- a/docs/adrs/002-framework.md
+++ b/docs/adrs/002-framework.md
@@ -13,7 +13,7 @@ Doing so requires us to be quite deliberate on what our web stack looks like.
 
 ### History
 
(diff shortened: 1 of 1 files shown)


docs/adrs/002-framework.md
```diff
@@ -0,0 +1,53 @@
...
+- can be run by PD (penumbra rpc node)
+  - statically rendered
+  - not requiring special server config
+  - can be run by generic static file server
+
+Doing so requires us to be quite deliberate on what our web stack looks like.
+
+### History
+
+We first began the app as a Next.js app. However, it's very server-focused and does not support patterns like data loaders that are quite helpful for
```
jessepinho, line 16:
```suggestion
We first began the app as a Next.js app. However, Next.js is very server-focused and does not support patterns like data loaders that are quite helpful for
```

jessepinho: (because it originally sounded like _the app_ is very server-focused)

docs/adrs/002-framework.md
```diff
@@ -0,0 +1,53 @@
...
+
+Our output build is still a single `index.html`. However, the hash effectively performs the re-direct (route after `#` not sent to the server).
+
+Afterward, we noticed [a few drawbacks to the hash router](https://github.com/penumbra-zone/web/issues/310#issuecomment-1904608594).
+It's not _really_ idiomatic web routing. We were also not doing pre-rendering or code-splitting (not as performant as we could be). The idea was discussed to migrate
+to a "meta-framework" that would handle all of this for us. However, we discovered code-splitting by route is _not_ privacy preserving.
+It allows any network adversary to watch a user's activity as they navigate the app (since the requests are predictable and content-free, they may leak paths through the length of the encrypted data, which TLS does not protect).
+
+If we are not code-splitting, pre-rendering html by route also doesn't make sense.
+Unfortunately, the better performance characteristics is not compatible with our stated philosophical goals.
```
jessepinho, line 38:
```suggestion
Unfortunately, the better performance characteristics are not compatible with our stated philosophical goals.
```

jessepinho (approved):
LGTM! other than a couple tiny comments (sorry, grammar stickler here).

So, this means we can't even do code splitting, since that would potentially reveal which _routes_ the user has navigated to? i.e., even that is considered private data — the fact that a user has navigated to e.g. the Staking page? Dang 😳

PR #574, opened 2024-02-19
Remove exact cargo.lock requirement

grod220:
- fixes cargo tree package target
- removes lock file update requirement (otherwise will throw on minor deps bump on crates.io)
- runs on main


turbocrime:
i think this is a bad idea. we want the ability to know exactly what we're building in CI, and what we're shipping. additionally, CI cacheing is based on the cargo lock hash.

Issue #139, opened 2023-10-25
No global import of testnet constants


grod220:
At the moment, a number of places are directly importing the testnet constants. We likely need to have an alternative approach than this kind of hardcoding method.

grod220:
Duplicate to: https://github.com/penumbra-zone/web/issues/900

PR #578, opened 2024-02-20
inject literal config values

turbocrime:
use webpack config to inject literals and environment variables

for example with `MINIFRONT_URL=https://localhost:5173 pnpm dev`

the typescript

```ts
console.log(
  'displaying injected literals',

  process.env['NODE_ENV'],

  PRAX,
  PRAX_ORIGIN,

  IDB_VERSION,
  USDC_ASSET_ID,

  MINIFRONT_URL,
  DEFAULT_GRPC_URL,
);

```

may become the javascript

```js
console.log(
  'displaying injected literals',
  'development',
  'lkpmkhpnhknhmibgnmmhdhgdilepfghe',
  'chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe',
  23,
  'reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg=',
  'https://localhost:5173',
  'https://grpc.testnet.penumbra.zone/',
);
```


apps/extension/webpack.config.js
grod220, line 1: Can we convert this to a typescript file? `webpack.config.ts`?

packages/constants/config.ts
grod220, line 1: I think it's worthwhile keeping a `prod` an `testnet` config that the web app webpack.config.js can import. It's helpful to keep our hardcoded registries and parameter stuff in one place.

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,146 @@
...
+      (argv.mode == 'development' && process.env.MINIFRONT_URL) || MINIFRONT_URL,
+    ),
+
+    DEFAULT_GRPC_URL: JSON.stringify(
+      // you may want https://grpc.testnet-preview.penumbra.zone/
+      (argv.mode == 'development' && process.env.PENUMBRA_NODE_PD_URL) || PENUMBRA_NODE_PD_URL,
+    ),
+  };
+
+  console.log('webpack', { argv, definitions });
```
grod220, line 38: can remove

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,146 @@
...
+
+export default (env, argv) => {
+  // types declared in prax.d.ts
+  const definitions = {
+    // process.env.NODE_ENV is automatically provided by DefinePlugin
+
+    PRAX: JSON.stringify(PRAX),
+    PRAX_ORIGIN: JSON.stringify(`chrome-extension://${PRAX}`),
+
+    IDB_VERSION: JSON.stringify(IDB_VERSION),
```
grod220, line 25: Is this a type mismatch if `IDB_VERSION` is expect to be a number?

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,146 @@
...
+const __dirname = new URL('.', import.meta.url).pathname;
+const srcDir = path.join(__dirname, 'src');
+
+export default (env, argv) => {
+  // types declared in prax.d.ts
+  const definitions = {
+    // process.env.NODE_ENV is automatically provided by DefinePlugin
+
+    PRAX: JSON.stringify(PRAX),
+    PRAX_ORIGIN: JSON.stringify(`chrome-extension://${PRAX}`),
```
grod220, line 23: the stringify's feel unnecessary if they are supposed to be strings already

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,146 @@
...
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+import webpack from 'webpack';
+
+const PRAX = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe';
+
+const IDB_VERSION = 23;
+const USDC_ASSET_ID = 'reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg=';
+
+const MINIFRONT_URL = 'https://app.testnet.penumbra.zone/';
+const PENUMBRA_NODE_PD_URL = 'https://grpc.testnet.penumbra.zone/';
```
grod220, line 12: See my other comment, think we should import these

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,146 @@
...
+        // Required by the `bip39` library
+        Buffer: ['buffer', 'Buffer'],
+      }),
+      new webpack.IgnorePlugin({
+        // Not required by the `bip39` library, but very nice
+        checkResource(resource) {
+          return /.*\/wordlists\/(?!english).*\.json/.test(resource);
+        },
+      }),
+      new webpack.DefinePlugin(definitions),
```
grod220, line 116: We could just do the prod/dev diff here:

```typescript
new webpack.DefinePlugin(argv.mode == 'prod' ? prodDefinitions : devDefinitions)
```

grod220 (commented):
Appears to address https://github.com/penumbra-zone/web/issues/139 👍

apps/extension/package.json
```diff
@@ -5,9 +5,9 @@
   "type": "module",
   "scripts": {
     "download-keys": "tsx src/utils/download-proving-keys.ts",
-    "dev": "webpack --config webpack/webpack.dev.js --watch",
+    "dev": "webpack --watch --mode=development -d inline-source-map",
     "clean": "rm -rfv dist bin",
-    "build": "webpack --config webpack/webpack.prod.js",
+    "build": "webpack --mode=development -d inline-source-map",
```
grod220, line 10: build should be `production` no?

apps/extension/webpack.config.js
turbocrime: we could, but it's really annoying, and i don't want to do it in this pr

apps/extension/webpack.config.js
turbocrime: this echoes webpack parameters including environment variables that are affecting config to the terminal, which i think is good.

apps/extension/webpack.config.js
turbocrime: it's injected as a literal string `23` which is a number in javascript

apps/extension/webpack.config.js
turbocrime: defineplugin needs a string to inject https://webpack.js.org/plugins/define-plugin/

apps/extension/package.json
turbocrime: this is a script for developer convenience. if a develper is running `pnpm build` at the terminal, they are probably doing development, and want development features. in the unsual case they might want something else, the developer can run webpack directly and more explicitly.

in this PR, when CI makes a production build, it uses 

https://github.com/penumbra-zone/web/blob/43bf771a567a3705554fc13cd1ec9f038e9589e3/.github/workflows/extension-publish.yml#L41-L42

apps/extension/webpack.config.js
turbocrime: now using dotenv to import

apps/extension/webpack.config.js
turbocrime: removed any prod/dev switching

apps/extension/package.json
turbocrime: okay, i changed this to just run `webpack` with no options, which defaults to `--mode=production` with no `devtool` (equivalent to before). the change to `extension-publish.yml` was reverted

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,138 @@
...
+// environment variables, you will override those defaults.
+dotenv.config();
+
+const definitions = {
+  // process.env.NODE_ENV is automatically provided by DefinePlugin
+
+  PRAX: JSON.stringify(process.env.PRAX),
+  PRAX_ORIGIN: JSON.stringify(`chrome-extension://${process.env.PRAX}`),
+
+  IDB_VERSION: JSON.stringify(Number(process.env.IDB_VERSION)),
```
grod220, line 17: Why would we need to go from string -> number -> string?

grod220 (approved):
Just a few questions, but feel free to merge after

apps/extension/webpack.config.js
turbocrime: we could do it directly, but this provides a little bit of type safety, and prevents any creative escape sequence adventures

apps/extension/webpack.config.js
```diff
@@ -0,0 +1,138 @@
...
+
+// Loads default vars from `.env` file in this directory.  If you set
+// environment variables, you will override those defaults.
+dotenv.config();
+
+const definitions = {
+  // process.env.NODE_ENV is automatically provided by DefinePlugin
+
+  PRAX: JSON.stringify(process.env.PRAX),
+  PRAX_ORIGIN: JSON.stringify(`chrome-extension://${process.env.PRAX}`),
```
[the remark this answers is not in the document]
turbocrime: they are not equal - again please see defineplugin docs https://webpack.js.org/plugins/define-plugin/#usage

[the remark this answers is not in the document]
turbocrime: tldr it injects a *code string* - so directly defining a string as 

```
SOME_DEFINE: `chrome-extension://${process.env.PRAX}`
```

would transform the code

```
console.log("Some define:", SOME_DEFINE);
```

into 

```
console.log("Some define:", chrome-extension://whatever);
```

when what you want is 


```
console.log("Some define:", "chrome-extension://whatever");
```

PR #580, opened 2024-02-20
Fix the extension background

jessepinho:
As @turbocrime and I were pairing on a separate issue, we realized that the extension was using some messy markup with absolute-positioned elements just to show a semitransparent background image. @turbocrime figured out a way to show a semitransparent background image by layering a solid color over it at 80% opacity, so this PR implements that to simplify the markup.

This PR also refactors the "successful login" extension page to use the `<SplashPage />` component, and fixes a z-index issue there too.

### Logged-out extension
image

### Logged-in extension
image

### Successful login
image

### Logged-out extension
image

### Logged-in extension
Note that the Penumbra rays are shifted down a bit from before, but I assume that's OK.

image

### Successful login
image

### DIFF
--- a/packages/ui/components/ui/splash-page.tsx
+++ b/packages/ui/components/ui/splash-page.tsx
@@ -1,12 +1,20 @@
 import { ReactNode } from 'react';
-import { Card, CardContent, CardHeader, CardTitle } from './card';
+import { Card, CardContent, CardDescription, CardHeader, CardTitle } from './card';
 import { CompressedVideoLogo } from './logo/compressed-video';
 import { FadeTransition } from './fade-transition';
 
-export const SplashPage = ({ title, children }: { title: string; children: ReactNode }) => {
+export const SplashPage = ({
+  title,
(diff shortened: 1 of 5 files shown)


packages/ui/components/ui/splash-page.tsx
```diff
@@ -1,12 +1,20 @@
...
+  children,
+}: {
+  title: string;
+  description?: ReactNode;
+  children: ReactNode;
+}) => {
   return (
     <FadeTransition>
-      <div className='absolute inset-0 flex w-screen items-center justify-center'>
+      <div className='absolute inset-0 z-[-1] flex w-screen items-center justify-center'>
```
jessepinho, line 17: Fixes an issue where, for some reason, the background video was floating in front of the page content.

grod220 (approved):

grod220:
About to cut a release of the extension. Definitely would be good to get this fix in before. So going to merge 👍

PR #589, opened 2024-02-21
add tests for notes-for-voting

Valentine1898:
This test does not really test anything, because all the filtering logic takes place in the storage package and is covered by the tests there. 

Should we refactor to move the filtering logic to the rpc layer?

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/notes-for-voting.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/notes-for-voting.test.ts
@@ -1,0 +1,71 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import {
+  NotesForVotingRequest,
+  NotesForVotingResponse,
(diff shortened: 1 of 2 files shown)


grod220:
> Should we refactor to move the filtering logic to the rpc layer?

I think this depends. Is the view rpc method the only one that is requiring this filtering? Or do we think it will be used by something else in the future? If we anticipate it being shared later, we should leave it in the db layer. If not, we should move it to the rpc method.

Valentine1898:
> I think this depends. Is the view rpc method the only one that is requiring this filtering? Or do we think it will be used by something else in the future? If we anticipate it being shared later, we should leave it in the db layer. If not, we should move it to the rpc method.

In that case, I think we have reason to leave all the logic in the db layer. 
Digging into rust, I found that `notes_for_voting` is also used to build transaction plan (wasm planner currently ignores this)

grod220 (approved):
Sounds good to me

PR #590, opened 2024-02-21
Show a user's unstaked tokens on the validators page

### LINKED ISSUE #586: Show a user's unstaked tokens on the validators page
Per @hdevalence 's [comment](https://github.com/penumbra-zone/web/pull/569#issuecomment-1957738474) on #569:

> The first block should be titled "Unstaked Tokens" and have one row, showing the user's balance of the UM token.

"Unstaked tokens" = a user's balance of penumbra. This should be displayed on a per-account basis, like how the Assets table shows assets by account ([context](https://discord.com/channels/824484045370818580/1044457038886998137/1209982003240505424)).

jessepinho:
image

This is the first step of reorganizing the Validators page to focus on tokens. See the tracking issue #463 for what's coming next.

This PR just displays the user's penumbra balances for any account that holds penumbra. In the upcoming PRs, I'll add delegation and unbonding tokens to each account, but I wanted to keep these PRs bite-sized.

Closes #586

### DIFF
--- a/apps/webapp/src/components/staking/validators-table.tsx
+++ b/apps/webapp/src/components/staking/validators-table.tsx
@@ -1,4 +1,15 @@
-import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from '@penumbra-zone/ui';
+import {
+  Card,
+  CardContent,
+  CardHeader,
+  CardTitle,
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@penumbra-zone/ui';
 import { getValidator } from '@penumbra-zone/types';
 import { useValidatorInfos } from './use-validator-infos';
 import { Oval } from 'react-loader-spinner';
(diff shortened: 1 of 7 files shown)


apps/webapp/src/components/staking/validators-table.tsx
```diff
@@ -14,41 +25,48 @@ export const ValidatorsTable = () => {
...
-            <TableCell colSpan={HEADERS.length}>
-              There was an error loading validators. Please reload the page.
-            </TableCell>
-          </TableRow>
-        )}
+    <Card>
+      <CardHeader>
+        <CardTitle>Active validators</CardTitle>
+      </CardHeader>
+      <CardContent>
```
jessepinho, line 32: Only changes to this file are adding the `<Card />` wrapper around the table.

grod220 (commented):
Code looks good to me. But growing the page down with each additional piece of content in this tracking issue feels like it will get noisy fast.

Think our tab component is a good candidate for doing sub-routing on this page:

Screenshot 2024-02-22 at 2 24 26 PM

These tabs should match url routes like:

```
/staking/unstaked
/staking/staked
/staking/unbonding
/staking/validators
```

hdevalence:
We should keep each category of staked funds on the same view, not split them across tabs. If we need to split up we should be splitting up by account, rather than interleaving, like I suggested in Discord.

grod220:
> We should keep each category of staked funds on the same view, not split them across tabs.

Isn't the screenshot an example of two categories in the same view?

Issue #542, opened 2024-02-15
Issue delegation actions


grod220:
Transactions for:
  - Delegate to a validator
  - Undelegate from validator (show unbonding status)
  - Undelegate claim (after unbonding)

PR #591, opened 2024-02-21
Show staked tokens on the staking page

### LINKED ISSUE #587: Show a user's staked tokens on the validators page
Per @hdevalence 's [comment](https://github.com/penumbra-zone/web/pull/569#issuecomment-1957738474) on #569:

> The second block should be titled "Staked Tokens" and have one row for each delegation token for which the user has a nonzero balance. If there are no rows, the block should be omitted. Each row should show the amount of delegation tokens, the equivalent amount of staking tokens and USD (this functionality being provided by the equivalent_values feature in the valueview, which the extension needs to but does not yet implement), information about the validator, and buttons for "Delegate" and "Undelegate".

This should be displayed on a per-account basis, like how the Assets table shows assets by account ([context](https://discord.com/channels/824484045370818580/1044457038886998137/1209982003240505424)).

### LINKED ISSUE #588: Show a user's unbonding tokens on the validators page
Per @hdevalence's [comment](https://github.com/penumbra-zone/web/pull/569#issuecomment-1957738474) on #569:

> The third block should be titled "Unbonding Tokens" and have one row for each unbonding token for which the user has a nonzero balance. If there are no rows, the block should be omitted. Each row should show the amount of unbonding tokens (initially we can ignore penalties and treat these as 1:1 for UM tokens), information about the validator, and a time-based estimate of when unbonding will finish. The row should have a button labeled "Claim" that does the UndelegateClaim, this should be disabled until unbonding is believed to be finished.

This should be displayed on a per-account basis, like how the Assets table shows assets by account ([context](https://discord.com/channels/824484045370818580/1044457038886998137/1209982003240505424)).

jessepinho:
🚨 300+ line changes are test files

image

## In this PR
- Build a hook that provides the data structure needed for rendering validators on a per-account basis.
- Reorganize some of the components related to rendering validators.
- Add tests for `assetPatterns`, and fix their regex for security implications.

## Punting for now
- The delegate / undelegate buttons don't do anything yet (but will after #542).
- The delegation/unbonding tokens currently display quite messily. I'll need to revisit this, but this PR is already a bit large so wanted to get it merged first.

Closes #587
Closes #588


apps/webapp/src/components/staking/accounts/staking-actions.tsx
```diff
@@ -0,0 +1,60 @@
...
+  unbondingTokens,
+  canDelegate,
+}: {
+  delegationTokens?: ValueView;
+  unbondingTokens?: ValueView;
+  canDelegate: boolean;
+}) => {
+  const hasTokens = !!delegationTokens || !!unbondingTokens;
+
+  const handleClickAction = () => alert('Not yet implemented; coming soon!');
```
jessepinho, line 21: #542

apps/webapp/src/components/staking/validator-info-row.tsx
```diff
@@ -27,12 +25,8 @@ export const ValidatorInfoRow = ({
       )}
     </TableCell>
 
-    {/** @todo: Render an icon for each state */}
-    <TableCell>{getStateLabel(validatorInfo)}</TableCell>
-
-    {/** @todo: Render an icon for each state */}
-    <TableCell>{getBondingStateLabel(validatorInfo)}</TableCell>
```
jessepinho, line 34: Per @hdevalence these aren't really needed

apps/webapp/src/components/staking/validators-table.tsx
```diff
@@ -1,72 +1,68 @@
...
-export const ValidatorsTable = () => {
-  const { loading, error, validatorInfos, votingPowerByValidatorInfo } = useValidatorInfos();
+interface ValidatorsTableProps extends ValidatorInfoContext {
+  /**
+   * Content to display inside the Staking cell.
+   */
+  renderStakingActions: (validatorInfo: ValidatorInfo) => ReactNode;
+}
 
+export const ValidatorsTable = ({
```
jessepinho, line 18: Two main changes in this file:

1. Remove the `<Card>` wrapper, since that's now provided by either `<AllValidators />` or `<Account />`.
2. Accept props for the validator infos/etc., rather than grabbing them from the context. This is needed because the parent components may be filtering the validator infos (such as the `<Account />` component, which only renders the validator infos for a given account).

apps/webapp/src/components/staking/accounts/account.tsx
```diff
@@ -1,42 +0,0 @@
-import { getDisplayDenomFromView } from '@penumbra-zone/types';
```
jessepinho, line 1: Moved to `account/index.tsx`

packages/constants/assets/index.ts
```diff
@@ -20,15 +22,34 @@ export interface UnbondingCaptureGroups {
...
+    /^delegation_(?<bech32IdentityKey>penumbravalid1(?<id>[a-zA-HJ-NP-Z0-9]+))$/,
   ),
   proposalNft: new RegExp(/^proposal_/),
+  /**
+   * Unbonding tokens have only one denom unit, which is the base denom. Hence
+   * the extra `u` at the beginning.
+   */
   unbondingToken: new RegExp(
-    /.*unbonding_epoch_(?<epoch>[0-9]+)_penumbravalid1(?<id>[a-zA-HJ-NP-Z0-9]+)$/,
+    /^uunbonding_epoch_(?<epoch>[0-9]+)_(?<bech32IdentityKey>penumbravalid1(?<id>[a-zA-HJ-NP-Z0-9]+))$/,
```
grod220, line 52: Can you add `bech32IdentityKey` to `interface UnbondingCaptureGroups`?

grod220 (commented):
I agree, the design feels a bit clunky. But honestly, no obvious notes from me how to improve it. We don't really have reference designs for the direction we are going.

In general, is the idea that the top boxes will show you who you've delegated to and those buttons allow you to delegate more to them specifically? When you click that button, will that give you a dialog box (or send you to a page) that prefills that validator in? Is a table the best design if the width doesn't allow all of the data to fit?

Overall, I think I'm far more accustomed to staking pages that say, look like Dimension's:
Screenshot 2024-02-23 at 12 51 04 PM

apps/webapp/src/components/staking/accounts/account/helpers.ts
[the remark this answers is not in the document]
jessepinho: Yeah I'd thought about that too, but this feels different because it's actually converting the value from one type to another, kind of like the `translators` directory. I don't know ... It certainly would be quite ergonomic to be able to chain this to a series of getters. Maybe there could be a class of getters that starts with `as`, to indicate that they transform a value rather than just getting it? e.g., `asBech32String`?

packages/constants/assets/index.ts
jessepinho: yep, done

apps/webapp/src/components/staking/validator-info-provider.tsx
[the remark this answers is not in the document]
jessepinho: Whoops 🤦🏻‍♂️ Will do

apps/webapp/src/components/staking/validators-table.tsx
```diff
@@ -1,72 +1,68 @@
...
-    </Card>
+        {showValidators &&
+          validatorInfos.map(validatorInfo => (
+            <ValidatorInfoRow
+              key={getValidator(validatorInfo).name}
+              loading={loading}
+              validatorInfo={validatorInfo}
+              votingPowerByValidatorInfo={votingPowerByValidatorInfo}
+              staking={renderStakingActions(validatorInfo)}
+            />
```
hdevalence, line 63: We should work towards the design in https://github.com/penumbra-zone/web/pull/569#issuecomment-1957738474 , this should be a component for rendering a delegation token, and it should be the same component as is used for the user's delegations.

jessepinho:
@hdevalence, sort of :) I started with the plan of creating a component that accepted both a value view and validator info, per your comment. However, I realized that it would mean that validators would be displayed in duplicate if a user has both unbonding tokens and delegation tokens for a given validator.

So instead, I grouped the delegation and unbonding tokens for a given validator together with the relevant validator info, so that each validator is only rendered once per account that has delegation OR unbonding tokens for it.

That said, is your concern more with the visual presentation of the tokens + validators? For the moment, I was just using the existing validator info rows from my previous PR, and then adding the token info in the rightmost column. I could try to redesign it a bit to highlight the token data more than the validator info, but that might be tricky since the only information to display about delegation/unbonding tokens is their balances.

jessepinho:
Sorry, one more note @hdevalence — I _would_ like to redesign the ValueViewComponent for delegation/unbonding tokens specifically, so it's a bit more readable and portable. This design isn't very usable:
image
But to me, that's a separate concern from how we group `ValueViewComponent` with validator info in this PR.

packages/constants/assets/index.ts
jessepinho: Yep, done

apps/webapp/src/state/staking.test.ts
```diff
@@ -0,0 +1,96 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
jessepinho, line 1: This file is pretty much moved from `use-validator-infos.test.ts`, but with some changes to account for using Zustand instead of context

apps/webapp/src/components/staking/validator-info-provider.tsx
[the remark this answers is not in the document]
jessepinho: @grod220 OK, I think this was the last step. Commit is [here](https://github.com/penumbra-zone/web/pull/591/commits/931d0fc17e6ceae44b3fd08f19068317dddff7f2).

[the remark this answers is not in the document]
jessepinho: @grod220 ah crap, hang on, gotta fix a few issues I missed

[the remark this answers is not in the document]
jessepinho: @grod220 OK ready now, see [here](https://github.com/penumbra-zone/web/pull/591/commits/50bf14310c4799e2bca2e3daa826a63679956a0a)

jessepinho:
> In general, is the idea that the top boxes will show you who you've delegated to and those buttons allow you to delegate more to them specifically? When you click that button, will that give you a dialog box (or send you to a page) that prefills that validator in? Is a table the best design if the width doesn't allow all of the data to fit?

Yeah, that's the idea. That said, I was expecting to do a bit of design refinement as the final step. To your point, it looks quite squished together. I could, for example, move the educational panel somewhere else to let the table be full-width. But for now, I'm just focusing on functionality, and then refining the design afterward. That OK?

hdevalence:
General comment: in retrospect, it would have been good to have done more upfront design on this, so we overran our headlights a bit. There are a bunch of different points here, I'll comment on each of them separately and then maybe we could have a synchronous session to pair on it a bit!

hdevalence:
> However, I realized that it would mean that validators would be displayed in duplicate if a user has both unbonding tokens and delegation tokens for a given validator. So instead, I grouped the delegation and unbonding tokens for a given validator together with the relevant validator info, so that each validator is only rendered once per account that has delegation OR unbonding tokens for it.

Yeah, this was an intended part of the design, for two reasons:

1. The common case is that a user doesn't have any unbonding tokens, so keeping them separate allows us to not display them.
2. The unbonding tokens are actually different from delegation tokens -- they're not getting any yield, they're denominated differently, they have different actions to take on them -- so they should be displayed separately.

hdevalence:
> Sorry, one more note @hdevalence — I would like to redesign the ValueViewComponent for delegation/unbonding tokens specifically, so it's a bit more readable and portable. This design isn't very usable: (img) But to me, that's a separate concern from how we group ValueViewComponent with validator info in this PR.

Yes and no. The design there isn't very usable, I agree, but the concern of "how do we represent delegation tokens" is very much relevant to this work, since the whole problem we are trying to solve here (in creating the staking page) is how to render delegation tokens.

One of the key points is that we don't necessarily trust the validator descriptions, since they're self-reported. The reason that text is showing the Bech32 encoding is because it's not a known validator, and we haven't registered metadata authorizing its name. So if we showed the name to users, they could be scammed into thinking that one validator's tokens were another's.

So while I agree that the specifics of the the `ValueViewComponent` changes are separate, they're related conceptually. This work is difficult because it's the first time that we've tried to grapple with the underlying (technical) design realities that in Penumbra, all the user state is represented by structured tokens, and we need to design a way to present those tokens to the user. The design work we overran was doing more detailed planning on how that should happen.

hdevalence:
sketch, bringing this closer to the dymension UI that @grod220 posted and keeping it token focused

Screenshot 2024-02-23 at 12 10 51 PM

- for the "active validators" this would show 0 balances instead, but that seemed complicated to make in a sketch
- we'd want to have the delegate/undelegate buttons activated or not internally to the component based on the value
- the logo would be read from the metadata, or maybe we could do some kind of custom identicon for delegation tokens with a colored border keyed by the full identity string?

hdevalence:
Updated sketch after pairing with @jessepinho:

Screenshot 2024-02-23 at 7 22 01 PM

Notes:

- Each row is an opaque component like `DelegationViewComponent` or something that takes a `ValueView` and a `ValidatorInfo`. Later, we'll have the extension fill in the `ValidatorInfo` as the `Any` in the `ValueView`, so we won't have to do any loading work.
- We render a list of delegation tokens, to highlight that each validator is 1:1 related to a delegation token, and that in Penumbra, all staking is done through LSTs. The component can treat a missing value as 0 value, in which case we just drop the balance display.
- The value is pulled out of the equivalent values list in the `ValueView`
- The icon for the delegation token is pulled from its `Metadata`
- Unbonding tokens are shown as the amount of UM unbonding (since unbonding tokens are 1:1 with UM in the happy path), together with a countdown timer and a claim button that's enabled when unbonding is complete (the timer can be inaccurate, an estimate is good enough)

grod220 (approved):

Issue #513, opened 2024-02-14
Create standalone static site for node status page


hdevalence:
As part of https://github.com/penumbra-zone/penumbra/pull/3522 , where we'd like to have the `pd` RPC endpoint serve browser-useful content, it would be useful to have a static landing page we can install as the root content.

This should be a static site with the following content:

- A progress bar for the node's sync status, using [this RPC](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.util.tendermint_proxy.v1#penumbra.util.tendermint_proxy.v1.TendermintProxyService.GetStatus) to get the [`SyncInfo` data](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.util.tendermint_proxy.v1#penumbra.util.tendermint_proxy.v1.SyncInfo). (It could update on a timer, say 5s). Since this is intended to be served off the RPC, the client should be aimed at the same HTTP endpoint serving the content.
- Below that, a link labeled "Frontend -->" that goes to `/app/`, where we will have installed minifront.

grod220:
This seems to be a good candidate to use a react "meta-framework". Related: https://github.com/penumbra-zone/web/issues/310#issuecomment-1904608594

PR #592, opened 2024-02-22
Add node status page

grod220:
Closes https://github.com/penumbra-zone/web/issues/513

https://github.com/penumbra-zone/web/assets/16624263/45abd9cd-d463-4231-a1d7-26298ff6701b


hdevalence:
This looks amazing!

apps/node-status/src/components/frontend-referral.tsx
```diff
@@ -0,0 +1,14 @@
+import { Button } from '@penumbra-zone/ui';
+import { devFrontend, prodFrontend } from '../constants.ts';
+
+export const FrontendReferral = () => {
+  const onClickHandler = () => {
+    window.open(import.meta.env.MODE === 'production' ? prodFrontend : devFrontend);
```
jessepinho, line 6: why not use `process.env.NODE_ENV`? Not available in Vite?

apps/node-status/src/fetching/refetch-hook.ts
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    const interval = setInterval(revalidate, REFETCH_INTERVAL);
+    return () => clearInterval(interval); // Clear the interval when the component is unmounted
+  }, [revalidate]);
+};
+
+const VISIBILITY_STATE_CHANGE_DELAY = 800;
+
+// Meant to slow down the state transition from loading to idle so the UI can show a discernible spinner
+export const useDeceleratedFetchState = () => {
```
jessepinho, line 18: This hook name threw me off a bit. What about something like `useDelayedIsLoading`? Hm... maybe it's not better. It just took me a minute to realize what this hook was actually doing.

apps/node-status/src/components/frontend-referral.tsx
grod220: Because this code will be executed in a browser environment, getting env variables that way will throw. Vite provides this as a kind analogous way to access env variables with a client side app.

apps/node-status/src/fetching/refetch-hook.ts
grod220: Renamed 👍

apps/node-status/src/components/node-info.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <span className='ml-2'>App: {nodeInfo.protocolVersion.app.toString()}</span>
+        </div>
+      )}
+      <div className='flex flex-col'>
+        <strong>Listen Address</strong>
+        <span className='ml-2'>{nodeInfo.listenAddr}</span>
+      </div>
+      <div className='flex flex-col'>
+        <strong>Channels</strong>
+        <span className='ml-2'>{uint8ArrayToBase64(nodeInfo.channels)}</span>
```
grod220, line 46: Swapping to `uint8ArrayToString`

Issue #495, opened 2024-02-13
IDB transaction is automatically closed while streaming transactionInfo


Valentine1898:
Issue is related to #481 

- If we only have 1 transaction, we won't get an error  
- If we have 2 or more transactions, we will get the error `Failed to execute 'continue' on 'IDBCursor': The transaction has finished `
-  transaction is always closed after the second record is transferred 
- `IterateAssetsMetadata()` tests do not cause an error 

https://github.com/jakearchibald/idb?tab=readme-ov-file#transaction-lifetime 
Most likely, there is an await call somewhere at the transport level that causes the IDB transaction to close prematurely

grod220:
Running into this for `unclaimedSwaps` endpoint as well. We should create a test that replicates this error so when its fixed, we are defended from it coming back.

grod220:
Yikes. It looks like our internal messaging is not compatible with indexdb async iterables due to its restrictions on using async in between calls. Take a look at the call stack on the second iteration:

<img width="1017" alt="Screenshot 2024-02-13 at 11 46 42 AM" src="https://github.com/penumbra-zone/web/assets/16624263/c50ff829-0cce-47dd-a709-32bd70494a41">

It appears an await gets added to pass the result on to the client, which kills the idb connection.

grod220:
For example, in `unclaimed-swaps.ts` this fixes it:

```typescript
  const swaps: { swap: SwapRecord }[] = [];
  for await (const swap of indexedDb.iterateSwaps()) {
    if (swap.heightClaimed !== 0n) continue;
    swaps.push({ swap });
  }

  for (const swap of swaps) {
    yield swap;
  }
```

Which is sad, because this basically is what we had before. This indexeddb restriction is quite unfortunate. It prevents us from having a true end-to-end way of streaming database results. I wonder if changing our transport streaming to async iterables would help 🤔 . But doesn't seem to be worth it at this stage.

turbocrime:
> Yikes. It looks like our internal messaging is not compatible with indexdb async iterables due to its restrictions on using async in between calls.
> [...]
> It appears an await gets added to pass the result on to the client, which kills the idb connection.

it is frustrating that we can't await while iterating idb - but i think there would be an effective boundary at chrome runtime transport. i could try and refactor to address this. removing the rethrow wrapper is the most direct step. we'd have to do the try-catch wrapper in every impl function directly but i think this wouldn't be too bad

turbocrime:
[...]

PR #593, opened 2024-02-22
add tests for owned-position-ids

Valentine1898:


packages/router/src/grpc/view-protocol-server/owned-position-ids.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+    if (req.tradingPair && !req.tradingPair.equals(position.phi?.pair)) {
+      continue;
+    }
+    responses.push({ positionId: PositionId.fromJson(positionRecord.id) });
+  }
+  // direct streaming from indexed-db is not supported because it leads to premature closing of IDB transaction
+  for (const response of responses) {
+    yield response;
+  }
```
Valentine1898, line 28: `ownedPositionIds` was broken for the same reason as #495 
At the same time, I can't do the temporary buffer trick for both `swaps` and `notes` 
https://github.com/penumbra-zone/web/blob/ad3eab64a80f0cdcee3304168c9d46738edfd100/packages/storage/src/indexed-db/index.ts#L207-L216
Because the `POSITIONS` table stores `PositionRecord`, which is not a buf message type. 

Also, I think the buffer solution, although it looks cleaner, results in reading all indexed-db records into memory, just as if we just used `getAll()`

grod220: Ah, I see. We might as well do a `getAll()`. 

Actually, @turbocrime, can you weigh in here? I'm not sure there is a case to continue doing `streamToGenerator` as it just loads everything into memory all at once anyway.

turbocrime: streamToGenerator does not load everything into memory all at once - it streams asynchronously. additionally, streamToGenerator and readable streams are not restricted to protobuf message types.

i wrote idbcursorsource to use a protobuf message type, which i guess was a restrictive choice. but it can be rewritten to support more things, or another simple stream source can be written.

it may also be possible to build indicies for this table that allow more specific queries.

turbocrime: if you store jsonified position in idb instead of creating `PositionRecord`, which contains only a jsonified position and a copy of the table key, the pattern is the same as the other functions.

turbocrime: you could write a stream like

```ts
  async *getOwnedPositionIds(
    positionState: PositionState | undefined,
    tradingPair: TradingPair | undefined,
  ): AsyncGenerator<PositionId, void> {
    yield* streamToGenerator(
      new ReadableStream({
        start: async cont => {
          let cursor = await this.db.transaction('POSITIONS').store.openCursor();
          while (cursor) {
            const position = Position.fromJson(cursor.value.position);
            if (positionState && !positionState.equals(position.state)) continue;
            if (tradingPair && !tradingPair.equals(position.phi?.pair)) continue;
            cont.enqueue(PositionId.fromJson(cursor.value.id));
            cursor = await cursor.continue();
          }
          cont.close();
        },
      }),
    );
  }
  ```

Comment (Valentine1898) [reply packages/router/src/grpc/view-protocol-server/owned-position-ids.ts]: Okay, `streamToGenerator` does not cause all records to be loaded into memory. 
I implemented getOwnedPositionIds() following the example from @turbocrime , and modified the tests.

**Comment by Valentine1898** [comment]
rebased on main, awaiting review

File: packages/storage/src/indexed-db/index.ts [hunk 2]
```diff
@@ -365,18 +365,26 @@ export class IndexedDb implements IndexedDbInterface {
...
+          if (positionState && !positionState.equals(position.state)) {
+            cursor = await cursor.continue();
+            continue;
+          }
+          if (tradingPair && !tradingPair.equals(position.phi?.pair)) {
+            cursor = await cursor.continue();
+            continue;
+          }
+          cont.enqueue(PositionId.fromJson(cursor.value.id));
+          cursor = await cursor.continue();
```
turbocrime, line 383:
```suggestion
          if (
            (!positionState || positionState?.equals(position.state)) &&
            (!tradingPair || tradingPair?.equals(position.phi?.pair))
          )
            cont.enqueue(PositionId.fromJson(cursor.value.id));
          cursor = await cursor.continue();
```

turbocrime (approved):

PR #596, opened 2024-02-22
import lint

turbocrime:
this has been bugging me and now it can bug everyone


packages/eslint-config-custom/index.js
turbocrime, line 1: the rules

PR #599, opened 2024-02-23
Use 'vitest' instead of 'vitest run' to allow watching

jessepinho:
`vitest` defaults to watch mode when in development, but non-watch mode in CI. `vitest run` always exits after a run.

This PR switches from using `vitest run` to `vitest`, so that we automatically use watch mode in local development when running `pnpm test` inside any package directory.


TalDerei (approved):
Looks like `vitest` aliases to `vitest watch` in development mode according to the documentation. It might also be worth experimenting with the `--coverage` flag for a more extensive coverage report.

turbocrime:
this broke our `turbo.json` config, which expects the `test` script in every package to be non-persistent and be cacheable

PR #604, opened 2024-02-26
Refactor/redesign of staking page

jessepinho:
Per some convos with Henry, we decided the Staking page should be refactored and redesigned to treat validators as tokens. We want users to intuit that validators essentially _ARE_ tokens, from a delegator's point of view.

To that end, the staking page now renders a list of `ValueView`s instead of a list of `ValidatorInfo`s. Each `ValueView` that contains a delegation token includes, in its `extendedMetadata` key, a `ValidatorInfo` of the validator represented by the token. For those validators for which the current account holds no delegation tokens, `ValueView`s with zero balances are generated on the fly.

We also no longer separate out "all validators" from "validators you've staked." There's now just a single list of delegation tokens (AKA validators), which each show the user's balance if the user has one.

image

image

## In this PR
- Create a `getDelegationsForAccount` fetcher. This will eventually be replaced by an RPC method, but for now, it gets all validators and then yields a `ValueView` for each. The `ValueView` contains the `ValidatorInfo` in its `extendedMetadata` key, as well as the balance of the delegation token for that validator. (If the account has no delegation tokens, the `ValueView` is generated on the fly with a zero balance.)
- Refactor the staking page to be based on a list of `ValueView`s with `ValidatorInfo`s embedded in them.

## Later PRs
- Show unbonding amount at top of page.
- Actually issue delegate/undelegate tokens when clicking those buttons.
- Move `getDelegationsForAccount` into an RPC call.


apps/webapp/src/components/staking/account/delegation-value-view/staking-actions.tsx
```diff
@@ -0,0 +1,41 @@
+import { Button } from '@penumbra-zone/ui';
+
+/**
+ * Renders Delegate/Undelegate buttons for a validator.
+ */
+export const StakingActions = ({
```
jessepinho, line 6: image

apps/webapp/src/components/staking/account/delegation-value-view/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Renders a `ValueView` that contains a delegation token, along with the
+ * validator that the token is staked in.
+ *
+ * @todo: Depending on the outcome of
+ * https://github.com/penumbra-zone/penumbra/issues/3882, we may be able to
+ * remove `votingPowerAsIntegerPercentage`.
+ */
+export const DelegationValueView = ({
```
jessepinho, line 15: image

apps/webapp/src/components/staking/account/delegation-value-view/validator-info-component.tsx
```diff
@@ -0,0 +1,77 @@
...
+import {
+  Identicon,
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@penumbra-zone/ui';
+import { CopyToClipboardIconButton } from '@penumbra-zone/ui/components/ui/copy-to-clipboard-icon-button';
+
+export const ValidatorInfoComponent = ({
```
jessepinho, line 17: image

apps/webapp/src/fetchers/balances/index.ts
```diff
@@ -35,8 +35,8 @@ const getBalances = ({ accountFilter, assetIdFilter }: BalancesProps = {}) => {
   return streamToPromise(iterable);
 };
 
-export const getAssetBalances = async (): Promise<AssetBalance[]> => {
-  const balances = await getBalances();
+export const getAssetBalances = async (balanceProps?: BalancesProps): Promise<AssetBalance[]> => {
```
jessepinho, line 38: `getAssetBalances` now takes optional arguments so that we can filter by address index or asset ID if desired

apps/webapp/src/fetchers/staking.ts
```diff
@@ -1,6 +1,84 @@
...
+      yield withValidatorInfo;
+    } else {
+      yield new ValueView({
+        valueView: {
+          case: 'knownAssetId',
+          value: {
+            amount: {
+              hi: 0n,
+              lo: 0n,
+            },
```
jessepinho, line 77: Not sure what to populate `metadata` with here, so I left it `undefined`. Any suggestions? (I'd prefer to punt that to a future PR, though.)

apps/webapp/src/components/staking/account/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const getVotingPowerAsIntegerPercentage = (
+  votingPowerByValidatorInfo: Record<string, VotingPowerAsIntegerPercentage>,
+  delegation: ValueView,
+) =>
+  votingPowerByValidatorInfo[
+    bech32IdentityKey(getIdentityKeyFromValidatorInfo(getValidatorInfoFromValueView(delegation)))
+  ];
+
+export const Account = () => {
```
jessepinho, line 23: image

packages/ui/components/ui/tx/view/value.tsx
```diff
@@ -41,7 +41,7 @@ export const ValueViewComponent = ({
       <div className='flex items-center gap-1 text-base font-bold'>
         {showIcon && <AssetIcon metadata={metadata} />}
         {showValue && <span>{formattedAmount}</span>}
-        {showDenom && <span>{symbol}</span>}
+        {showDenom && <span className='truncate'>{symbol}</span>}
```
jessepinho, line 44: Eventually, we'll probably want to re-label delegation tokens to be briefer, but this is good enough for now.

## Before
image


## After
image

apps/webapp/src/components/staking/account/delegation-value-view/validator-info-component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <TooltipProvider delayDuration={0}>
+      <div className='flex items-center gap-4'>
+        <div className='shrink-0'>
+          <Identicon
+            uniqueIdentifier={bech32IdentityKey(identityKey)}
+            size={48}
+            className='rounded-full'
+            type='gradient'
+          />
```
hdevalence, line 40: can this use the same identicon logic as the `ValueView` uses for rendering the delegation token in other places, so that the icon is the same regardless of whether it's rendered in the context of the staking page or in an asset list?

also, that way if we have a pre-registered Metadata for that delegation token (eg the validator operator sent us a PR) then we'll automatically use any logo they set.

apps/webapp/src/fetchers/staking.ts
hdevalence: ideally it should be the token metadata for the delegation, from the extension's token registry -- and we'd really like that info to take priority over the validator info if possible, because it's data we check in to the extension's source code rather than just something someone posted to the chain

hdevalence:
This looks amazing!

apps/webapp/src/fetchers/staking.ts
grod220: You can do

```
const { denomMetadata } = await viewClient.assetMetadataById({ assetId });
```

but it's not super clear to me where you'd get this assetId if it's in the `!delegation` part of the clause.

packages/ui/components/ui/tx/view/value.tsx
grod220: Should we make this a bit bigger? Usually asset icons are the size of the text

Screenshot 2024-02-27 at 9 09 46 AM

grod220 (approved):
Great job!

packages/ui/components/ui/tx/view/value.tsx
jessepinho: Whoops, will fix

jessepinho: image
[Fixed](https://github.com/penumbra-zone/web/pull/604/commits/d1ecc70552d251a71eb4af362d9848641d0c4d48)

apps/webapp/src/components/staking/account/delegation-value-view/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  votingPowerAsIntegerPercentage?: number;
+  /**
+   * Whether the user can delegate to this validator (i.e., whether the user has
+   * a nonzero balance of the staking token with which to stake in this
+   * validator).
+   */
+  canDelegate: boolean;
+}) => {
+  const delegationBalance = getAmount.optional()(valueView);
+  const canUndelegate = delegationBalance ? !!joinLoHiAmount(delegationBalance) : false;
```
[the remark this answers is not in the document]
jessepinho: Not exactly — I want `canUndelegate` to be a `boolean` type. The second line you wrote makes it a `bigint | false` type, which would then cause the compiler to complain when it's passed to `<StakingActions />` below:
image

The idea is, we can only undelegate if A) we have a non-undefined delegation balance and B) that delegation balance is greater than 0.

apps/webapp/src/components/staking/account/delegation-value-view/validator-info-component.tsx
```diff
@@ -0,0 +1,81 @@
...
+  votingPowerAsIntegerPercentage,
+}: {
+  validatorInfo: ValidatorInfo;
+  votingPowerAsIntegerPercentage?: number;
+}) => {
+  const validator = getValidator(validatorInfo);
+  const identityKey = getIdentityKeyFromValidatorInfo(validatorInfo);
+
+  return (
+    <TooltipProvider delayDuration={0}>
```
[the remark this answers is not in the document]
jessepinho: Good call. [Done](https://github.com/penumbra-zone/web/pull/604/commits/bb71a99ac029f5d1de0fa8c7c6fc97c4fa74e2d2).

apps/webapp/src/components/staking/account/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Card, CardContent, CardHeader, CardTitle } from '@penumbra-zone/ui';
+import { ValueView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
+
+const getVotingPowerAsIntegerPercentage = (
+  votingPowerByValidatorInfo: Record<string, VotingPowerAsIntegerPercentage>,
+  delegation: ValueView,
+) =>
+  votingPowerByValidatorInfo[
+    bech32IdentityKey(getIdentityKeyFromValidatorInfo(getValidatorInfoFromValueView(delegation)))
```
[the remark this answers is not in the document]
jessepinho: Whoops, [fixed](https://github.com/penumbra-zone/web/pull/604/commits/8dd9756d3395a6c06c61fa7c32e6297ee513b671).

apps/webapp/src/components/staking/account/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const Account = () => {
+  const { account, delegationsByAccount, unstakedTokensByAccount, votingPowerByValidatorInfo } =
+    useStore(stakingSelector);
+  const unstakedTokens = unstakedTokensByAccount.get(account);
+  const delegations = delegationsByAccount.get(account) ?? [];
+
+  return (
+    <div className='flex flex-col gap-4'>
+      {!!unstakedTokens && (
```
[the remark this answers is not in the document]
jessepinho: Not strictly. But I've generally adopted the habit of using `!!` to coerce values to booleans in templates. Otherwise, you get the occasional unintended markup. For example, let's say you have a numeric value `balance`, and you write the following markup:

```tsx
<>
  {balance && <span>Balance: {balance}penumbra</span>}
</>
```

If `balance` is equal to `0`, then the following will be rendered:

```tsx
<>
  0
</>
```

...because the return value of a statement `somethingFalsey && somethingTruthy` is `somethingFalsey`. And since React will render any value to the page except for `null`, `false`, and `true` (IIRC), `somethingFalsey` (in this case, `0`) will get rendered to the page unintentionally. (See the "Pitfall" note [here](https://react.dev/learn/conditional-rendering#logical-and-operator-).)

Hope this makes sense, haha. I can definitely remove the `!!` if you want; I just feel like it's generally safer to include it.

apps/webapp/src/components/staking/layout.tsx
```diff
@@ -1,36 +1,25 @@
...
+import { useEffect } from 'react';
 import { useStore } from '../../state';
+import { stakingSelector } from '../../state/staking';
+import { throwIfExtNotInstalled } from '../../utils/is-connected';
+import { Account } from './account';
 
-export const StakingLoader: LoaderFunction = async (): Promise<BalancesByAccount[]> => {
+export const StakingLoader = () => {
   throwIfExtNotInstalled();
+  void useStore.getState().staking.loadUnstakedTokensByAccount();
```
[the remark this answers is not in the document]
jessepinho: Ah, good call. [Fixed](https://github.com/penumbra-zone/web/pull/604/commits/a37a3e8c45efad3f7e247fe4a86b63190a6bf787).

apps/webapp/src/fetchers/staking.ts
jessepinho: @grod220 Yeah, exactly — I'm not sure how to look that up since we don't actually have any tokens. I'll punt this down to a later PR and make a ticket for it.

jessepinho: [Ticket](https://github.com/penumbra-zone/web/issues/617)

Issue #305, opened 2023-12-14
dashboard/transactions should display syncing status


turbocrime:
transactions view is limited by available sync. the transactions view should prominently indicate if syncing is not complete.

hdevalence:
it would be cool to have the progress bar at the top of the dashboard. there is probably also a typescript library for doing EWMA estimation of speed and time remaining. the progress bar in pcli uses `indicatif`, which has an EWMA-based estimator of how much time will be left that seems reasonably accurate.

PR #600, opened 2024-02-23
Add block sync time estimate

grod220:
Part 1 of https://github.com/penumbra-zone/web/issues/305. Changes extension preview. Next PR will be the web app update.

Calculates time remaining on sync via _Exponential Weighted Moving Average_

https://github.com/penumbra-zone/web/assets/16624263/87d93357-3bf7-473a-8bf0-9f5ffb7af734

PR #607, opened 2024-02-26
Web app block sync UI v2

grod220:
Closes https://github.com/penumbra-zone/web/issues/305
Part 2 of: https://github.com/penumbra-zone/web/pull/600

Adding a more prominent block sync status to web app. 

- `LastBlockSynced` renamed to `FullSyncHeight` to align with protobuf naming
- Bumped the indexdb version so locally stored sync height reset (given new key)


apps/extension/.env
```diff
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=23
+IDB_VERSION=24
```
grod220, line 3: New key for sync height, so easiest to wipe state

apps/webapp/src/components/header/top-row.tsx
grod220, line 1: Largely all moved code

packages/ui/lib/toast/transaction-toast.tsx
```diff
@@ -49,6 +50,14 @@ export class TransactionToast {
...
   constructor(private transactionClassification: TransactionClassification) {}
 
+  private get label(): string {
+    return TRANSACTION_LABEL_BY_CLASSIFICATION[this.transactionClassification];
+  }
+
+  private get shortenedTxHash(): string {
+    return this._txHash ? shorten(this._txHash, 8) : '';
+  }
+
```
jessepinho, line 60: Was this an automatic change from your IDE? I feel like it's generally the norm to put private methods last.

packages/types/src/indexed-db.ts
```diff
@@ -48,7 +48,7 @@ export interface IndexedDbInterface {
   ): AsyncGenerator<IdbUpdate<DBTypes, StoreName>, void>;
   constants(): IdbConstants;
   clear(): Promise<void>;
-  getLastBlockSynced(): Promise<bigint | undefined>;
+  getfullSyncHeight(): Promise<bigint | undefined>;
```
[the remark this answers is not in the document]
jessepinho: (And this should be changed everywhere this method is referenced 😬 `packages/storage/src/indexed-db/index.ts`, `packages/router/src/grpc/view-protocol-server/test-utils.ts`, etc.)

packages/storage/src/indexed-db/index.ts
```diff
@@ -139,7 +139,7 @@ export class IndexedDb implements IndexedDbInterface {
     await this.u.updateAll(txs);
   }
 
-  async getLastBlockSynced() {
+  async getfullSyncHeight() {
     return this.db.get('LAST_BLOCK_SYNCED', 'last_block');
```
jessepinho, line 143: Shouldn't this key also be changed to `FULL_SYNC_HEIGHT`?

apps/webapp/src/components/header/sync-status-section.tsx
```diff
@@ -0,0 +1,24 @@
+import { useStream } from '../../fetchers/stream';
+import { useMemo } from 'react';
+import { viewClient } from '../../clients/grpc';
+import { BlockSyncStatus } from '@penumbra-zone/ui';
+
+export const SyncStatusSection = () => {
```
jessepinho, line 6: Totally not necessary to change the UI for this PR, but my feeling is that the block sync takes up a lot of above-the-fold space for something that probably should be more back-of-mind for users. Instagram, for example, has a bar that goes across the _very_ top of the screen (which we could also augment with the words "Blocks synced ✔️"). Again, not necessary for this PR, but just thinking about a future improvement.



https://github.com/penumbra-zone/web/assets/1121544/1736c09e-adf4-4cdd-99fd-55f6b6838b30

packages/ui/components/ui/block-sync-status/syncing-state.tsx
```diff
@@ -0,0 +1,99 @@
...
+        status='in-progress'
+        value={(fullSyncHeight / latestKnownBlockHeight) * 100}
+        background={background}
+      />
+      <p className='font-mono text-sand'>
+        {fullSyncHeight}/{latestKnownBlockHeight}
+      </p>
+      <p
+        className={cn(
+          '-mt-1 font-mono text-sm text-sand transition-all duration-300',
```
jessepinho, line 48:
```suggestion
          'mt-1 font-mono text-sm text-sand transition-all duration-300',
```

jessepinho (approved):
LGTM. Looks like a couple things need to be renamed, especially the `LAST_BLOCK_SYNCED` IDB key and the `getfullSyncHeight` method (due to capitalization). Feel free to self-merge after those changes! Nice work.

apps/webapp/src/components/header/sync-status-section.tsx
grod220: Ah, you are so right. Relocating to the top.

PR #610, opened 2024-02-26
simplify workflows

turbocrime:
don't attempt any conditional CI, move all CI to buildjet. this reduces init and cache load/save times

### DIFF
--- a/.github/workflows/compile-wasm.yml
+++ b/.github/workflows/compile-wasm.yml
@@ -12,15 +12,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - id: compiled
-        uses: actions/cache@v3 # turbo cache, shared between buildjet and github
+        uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}
-            ${{ runner.os }}-turbo-
+          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}
       - uses: pnpm/action-setup@v2
-      - uses: buildjet/setup-node@v4 # node cache, distinct between buildjet and github
+      - uses: buildjet/setup-node@v4
         with:
           node-version: '21'
           cache: 'pnpm'
@@ -29,11 +27,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable # rust only on buildjet
         with:
           targets: wasm32-unknown-unknown
-      - uses: Swatinem/rust-cache@v2.7.3 # rust cache
-        with:
-          shared-key: wasm
-          cache-provider: 'buildjet'
-          workspaces: 'packages/wasm/crate'
       - uses: jetli/wasm-pack-action@v0.4.0 # preinstall wasm-pack
         with:
           version: 'latest'
--- a/.github/workflows/extension-publish.yml
+++ b/.github/workflows/extension-publish.yml
@@ -19,19 +19,17 @@ jobs:
 
   publish:
     environment: ext-publish
-    runs-on: ubuntu-latest
+    runs-on: buildjet-16vcpu-ubuntu-2204
     needs: compile
 
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/cache@v3
+      - uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-publish
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
       - uses: pnpm/action-setup@v2
-      - uses: actions/setup-node@v4
+      - uses: buildjet/setup-node@v4
         with:
           node-version: '21'
           cache: 'pnpm'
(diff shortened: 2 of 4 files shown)


.github/workflows/turbo-ci.yml
```diff
@@ -7,6 +7,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
grod220, line 12: What's this do?

grod220 (approved):

.github/workflows/turbo-ci.yml
turbocrime: cancel running jobs if it runs again

.github/workflows/compile-wasm.yml
```diff
@@ -12,15 +12,13 @@ jobs:
...
       - id: compiled
-        uses: actions/cache@v3 # turbo cache, shared between buildjet and github
+        uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}
-            ${{ runner.os }}-turbo-
+          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
```
jessepinho, line 18: Hmm, a couple notes:
1. [`hashFiles` can take multiple arguments](https://docs.github.com/en/actions/learn-github-actions/expressions#example-with-multiple-patterns), so you could rewrite the first part of the key as: `${{ hashFiles('**/Cargo.lock', 'pnpm-lock.yaml') }}`
2. But more importantly: Wouldn't it actually be better to cache based on the hash of all files in the `packages/wasm/crate` directory, so that we only compile WASM when any WASM files change? (And I'd remove the `github.ref` and `github.sha`, since those can change without the WASM code changing. Right?)

.github/workflows/extension-publish.yml
```diff
@@ -19,19 +19,17 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/cache@v3
+      - uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-publish
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
```
jessepinho, line 30: Same notes here

.github/workflows/turbo-ci.yml
```diff
@@ -109,20 +52,18 @@ jobs:
...
-        uses: actions/cache@v3
+        uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-built
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-built
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
```
jessepinho, line 64: Same notes here

.github/workflows/turbo-ci.yml
```diff
@@ -133,26 +74,51 @@ jobs:
...
+        uses: buildjet/cache@v3
         with:
           path: .turbo
-          key: ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-${{ matrix.test }}
-          restore-keys: |
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-built
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}
-            ${{ runner.os }}-turbo-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-${{ matrix.test }}
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
```
jessepinho, line 88: And here

.github/workflows/turbo-ci.yml
```diff
@@ -133,26 +74,51 @@ jobs:
...
+    runs-on: buildjet-16vcpu-ubuntu-2204
+    needs: turbo-compile
+    steps:
+      - uses: actions/checkout@v4
+      - id: tested
+        uses: buildjet/cache@v3
+        with:
+          path: .turbo
+          key: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-test:rust
+          restore-keys: ${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.ref }}-${{ github.sha }}-compiled
```
jessepinho, line 109: And here

jessepinho (commented):
Gabe already approved, so I'm not doing a thorough review — but just left a comment about the keys. Feel free to ignore if I'm missing something.

.github/workflows/compile-wasm.yml
turbocrime: i had trouble with getting `hashFiles` to work on multiple files, with directories, or with globs. it successfully returns an empty string on any failure, and i can't test these actions locally on my m3.

i agree such a change would be an improvement but i don't want to spend more time on this. PR it if you'd like

PR #614, opened 2024-02-27
ReadableStream polyfills

turbocrime:
remove iterable polyfill after https://groups.google.com/a/chromium.org/g/blink-dev/c/EseAQMDRppM

remove from polyfill after https://issues.chromium.org/issues/40267032


jessepinho (approved):
Nice! 👏🏻

Issue #525, opened 2024-02-15
Add Views to BalancesResponse


grod220:
Pairs with: https://github.com/penumbra-zone/penumbra/pull/3824

Web app currently takes assetId+amount response and transforms it to AddressView+ValueView:
https://github.com/penumbra-zone/web/blob/fa68d2d52fb135c661eb58ce6e96f69ec7324924/apps/webapp/src/fetchers/balances/index.ts#L44-L79

After new protos, this should move into balances view rpc method in extension.

While here, should delete `calculateUsdcValue()` in web app as extension now provides pricing.

PR #615, opened 2024-02-27
Web app -> Rpc for balances metadata

grod220:
Closes: https://github.com/penumbra-zone/web/issues/525

Now that the new protos are published https://github.com/penumbra-zone/penumbra/pull/3824, we can migrate a bunch of code relating to augmenting our asset balances with denom date to the extension.

As `AssetBalances` touches nearly everything, there are quite a few diffs.


apps/webapp/src/components/staking/accounts/account/use-validators-for-account.ts
grod220, line 1: Touches your staking code @jessepinho, so this may be out of date

apps/webapp/src/fetchers/balances/by-account.ts
```diff
@@ -2,28 +2,26 @@ import {
...
-  balances: AssetBalance[];
+  balances: BalancesResponse[];
 }
 
-const groupByAccount = (acc: BalancesByAccount[], curr: AssetBalance): BalancesByAccount[] => {
-  if (curr.address.addressView.case !== 'decoded') throw new Error('address is not decoded');
-  if (!curr.address.addressView.value.address) throw new Error('no address in address view');
-  if (!curr.address.addressView.value.index) throw new Error('no index in address view');
-
-  const index = curr.address.addressView.value.index;
```
grod220, line 18: Getter pattern for the win 💪

packages/router/package.json
```diff
@@ -16,8 +16,5 @@
     "@penumbra-zone/storage": "workspace:*",
     "@penumbra-zone/transport": "workspace:*",
     "@penumbra-zone/wasm": "workspace:*"
-  },
-  "devDependencies": {
-    "@types/chrome": "0.0.260"
```
grod220, line 21: Noticed these types deps in nearly all packages. Moved to top level .package.json,

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -1,81 +1,162 @@
...
-  const accounts: AccountMap = {};
+  const aggregator = new BalancesAggregator(ctx);
 
   for await (const noteRecord of indexedDb.iterateSpendableNotes()) {
     if (noteRecord.heightSpent !== 0n) continue;
+    await aggregator.add(noteRecord);
+  }
+
+  yield* aggregator.filteredResponses(req);
+};
```
grod220, line 45: Instead of one long function, abstracted this logic into a class below. Feels more readable, especially given the expanded functionality.

apps/webapp/src/state/send.test.ts
```diff
@@ -90,9 +88,9 @@ describe('Send Slice', () => {
 
     test('validate high enough amount validates', () => {
       const assetBalance = new Amount({ hi: 1n });
-      const state = produce(selectionExample, draft => {
-        draft.value.valueView.value!.amount = assetBalance;
-      });
```
Valentine1898, line 95: produce can only be called on things that are draftable: plain objects, arrays, Map, Set or classes that are marked with '[immerable]: true'
we can't use `immer` for `BalancesResponse`, so we modify it directly

Valentine1898 (approved):

Valentine1898:
@jessepinho I would appreciate your review of the changes related to staking

PR #618, opened 2024-02-27
refactor settings rpc page

turbocrime:
this now avoids clearing cache if the chainid at the new rpc endpoint remains the same

i refactored this when i expected to need to request permissions on a new endpoint after config change. that turned out to be unnecessary but i think this is better, provides some visual feedback, and prevents a long reset in some cases

### DIFF
--- a/apps/extension/src/routes/popup/settings/settings-rpc.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-rpc.tsx
@@ -1,34 +1,54 @@
-import { AppQuerier } from '@penumbra-zone/query';
+import { QueryService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/core/app/v1/app_connect';
+import { createGrpcWebTransport } from '@connectrpc/connect-web';
+import { createPromiseClient } from '@connectrpc/connect';
 import { FormEvent, useState } from 'react';
 import { Button, FadeTransition, Input } from '@penumbra-zone/ui';
-import { cn } from '@penumbra-zone/ui/lib/utils';
-import { useChainId } from '../../../hooks/chain-id';
+import { useChainIdQuery } from '../../../hooks/chain-id';
 import { ShareGradientIcon } from '../../../icons';
 import { SettingsHeader } from '../../../shared';
 import { useStore } from '../../../state';
 import { networkSelector } from '../../../state/network';
 import { internalSwClient } from '@penumbra-zone/router';
+import '@penumbra-zone/types/src/promise-with-resolvers';
 
 export const SettingsRPC = () => {
-  const { chainId, refetch } = useChainId();
+  const { chainId: currentChainId } = useChainIdQuery();
+  const [newChainId, setNewChainId] = useState(undefined as string | undefined);
   const { grpcEndpoint, setGRPCEndpoint } = useStore(networkSelector);
 
-  const [rpc, setRpc] = useState(grpcEndpoint ?? '');
-  const [rpcError, setRpcError] = useState(false);
+  const [rpcInput, setRpcInput] = useState(grpcEndpoint ?? '');
+  const [rpcError, setRpcError] = useState(undefined as string | undefined);
+  const [countdownTime, setCountdownTime] = useState(undefined as number | undefined);
 
-  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
-    e.preventDefault();
+  const countdown = (seconds: number) => {
+    const { promise, resolve } = Promise.withResolvers<undefined>();
+    setCountdownTime(seconds);
+    setInterval(() => {
+      if (!seconds) resolve(undefined);
+      setCountdownTime(--seconds);
+    }, 1000);
+    return promise;
+  };
+
+  const onSubmit = (evt: FormEvent<HTMLFormElement>) => {
+    evt.preventDefault();
     void (async () => {
-      const querier = new AppQuerier({ grpcEndpoint: rpc });
-
       try {
-        await querier.appParams();
-        await setGRPCEndpoint(rpc);
-        await internalSwClient.clearCache();
-        await refetch();
-        setRpcError(false);
-      } catch {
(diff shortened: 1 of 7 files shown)


apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -1,34 +1,54 @@
...
-        await setGRPCEndpoint(rpc);
-        await internalSwClient.clearCache();
-        await refetch();
-        setRpcError(false);
-      } catch {
-        setRpcError(true);
+        const { appParameters } = await createPromiseClient(
+          QueryService,
+          createGrpcWebTransport({ baseUrl: rpcInput }),
+        ).appParameters({});
```
grod220, line 40: Can we separate these to two lines?

```typescript
        const client = createPromiseClient(
          QueryService,
          createGrpcWebTransport({ baseUrl: rpcInput }),
        );
        const { appParameters } = await client.appParameters({});
```

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -1,34 +1,54 @@
...
 import { SettingsHeader } from '../../../shared';
 import { useStore } from '../../../state';
 import { networkSelector } from '../../../state/network';
 import { internalSwClient } from '@penumbra-zone/router';
+import '@penumbra-zone/types/src/promise-with-resolvers';
 
 export const SettingsRPC = () => {
-  const { chainId, refetch } = useChainId();
+  const { chainId: currentChainId } = useChainIdQuery();
+  const [newChainId, setNewChainId] = useState(undefined as string | undefined);
```
grod220, line 16:
```suggestion
  const [newChainId, setNewChainId] = useState<string>();
```

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -1,34 +1,54 @@
...
   const { grpcEndpoint, setGRPCEndpoint } = useStore(networkSelector);
 
-  const [rpc, setRpc] = useState(grpcEndpoint ?? '');
-  const [rpcError, setRpcError] = useState(false);
+  const [rpcInput, setRpcInput] = useState(grpcEndpoint ?? '');
+  const [rpcError, setRpcError] = useState(undefined as string | undefined);
```
grod220, line 20:
```suggestion
  const [rpcError, setRpcError] = useState<string>();
```

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -48,38 +68,34 @@ export const SettingsRPC = () => {
...
-          <Button
-            variant='gradient'
-            size='lg'
-            className='w-full'
-            type='submit'
-            disabled={rpc === grpcEndpoint || rpcError}
-          >
-            Save
-          </Button>
+          {countdownTime != null ? (
```
grod220, line 90:
```suggestion
          {countdownTime ? (
```

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -1,34 +1,54 @@
...
+          createGrpcWebTransport({ baseUrl: rpcInput }),
+        ).appParameters({});
+        if (!appParameters?.chainId) throw new Error('Endpoint did not provide a valid chainId');
+
+        setRpcError(undefined);
+        setNewChainId(appParameters.chainId);
+        await setGRPCEndpoint(rpcInput);
+        // TODO: show dialog, explain new chain
+        if (appParameters.chainId != currentChainId) void internalSwClient.clearCache();
+        await countdown(5).then(() => chrome.runtime.reload());
```
grod220, line 48:
```suggestion
        await countdown(5);
        chrome.runtime.reload();
```

grod220, line 48: what's the reasoning for `chrome.runtime.reload()`? Can you add a code comment why we need to restart the extension?

apps/extension/src/routes/popup/settings/settings-rpc.tsx
```diff
@@ -1,34 +1,54 @@
...
 
-  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
-    e.preventDefault();
-    void (async () => {
-      const querier = new AppQuerier({ grpcEndpoint: rpc });
+  const countdown = (seconds: number) => {
+    const { promise, resolve } = Promise.withResolvers<undefined>();
+    setCountdownTime(seconds);
+    setInterval(() => {
+      if (!seconds) resolve(undefined);
```
grod220, line 27: surprised we need to pass `undefined` to resolve, thought values were optional here

grod220 (approved):
Feel free to merge after addressing comments 👍

apps/extension/src/routes/popup/settings/settings-rpc.tsx
turbocrime: it would be optional if i used `Promise.withResolvers<void>` but eslint dislikes that

apps/extension/src/routes/popup/settings/settings-rpc.tsx
turbocrime: there could be active queries or running promises to the old endpoint anywhere in the extension. this ensures every connection is killed

apps/extension/src/routes/popup/settings/settings-rpc.tsx
turbocrime: the countdown currently displays a countdownTime of 0 at the end. if i made this change it would at that moment revert to a normal button again

apps/extension/src/routes/popup/settings/settings-rpc.tsx
turbocrime: updated the polyfill to default to a null parameter

PR #619, opened 2024-02-27
Include token metadata for zero-balance ValueViews of delegation tokens

### LINKED ISSUE #617: Include token metadata for zero-balance `ValueView`s of delegation tokens
Currently, when a user has no delegation tokens for a given validator, we don't show their balance at all:
image

We should show a balance of `0` in the middle of that UI, just like we show a nonzero balance for validators the user _has_ staked:
image

The problem is, [we're generating `ValueView`s on the fly](https://github.com/penumbra-zone/web/pull/604/files#diff-9bd971ea2f5dadcc555a9a848d1cea2a0d6df1a0069b9e5e490218058ac5adbcR70-R81) for validators we're not staked in. This means we don't have access to an asset ID for the delegation tokens we're supposed to show a zero balance for, and thus we can't fetch a fully fleshed out metadata object for the delegation tokens.*

## Possible solutions
- Fetch the metadata by a different method than asset ID, such as by display name. (This would likely require a new RPC method. Or, once #612 is done, the extension itself could take care of that more easily by looking for the right metadata in IndexedDB.)

\* Per @aubrika, when I asked if it's possible for a validator to exist when no tokens have been staked:
> https://github.com/penumbra-zone/penumbra/issues/2921 depends on if you are listing "declared" validators, "defined" validators, or both, I suppose. see linked issue for deets

Per @hdevalence:
> how does a validator end up on the chain? someone makes a tx with a ValidatorDefinition action that declares it. at that moment, the validator has been "created" on chain, but there's no delegation tokens yet. i believe that the on chain asset registry actually registers the delegation token at this point
> 
> however to become active, or to appear in a users balances, there must be delegation

jessepinho:
Previously we weren't rendering any balance for delegation tokens that the user doesn't hold any of. That's because there wasn't any metadata for those tokens, so `ValueViewComponent` rendered nothing for them.

## Before
image

## After
image

## In this PR
- Fetch asset metadata based on the validator's delegation token name, and use it to render the zero balances.
- Remove `customizeSymbol` from the block processor
  - Would especially like @grod220's thoughts on this. I made two changes to `customizeSymbol`: 1) I made it a pure function, returning a new copy of the `Metadata` with the symbol customized, rather than mutating the existing one; and 2) I moved it to `ValueViewComponent`, so that the customizing happens at the render stage rather than the block processing stage. This means that:
    - The webapp, rather than the extension, is now responsible for accurately displaying the type of token. This could be a downside if we are worried that other web apps won't realize that validator names could be spoofed. However, I think the benefits outweigh the downsides.
    - If/when we want to change the rendering of the symbol, we don't have to blow away all the IndexedDB data and rebuild. Rather, we can just change the `ValueViewComponent`'s rendering logic. Another way of saying this: _The concern of how to render a symbol rests within the rendering layer (where it belongs, I'd argue), rather than the processing layer._

## Punting for now
- I'd love to build out the `ValueViewComponent` a bit in the future so that it has a tooltip that shows much more information. Then, we could make the symbol much shorter, and show which validator the token is for on hover.

Closes #617

### DIFF
--- a/apps/webapp/src/fetchers/staking.ts
+++ b/apps/webapp/src/fetchers/staking.ts
@@ -6,6 +6,7 @@ import { stakingClient, viewClient } from '../clients/grpc';
 import { ValueView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import {
   bech32IdentityKey,
+  customizeSymbol,
   getDisplayDenomFromView,
   getIdentityKeyFromValidatorInfo,
   getValidatorInfo,
@@ -70,7 +71,7 @@ export const getDelegationsForAccount = async function* (addressIndex: AddressIn
 
       yield withValidatorInfo;
     } else {
-      const assetMetadataByIdResponse = await viewClient.assetMetadataById({
+      const { denomMetadata } = await viewClient.assetMetadataById({
         assetId: { altBaseDenom: getDelegationTokenBaseDenom(validatorInfo) },
       });
 
@@ -82,7 +83,7 @@ export const getDelegationsForAccount = async function* (addressIndex: AddressIn
               hi: 0n,
               lo: 0n,
             },
-            metadata: assetMetadataByIdResponse.denomMetadata,
+            metadata: denomMetadata ? customizeSymbol(denomMetadata) : undefined,
             extendedMetadata,
           },
         },
--- a/apps/webapp/src/fetchers/staking.ts
+++ b/apps/webapp/src/fetchers/staking.ts
@@ -2,7 +2,7 @@ import {
   AddressIndex,
   IdentityKey,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-import { stakingClient } from '../clients/grpc';
+import { stakingClient, viewClient } from '../clients/grpc';
 import { ValueView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import {
   bech32IdentityKey,
@@ -25,6 +25,9 @@ const isDelegationBalance = (balance: BalancesResponse, identityKey: IdentityKey
   return bech32IdentityKey(identityKey) === matchGroups.bech32IdentityKey;
 };
 
+const getDelegationTokenBaseDenom = (validatorInfo: ValidatorInfo) =>
+  `udelegation_${bech32IdentityKey(getIdentityKeyFromValidatorInfo(validatorInfo))}`;
+
 /**
  * Given an `AddressIndex`, yields `ValueView`s of the given address's balance
  * of delegation tokens. Each `ValueView` has an `extendedMetadata` property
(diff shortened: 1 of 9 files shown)


apps/webapp/src/components/staking/account/delegation-value-view/index.tsx
```diff
@@ -17,7 +17,7 @@ export const DelegationValueView = ({
   votingPowerAsIntegerPercentage,
   canDelegate,
 }: {
-  valueView?: ValueView;
+  valueView: ValueView;
```
jessepinho, line 20: Accidentally left this optional before

packages/types/src/customize-symbol.test.ts
```diff
@@ -8,25 +8,27 @@ describe('Customizing metadata', () => {
       display:
         'delegation_penumbravalid1fjuj67ayaqueqxg03d65ps5aah6m39u39qeacu3zv2cw3dzxssyq3yrcez',
     });
-    customizeSymbol(metadata);
-    expect(metadata.symbol).toBe('Delegated UM (fjuj67ayaqueqxg03d65ps5aa...)');
+
+    expect(customizeSymbol(metadata).symbol).toBe('Delegated UM (fjuj67ayaqueqxg03d65ps5aa...)');
```
jessepinho, line 12: `customizeSymbol` is now a pure function that returns a new (and modified) copy of the metadata

grod220:
>`customizeSymbol`... moved it to ValueViewComponent, so that the customizing happens at the render stage rather than the block processing

The symbol being set within the extension came from https://github.com/penumbra-zone/web/pull/558 by @hdevalence. Perhaps he can weigh in here. I have a feeling that we want to keep this in the extension because:
1. Security implications. In this case, it's a bit sensitive because validators could make up any name. If web apps, make the wrong call and show the validator created name as the token symbol, the user could be vulnerable to impersonation.
2. Convenience. Web apps ideally can just request metadata and display it without having to add additional logic.

hdevalence:
Yes, we need the symbol to come from the asset metadata registered in the extension's asset database. 

The only part of the self declared validator data we can actually rely on is the validator identity. So we can either have some "known delegation tokens" (as experimented with in that PR) or uniformly generate a symbol from the validator identity.

jessepinho:
OK @grod220, requesting re-review now that I've moved symbol customization back to the extension (and copied that logic in the staking fetcher). [Commit here](https://github.com/penumbra-zone/web/pull/619/commits/46d33ac40ca71bdda3362ff16e2bc125f8d1c7a2).

apps/webapp/src/fetchers/staking.ts
```diff
@@ -24,6 +25,9 @@ const isDelegationBalance = (assetBalance: AssetBalance, identityKey: IdentityKe
   return bech32IdentityKey(identityKey) === matchGroups.bech32IdentityKey;
 };
 
+const getDelegationTokenBaseDenom = (validatorInfo: ValidatorInfo) =>
+  `udelegation_${bech32IdentityKey(getIdentityKeyFromValidatorInfo(validatorInfo))}`;
+
```
grod220, line 30: I wonder if this should be with the getters? Not sure. 🤔

grod220 (approved):

apps/webapp/src/fetchers/staking.ts
jessepinho: Yeah, I'm actually planning on (sometime soon, when I have a break) writing a bunch of `asX` getters that transform objects into different types, such as `asBech32String` for identity keys. I'll punt that to a future PR though.

PR #620, opened 2024-02-27
Issue delegate actions

jessepinho:
So far, the Staking page has only rendered validators and our balances of tokens in each. 

This PR builds the logic and UI to actually issue delegate actions when the user clicks Delegate.

https://github.com/penumbra-zone/web/assets/1121544/07f83bb4-abba-463e-b95b-719ecb99e535

## In this PR
- Create a UI for delegating to a validator.
- Create a simple `<IdentityKeyComponent />` for rendering a validator's `IdentityKey`.
- Create a simple `<DelegateComponent />` for rendering a `Delegate` action.
- Update the transaction planner in JS and WASM to account for delegate actions.
- Fix issue where the extension was trying to load a proving key that didn't exist for actions that (apparently) don't even need the key, such as `delegate`.

## Future PRs
- Issue undelegate actions
- Prettify the `<DelegateComponent />`
- Fix issues where the super-long symbols for delegation tokens don't screw up the layout in the approval popup.

Relates to #542 (but doesn't close it, since that issue also includes undelegate actions).


apps/extension/src/wasm-build-action.ts
```diff
@@ -38,6 +39,10 @@ const workerListener = ({ data }: { data: ActionBuildRequest }) => {
 
 self.addEventListener('message', workerListener, { once: true });
 
+const actionTypeRequiresProvingKey = (
```
jessepinho, line 42: My logic for whether a proving key is required for a given action, is checking whether such a proving key even exists. Let me know if this is incorrect. I had to make this change because I was getting an error that no proving key existed for the `delegate` action. When I stopped trying to load the nonexistent proving key for that action, it Just Worked™.

apps/webapp/src/components/staking/account/delegation-value-view/index.tsx
```diff
@@ -15,19 +15,20 @@ import { getAmount, getValidatorInfoFromValueView, joinLoHiAmount } from '@penum
 export const DelegationValueView = ({
   valueView,
   votingPowerAsIntegerPercentage,
-  canDelegate,
```
jessepinho, line 18: Moving the `canDelegate` and `canUndelegate` logic down into the `<StakingActions />` component, since that component now know our balance of staking and delegation tokens and can thus determine for itself whether we can delegate or undelegate.

apps/webapp/src/components/staking/account/delegation-value-view/staking-actions.tsx
```diff
@@ -1,41 +0,0 @@
-import { Button } from '@penumbra-zone/ui';
```
jessepinho, line 1: Moved to `./staking-actions/index.tsx`

apps/extension/src/wasm-build-action.ts
```diff
@@ -38,6 +39,10 @@ const workerListener = ({ data }: { data: ActionBuildRequest }) => {
...
+  actionPlanType: Exclude<TransactionPlan['actions'][number]['action']['case'], undefined>,
```
grod220, line 43: Think this is a good candidate to extract:

```typescript
type ActionPlan = Exclude<TransactionPlan['actions'][number]['action']['case'], undefined>;

const actionTypeRequiresProvingKey = (actionPlanType: ActionPlan) =>
```

apps/extension/src/wasm-build-action.ts
grod220: @TalDerei might you know about this? Are there some actions that do not require proving keys? (aka delegation?)

apps/webapp/src/components/staking/account/delegation-value-view/staking-actions/form-dialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  return (
+    <Dialog open={!!action} onOpenChange={handleOpenChange}>
+      <DialogContent size='sm'>
+        {!!action && (
+          <>
+            <DialogHeader>{getDialogHeader(action, validator)}</DialogHeader>
+            <form className='flex flex-col gap-4 overflow-hidden px-4 pb-4' onSubmit={handleSubmit}>
+              <div className='flex flex-col'>
+                <div className='truncate'>{validator.name}</div>
```
grod220, line 74: I feel like this is too long for the dialog title. Perhaps this should be large/bold in the body.

apps/webapp/src/state/staking.ts
```diff
@@ -4,16 +4,28 @@ import { getDelegationsForAccount } from '../fetchers/staking';
...
-import { STAKING_TOKEN } from '@penumbra-zone/constants';
+import { STAKING_TOKEN, localAssets } from '@penumbra-zone/constants';
 import { AddressIndex } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { TransactionToast } from '@penumbra-zone/ui';
+import { authWitnessBuild, broadcast, getTxHash, plan, userDeniedTransaction } from './helpers';
+import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+import { BigNumber } from 'bignumber.js';
+
+const STAKING_TOKEN_DISPLAY_DENOM_EXPONENT = (() => {
+  const stakingAsset = localAssets.find(asset => asset.display === STAKING_TOKEN);
```
grod220, line 26: Not important for this PR, but I have a hunch we probably can't rely upon a direct `localAssets` import and need to query the extension for a list of assets it knows of. Need to have a thought about this. Don't think we have an RPC method to give us "all metadata you know of" (aka the `ASSETS` table).

packages/ui/components/ui/identity-key-component.tsx
```diff
@@ -0,0 +1,18 @@
+import { IdentityKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { CopyToClipboardIconButton } from './copy-to-clipboard-icon-button';
+import { bech32IdentityKey } from '@penumbra-zone/types';
+
+/**
+ * Renders an `IdentityKey` for a validator, along with a copy button.
+ */
```
grod220, line 7: Can you list an example one so we know what an identity key looks like?

grod220 (approved):
Minor comments. Nice job!

apps/webapp/src/state/staking.ts
grod220: Correction, we have `AssetsRequest` we could use. Think this is worthwhile to use here instead of the hardcoded registry.

turbocrime: do we actually need to query this way? there's only one staking token iirc and its id should be static, we will either have the md or could query by id

apps/extension/src/wasm-build-action.ts
TalDerei: > @TalDerei might you know about this? Are there some actions that do not require proving keys? (aka delegation?)

The 7 action types (excluding `nullifier_derivation` which is used for verifiable disclosure purposes, and won't be used by the web-code) in proving-keys.ts [require](https://github.com/penumbra-zone/penumbra/blob/a9694cfc997a893bf3a9a66a6822e2b078b861b4/crates/bin/pcli/tests/proof.rs#L99) proving keys fetched from [here](https://github.com/penumbra-zone/penumbra/tree/a9694cfc997a893bf3a9a66a6822e2b078b861b4/crates/crypto/proof-params/src/gen). These are the only actions that require proving keys, and there are other [actions](https://github.com/penumbra-zone/penumbra/blob/a9694cfc997a893bf3a9a66a6822e2b078b861b4/crates/core/transaction/src/plan/action.rs#L38) that don't.

hdevalence: Yes, most actions don't have proofs, so we should be handling the case that an action doesn't require a proving key.

Wondering if it might be cleanest to have a mapping of action type -> proving key name | null ?

apps/webapp/src/state/staking.ts
jessepinho: @grod220 I was thinking along the same lines as @turbocrime — that there's only one staking token, so no need to make a query. i.e., this can be hardcoded. Thoughts?

packages/ui/components/ui/identity-key-component.tsx
jessepinho: Does [this](https://github.com/penumbra-zone/web/pull/620/commits/442519949ead1c1d286f5b1fff39fcd033b1caa1) work? We don't usually construct identity keys ourselves — we just grab them off of `Validator` objects.

apps/webapp/src/components/staking/account/delegation-value-view/staking-actions/form-dialog.tsx
jessepinho: Did you mean to make this comment a few lines above, under the `<DialogHeader />`? 

I intentionally repeated the validator name in the body since it can be too long for the dialog header. That said, I could definitely shorten the header to just say "Delegate" or "Undelegate", rather than "Delegate _to_ [ValidatorName]"...?

jessepinho: image

For now, I'm shortening it to just the action per your comment. We can revisit it if needed.

apps/extension/src/wasm-build-action.ts
jessepinho: Good call, done

apps/webapp/src/state/staking.ts
jessepinho: Going to leave this as-is for now so I can merge, but let's revisit Monday if I'm missing something .

PR #625, opened 2024-02-28
workspace vitest config

turbocrime:
vitest updated and added some nice features for monorepo/workspace tests

`pnpm test` at the root uses turbo to call the test script of each package, which is just `vitest run`

`pnpm vitest` from the workspace root or in any package will watch tests with the appropriate config, and `pnpm vitest run` will one-shot them with the appropriate config

CI workflow is updated to no longer matrix, as test configs and environments now resolve properly to each test file without command-line arguments.

the vscode vitest extension now supports workspaces, and this config, so those of us running vscode get a nice watch ui with jump-to-failures and a debugger

Screenshot 2024-02-28 at 14 44 26


packages/ui/lib/toast/toast.test.ts
```diff
@@ -227,21 +227,17 @@ describe('Toast', () => {
...
+    const toast = new Toast().message('Hello, world!').description('Description here').render();
+    (['success', 'info', 'warning', 'error', 'loading'] as const).map(toastFn => {
+      toast[toastFn]().render();
+      expect(mockToastFn[toastFn]).toHaveBeenCalledTimes(1);
+      expect(mockToastFn[toastFn]).toHaveBeenCalledWith(
+        'Hello, world!',
+        expect.objectContaining({
+          description: 'Description here',
+        }),
+      );
```
turbocrime, line 240: updated these dynamically generated tests to just iterate within the test, because the vscode extension had trouble understanding what was happening here

turbo.json
```diff
@@ -10,6 +10,7 @@
       "outputs": ["dist/**"]
     },
     "download-keys": {
+      "inputs": ["src/utils/download-proving-keys.ts"],
```
turbocrime, line 13: also cacheing download properly now

apps/extension/vitest.config.ts
turbocrime, line 1: extension doesn't use vite, just vitest

packages/router/vitest.config.ts
turbocrime, line 1: router doesn't use vite, just vitest

packages/storage/src/chrome/migration.test.ts
```diff
@@ -147,7 +146,7 @@ describe('Storage migrations', () => {
         },
       ]);
 
-      await flushPromises();
+      await Promise.resolve();
```
grod220, line 149: If we remove `flushPromises` we've lost some documentation. Can you add that here?

packages/crypto/package.json
```diff
@@ -5,10 +5,7 @@
   "type": "module",
   "scripts": {
     "lint": "eslint \"**/*.ts*\"",
-    "test:browser": "vitest run --config browser.config.ts",
-    "test:browser-watch": "vitest --config browser.config.ts",
-    "test": "vitest run",
-    "test-watch": "vitest"
+    "test": "vitest run"
```
grod220, line 8: How do these tests know to run in a browser? Way back we had issues with the global `crypto` package being used in a nodejs environment.

grod220 (approved):

packages/storage/src/chrome/migration.test.ts
grod220: Looks like it was insta merged, but could we still address this one ⬆️

grod220:
Come to think of it, maybe we shouldn't have auto-merge 🤔. It makes sense after an approval (so you don't have to wait on the build), but before an approval it's possibly too hasty.

packages/storage/src/chrome/migration.test.ts
turbocrime: it seems unnecessary so i'll just make a PR that removes it.

packages/crypto/package.json
turbocrime: package-level config runs it in playwright

https://github.com/penumbra-zone/web/blob/ea1c3740e030a29705cdf3ac5e4dbd0c15e6cefe/packages/storage/vitest.config.ts#L1-L14

turbocrime:
ew, it cluttered up the commit log too. i expected a squash

Issue #612, opened 2024-02-27
Create new `StakingService` method that returns all validators as `ValueView`s


jessepinho:

hdevalence:
I think this should be in the view service, if we're expecting it to return info about the user's delegation tokens -- which the chain doesn't have.

How about `rpc DelegationBalances` with a request parameter that has an `all_active_validators` field?

jessepinho:
Thanks @hdevalence — I'll add this to the issue description. `all_active_validators` would determine whether to filter for just active ones, you mean?

jessepinho:
cc @grod220 — this is the second part of moving some staking work into Rust/the extension

hdevalence:
> Thanks @hdevalence — I'll add this to the issue description. `all_active_validators` would determine whether to filter for just active ones, you mean?

I was thinking it would be telling the view service whether to include zero balances at all, the confusion is maybe a sign my name is bad.

- default behavior: like `BalancesRequest` but only returns delegation tokens the user has
- with flag set to `true`: includes one entry for each validator in active set, regardless of balance

jessepinho:
@hdevalence Gotcha. But there wouldn't be a scenario where we'd only want the validators that the user has staked though, right? Wouldn't we always want to show all validators? 

Although, maybe you're thinking of frontends other than ours who might want to arrange that view differently. Anyway, it can't hurt to include that functionality.

hdevalence:
Yeah, exactly. My thought was that from the extension's point of view, the implementation involves querying the user's delegation tokens and then merging that with the validator list. So there's not much extra involved in having a flag to skip the merging, and maybe that's useful for a different use case.

Issue #601, opened 2024-02-23
Store all `ValidatorInfo`s in IDB


jessepinho:
Currently, validator infos are requested when a user navigates to the Staking page. This is a bit slow, and is also a privacy concern, since it indicates to the RPC node that a user is interested in staking.

Instead, we should store `ValidatorInfo`s in IndexedDB. Each time the epoch changes*, we should request updated validator info, since validator info only changes at epoch boundaries.** (This work should most likely be done in the block processor.)

We can then complete #612, using the validator infos in IDB to generate `ValueView`s.

\* Note that we'll want to take different approaches to updating validator info for when we're A) syncing blocks for the first time, vs. B) doing our regular syncs in the background.
A. When syncing blocks for the first time, we obviously don't want to fetch new validator info at every epoch boundary, since there could be thousands of epoch boundaries within a few years. Instead, we'll want to wait until syncing is done, then fetch validator info.
B. When doing regular syncs, we'll want to keep track of our last known epoch number, and then fetch new validator info when the current epoch changes.

\** New validators only start participating in consensus at the beginning of an epoch. New delegations or undelegations only take effect at epoch boundaries. And anytime a validator is tombstoned, jailed, etc., that triggers the end of the current epoch. Thus, **epochs constitute the smallest, most indivisible measurement of a validator's lifespan**, so we only ever need to fetch new validator info when the epoch changes.

jessepinho:
cc @grod220 — this is the first part of moving some staking work into Rust/the extension

PR #627, opened 2024-02-29
Show an account switcher on the Staking page

jessepinho:
Previously, you could only delegate, view validators, etc. from account #0. Now you can switch accounts.

Note that at the moment, it makes a request to the RPC node every time you switch accounts. This will no longer be the case once #601 and #612 are done.

https://github.com/penumbra-zone/web/assets/1121544/eeac65ab-2727-4f96-b584-1102b4088d4b

## In this PR
- Extract a new `<AccountSwitcher />` component from the `<SelectAccount />` component.
- Use it at the top of the account view.

## Future PRs
- Put `account` state in the query string? But that could require a bit of refactoring, so I'll leave that for a future improvement.

### DIFF
--- a/packages/ui/components/ui/account-switcher.tsx
+++ b/packages/ui/components/ui/account-switcher.tsx
@@ -1,0 +1,80 @@
+import { ArrowLeftIcon, ArrowRightIcon } from 'lucide-react';
+import { cn } from '../../lib/utils';
+import { Button } from './button';
+import { Input } from './input';
+import { useState } from 'react';
+
+const MAX_INDEX = 2 ** 32;
+
+/**
+ * Renders arrows with which to switch between accounts by index, from 0 to
+ * 2^32.
+ */
+export const AccountSwitcher = ({
+  account,
+  onChange,
+}: {
+  account: number;
+  onChange: (account: number) => void;
+}) => {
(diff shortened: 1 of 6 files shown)


packages/ui/components/ui/account-switcher.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Input } from './input';
+import { useState } from 'react';
+
+const MAX_INDEX = 2 ** 32;
+
+/**
+ * Renders arrows with which to switch between accounts by index, from 0 to
+ * 2^32.
+ */
+export const AccountSwitcher = ({
```
jessepinho, line 13: Extracted from `SelectAccount`

grod220 (approved):
Codewise LGTM 👍 

_== Side note: design thoughts ==_

Let's say a user has delegation tokens in account 1234. Our current design requires them to have foreknowledge of where they already are. We aren't really providing an overview for all their accounts.

Isn't it also odd that the user can toggle through accounts for which we know they have zero delegation tokens in and zero penumbra tokens to be able to delegate?

jessepinho:
@grod220 ah, good point — Henry had actually mentioned that the switcher should be extended to allow filtering of accounts ,so that we can only show accounts that have balances in either staking, delegation, or unbonding tokens. I'll do this in a follow-up PR.

PR #631, opened 2024-02-29
fix: do not render no memo as 'Invalid MemoView'

llllvvuu:
`swap` and `swapClaim` transactions have no memo.


packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -12,7 +12,9 @@ export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
 
   return (
     <div className='flex flex-col gap-8'>
-      <MemoViewComponent memo={txv.bodyView.memoView} />
+      {txv.bodyView.memoView?.memoView.case !== undefined && (
```
llllvvuu, line 15: @turbocrime I had to change back to `memoView.case` instead of `memoView` here, since `new MemoView` automatically inserts `{ case: undefined }` (I guess it is a gRPC codegen thing)

llllvvuu: actually just kidding, I was able to fix this by re-arranging the conditional in `perspective.ts`

llllvvuu:
> Nice job! Can you rebase before merge?

Sure, rebased! (I can't merge as an external contributor)

grod220:
Thanks for your contribution!

PR #632, opened 2024-02-29
Show unbonding tokens on the Staking page

### LINKED ISSUE #588: Show a user's unbonding tokens on the validators page

jessepinho:
image
image

## In this PR
- Display the amount of unbonding tokens the current account has at the top of the staking page, along with a tooltip showing the individual unbonding tokens.
- Create new `Stat` component to show a stat at the top of the staking page.

## Future PRs
- [Shorten the symbol for unbonding tokens](https://discord.com/channels/824484045370818580/883455571003576320/1212870181324460072)
- When that's done, render unbonding tokens in a table that also include their epoch

Closes #588


apps/webapp/src/components/staking/account/header/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      metadata: STAKING_TOKEN_METADATA,
+    },
+  },
+});
+
+/**
+ * The header of the account view, with an account switcher and balances of
+ * various token types related to staking.
+ */
+export const Header = () => {
```
jessepinho, line 36: image

apps/webapp/src/components/staking/account/header/stat.tsx
```diff
@@ -0,0 +1,13 @@
+import { ReactNode } from 'react';
+
+/**
+ * A statistic that shows up at the top of the staking page.
+ */
+export const Stat = ({ label, children }: { label: string; children: ReactNode }) => {
```
jessepinho, line 6: image

apps/webapp/src/components/staking/account/header/index.tsx
grod220: Not related to this PR, but I wonder why the account switcher appears to make a request every time it's toggled. I feel like we should only need to request all delegations once and just have the UI show filtered results of that.

apps/webapp/src/state/staking.ts
```diff
@@ -158,21 +179,55 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+    }>(
       (prev, curr) => {
-        const unstakedTokens = curr.balances.find(
-          ({ balanceView }) => getDisplayDenomFromView(balanceView) === STAKING_TOKEN,
+        prev.unstakedTokensByAccount.set(
+          curr.index.account,
+          curr.balances.find(
+            ({ balanceView }) => getDisplayDenomFromView(balanceView) === STAKING_TOKEN,
+          )?.balanceView,
         );
```
grod220, line 194: I find this a bit more readable unnested:

```typescript
const foundStakingToken = curr.balances.find(
   ({ balanceView }) => getDisplayDenomFromView(balanceView) === STAKING_TOKEN)?.balanceView;
prev.unstakedTokensByAccount.set(curr.index.account, foundStakingToken);
```

apps/webapp/src/state/staking.ts
```diff
@@ -158,21 +179,55 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
       state.staking.votingPowerByValidatorInfo = getVotingPowerByValidatorInfo(validatorInfos);
     });
   },
-  loadUnstakedTokensByAccount: async () => {
+  loadUnstakedAndUnbondingTokensByAccount: async () => {
     const balancesByAccount = await getBalancesByAccount();
-    const unstakedTokensByAccount = balancesByAccount.reduce<Map<number, ValueView | undefined>>(
+    const { unstakedTokensByAccount, unbondingTokensByAccount } = balancesByAccount.reduce<{
```
grod220, line 184: What do you think about splitting this up into two reduces? One for unstaked and one for unbonding? Feel it may be a bit more understandable as two helper functions.

packages/constants/assets/index.ts
```diff
@@ -2,7 +2,14 @@ import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/
...
 import { JsonValue } from '@bufbuild/protobuf';
 
+export const localAssets: Metadata[] = LocalAssetRegistry.map(a =>
+  Metadata.fromJson(a as JsonValue),
+);
+
 export const STAKING_TOKEN = 'penumbra';
+export const STAKING_TOKEN_METADATA = localAssets.find(
+  metadata => metadata.display === STAKING_TOKEN,
+);
```
grod220, line 12: Maybe add a `!` as we guarantee it's there? It would be nice if Typescript had the Rust-like: `.ok_or_else(Error::Something("no there"))`

grod220 (approved):

apps/webapp/src/state/staking.ts
jessepinho: Ugh, yeah, I did it this way for performance reasons. If we split it into two `reduce`s, we'd have to make two passes over `balancesByAccount`, which _could_ be a pretty large array. 

Now that I write this, though, I realize most people will probably just use account #0 (right?), so this performance concern is probably an edge case. I'll split it into two helpers.

apps/webapp/src/state/staking.ts
jessepinho: Good call, done

apps/webapp/src/state/staking.ts
jessepinho: [Done](https://github.com/penumbra-zone/web/pull/632/commits/8f90a65b42cb951fbf021d906136221a3676c551)

packages/constants/assets/index.ts
jessepinho: Whoops, good call. I'll do that and throw an error if it's falsey.

jessepinho: JK re: throwing an error — TypeScript complained if I checked for its presence after already asserting its presence. But I've at least added the `!`.

jessepinho:
@grod220:

> Not related to this PR, but I wonder why the account switcher appears to make a request every time it's toggled. I feel like we should only need to request all delegations once and just have the UI show filtered results of that.

This will soon be mitigated once #601 and #612 are done, since validator info will load from IndexedDB rather than from the RPC node.

That said, it _is_ required that the account switcher make a request each time it's toggled, because of the way we're now representing validators as `ValueView`s with the current address's balance of that validator's token. That is, we're no longer treating validators as a first class type of data; rather, _delegation token `ValueView`s with `ValidatorInfo` embedded in their `extendedMetadata`_ are how we represent validators. This is the "token-centric" approach that @hdevalence suggested.

The reason we can't request all delegations at once and then filter by address is that we need to make sure we're showing all validators for each address, _including validators that the address has no delegation tokens for_. This means generating zero-balance `ValueView`s for each validator that the address has no delegation tokens for, when that address is selected on the staking page. There are 2^32 addresses in a Penumbra wallet, so we can't generate on-the-fly zero-balance `ValueView`s for every address; instead, we just generate them for the address currently being viewed.

Of course, the approach I've outlined isn't strictly necessary for our code _as it presently exists_, since the webapp is currently A) requesting validators via the `ValidatorInfo` RPC method, then B) merging those with `ValueView`s of the user's balances of each delegation token. But I want to move that merging logic into an RPC method for ease of use (#612), since presumably a lot of clients will want to do the same thing.

PR #636, opened 2024-02-29
Filter out assets that are not available for swap from asset selector

### LINKED ISSUE #485: Filter out LpNft's from swap asset selector
When selecting an asset to swap, we see LpNfts included:

Per @grod220 's [comment](https://github.com/penumbra-zone/web/pull/476#pullrequestreview-1870071861):

> We should not have lpnft's in that selector drop down box at all.

This ticket is to filter for only spendable assets wherever we have asset selectors. We'll likely want to use [the `assetPatterns` const](https://github.com/penumbra-zone/web/blob/5f5e98bf6e85dc58f19df83977f6aee748a6ca87/packages/constants/assets/index.ts).

Valentine1898:
close #485

### DIFF
--- a/apps/webapp/src/components/swap/swap-loader.tsx
+++ b/apps/webapp/src/components/swap/swap-loader.tsx
@@ -9,8 +9,13 @@ import {
 import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { fetchUnclaimedSwaps } from '../../fetchers/unclaimed-swaps';
 import { viewClient } from '../../clients/grpc';
-import { getSwapAsset1, getSwapAsset2, uint8ArrayToBase64 } from '@penumbra-zone/types';
-import { localAssets } from '@penumbra-zone/constants';
+import {
+  getDisplayDenomFromView,
+  getSwapAsset1,
+  getSwapAsset2,
+  uint8ArrayToBase64
+} from '@penumbra-zone/types';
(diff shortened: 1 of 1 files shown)


Valentine1898:
image
image

I think we should filter not only `lpNft` but also `proposalNft`, `unbondingToken`, `votingReceipt`, obviously they can't be swapped.

apps/webapp/src/components/swap/swap-loader.tsx
```diff
@@ -26,13 +31,18 @@ export interface SwapLoaderResponse {
 const getAndSetDefaultAssetBalances = async () => {
   const assetBalances = await getBalances();
 
+  // filter assets that are not available for swap
+  const filteredAssetBalances = assetBalances.filter(b =>
+      [assetPatterns.lpNft, assetPatterns.proposalNft, assetPatterns.unbondingToken, assetPatterns.votingReceipt]
```
grod220, line 36: I think there could potentially be a market for unbonding tokens. Think we should keep that one.

Valentine1898: It's actually kind of weird. 
We will have 100 different liquid staking tokens for 100 different validators, and for each of these liquid staking tokens we can have many more unbonding tokens for each epoch  :exploding_head:

Valentine1898: But in general I agree, removed unbonding tokens from filtering

grod220 (approved):

Issue #595, opened 2024-02-22
Minifront version footer


hdevalence:
It would be convenient if all the pages in minifront had an unobtrusive footer with a version string like

minifront v2.3.4

This could be greyed out etc but that way if copies of it proliferate (eg fullnode embedding) there's a way for users to see what version they're using

PR #638, opened 2024-02-29
Minifront version footer

grod220:
Closes https://github.com/penumbra-zone/web/issues/595

Creates a vite plugin for minifront that injects the commit hash + date into the frontend to convey version context. Quite helpful in cases where bundled node frontends are behind the deployed web app.

I find this better than publishing a version like `4.1.2` as that would require manual updates while this is automatic.

In context of page:
Screenshot 2024-02-29 at 4 14 30 PM

Zoomed in:
Screenshot 2024-02-29 at 4 14 33 PM


apps/webapp/src/utils/commit-hash-vite-plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const PATH_TO_DECLARATION_FILE = 'src/vite-env.d.ts';
+const COMMIT_VAR_DECLARATION = `declare const __COMMIT_HASH__: string;\n`;
+const COMMIT_DATE_DECLARATION = `declare const __COMMIT_DATE__: string;\n`;
+
+// Vite plugin used to inject the current commit hash + commit date into React
+// so the user can be informed the version of minifront they are using
+export const commitHashPlugin = (): Plugin => {
+  const commitHash = execSync('git rev-parse HEAD').toString().trim();
+  const commitDate = execSync('git log -1 --format=%cI').toString().trim();
```
Valentine1898, line 13: Will this work for a deploy to firebase?

Valentine1898 (approved):

turbocrime:
let's use same env for extension build and manifest version

grod220:
> let's use same env for extension build and manifest version

Sorry, can you clarify? Do you mean that we should add a plugin to the extension vite build to inject the manifest version into a react component there?

grod220:
> oh, you're mutating typescript?

The plugin just ensures the declaration files have the correct env vars

> why not env

Are you saying the plugin should create an env file with the git commit + date as env vars? We still need the declaration file though.

apps/webapp/src/types/vite-env.d.ts
grod220, line 1: should delete

apps/webapp/src/utils/commit-hash-vite-plugin.ts
grod220: Indeed, here's an example of the build output with the commit in the bundle (tested in github docker container)

Screenshot 2024-03-01 at 11 44 54 AM

PR #641, opened 2024-02-29
eslint eqeqeq

turbocrime:
enforce triple-equals over coercive double-equals.

exception for null comparisons - should null comparisons be excepted?


grod220 (approved):
I would not give null an exception though. I think it's pretty common in javascript-land to never use a `==`.

PR #643, opened 2024-02-29
prevent conversations about brace style

turbocrime:
i don't care about the particular rule we use, this is a suggestion. i'll accept anything as long as we all agree to never go back.

unfortunately brace rules greatly increase eslint exec time.

### DIFF
--- a/packages/eslint-config-custom/index.js
+++ b/packages/eslint-config-custom/index.js
@@ -27,6 +27,7 @@ module.exports = {
       },
     ],
     eqeqeq: ['error', 'always', { null: 'ignore' }],
+    curly: ['error', 'multi-or-nest'],
     'import/no-useless-path-segments': ['error', { noUselessIndex: true }],
     'import/no-relative-packages': 'error',
     'import/no-self-import': 'error',
--- a/packages/eslint-config-custom/index.js
+++ b/packages/eslint-config-custom/index.js
@@ -26,6 +26,7 @@ module.exports = {
         mjs: 'never',
       },
     ],
+    eqeqeq: ['error', 'always', { null: 'ignore' }],
     'import/no-useless-path-segments': ['error', { noUselessIndex: true }],
     'import/no-relative-packages': 'error',
     'import/no-self-import': 'error',
--- a/apps/extension/src/routes/page/onboarding/generate.tsx
+++ b/apps/extension/src/routes/page/onboarding/generate.tsx
@@ -28,9 +28,9 @@ export const GenerateSeedPhrase = () => {
 
   // On render, generate a new seed phrase
   useEffect(() => {
-    if (!phrase.length) {
+    if (!phrase.length) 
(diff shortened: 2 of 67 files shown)


packages/eslint-config-custom/index.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
       },
     ],
     eqeqeq: ['error', 'always', { null: 'ignore' }],
+    curly: ['error', 'multi-or-nest'],
```
turbocrime, line 30: the rule

turbocrime:
lmao brace lint rules oom the smallest buildjet runner

grod220:
Omitting brackets are kinda a "liberal" thing in javascript. A lot of folks think you should never do it. In my opinion, I think they are occasionally more readable though. Personally, I think this could be condensed and be just as readable:

```typescript
if (conditionA) {
   throw new Error("issue with conditionA")
}

if (conditionB) {
   throw new Error("issue with conditionB")
}
```

with this:

```typescript
if (conditionA) throw new Error("issue with conditionA")
if (conditionB) throw new Error("issue with conditionB")
```

I find that to be the case for early returns too.

```typescript
const someFunc = (valueView) => {
   if (!valueView.case) return "Unknown Case"
   ...
}
```
In my personal opinion only one line return statements and thrown errors can omit brackets (it's optional though)---all others should be required. Unfortunately, I think this would require a custom lint rule.

I think our options are:
- We do nothing and enforce this in PR reviews (not a bad choice if this issue is taking too much time)
- We require brackets all of the time (adjust rule to require that)
- Allow optional omitting for one line returns/throws

Maybe it's overkill but chatgpt says this would work in the case I am describing:

```typescript
module.exports = {
  meta: {
    type: "suggestion",
    docs: {
      description: "Enforce curly braces for control statements, allowing exceptions for single-line return or throw statements.",
      category: "Stylistic Issues",
      recommended: true,
    },
    fixable: "code", // Indicates that this rule supports automatic fixing
    schema: [], // No options required for this rule
  },

  create(context) {
    // Checks if the node is a return or throw statement on a single line
    function isSingleLineReturnOrThrow(node) {
      return (node.type === 'ReturnStatement' || node.type === 'ThrowStatement') &&
             node.loc.start.line === node.loc.end.line;
    }

    // Determines if the node requires braces
    function requiresBraces(node) {
      return node.type !== 'BlockStatement' && !isSingleLineReturnOrThrow(node);
    }

    // Reports if a node that requires braces doesn't have them
    function reportIfNoBraces(node) {
      if (requiresBraces(node)) {
        context.report({
          node,
          message: "Control statements must be enclosed in curly braces.",
        });
      }
    }

    return {
      IfStatement(node) {
        reportIfNoBraces(node.consequent);
      },
      WhileStatement(node) {
        reportIfNoBraces(node.body);
      },
      ForStatement(node) {
        reportIfNoBraces(node.body);
      },
      // Additional control statements like ForInStatement, ForOfStatement, and DoWhileStatement could be added here
    };
  },
};

```

PR #644, opened 2024-03-01
Declare type for Array.fromAsync

turbocrime:
this is already in chrome, but not known to typescript.

it fully replaces streamToPromise.


packages/polyfills/streamToPromise.test.ts
```diff
@@ -0,0 +1,53 @@
+import { beforeEach, describe, expect, test } from 'vitest';
+
+import Array from './Array.fromAsync';
+
+// eslint-disable-next-line @typescript-eslint/unbound-method
+const streamToPromise = Array.fromAsync;
+
+describe('streamToPromise()', () => {
```
jessepinho, line 8: Hm, I feel like these tests are no longer needed — even the ones below under `old streamToPromise behavior`. You're essentially testing something that's part of the Chrome JavaScript API, which should already be well-tested and guaranteed to work. No need to retest it ourselves.

If anything, if there are places where we're using `Array.fromAsync` that _does_ depend on old `streamToPromise` behavior, we should write tests to make sure it's accounted for the fact that `Array.fromAsync` works differently.

packages/polyfills/Array.fromAsync.test.ts
```diff
@@ -0,0 +1,73 @@
+import { expect, describe, it } from 'vitest';
+
+import Array from './Array.fromAsync';
+import './ReadableStream[Symbol.asyncIterator]';
+
+describe('Array.fromAsync', () => {
```
jessepinho, line 6: Tests not needed

jessepinho (changes requested):
Added a couple comments — it seems that the tests you have here aren't really needed, since they're retesting a core API. And can we just redefine `Array` globally, rather than re-exporting it and thus having to import a specific "implementation" of it?

turbocrime:
> it seems that the tests you have here aren't really needed, since they're retesting a core API.

it was useful to demonstrate

1) the actual behavior of that api, since i always get questions every time i do something like this
2) the behavior of the api matches our use-case, by applying the old streamToPromise tests to the browser api

packages/polyfills/Array.fromAsync.ts
```diff
@@ -0,0 +1,12 @@
...
+
+type ArrayWithAsync = typeof Array & {
+  fromAsync<T, U = T>(
+    arrayLike: ArrayLike<T> | AsyncIterable<T>,
+    mapfn?: (v: T, k: number) => U,
+    thisArg?: any,
+  ): Promise<U[]>;
+};
+
+export default Array as ArrayWithAsync;
```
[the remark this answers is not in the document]
turbocrime: typescript exposes `Array` as a `var`, which does not support declaration merging

packages/polyfills/streamToPromise.test.ts
turbocrime: maintaining the old tests was useful to demonstrate the api satisfies the behavior we needed

turbocrime:
not making changes

jessepinho (approved):

PR #650, opened 2024-03-02
Account for delegate/undelegate actions in the public/private views of a transaction

jessepinho:
As part of my work on the staking page, this makes sure that the public/private views of a transaction account for delegate and undelegate actions. Without this change, those actions would be displayed in correctly when, e.g., clicking the "Public" view in the approval popup.


grod220 (approved):
LGTM

PR #651, opened 2024-03-02
Allow users to undelegate from a validator

### LINKED ISSUE #542: Issue delegation actions

jessepinho:
This builds out the UI for undelegating from a validator.

https://github.com/penumbra-zone/web/assets/1121544/edddcac9-bdba-40ec-9101-d376028a75a8

Closes #542

### DIFF
--- a/apps/webapp/src/state/staking.ts
+++ b/apps/webapp/src/state/staking.ts
@@ -227,10 +227,9 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
     try {
       const transactionPlan = await plan(assembleUndelegateRequest(get().staking));
 
-      // Reset form _after_ building the transaction form, since it depends on
-      // the state.
+      // Reset form _after_ assembling the transaction planner request, since it
+      // depends on the state.
       set(state => {
-        state.staking.amount = '';
         state.staking.action = undefined;
         state.staking.validatorInfo = undefined;
       });
@@ -255,6 +254,10 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
       } else {
         toast.onFailure(e);
       }
+    } finally {
+      set(state => {
+        state.staking.amount = '';
+      });
     }
   },
   loading: false,
--- a/apps/webapp/src/state/staking.ts
+++ b/apps/webapp/src/state/staking.ts
@@ -4,12 +4,15 @@ import { getDelegationsForAccount } from '../fetchers/staking';
 import {
   VotingPowerAsIntegerPercentage,
   getAmount,
+  getAssetIdFromValueView,
   getDisplayDenomExponent,
+  getDisplayDenomExponentFromValueView,
   getDisplayDenomFromView,
   getRateData,
   getValidatorInfoFromValueView,
   getVotingPowerByValidatorInfo,
   getVotingPowerFromValidatorInfo,
+  isDelegationTokenForValidator,
   joinLoHiAmount,
   toBaseUnit,
 } from '@penumbra-zone/types';
@@ -50,6 +53,7 @@ export interface StakingSlice {
    */
   loadUnstakedTokensByAccount: () => Promise<void>;
   delegate: () => Promise<void>;
+  undelegate: () => Promise<void>;
   loading: boolean;
   error: unknown;
   votingPowerByValidatorInfo: Record<string, VotingPowerAsIntegerPercentage>;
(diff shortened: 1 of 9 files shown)


packages/wasm/src/tx-planner.ts
```diff
@@ -58,8 +61,12 @@ export class TxPlanner {
     this.wasmPlanner.ics20_withdrawal(withdrawal.toJson());
   }
 
-  delegate(delegation: TransactionPlannerRequest_Delegate): void {
-    this.wasmPlanner.delegate(delegation.amount?.toJson(), delegation.rateData?.toJson());
+  delegate(delegate: TransactionPlannerRequest_Delegate): void {
+    this.wasmPlanner.delegate(delegate.amount?.toJson(), delegate.rateData?.toJson());
+  }
```
jessepinho, line 66: Only change here was renaming the `delegation` arg to `delegate` for consistency with the type name

apps/webapp/src/state/staking.ts
```diff
@@ -176,11 +180,8 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
     });
   },
   onSubmit: () => {
-    /**
-     * @todo: Add switching logic depending on the value of `action`, once
-     * undelegation is implemented.
-     */
-    void get().staking.delegate();
+    if (get().staking.action === 'delegate') void get().staking.delegate();
+    else void get().staking.undelegate();
```
grod220, line 184: Think the buttons should just call `delegate()` / `undelegate()` fns directly

apps/webapp/src/state/staking.ts
```diff
@@ -219,6 +220,43 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+      toast.txHash(txHash);
+      const { detectionHeight } = await broadcast({ transaction, awaitDetection: true }, status =>
+        toast.onBroadcastStatus(status),
+      );
+      toast.onSuccess(detectionHeight);
+
+      // Reload delegation tokens and unstaked tokens to reflect their updated
+      // balances.
+      void get().staking.loadDelegationsForCurrentAccount();
+      void get().staking.loadUnstakedTokensByAccount();
```
grod220, line 251: Will the `void` cause it to exit this try/catch and not be handled if errored? 🤔

apps/webapp/src/state/staking.ts
```diff
@@ -219,6 +220,43 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
     }
   },
+  undelegate: async () => {
+    const toast = new TransactionToast('undelegate');
+    toast.onStart();
+
+    try {
+      const transactionPlan = await plan(assembleUndelegateRequest(get().staking));
+
+      // Reset form _after_ building the transaction form, since it depends on
```
grod220, line 230: Is it after the build or after planning?

apps/webapp/src/state/staking.ts
```diff
@@ -219,6 +220,43 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+      // the state.
+      set(state => {
+        state.staking.amount = '';
+        state.staking.action = undefined;
+        state.staking.validatorInfo = undefined;
+      });
```
grod220, line 236: If this throws an error, the user has to re-enter the data. Do we want this? If not, think this could go in the `finally` block.

grod220 (approved):

apps/webapp/src/state/staking.ts
jessepinho: [Done](https://github.com/penumbra-zone/web/pull/651/commits/0e2b8edca18ae9b9da92121f8678157ef3e81a7b)

apps/webapp/src/state/staking.ts
jessepinho: Nope — [void simply has the effect of making the statement evaluate to `undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void), which our linter requires when we're not doing anything with a promised return value.

apps/webapp/src/state/staking.ts
jessepinho: Whoops, I had a typo — the second "form" wasn't supposed to be there. I also reword the comment — it now reads:
```ts
      // Reset form _after_ assembling the transaction planner request, since it
      // depends on the state.
```

apps/webapp/src/state/staking.ts
jessepinho: Good call — I reset the `amount` in the `finally` block, but left the `action`/`validatorInfo` here at the top, so that the form dialog closes as soon as they submit it.

PR #660, opened 2024-03-02
declare Map.groupBy type

turbocrime:
another standard api already shipped to our target platform with no ts definition

once again can't just declare, but must declare and re-export, because typescript does not support declaration merging for vars

### DIFF
--- a/packages/polyfills/Map.groupBy.ts
+++ b/packages/polyfills/Map.groupBy.ts
@@ -1,0 +1,11 @@
+/** @see https://tc39.es/proposal-array-grouping/#sec-map.groupby */
+// already shipped in chrome, no type definition yet
+
+type MapGroupBy = <T, G>(
+  items: Iterable<T>,
+  callbackfn: (value: T, index: number) => G,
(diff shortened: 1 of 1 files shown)


turbocrime:
there's also an Object.groupBy but i wasn't about to use that

grod220 (commented):
Two things:

- Will this replace some code we've written manually? Aka, what do we need this for?
- I noticed that the casing in this package is not using kebab case (the standard in the repo). Think it would be helpful to align these.

turbocrime:
> Two things:
> 
>     * Will this replace some code we've written manually? Aka, what do we need this for?
> 
>     * I noticed that the casing in this package is not using kebab case (the standard in the repo). Think it would be helpful to align these.

this is used in connected site settings.

this isn't using kebab case because i am naming items in the polyfills package explictly as the symbol they declare or polyfill, so there is no misunderstanding of what they are exporting or declaring.

grod220 (approved):
The file names look a bit odd, but I suppose we can live with this

PR #663, opened 2024-03-04
wasm: simplify transaction planner

### LINKED ISSUE #609: Simplify `TransactionPlanner` implementation by collapsing unnecessary abstraction
Currently, the extension's wasm code has its own fork of the `Planner` code from the Rust view crate, which it uses to handle the `TransactionPlanner` RPC method. This leads to an enormous amount of unnecessary complexity, originally in service of code sharing that wasn't possible anyways.

To fix this, we should collapse the planner implementation into the `TransactionPlanner` RPC implementation, so that that RPC is handled directly and is the only way to plan transactions, and eliminate the `Planner` API which is otherwise unused.

hdevalence:
Closes #609

This commit gets most of the way to dramatically simplifying the transaction planner.  It completely eliminates the fork of the Rust crate's `Planner` API, and focuses on what the web extension actually needs, namely, serving `TransactionPlanRequest`s.

The only thing remaining is cleaning up the test code, which was written against the previous API and much of which doesn't make sense any more.

Along the way, it also fixes a bug in the existing web extension code, which would result in making unnecessary spends sometimes (always spending all available notes, even if unnecessary).

One observation growing out of this (cc @zbuc) is that we should change the way gas costs are computed on-chain, removing the transaction-wide byte costs and making the gas cost of the transaction just be the sum of the gas costs of its constituent actions. (In other words, we'd attribute the block space costs of each action to that action, and make the surrounding transaction contents free, rather than what we currently do, which is have the block space costs attributed to the transaction as a whole and make it free for each action). The reason is that this makes it very difficult to do "incremental" reasoning about the gas costs of the transaction as actions are added or removed.

We could also later backport much of this simplification to the Rust planner.


packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
 
-    /// Set the expiry height for the transaction plan.
-    pub fn expiry_height(&mut self, expiry_height: u64) -> &mut Self {
-        self.plan.transaction_parameters.expiry_height = expiry_height;
-        self
-    }
+    fn gas_estimate(&self) -> Gas {
+        // TODO: this won't include the gas cost for the bytes of the tx itself
+        // so this gas estimate will be an underestimate, but since the tx-bytes contribution
+        // to the fee is ideally small, hopefully it doesn't matter.
```
hdevalence, line 66: cc @zbuc -- we could fix this in the protocol by attributing the block space costs to the action, and get a perfect estimate

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
+    change_address: JsValue,
+    // all of these params should be removed
+    // TODO: read directly from IDB
+    fmd_params: JsValue,
+    // TODO: read directly from IDB
+    sct_params: JsValue,
+    // TODO: read directly from IDB
+    gas_prices: JsValue,
+    // TODO: read directly from IDB
+    chain_id: JsValue,
```
hdevalence, line 125: To simplify the API, all of these parameters should be read directly out of IDB, but I didn't know how to do that.

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
+) -> WasmResult<JsValue> {
+    utils::set_panic_hook();
+
+    let request: TransactionPlannerRequest = serde_wasm_bindgen::from_value(request)?;
+    // TODO: should we take an FVK to derive this from the data in the request?
+    let change_address: Address = serde_wasm_bindgen::from_value(change_address)?;
```
hdevalence, line 131: I think it'd be a better API if we passed an FVK instead of the change address, but I wasn't sure how to do that and the changes were quite large.

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
-}
-
-impl<R: RngCore + CryptoRng> Debug for Planner<R> {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Builder")
-            .field("balance", &self.balance)
-            .field("plan", &self.plan)
-            .finish()
-    }
+struct ActionList {
```
hdevalence, line 33: This was intended as just an action list but grew to become something like a very very barebones planner-backbone. However, I think it's probably good to keep decoupled from the Rust view server implementation, considering how we ended up here.

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
+        actions.push(rate_data.build_undelegate(value.amount).into());
     }
 
-    /// Cast a validator vote in this transaction.
-    pub fn validator_vote(&mut self, vote: ValidatorVote) -> &mut Self {
-        self.action(ActionPlan::ValidatorVote(vote));
-        self
+    /*
+    for tpr::UndelegateClaim { .. } in request.undelegation_claims {
+        // need to wait for a new release of monorepo
```
hdevalence, line 264: We need to wait for the web repo to consume a new release of the core Rust crates to pick up these proto definitions. But when we do, this will be the place to process them.

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
-                return Err(anyhow!(
-                    "can't vote on proposal {} because no delegation notes were staked to an active validator when voting started",
-                    proposal
-                ));
-            }
+    let fee_tier = match request.fee_mode {
+        None => FeeTier::default(),
+        Some(tpr::FeeMode::AutoFee(tier)) => tier.try_into()?,
+        Some(tpr::FeeMode::ManualFee(_)) => {
+            return Err(anyhow!("Manual fee mode not yet implemented").into());
```
hdevalence, line 294: We could leave this unimplemented presumably indefinitely -- at least until someone runs across the error (thus showing that there's actually a use case for it).

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
 
-        // For any remaining provided balance, make a single change note for each
-        for value in self.balance.provided().collect::<Vec<_>>() {
-            self.output(value, self_address);
-        }
+        // Recompute the change outputs, without accounting for fees.
+        actions.refresh_change(change_address);
+        // Now re-estimate the fee of the updated transaction and adjust the change if possible.
+        let fee = actions.fee_estimate(&gas_prices, &fee_tier);
+        actions.adjust_change_for_fee(fee);
```
hdevalence, line 350: This first updates all the planned change outputs, so they're in place and counting towards the gas and hence the fee estimate, and _then_ adjusts the change note for the staking token to fund as much of the fee estimate as possible. 

So, if we've spent more than is required, we compute all the necessary change, and then "spend out of the change note" to cover fees.

Or, if we've spent less than is required, we'll set the value of the change output to 0, so we have a smaller imbalance to try to repair in the next iteration of the loop.

packages/wasm/crate/src/planner.rs
```diff
@@ -1,591 +1,389 @@
...
-        self.balance += Value {
-            amount: excess_fee_spent,
-            asset_id: *STAKING_TOKEN_ASSET_ID,
+    let mut iterations = 0usize;
+
+    while let Some(required) = actions
+        .balance_with_fee_estimate(&gas_prices, &fee_tier)
+        .required()
+        .next()
+    {
```
hdevalence, line 330: cc @zbuc, I think this gives us a way out of the horrible over-estimate-then-adjust-then-hope method we had to do previously. I am pretty sure it would work well enough, but also that it would work perfectly if we changed the gas schedule so that the gas cost of the `Transaction` was _exactly_ the sum of the gas of its `Action`s.

packages/wasm/crate/src/storage.rs
```diff
@@ -104,39 +106,48 @@ impl IndexedDBStorage {
...
+            }
+
+            total += record.note.amount();
+            records.push(record);
+
+            if let Some(amount_to_spend) = amount_to_spend {
+                if total >= amount_to_spend {
+                    break;
+                }
+            }
```
hdevalence, line 147: This fixes a bug in the previous implementation -- if you've ever looked at a transaction and noticed that it spent more notes than necessary, this is why. (The behavior is however no longer used by the current planner implementation).

packages/wasm/src/planner2.ts
```diff
@@ -0,0 +1,46 @@
...
+        idbConstants,
+        request,
+        changeAddress,
+        fmdParams,
+        sctParams,
+        gasPrices,
+        chainId,
+    )) as JsonValue;
+    */
+    const jsonPlan = ret as JsonValue;
```
hdevalence, line 43: this needs cleanup, I have no idea why `gasPrices` is marked in the console as "Object" rather than "Message", but putting `.toJson()` on it causes everything to blow up (and there's some leftover code from along the way to discovering that).

Valentine1898:
Initially, I assumed we were moving towards using wasm only in places where it was critically needed 
- the use of specific cryptography 
- heavy code duplication 
I also assumed that we were moving towards avoiding reading and writing to IndexedDb for wasm. For reasons such as:
- we still don't know how to test wasm with indexedDb global 
- we don't have good control over everything that happens in wasm 
I have an idea that instead of having one big wasm function that turns `TransactionPlannerRequest` to `TransactionPlan` (and which is kind of a black box for TS) we can have many small wasm wrapper functions over things we can't do in TS. 
For example 
```rust
for tpr::Swap {
        value,
        target_asset,
        fee,
        claim_address,
    } in request.swaps
    {
        let value: Value = value
            .ok_or_else(|| anyhow!("missing value in swap"))?
            .try_into()?;
        let target_asset = target_asset
            .ok_or_else(|| anyhow!("missing target asset in swap"))?
            .try_into()?;
        let fee = fee
            .ok_or_else(|| anyhow!("missing fee in swap"))?
            .try_into()?;
        let claim_address = claim_address
            .ok_or_else(|| anyhow!("missing claim address in swap"))?
            .try_into()?;

        // Determine the canonical order for the assets being swapped.
        // This will determine whether the input amount is assigned to delta_1 or delta_2.
        let trading_pair = TradingPair::new(value.asset_id, target_asset);

        // If `trading_pair.asset_1` is the input asset, then `delta_1` is the input amount,
        // and `delta_2` is 0.
        //
        // Otherwise, `delta_1` is 0, and `delta_2` is the input amount.
        let (delta_1, delta_2) = if trading_pair.asset_1() == value.asset_id {
            (value.amount, 0u64.into())
        } else {
            (0u64.into(), value.amount)
        };

        // If there is no input, then there is no swap.
        if delta_1 == Amount::zero() && delta_2 == Amount::zero() {
            return Err(anyhow!("No input value for swap").into());
        }

        // Create the `SwapPlaintext` representing the swap to be performed:
        let swap_plaintext = SwapPlaintext::new(
            &mut OsRng,
            trading_pair,
            delta_1,
            delta_2,
            fee,
            claim_address,
        );

        let swap = SwapPlan::new(&mut OsRng, swap_plaintext).into();

        actions.push(swap);
    }
``` 
We can replicate all this logic on TS, but we can't replicate SwapPlan::new() because it uses cryptography to generate proof_blinding_r and proof_blinding_s
https://github.com/penumbra-zone/penumbra/blob/5863116e3a2e0a0ee379eb22818316e40c4be275/crates/core/component/dex/src/swap/plan.rs#L26-L37
The idea is that we create a wasm wrapper for SwapPlan::new(), but move all the rest of the logic to TS. 
In this case wasm planner will become a set of small helper functions (obviously there will be quite a lot of them), but most of the logic will be moved to TS (hopefully all indexed-db interaction) and covered by tests.

hdevalence:
The problem is that the transaction planner needs to do fee estimation, and because of that it needs to do exact arithmetic on `Value`s, which is not really possible to do separately from the Rust code. So it's better that the entire thing be on the Rust side, where at least it can be simple, straight-line code.

packages/wasm/crate/src/planner.rs
zbuc: It doesn't look like we're even including any transaction-level byte cost right now FWIW:

```rust
impl GasCost for Transaction {
    fn gas_cost(&self) -> Gas {
        self.actions().map(GasCost::gas_cost).sum()
    }
}
```

packages/wasm/crate/src/planner.rs
zbuc: I don't think this necessarily fixes it because:

1. Right now the gas cost of the `Transaction` **is** exactly the sum of the gas of its `Action`s
2. The reason we need to do the over-estimate-then-adjust dance is because paying the fee can involve adding an unknown number of additional `Spend` actions to the transaction, depending on the available notes. So we would still have an unknown number of additional `Spend` actions (to pay fees) that we recursively have to pay fees for.

packages/router/src/grpc/view-protocol-server/transaction-planner.ts
```diff
@@ -1,78 +1,25 @@
...
-    querier,
   } = await services.getWalletServices();
+
   const fmdParams = await indexedDb.getFmdParams();
   if (!fmdParams) throw new ConnectError('FmdParameters not available', Code.FailedPrecondition);
   const { chainId, sctParams } = (await indexedDb.getAppParams()) ?? {};
   if (!sctParams) throw new ConnectError('SctParameters not available', Code.FailedPrecondition);
   if (!chainId) throw new ConnectError('ChainId not available', Code.FailedPrecondition);
+  const gasPrices = await indexedDb.getGasPrices();
+  if (!gasPrices) throw new ConnectError('Gas prices is not available', Code.FailedPrecondition);
```
Valentine1898, line 19: It probably makes sense to keep these checks even though wasm will read this data directly from indexed-db. This way we can return an error with FailedPrecondition code to clients

Valentine1898:
✅ wasm reads all necessary data directly from indexed-db 
✅ pass FVK to wasm instead of passing refundAddress 
✅ cleanup  

This works for most transactions, but when trying to spend a penumbra (UM) token I get an error 
ConnectError: [unknown] Tendermint: failed to deliver transaction: check_stateless failed: a spend proof did not verify: spend proof did not verify

❌ wasm tests are still broken

Valentine1898:
image
The error seems to be related to zero spend

hdevalence:
@Valentine1898 good catch. I have an idea of a clean way to handle it, I'll push a commit onto the branch

Valentine1898:
image
I'm not sure at all if this error is related to this PR, because I found in indexed-db note with zero amount, maybe it's a view service issue

hdevalence:
@Valentine1898 I think the problem with zero valued notes is that they're treated differently in the proofs -- there's a capability to do "dummy spends" at the protocol level, I think this makes spending a zero-valued note in the normal way not possible. @redshiftzero would know the exact details.

In the commit I pushed, we filter them out at the same time as prioritizing notes.

Valentine1898:
> @Valentine1898 I think the problem with zero valued notes is that they're treated differently in the proofs -- there's a capability to do "dummy spends" at the protocol level, I think this makes spending a zero-valued note in the normal way not possible. @redshiftzero would know the exact details.


I got a zero note as a result of a swap claim, I thought there used to be zero note filtering somewhere at the block scanner layer, but I haven't found anything like it

Valentine1898:
> In the commit I pushed, we filter them out at the same time as prioritizing notes.

That fixed the error, awesome

Valentine1898:
https://github.com/penumbra-zone/penumbra/blob/d2ec71f38f73cb2f1e8c57a42c5744f706e64adb/crates/core/component/shielded-pool/src/note_payload.rs#L26-L31
Found it, zero notes are legal

Valentine1898 (approved):
✅ wasm tests fixed  

In general, feel free to merge. 
Not sure if we should wait for https://github.com/penumbra-zone/penumbra/issues/3940 and https://github.com/penumbra-zone/penumbra/issues/3941

turbocrime:
oh excellent! thank you

PR #664, opened 2024-03-04
display token symbol in asset selector

hdevalence:


grod220 (approved):
Some unused vars to delete, but otherwise LGTM

grod220:
Discussed with Henry about taking this over

Issue #138, opened 2023-10-25
Npm package publishing


grod220:
We need a strategy/method on how we publish our `@penumbra-zone/<package>` 's. How often should we do this? Is there a way CI/CD can automatically detect diffs of packages and publish the new version?

TalDerei:
Referenced in #218. The NPM publishing workflow exists in a state where the web app and extension are synchronized, and the dialogue between them hasn't been the source of breaking changes. The combination of protobuf spec changes and new testnet deployments, without publishing an updated extension, was the major source of breakage. Finalizing the protobuf spec https://github.com/penumbra-zone/penumbra/issues/3720 helps, but there are still pain points related to web-related breaking changes if there's a mismatch between web app and state of the extension. 

One track we've discussed is a multi-branch workflow where 

@grod220 cc

TalDerei:
Referenced in https://github.com/penumbra-zone/web/issues/218. The NPM publishing workflow exists in a state where the web app and extension are synchronized, and the dialogue between them hasn't been the source of breaking changes. The combination of protobuf spec changes and new testnet deployments, without publishing an updated extension, was the major source of breakage. Finalizing the protobuf spec https://github.com/penumbra-zone/penumbra/issues/3720 helps, but there are still pain points related to web-related breaking changes if there's a mismatch between web app and state of the extension.

One track we've discussed is a multi-branch workflow where we have a development branch that we merge things into instead of main acting as the deployment branch. The main branch would act as what's in production, and the dev branch would be what's in preview. This would increase the complexity of the workflow, but it's one method to reduce the number of breaking changes. @conorsch iirc, you already do this implicitly when we release a point release in the monorepo. 

Another track would be an additional manual github action, where the GitHub action would be a production web app deployment. We’d have the existing continuous deployments going to a preview, and the secondary one would be a manual action whenever we publish or there's an extension update. This has also been recently discussed as a viable option. 

@grod220 cc

grod220:
At the moment, given things are in flux, we are simply deploying the web app + extension manually via github actions. However, maybe after a month or two, we should return to continuous deployment of the web app (assuming the messaging api is stable).

This issue really had in mind third-party wallets (keplr, leap) that we eventually will want to expose all of our work. We want them (with fairly little effort) to be able to import and use these packages to enable using Penumbra for their users. However, at the moment, we aren't publishing them to npm.

grod220:
[...]

PR #672, opened 2024-03-05
Add changeset support (npm publishing part 1)

grod220:
Closes https://github.com/penumbra-zone/web/issues/138

We have a few consumers interested in making requests to the extension. That means we need to start publishing our packages to npm 📦.  Followed the recommendations from [turborepo docs](https://turbo.build/repo/docs/handbook/publishing-packages/versioning-and-publishing) in using changesets for versioning+publishing.

- Marked packages we don't anticipate publishing with private
- Uses changeset github action
- Added deployment docs
- Adds compilation for a few packages so they can serve as independent packages

Specifically targeting `client` package to publish first. However, need to handle types, getters, constants, and transport-dom as well due to dep tree.


turbocrime:
it seems unnecessary to bother with CJS support. both node and the browser have full ESM support now

.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   // This tells ESLint to load the config from the package `eslint-config-custom`
-  extends: ['custom'],
+  extends: ['@penumbra-zone/eslint-config-custom'],
```
grod220, line 4: Note: packages published externally need to have their dependencies published as well

apps/extension/src/approve-origin.ts
```diff
@@ -4,8 +4,8 @@ import { errorFromJson } from '@connectrpc/connect/protocol-connect';
 import { localExtStorage } from '@penumbra-zone/storage';
 import { OriginApproval, PopupType } from './message/popup';
 import { popup } from './popup';
-import Map from '@penumbra-zone/polyfills/Map.groupBy';
-import { UserChoice } from '@penumbra-zone/types/src/user-choice';
+import Map from '@penumbra-zone/polyfills/src/Map.groupBy';
```
grod220, line 7: All polyfill files moved into /src folder for organizational purposes (did this with a few packages). Most of the time created barrel files, but it doesn't really fit so well with polyfills given some of them produce side effects upon importing.

packages/constants/src/chains.ts
```diff
@@ -1,5 +1,10 @@
 // Canonical data source: https://github.com/cosmos/chain-registry/tree/master
-import { Chain } from '@penumbra-zone/types';
+export interface Chain {
+  displayName: string;
+  chainId: string;
+  ibcChannel: string;
+  iconUrl: string;
+}
```
grod220, line 7: We had a hidden cyclical dependency

packages/crypto/package.json
```diff
@@ -1,15 +1,26 @@
...
+    "build": "pnpm clean && tsc && vite build",
+    "clean": "rm -rfv dist",
     "lint": "eslint \"**/*.ts*\"",
     "test": "vitest run"
   },
   "dependencies": {
     "bip39": "^3.1.0",
-    "crypto-js": "^4.2.0"
+    "crypto-js": "^4.2.0",
+    "@penumbra-zone/types": "workspace:*"
```
grod220, line 23: Two reasons why `@penumbra-zone/types` was put back in the local directories:
- Cyclical dependencies (not detected when in parent package.json)
- Turbo's automatic detection of dependencies did not work without it

packages/router/src/grpc/view-protocol-server/fmd-parameters.test.ts
```diff
@@ -23,7 +23,9 @@ describe('FmdParameters request handler', () => {
       getFmdParams: vi.fn(),
     };
     mockServices = {
-      getWalletServices: vi.fn(() => Promise.resolve({ indexedDb: mockIndexedDb })),
+      getWalletServices: vi.fn(() =>
+        Promise.resolve({ indexedDb: mockIndexedDb }),
+      ) as MockServices['getWalletServices'],
```
grod220, line 28: Given we have to build all these packages, type errors are now getting caught that went unchecked before

packages/router/src/grpc/view-protocol-server/ephemeral-address.ts
```diff
@@ -1,9 +1,9 @@
-import type { Impl } from '.';
+import type { ViewImpl } from '.';
 import { servicesCtx } from '../../ctx';
 
 import { getEphemeralByIndex } from '@penumbra-zone/wasm';
 
-export const ephemeralAddress: Impl['ephemeralAddress'] = async (req, ctx) => {
+export const ephemeralAddress: ViewImpl['ephemeralAddress'] = async (req, ctx) => {
```
grod220, line 6: Easier to re-export in barrel file if they aren't all named the same `Impl` var

packages/services/package.json
```diff
@@ -1,16 +1,27 @@
 {
   "name": "@penumbra-zone/services",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
grod220, line 3: These packages might have strange version numbers, but that's just because I've been publishing to test it out. Don't think we need to start on a clean 1.0.0 anyway.

packages/client/package.json
```diff
@@ -1,9 +1,23 @@
...
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "scripts": {
+    "build": "pnpm clean && tsc && vite build",
```
grod220, line 15: Conundrum, vite build is not idempotent for some reason. Exclude in tsconfig is not respected and therefore tries to re-bundle dist.

```
{
  "extends": "@penumbra-zone/tsconfig/base.json",
  "exclude": ["node_modules"],
  "compilerOptions": {
    "outDir": "dist"
  }
}

```

.github/workflows/turbo-ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   turbo-lint:
     name: Lint
-    runs-on: buildjet-2vcpu-ubuntu-2204
+    runs-on: buildjet-4vcpu-ubuntu-2204
```
grod220, line 33: We need more ram now...

.eslintrc.js
turbocrime: there's a couple options here which don't require publishing an eslint config package

1. transition to eslint flat config, so it becomes a workspace dependency
2. programmatically remove the eslint devDeps with a `prepack` configuration

packages/client/package.json
turbocrime: calling `vite build` manually demonstrates that it is in fact idempotent. it's actually a `tsc` problem - the typescript compiler is refusing to clobber existing output

after some experimenting, it seems that `tsc` is unnecessary, and `vite build` by itself doesn't complain. if we're using `vite-plugin-dts` we should either not call `tsc` or use `noEmit` in tsconfig anyway

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime, line 1: barrel import change causes this file to grow from 1574 bytes to _399,176 bytes_ (between each of these `wc` i'm changing branches and rebuilding in another terminal)
Screenshot 2024-03-12 at 14 20 55

these injected scripts are very important to minimize. due to the specified requirements of the connection procedure, we unconditionally inject these scripts into every page the user visits.

turbocrime (changes requested):
it's wild that turbo doesn't handle top-level workspace dependencies, but maybe that's just a sign that our `types` package is overgrown.

i have several suggestions and i'll either PR some changes here or make a separate example branch.

grod220:
A new starting point in attempt to publish `client` package. Required:
- client
- constants
- getters
- polyfills
- transport-dom

Reverted
- crypto
- estlint-custom-config
- query
- router
- services
- storage
- tailwind-config
- transport-chrome
- ts-config

Surprised that eslint/ts-config is no longer giving publishing issues. I suppose it's because they are parent package.json deps now? Can't remember the issue before.

packages/client/package.json
grod220: I think we should keep tsc. It has managed to find type errors in a few packages that are otherwise going unnoticed.

grod220:
Removed the barrel files for these packages we are now exposing. Thought, it kinda feels... wrong. I feel like there is major issue in javascript land that a clean API and performant imports are not compatible. Regardless, this is something perhaps we can explore ongoingly.

apps/extension/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "tsconfig/extension.json",
+  "extends": "@penumbra-zone/tsconfig/extension.json",
```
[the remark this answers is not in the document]
grod220: Turns out we don't need to, think this may have been an issue before because a package explicitly set it as a dep rather than inherited it

packages/client/package.json
```diff
@@ -1,9 +1,24 @@
...
+  "description": "Package for connecting to Prax extension",
+  "type": "module",
+  "files": [
+    "dist"
+  ],
+  "exports": {
+    ".": "./dist/prax.js",
+    "./global": "./dist/global.js",
+    "./src/global": "./dist/global.js",
+    "./get-port": "./dist/get-port.js"
```
grod220, line 13: So this is kinda the demo of how we can make external packages a bit nicer.

`"./src/global": "./dist/global.js",` <-- the way that internal packages require the exports
`"./global": "./dist/global.js",` <-- external packages get the ability to lose the `/src/` part of the import
`".": "./dist/prax.js",` <-- the default `.` is the only harmonized internal/external import that drops the need for `/src/`. In this case, gave that honor to prax.js as it's likely the most common import for this package.

turbocrime: from my example

https://github.com/penumbra-zone/web/blob/b1a2a334fef98ddbcd92134eafcc47cf9a0b6f55/packages/client/package.json#L10-L28

the types package, which is private in example and needs no publication, remains unchanged

packages/client/package.json
turbocrime: ok - then let's use tsconfig noEmit

packages/router/src/grpc/view-protocol-server/ephemeral-address.ts
turbocrime: the type is for local use, so we should probably break this out into another file that isn't exported from the package

also, now that we're no longer doing barrel this change could revert

packages/router/src/grpc/view-protocol-server/fmd-parameters.test.ts
turbocrime: #680

package.json
```diff
@@ -17,7 +17,9 @@
     "format-check": "prettier --check .",
     "format-check:rust": "turbo run format-check:rust",
     "playwright-install": "playwright install",
-    "all-check": "pnpm install && pnpm compile && pnpm format-check && pnpm lint && pnpm test && pnpm download-keys && pnpm build && pnpm format-check:rust && pnpm lint:rust && pnpm test:rust"
+    "changeset": "changeset",
+    "changeset:publish": "changeset publish",
+    "all-check": "pnpm install && pnpm clean && pnpm compile && pnpm format-check && pnpm lint && pnpm test && pnpm download-keys && pnpm build && pnpm format-check:rust && pnpm lint:rust"
```
turbocrime, line 22: i think having clean in all-check makes dev difficult because it blows away keys and results in a lot of redownload

apps/extension/src/message/popup.ts
```diff
@@ -1,12 +1,12 @@
...
 import type { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1/custody_pb';
-import type {
+import {
   InternalMessage,
   InternalRequest,
   InternalResponse,
 } from '@penumbra-zone/types/src/internal-msg/shared';
-import type { Jsonified } from '@penumbra-zone/types/src/jsonified';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
+import { Jsonified } from '@penumbra-zone/types/src/jsonified';
```
turbocrime, line 9: i often used type imports to ensure they generated no actual code effect, we shouldn't convert type imports to normal imports

apps/extension/src/popup.ts
```diff
@@ -1,10 +1,7 @@
 import { sessionExtStorage } from '@penumbra-zone/storage';
-import type {
```
turbocrime, line 2: type import conversion seems to occur in many places

apps/minifront/src/state/staking/index.test.ts
```diff
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { StoreApi, UseBoundStore, create } from 'zustand';
+import { create, StoreApi, UseBoundStore } from 'zustand';
```
turbocrime, line 2: it seems like a good chunk of these changes are due to import sorting, can you turn that off

grod220:
> this still seems overcomplicated with many unnecessary changes

Most of these changes were due to the fact we are now building and it exposed a number of typing issues.

> and we really don't need to be publishing internal packages like tsconfig

It is not, it's a private package

apps/extension/src/popup.ts
grod220: Yep, will fix

apps/minifront/src/state/staking/index.test.ts
grod220: lol definitely not

packages/types/src/translators/address-view.test.ts
```diff
@@ -9,7 +9,7 @@ describe('asOpaqueAddressView()', () => {
   describe('when the address view is visible', () => {
     const addressView = new AddressView({
       addressView: {
-        case: 'visible',
+        case: 'decoded',
```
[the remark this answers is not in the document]
grod220: A type error that's been fixed

packages/client/package.json
grod220: The types package is a dependency of this package. If we do not publish it, it will throw an error when the external app goes to npm install it

turbocrime:
the polyfill publish is broken

turbocrime (approved):
merge it create issues for further concerns

PR #673, opened 2024-03-05
Small refactor to how proving keys are stored and accesssed

jessepinho:
While working on another branch, I recalled @hdevalence 's comment that we could store proving keys as a map of each action type to the proving key required for that action. Accessing them from an array was a bit cumbersome, so I converted them to an object to make it explicit which key is needed for which action.

This PR also removes the nullifier derivation proving key, which [Henry said](https://discord.com/channels/824484045370818580/883455571003576320/1214670615622066206) is not actually needed on web.

### DIFF
--- a/apps/extension/src/utils/download-proving-keys.ts
+++ b/apps/extension/src/utils/download-proving-keys.ts
@@ -1,31 +1,42 @@
 import fetch from 'node-fetch';
 import * as fs from 'fs';
 import * as path from 'path';
-import { provingKeys } from '@penumbra-zone/types/src/proving-keys';
+import { ProvingKey, provingKeysByActionType } from '@penumbra-zone/types/src/proving-keys';
 
-const VERSION_TAG = 'v0.68.0';
+const main = () => {
+  const VERSION_TAG = 'v0.68.0';
 
-const githubSourceDir = `https://github.com/penumbra-zone/penumbra/raw/${VERSION_TAG}/crates/crypto/proof-params/src/gen/`;
+  const githubSourceDir = `https://github.com/penumbra-zone/penumbra/raw/${VERSION_TAG}/crates/crypto/proof-params/src/gen/`;
 
-const binDir = path.join('bin');
+  const binDir = path.join('bin');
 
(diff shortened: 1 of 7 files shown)


packages/types/src/utility.ts
```diff
@@ -6,6 +6,3 @@ export const isEmptyObj = <T>(input: T): input is T & EmptyObject => {
   }
   return false;
 };
-
-export const camelToSnakeCase = (str: string) =>
```
jessepinho, line 10: No longer used anywhere

apps/extension/src/utils/download-proving-keys.ts
grod220, line 1: Not a big deal, but this file feels like it could be wrapped in a main() or something instead of being all top level

packages/types/src/proving-keys.ts
```diff
@@ -1,9 +1,32 @@
...
-];
+import { Action } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+
+type ActionType = Exclude<Action['action']['case'], undefined>;
+export interface ProvingKey {
+  keyType: string;
+  file: string;
+}
+
+export const provingKeysByActionType: Record<ActionType, ProvingKey | null> = {
```
grod220, line 9: Can we convert the `null`'s to `undefined`? Not sure how you feel, but I lean toward the opinion that null was a mistake in the javascript language.

grod220: Another option is to make this record a partial so it doesn't need them all enumerated. But perhaps it's better that we see them all? 🤷

grod220 (approved):

packages/types/src/proving-keys.ts
jessepinho: Ah, good call — I made them undefined. I like the idea that, even for us coders' sake, we see all the action names and what keys are required for each, even if the answer is "none."

turbocrime:
ah, i accidentally did some parallel work on this. let me see how it compares

apps/extension/src/utils/download-proving-keys.ts
jessepinho: Cool, done.

PR #674, opened 2024-03-05
Split getters package

turbocrime:
for consumption 🤒


apps/webapp/package.json
```diff
@@ -32,6 +32,7 @@
     "zustand": "^4.5.1"
   },
   "devDependencies": {
+    "@penumbra-zone/getters": "workspace:*",
```
grod220, line 35: This should be a normal dependency given it's a runtime package and not really dev tooling

packages/getters/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "1.0.0",
+  "description": "Convenience getters for the deeply nested optionals of Penumbra's protobuf types",
+  "type": "module",
+  "scripts": {
+    "lint": "eslint \"**/*.ts*\"",
+    "test": "vitest run"
+  },
+  "devDependencies": {
+    "bech32": "^2.0.0"
+  }
```
grod220, line 12: Think this should be a normal dep too

packages/getters/asset.ts
turbocrime, line 1: notably, this file does not contain a gitter but would have caused a circular dependency. so i just moved it to getters

grod220 (approved):
Think a passing build + moving the package to the normal dep section and it's good to merge

jessepinho (approved):
Nice! 

I'd personally put the getter files (except `index.ts`) underneath a `/src` directory in that package. But up to you.

grod220:
> I'd personally put the getter files (except index.ts) underneath a /src directory in that package. But up to you.

Ah yes, we should do that

PR #675, opened 2024-03-05
move proving key logic into wasm crate

turbocrime:
some commentary on jesse's changes, deferred so they can merge


jessepinho (approved):
Good call 👍🏻

jessepinho (approved):
Cool, still LGTM

PR #679, opened 2024-03-06
Filter the `<AccountSwitcher />` on the staking page to only allow switching between relevant accounts

jessepinho:
https://github.com/penumbra-zone/web/assets/1121544/bbf8799d-eee7-432e-98a6-29121309775c

Currently, the account switcher allows users to switch between any of their accounts, between 0 and 2^32. 

In some contexts, though, we want to restrict which accounts the user should be able to switch between. For example, on the staking page, we only want to allow the user to switch between accounts that have unstaked UM tokens, delegation tokens, and/or unbonding tokens. (Otherwise, there's no staking-related action the user could take on that account.)

This PR adds a `filter` prop to `<AccountSwitcher />` to make that functionality possible, then uses that prop in the staking page.

### DIFF
--- a/packages/ui/components/ui/account-switcher.tsx
+++ b/packages/ui/components/ui/account-switcher.tsx
@@ -54,18 +54,21 @@ export const AccountSwitcher = ({
 
   return (
     <div className='flex items-center justify-between'>
-      <Button variant='ghost' className={cn('hover:bg-inherit', account === 0 && 'cursor-default')}>
-        {shouldShowPreviousButton ? (
+      {shouldShowPreviousButton ? (
+        <Button
+          variant='ghost'
+          className={cn('hover:bg-inherit', account === 0 && 'cursor-default')}
+        >
           <ArrowLeftIcon
             aria-label='Previous account'
             role='button'
             onClick={handleClickPrevious}
             className='size-6 hover:cursor-pointer'
           />
-        ) : (
-          <span className='size-6' />
-        )}
-      </Button>
+        </Button>
+      ) : (
+        <span className='size-6' />
+      )}
       <div className='select-none text-center font-headline text-xl font-semibold leading-[30px]'>
         <div className='flex flex-row flex-wrap items-end gap-[6px]'>
           <span id='AccountSwitcher__label'>Account</span>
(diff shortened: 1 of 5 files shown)


packages/ui/components/ui/account-switcher.tsx
```diff
@@ -13,42 +13,85 @@ const MAX_INDEX = 2 ** 32;
...
             className='size-6 hover:cursor-pointer'
           />
-        ) : (
-          <span className='size-6' />
-        )}
-      </Button>
+        </Button>
+      ) : (
+        <span className='size-6' />
+      )}
```
jessepinho, line 71: I also noticed here that we weren't hiding the `Button` element, but only the arrow element — meaning that, when we shouldn't have the Previous button at all, we were still rendering an invisible one 😆 

Same for the Next button at the bottom of this file.

packages/ui/components/ui/account-switcher.tsx
```diff
@@ -13,42 +13,85 @@ const MAX_INDEX = 2 ** 32;
 export const AccountSwitcher = ({
   account,
   onChange,
+  filter,
 }: {
   account: number;
   onChange: (account: number) => void;
+  /**
+   * An array of address indexes to switch between, if you want to limit the
+   * options. Must be pre-sorted! `AccountSwitcher` won't sort this for you.
```
grod220, line 22: But it looks like it's being sorted with `[...filter].sort()` no?

packages/ui/components/ui/account-switcher.tsx
```diff
@@ -13,42 +13,85 @@ const MAX_INDEX = 2 ** 32;
...
 }) => {
   const [inputCharWidth, setInputCharWidth] = useState(1);
 
+  const sortedFilter = useMemo(() => (filter ? [...filter].sort() : undefined), [filter]);
+
+  const handleClickPrevious = () => {
+    if (sortedFilter) {
+      const previousAccount = sortedFilter[sortedFilter.indexOf(account) - 1];
+      if (previousAccount !== undefined) onChange(previousAccount);
+    } else onChange(account - 1);
```
grod220, line 34: Think we should have brackets here. I was at first thought the else corresponded to the if clause above.

packages/ui/components/ui/account-switcher.tsx
```diff
@@ -13,42 +13,85 @@ const MAX_INDEX = 2 ** 32;
...
-        )}
-      </Button>
+        </Button>
+      ) : (
+        <span className='size-6' />
+      )}
       <div className='select-none text-center font-headline text-xl font-semibold leading-[30px]'>
         <div className='flex flex-row flex-wrap items-end gap-[6px]'>
-          <span>Account</span>
+          <span id='AccountSwitcher__label'>Account</span>
```
grod220, line 74: Is the id needed?

grod220 (approved):
Good tests!

packages/ui/components/ui/account-switcher.tsx
jessepinho: Whoops, outdated comment. will fix

packages/ui/components/ui/account-switcher.tsx
jessepinho: Whoops, left that in accidentally from when I was using `aria-labelledby`. Will remove.

packages/ui/components/ui/account-switcher.tsx
jessepinho: Dang it, I think I forgot to self-review this. My bad. Will fix.

PR #685, opened 2024-03-06
npm packages

turbocrime:
suggestions for @grod220 #672 

entry to vite-built packages is not complete, it renames them or something


config/tsconfig/base.json
turbocrime, line 1: see https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler

also, strictNullChecks is already enabled by strict

apps/extension/src/utils/rethrow-impl-errors.ts
turbocrime, line 1: this file was weirdly reaching deep into connectrpc's cjs-specific exports to pull out some non-exported types. i just used an equivalent type - this isn't actually a refactor.

config/tailwind-config/package.json
turbocrime, line 1: moved internal-only 'config' packages to a config dir, un-namespaced tailwind to match other configs

apps/node-status/package.json
turbocrime, line 1: every package with a 'build' script now has a corresponding 'clean' script

packages/ui/package.json
turbocrime, line 1: this is only complex because the package is complex :/

grod220:
Looks like our branches diverged quite a bit. Think it's best to keep `changset-npm` as the canonical branch, but will definitely consult this work as I go.

PR #690, opened 2024-03-06
add tests for transaction-planner

Valentine1898:
The simplification of wasm transaction planner https://github.com/penumbra-zone/web/issues/609 is already finished, so it no longer makes sense for us to hold off on these tests

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/transaction-planner.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/transaction-planner.test.ts
@@ -1,0 +1,127 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+import { createContextValues, createHandlerContext, HandlerContext } from '@connectrpc/connect';
+import { ViewService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/view/v1/view_connect';
+import { servicesCtx } from '../../ctx';
+import { IndexedDbMock, MockServices, ViewServerMock } from '../test-utils';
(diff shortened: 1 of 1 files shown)


packages/router/src/grpc/view-protocol-server/transaction-planner.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import { FmdParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/shielded_pool/v1/shielded_pool_pb';
+import { AppParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/app/v1/app_pb';
+import { SctParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/sct/v1/sct_pb';
+import { GasPrices } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1/fee_pb';
+import { transactionPlanner } from './transaction-planner';
+
+const mockPlanTransaction = vi.hoisted(() => vi.fn());
+vi.mock('@penumbra-zone/wasm', () => ({
+  planTransaction: mockPlanTransaction,
+}));
```
Valentine1898, line 17: We didn't get rid of the indexed-db global in wasm and we don't have that goal anymore, so we have no choice but to use mock here

grod220:
Agree, tests in the crate will be far more meaningful. Could you create a tracking issue with a proposal of tests we can write for the crate to increase coverage on that side?

grod220 (approved):

Valentine1898:
> Agree, tests in the crate will be far more meaningful. Could you create a tracking issue with a proposal of tests we can write for the crate to increase coverage on that side?

Yep https://github.com/penumbra-zone/web/issues/691

PR #696, opened 2024-03-06
Update json viewer pkg

grod220:
We've been getting peer dependency errors:

Screenshot 2024-03-06 at 4 42 33 PM

Turns out our json viewer is quite out of date and [unmaintained](https://github.com/mac-s-g/react-json-view).

This PR updates to one that looks like the successor to this library. Was able to achieve a nearly identical design. Uses a base16 color theming thing. Kinda strange, but 🤷 
Screenshot 2024-03-06 at 5 33 12 PM

Fyi, these are peer dependencies of this lib:
```
"@emotion/react": "^11.11.4",
"@emotion/styled": "^11.11.0",
"@mui/material": "^5.15.12",
```


turbocrime:
i've been looking at this too and thought about https://www.npmjs.com/package/react-json-view-lite which is much smaller - we'd save over 1mb by using this instead, which is important as we're serving the entire frontend as a blob.

packages/ui/components/ui/json-viewer.tsx
```diff
@@ -0,0 +1,40 @@
+import { JsonViewer as TextaJsonViewer, NamedColorspace } from '@textea/json-viewer';
+
+import type { JsonObject, JsonValue } from '@bufbuild/protobuf';
+
+export const customTheme: NamedColorspace = {
+  scheme: 'custom',
+  author: 'penumbra',
+  base00: '#000000', // Background Color
+  base01: '#ffffff',
+  base02: '#212121', // Nested section border color
```
jessepinho, line 10: is it possible to use `--var`s?

packages/ui/package.json
```diff
@@ -11,6 +11,9 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
+    "@emotion/react": "^11.11.4",
+    "@emotion/styled": "^11.11.0",
+    "@mui/material": "^5.15.12",
```
jessepinho, line 16: Hm, are these required? These are kinda hefty packages that will require a lot of CPU time for just a JSON viewer.

jessepinho (approved):
LGTM! Just a couple small notes

grod220:
I'll check it out 👀

grod220:
With `react-json-view-lite`, we have to do more work on the style-side of things, but wasn't too bad adding the tailwind classes. Got it to look quite close to what we have now.

jessepinho (approved):
LGTM 🔥

PR #698, opened 2024-03-06
Fix logic for whether we're currently catching up

jessepinho:
It's currently backwards.

### DIFF
--- a/packages/router/src/grpc/view-protocol-server/status.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/status.test.ts
@@ -28,7 +28,9 @@ describe('Status request handler', () => {
     };
 
     mockServices = {
-      getWalletServices: vi.fn(() => Promise.resolve({ indexedDb: mockIndexedDb })),
+      getWalletServices: vi.fn(() =>
+        Promise.resolve({ indexedDb: mockIndexedDb }),
+      ) as MockServices['getWalletServices'],
       querier: {
         tendermint: mockTendermint,
       },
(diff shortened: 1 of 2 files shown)


packages/router/src/grpc/view-protocol-server/status.test.ts
```diff
@@ -28,7 +28,9 @@ describe('Status request handler', () => {
     };
 
     mockServices = {
-      getWalletServices: vi.fn(() => Promise.resolve({ indexedDb: mockIndexedDb })),
+      getWalletServices: vi.fn(() =>
+        Promise.resolve({ indexedDb: mockIndexedDb }),
+      ) as MockServices['getWalletServices'],
```
jessepinho, line 33: Fixes a TypeScript complaint

packages/router/src/grpc/view-protocol-server/status.ts
```diff
@@ -9,7 +9,7 @@ export const status: Impl['status'] = async (_, ctx) => {
   if (!fullSyncHeight) throw new Error('Last block synced not in db');
 
   return {
-    catchingUp: fullSyncHeight === latestBlockHeight,
+    catchingUp: fullSyncHeight !== latestBlockHeight,
```
grod220, line 12: Ah good catch. It's an edge case (maybe an error somewhere else if so), but should we do:

```
catchingUp: fullSyncHeight < latestBlockHeight,
```

grod220 (approved):

packages/router/src/grpc/view-protocol-server/status.ts
Valentine1898: https://github.com/penumbra-zone/penumbra/blob/d3bbe4c920dd03981ed883d347ed579dae6af363/crates/view/src/service.rs#L326-L334
rust view service also returns `catchingUp=false` if the difference is only 1 block

Valentine1898 (approved):

packages/router/src/grpc/view-protocol-server/status.ts
grod220: ah, nice find

PR #699, opened 2024-03-06
Store `Epoch`s in IndexedDB and serve from extension

### LINKED ISSUE #601: Store all `ValidatorInfo`s in IDB

jessepinho:
This started as a PR to address #601 (serving validator info from the extension), but I quickly realized I'd need to store epochs in IndexedDB first. So this PR takes care of that.

## In this PR
- Create a new `EPOCHS` table in IndexedDB.
- Since we're storing epochs locally anyway, I also created an `SctService` impl in the extension. (It only has one method, `GetEpochByHeight`.)

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -236,7 +236,8 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.saveTransactionInfos(compactBlock.height, relevantTx);
       }
 
-      if (compactBlock.epochRoot) nextEpochStartHeight = compactBlock.height + 1n;
+      const isLastBlockOfEpoch = !!compactBlock.epochRoot;
+      if (isLastBlockOfEpoch) nextEpochStartHeight = compactBlock.height + 1n;
     }
   }
 
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -142,9 +142,8 @@ export class BlockProcessor implements BlockProcessorInterface {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
-      const isNewEpoch = nextEpochStartHeight === compactBlock.height;
-      if (isNewEpoch) {
-        await this.indexedDb.addEpoch(nextEpochStartHeight!);
+      if (nextEpochStartHeight === compactBlock.height) {
+        await this.indexedDb.addEpoch(nextEpochStartHeight);
         nextEpochStartHeight = undefined;
       }
 
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -133,6 +133,8 @@ export class BlockProcessor implements BlockProcessorInterface {
     const startHeight = fullSyncHeight ? fullSyncHeight + 1n : 0n;
     const latestBlockHeight = await this.querier.tendermint.latestBlockHeight();
 
+    let nextEpochStartHeight: bigint | undefined = startHeight === 0n ? 0n : undefined;
+
     // this is an indefinite stream of the (compact) chain from the network
     // intended to run continuously
     for await (const compactBlock of this.querier.compactBlock.compactBlockRange({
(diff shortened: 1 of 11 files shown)


packages/query/src/block-processor.ts
```diff
@@ -133,13 +133,19 @@ export class BlockProcessor implements BlockProcessorInterface {
     const startHeight = fullSyncHeight ? fullSyncHeight + 1n : 0n;
     const latestBlockHeight = await this.querier.tendermint.latestBlockHeight();
 
+    let nextEpochStartHeight: bigint | undefined = startHeight === 0n ? 0n : undefined;
+
```
Valentine1898, line 137: This is probably a very rare case and I won't be able to reproduce it, but if we stop sync after saving the last block of an epoch and then start sync again, we miss an epoch.

jessepinho: Damn, you're right. What do you think — maybe I should store `nextEpochStartHeight` in IndexedDB? 😩

Valentine1898: I had an idea to check the height of the epoch change by condition `height % epochDuration === 0`, but this is probably bad, `epochDuration` can change

Valentine1898: It would be convenient if the compact block had some indicator that determines the first block of the epoch, not the last one

jessepinho: > I had an idea to check the height of the epoch change by condition height % epochDuration === 0, but this is probably bad, epochDuration can change

Yeah, that won't work — an epoch can stop short if e.g., a validator is slashed.

> It would be convenient if the compact block had some indicator that determines the first block of the epoch, not the last one

YEP 😅

jessepinho: Just had a thought — I could, here at the end, insert the epoch into the database, since I'll already know it's start height. That'd work, right?

Valentine1898: > Just had a thought — I could, here at the end, insert the epoch into the database, since I'll already know it's start height. That'd work, right?

Yes

But you must update the validator info in the first epoch block, in any case the block processor must be able to detect the first epoch block

jessepinho: Got it. I'll work on this. Thanks for flagging this issue!

jessepinho: @Valentine1898 actulaly, I'll do that as part of my next PR where I store validator infos in the database, if that's OK with you (since this is SUCH an edge case). Mind approving this in the meantime?

Valentine1898 (approved):

packages/query/src/block-processor.ts
Valentine1898: Uh, yeah, sure

PR #700, opened 2024-03-07
WIP: Store validators in IndexedDB and serve StakingService from the extension

### LINKED ISSUE #601: Store all `ValidatorInfo`s in IDB

jessepinho:
Closes #601


packages/query/src/block-processor.ts
```diff
@@ -142,10 +148,6 @@ export class BlockProcessor implements BlockProcessorInterface {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
-      if (nextEpochStartHeight === compactBlock.height) {
-        await this.indexedDb.addEpoch(nextEpochStartHeight);
-        nextEpochStartHeight = undefined;
-      }
```
jessepinho, line 148: This is now moved to the end of this loop, to avoid the edge case @Valentine1898 discovered [here](https://github.com/penumbra-zone/web/pull/699#discussion_r1515328606).

packages/query/src/queriers/staking.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class StakingQuerier implements StakingQuerierInterface {
+  private readonly client: PromiseClient<typeof StakingService>;
+
+  constructor({ grpcEndpoint }: { grpcEndpoint: string }) {
+    this.client = createClient(grpcEndpoint, StakingService);
+  }
+
+  allValidatorInfos(): AsyncIterable<ValidatorInfoResponse> {
+    return this.client.validatorInfo({ showInactive: true });
```
jessepinho, line 15: Including inactive ones when saving to our local database, since we're now serving the `ValidatorInfo` RPC method from the extension, and may receive requests for inactive validators.

packages/storage/src/indexed-db/index.ts
```diff
@@ -469,6 +473,29 @@ export class IndexedDb implements IndexedDbInterface {
     return epoch;
   }
 
+  /**
+   * Inserts the validator info into the database, or updates an existing
+   * validator info if one with the same identity key exists.
+   */
+  async upsertValidatorInfo(validatorInfo: ValidatorInfo): Promise<void> {
```
grod220, line 480: Is upsert a common term? Does it convey something different than "update"?

grod220 (approved):

packages/query/src/queriers/staking.ts
jessepinho: good call, done

apps/extension/src/impls.ts
```diff
@@ -32,7 +33,6 @@ const penumbraProxies = [
   DexSimulationService,
   GovernanceService,
   ShieldedPoolService,
-  StakeService,
```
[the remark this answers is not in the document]
jessepinho: done

packages/query/src/block-processor.ts
```diff
@@ -236,8 +238,12 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.saveTransactionInfos(compactBlock.height, relevantTx);
       }
 
+      if (compactBlock.height > latestKnownBlockHeight)
+        latestKnownBlockHeight = compactBlock.height;
+
       const isLastBlockOfEpoch = !!compactBlock.epochRoot;
-      if (isLastBlockOfEpoch) nextEpochStartHeight = compactBlock.height + 1n;
+      if (isLastBlockOfEpoch)
+        void this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
```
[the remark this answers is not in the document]
jessepinho: Ah, got it — fixed

packages/query/src/block-processor.ts
```diff
@@ -236,8 +238,12 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.saveTransactionInfos(compactBlock.height, relevantTx);
       }
 
+      if (compactBlock.height > latestKnownBlockHeight)
+        latestKnownBlockHeight = compactBlock.height;
```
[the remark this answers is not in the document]
jessepinho: Yep, exactly. I'll add a comment.

packages/query/src/block-processor.ts
```diff
@@ -365,4 +371,36 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+    // only get updated validator infos once we're within the latest known
+    // epoch.
+    if (nextEpochIsLatestKnownEpoch) void this.updateValidatorInfos();
+  }
+
+  private async updateValidatorInfos(): Promise<void> {
+    for await (const validatorInfoResponse of this.querier.staking.allValidatorInfos()) {
+      if (!validatorInfoResponse.validatorInfo) continue;
+
+      // Await the upsert. This makes it possible for users of this method to
```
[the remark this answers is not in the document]
jessepinho: update/insert — i.e., if updates it if it already exists, or inserts it if it doesn't

packages/storage/src/indexed-db/index.ts
jessepinho: See my other comment

jessepinho: (And yeah, it's pretty common at least in the SQL DB world.)

apps/extension/src/impls.ts
```diff
@@ -50,6 +50,8 @@ export const rpcImpls = [
   // @ts-expect-error TODO: accept partial impl
   [CustodyService, rethrowImplErrors(CustodyService, custodyImpl)],
   [SctService, rethrowImplErrors(SctService, sctImpl)],
+  // @ts-expect-error TODO: accept partial impl
+  [StakingService, rethrowImplErrors(StakingService, stakingImpl)],
```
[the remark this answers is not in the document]
jessepinho: Looks like this breaks some tests — I'll do this in a follow-up PR right after this

PR #704, opened 2024-03-07
Fix typings for impls

jessepinho:
Per @grod220 's comment [here](https://github.com/penumbra-zone/web/pull/700#discussion_r1516027514)


grod220 (approved):
LGTM

PR #705, opened 2024-03-07
Fix slow loading on Staking page

### LINKED ISSUE #701: Results from extension service stream slowly to minifront
While working on #601, I expected that, once `StakingService` was hosted inside the extension, we'd get much faster results on the staking page since we were no longer loading validator info from the RPC node, but rather from our own IndexedDB database.

However, that was not the case:

https://github.com/penumbra-zone/web/assets/1121544/2c072997-7908-46e9-9f27-66b34130fcbc

It took about 23 seconds to load all validator infos!

I wondered if the results were streaming slowly from the database, so I added a log statement to the `ValidatorInfo` impl so I'd see how long it was taking on the extension side. But it loaded all 446 rows almost instantaneously:

https://github.com/penumbra-zone/web/assets/1121544/70b66158-9856-4e64-b581-2cc83a83d3da

I also drastically simplified the code inside the `for` loop in the [staking fetcher](https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/fetchers/staking.ts), but to no avail. I also simplified the code in the Staking slice — again, no results.

The only thing I can imagine is that the results are taking a while to get from the extension to minifront. Something about the channel transport, perhaps?

jessepinho:
I created 400 validators (via #706) and tested how long it took to render them all on the staking page. On `main`, it took about 27 seconds. With this PR's changes, it takes about 1.5 seconds.

Originally, I'd thought that there must have been an issue with getting the data from extension to the web page — i.e., that the channel transports were slow or something. But with some testing, I realized our performance issues were more related to rendering. So I used the React Profiler to investigate, and came up with some optimizations.

## In this PR
- Split out a `<Delegations />` component from the `<Account />` component. That way, only the `<Delegations />` component re-renders as more validators are loaded, and `<Account />`/`<Header />` aren't forced to re-render each time.
- `memo()`ize the `<DelegationValueView />` component. This had about a half-second improvement, after other improvements were made.
- Hide `<DelegationValueView />`'s tooltips until all loading is complete.
  - Per the [React Profiler](https://react.dev/learn/react-developer-tools), the tooltips were re-rendering frequently while the validators were still loading, causing poor performance.

## Performance takeaways that we can use going forward
- **Use component-specific Zustand selectors, rather than slice-specific Zustand selectors, wherever possible.**

    Currently, all of our Zustand slices export a selector that selects the entire slice. This isn't great for performance, since it means that _every_ component that uses this selector will have to re-render when _any_ property in the slice changes, even if the given component is not using that particular property.
    
    Instead, we should use Zustand selectors that only return the properties that the given component needs. That way, the component will only re-render when the properties it cares about change.
- **Use inline selectors when a component only needs a single property from Zustand state.**

  See [this example](https://github.com/penumbra-zone/web/pull/705/files?diff=split&w=1#diff-964bf15a0e1f337ecd16caf03b5b5878828b4ad2e6c95004a85affe5ce2390a7R25-R27). Since we're only using a single property from the state, we can just inline the selector, and the component will only re-render when the output of that selector changes.
- **Use `useShallow()` when a component needs multiple properties from Zustand state.**
    If we need multiple properties from the state, simply using `useStore()` with a selector that grabs the properties we need will result in unnecessary rerenders:

    ```tsx
    const MyComponent = () => {
      const state = useStore(state => ({
        prop1: state.mySlice.prop1,
        prop2: state.mySlice.prop2,
      }))
      // ^ this object gets defined on every call, so `state` will update when _any_ state property
      // changes, resulting in unnecessary re-renders
    }
    ```

    Instead, we should use [`useShallow()`](https://github.com/pmndrs/zustand/blob/main/docs/guides/prevent-rerenders-with-use-shallow.md) inside of `useStore()`. `useShallow()` iterates over each _top-level_ property in the object, and only triggers a re-render when one of those properties changes:

    ```tsx
    const MyComponent = () => {
      const state = useStore(useShallow(state => ({
        prop1: state.mySlice.prop1,
        prop2: state.mySlice.prop2,
      })))
      // ^ `useShallow()` iterates over this object, and only triggers a re-render if one of the top-level properties
      // changes.
    }
    ```

    For code cleanliness, we'll probably want to extract these selectors to above the component definition:

    ```tsx
    const myComponentSelector = (state: AllSlices) => ({
        prop1: state.mySlice.prop1,
        prop2: state.mySlice.prop2,
    })

    const MyComponent = () => {
      const state = useStore(useShallow(myComponentSelector))
    }
    ```

    For maintainability, these selectors should be colocated with the component rather than being exported from the slice file.

Closes #701


apps/minifront/src/components/staking/account/delegation-value-view/index.tsx
```diff
@@ -12,43 +13,46 @@ import { getValidatorInfoFromValueView } from '@penumbra-zone/getters';
...
-  valueView: ValueView;
-  votingPowerAsIntegerPercentage?: number;
-  /**
-   * A `ValueView` representing the address's balance of staking (UM) tokens.
-   * Used to show the user how many tokens they have available to delegate.
-   */
-  unstakedTokens?: ValueView;
-}) => {
-  const validatorInfo = getValidatorInfoFromValueView(valueView);
+export const DelegationValueView = memo(
```
jessepinho, line 16: This `memo()` call is the only change to this file. Everything else is just indentation changes.

apps/minifront/src/components/staking/account/delegation-value-view/staking-actions/index.tsx
```diff
@@ -30,17 +41,7 @@ export const StakingActions = ({
...
-    amount,
-    onClickActionButton,
-    delegate,
-    undelegate,
-    onClose,
-    setAmount,
-    validatorInfo: selectedValidatorInfo,
-  } = useStore(stakingSelector);
-
+  const state = useStore(useShallow(stakingActionsSelector));
```
jessepinho, line 44: There are a number of ways we could handle the syntax here. Using `state`, rather than destructuring the properties, felt cleanest. But we could also do e.g.,:

```suggestion
  const {
    action,
    amount,
    delegate,
    onClickActionButton,
    onClose,
    setAmount,
    undelegate,
    validatorInfo,
  } = useStore(useShallow(stakingActionsSelector));
```

I kind of like using `state`, though, since it makes it very clear in the rest of the component whether a value is coming from props vs. from state.

apps/minifront/src/state/staking.ts
```diff
@@ -162,26 +169,67 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+        );
 
+        set(state => {
+          state.staking.delegationsByAccount.set(addressIndex.account, sortedDelegations);
+        });
+
+        delegationsToFlush = [];
+      }
+
+      requestAnimationFrame(() => flushToState());
```
grod220, line 207: Oh interesting. I've only seen `requestAnimationFrame` in use when dealing with wanting to paint to the screen no faster than the refresh rate of the device. Regardless, this is still quite fast. Given there is `DELEGATIONS_FLUSH_TO_STATE_INTERVAL`, shouldn't that be the thing that controls the frequency of the re-render? It seems we have two levers that control throttling. What happens if you remove `requestAnimationFrame`?

apps/minifront/src/state/staking.ts
```diff
@@ -130,6 +130,13 @@ const byBalanceAndVotingPower = (valueViewA: ValueView, valueViewB: ValueView):
   return byVotingPower;
 };
 
+/**
+ * To optimize performance while streaming potentially hundreds of
+ * delegations/validators in rapid succession, we only "flush" to state at a
+ * fixed interval, so as to prevent React from having to re-render too often.
+ */
+const DELEGATIONS_FLUSH_TO_STATE_INTERVAL = 50;
```
grod220, line 138: is 50 too high? It seems to be taking a while on my machine

https://github.com/penumbra-zone/web/assets/16624263/72f28bdc-713e-42f4-ac3d-3bf78bb33498

apps/minifront/src/state/staking.ts
```diff
@@ -162,26 +169,67 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+     * frequently. Rendering happens synchronously, which means that the `for`
+     * loop below has to wait until rendering is done before moving on to the
+     * next delegation. Thus, the staking page loads super slowly if we render
+     * delegations as soon as we receive them.
+     *
+     * To resolve this performance issue, we instead queue up a number of
+     * delegations (up to `DELEGATIONS_FLUSH_TO_STATE_INTERVAL`), and then flush
+     * them to state in batches.
+     */
+    const flushToState = (ignoreInterval = false) => {
```
grod220, line 190: I wonder if there is a more generic throttle helper function that could be made here 🤔

grod220 (approved):
Very nice fixes! Definitely worthwhile giving a demo on this on Thursday. It also leads me to believe that zustand is kinda not as performant by default then say mobx (optimal rendering built in).

apps/minifront/src/state/staking.ts
```diff
@@ -162,26 +169,67 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
...
+      if (isFirstDelegation) {
+        requestAnimationFrame(() => flushToState());
+        isFirstDelegation = false;
+      }
     }
 
+    // One last flush, in case the number of remaining delegations is less than
+    // `FLUSH_INTERVAL`.
+    flushToState(true);
+    allDelegationsLoaded = true;
```
grod220, line 224: Given `allDelegationsLoaded = true;` is right before this exits, I at first was going to question whether it needs to exist at all. But given it's not set true anywhere else, does that mean we have a potential infinite loop here? flushToState() seems to run infinitely (?).

apps/minifront/src/state/staking.ts
jessepinho: ~Interesting... I was only testing this against my local devnet, which was more or less "infinitely" fast since there was no load on the server and no network lag time. So I only needed to optimize for _rendering_ performance, rather than _network_ performance, which made `50` an optimal interval.~

~But then I connected to `grpc2` and saw results more similar to yours. So I guess it's a balance between optimizing for rendering vs. network performance. I lowered the interval just now to `5` and the results seem quite a bit better, but still way slower than when connected to devnet. Perhaps a future optimization would be to cache the results between loads and only update balances.~

Wait a sec, this is weird. Loading speeds should be the same regardless of which RPC is being used, since validators are cached locally in IndexedDB. Going to keep investigating...

jessepinho: @grod220 Think I figured it out. Can you try checking out [`jessepinho/metadata-fix`](https://github.com/penumbra-zone/web/pull/719) and testing that? It will likely be slow again the first time, then much faster on subsequent loads.

It was performing slowly because it had to fetch asset metadata for each delegation token. On my branch above, it saves these metadatas to the database as it fetches them, so that subsequent loads are faster.

(If this works, please approve [the PR](https://github.com/penumbra-zone/web/pull/719) so I can merge that branch into this one.)

apps/minifront/src/state/staking.ts
jessepinho: [This line](https://github.com/penumbra-zone/web/pull/705/files#diff-7c5ec26a35f650875c894552352cecdc5e1042768e81d382bffa45e0a4dc0c1eR196) stops the loop if `allDelegationsLoaded` is true.

grod220 (approved):
bitcoin-blockchain

PR #706, opened 2024-03-08
Create scripts for creating and delegating to validators

jessepinho:
I create these to make it easier for myself to test staking-related functionality on my local devnet, by spinning up hundreds of validators and delegating to them in bulk.

Not sure these should actually exist in the web repo, or anywhere at all other than for my personal use. The delegation script specifically doesn't even work super well. But I didn't want to lose track of these. Thoughts? Should these be in core?

### DIFF
--- a/scripts/create-validators.sh
+++ b/scripts/create-validators.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+# Creates wallets and validators for testing with.
+#
+# Usage:
+# sh scripts/create-validators -n 100 -b /tmp/pcli-validators
+#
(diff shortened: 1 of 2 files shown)


grod220 (approved):
I have no clue how to interpret the shell script. Though the comments help! Perhaps scrips should be written in node or Rust 🤔 ? Regardless, I think it's fine to add in case other folks want to use. Doesn't hurt.

jessepinho:
Ugh, yeah, I hate Bash too. It just seemed like the most natural choice since almost everything the script does is a command line command.

Issue #695, opened 2024-03-06
wasm compilation 


TalDerei:
Error on a fresh installation of the repository

<img width="1728" alt="Screenshot 2024-03-06 at 8 16 50 AM" src="https://github.com/penumbra-zone/web/assets/70081547/3b6bba81-77cf-4657-8bf2-ebe1452ebb0d">

grod220:
What steps did you do to get here? This looks like it's missing the pnpm compile stage.

TalDerei:
`pnpm i && pnpm` dev at root. Assumed `pnpm compile` is implicitly executed in there.

grod220:
Can you share the full log in a pastebin or gist?

grod220:
Ah, it seems the dev task is not waiting for compile to finish. I think because in `turbo.json`, the `compile` dependency tells turborepo to look for a task named "compile" in the same workspace where the task is defined. Because our apps don't have anything defined there, it runs in parallel rather than waiting.

I think that means we need to rewrite the build steps so something like:

```diff
    "dev": {
      "dotEnv": [".env"],
      "env": ["PRAX", "IDB_VERSION", "USDC_ASSET_ID", "MINIFRONT_URL", "PENUMBRA_NODE_PD_URL"],
-    "dependsOn": ["download-keys", "compile"],
+    "dependsOn": ["^build"],
      "outputs": ["dist/**"],
      "persistent": true,
      "cache": false
    },
```

This says, "in all workspaces call pnpm dev only after it's dependencies have pnpm build called on them first". This would trigger the wasm workspaces build task, which depends on compiling first. Think this will fix the issue.

PR #707, opened 2024-03-08
turbo config

TalDerei:
references #695


grod220 (approved):

grod220:
Because some of our internal packages will be converted to external, we'll very likely need to revise this in the npm PR (given packages need to be built before consuming now). We may need to do package specific syntax for this file.

Issue #341, opened 2024-01-03
simplify third-party consumption of services


turbocrime:
# Goal: ConnectRPC/Protobuf `ServiceImpl`s should be third-party consumable

my notes on this goal. read this as a conceptual sketch - suggestions,
counterpoints, and disagreements welcome

## brief

in the interest of an open ecosystem, we should make our service implementations
directly useful to developers of complementary and alternative tools and
wallets.  this requires designing our implementations in such a way that they
can operate as a 'drop-in' component within diverse environments.

## current

our implementations are tightly coupled to our browser extension -
they directly message and control lifecycle of extension documents, directly
access extensionStorage and localStorage, and directly query indexeddb.

some functionality and types are imported from other packages, such as our types
package, greatly obfuscating the dependency graph.

## breakdown

1. isomorphic code targeting standard browser DOM first, node second
    * never call chrome.runtime
2. minimize package deps
    * depend on: buf registry and connectrpc
    * avoid depending on: our other packages, general npm packages
3. eliminate indexeddb queries, extension storage queries, and local storage queries
    * abstract all details of our storage formats and interfaces
    * ok to query pb-specified services
    * use contextValues to inject data not present in request
4. eliminate most imported types
    * buf registry should provide most types
    * convenience types should be defined locally and not exported
5. eliminate most imported code
    * provide a minimal util or convenience module per service, for functionality duplicated between methods
    * provide type guards per service, to cover specific pb-spec shortcomings

### HandlerContext and contextValues

much of this can be achieved by using HandlerContext to access contextValues,
instead of directly interacting with other parts of the extension.  this makes
awareness of the runtime the responsibility of the runtime adapter.

in routing-refactor, contextValues were used to simply inject broad storage
access, to maintain existing use. now that routing-refactor has landed, it would
be a good first step to begin abstracting storage access into context values.

it may be possible to use 'interceptors' configured by the adapter to inject
things like interactive approval.

### offscreen.html

in a chrome extension, offscreen.html is the appropriate location to run code
targeting the standard browser dom.

this may seem like a drastic choice, but we are already entering services via
the chrome runtime, which is the way into offscreen, so this would involve
minimal change. either the connection manager or the adapter would become the
new boundary. service implementations themselves would continue to operate
identically, except for offscreenClient, which would no longer be necessary -
handlers could just call wasm directly as needed.

offscreenClient would be removed, and the only offscreen lifecycle management
necessary would become launch upon first client connection, and terminate upon
last client close.

[...]

Issue #488, opened 2024-02-09
Indexed-db global refactoring


Valentine1898:
Theoretically, there is a way in which we can get rid of Indexed-db global simply by passing all data as parameters to the wasm function. 
1. `view_server.rs` 
View server (aka block scanner) uses `Indexed-db` only to save and read `advices`. 
The solution is that we should add `advices` discovery as one of the conditions for calling `flush_updates()`. Currently we only call `flush_updates()` when we detect new notes and swaps. Also in this case we will need to add advices to `ScanBlockResult` and also pass advices already stored in `indexed-db` when initializing view server. 
My assumptions:
- we won't lose in scanning speed because we won't actually call `flush_updates()` more often than we do now because in blocks where we detect advices we are also guaranteed to detect swaps and call `flush_updates()` anyway
- We will not pass more data during initialization than we do now. We can easily recognize already used `advices` from unused ones. Used `advices` can be ignored and not passed during view server initialization, and considering that swap claim will almost always automatically occur in the next block after swap, in practice advices will be unused for only one block, and will almost never be passed during view server initialization.
2. `tx.rs`
We use reading from `indexed-db` to construct transaction view and transaction perspective. 
The point is that only during the `TxV` and `TxP` generation itself we can determine which `notes` and `assets` data we need. This is the main reason why we have to read data from `indexed-db` directly and we can't prepare it on TS side and pass it as parameter, and passing all `notes` and `swaps` records as parameters seems like a very bad idea.
The solution is to decompose the `transaction_info()` wasm function into several smaller helper wasm functions, and move the TxV and TxP generation itself to the TS. 
- This does not lead to additional code duplication.  We already have 2 different functions for `TxV` and `TxP` generation for wasm and pcli. 
3. `wasm_planner.rs` 
For `swap_claim()`, there is no difficulty to pass `SwapRecord` as a parameter, having previously prepared the data on the TS side. 
For `plan()`, we can move the implementation of this function to the TS side. 
First we call `planner.notes_requests()` and get `NotesRequest` 
then, based on the received `NotesRequest`, we prepare notes data on the TS side.  
And finally call `plan_with_spendable_and_votable_notes()` with already prepared data. 

I think I could quickly test the proposed solution for View server, and if it works, then other steps to get rid of Indexed-db are definitely possible.

TalDerei:
@grod220 should we consider closing this as "not planned"?

grod220:
Yes, with the new Storage generic, think most of my qualms with this are addressed

PR #710, opened 2024-03-09
Offscreen RPC

turbocrime:
# offscreen rpc

did some exploratory work today attempting to host rpc in the offscreen document.

## why

this is intended to make progress towards #341

targeting native dom will simplify our architecture, make our rpc implementations more portable, and also force us to abstract our context interfaces, which will make our implementations less prax-specific.

simplification alone is enough of a motivator. what actually took most of my time today was pairing with @jessepinho on worker errors. the current architecture is nearly impossible to debug: build logic is split across five documents and several simultaneous wasm runtimes, through more layers of ser/de than i want to count ;_;

## current situation

welcome to hell!! welcome to hell!!

```mermaid
graph TB

Minifront

subgraph Prax



        style storage color:red,stroke:red
        subgraph storage
            pk((Password Key))
            subgraph localExt
            wallets((wallets))
            idbConstants
            grpcEndpoint
            end
        end

        page[Extension Pages]

        subgraph service_worker.ts
            init>services.initialize]
            listeners{{Listeners}}
            SessionManager

            subgraph Services
            style Services fill:green
                queriers
                subgraph WalletServices

                    subgraph ViewServer
                        fvk((Full Viewing Key))
                        esp((Encrypted Seed Phrase))
                        wview[/WASM ViewServer\]
                    end

                    idb{{IndexedDbInterface}}
                    BlockProcessor 
                end
            end

            subgraph router [ConnectRouter]
            style router fill:purple
                subgraph impls
                view{View RPC}
                custody{Custody RPC}
                proxy{Proxy RPC}

                wplan[/WASM Planner\]
                wbuild[/WASM Builder\]
                offscreenClient{{Offscreen Client<br>not context, not an impl, but here anyway :C}}
                end
            end
        end



    subgraph offscreen
        spawnWorker
        w1[\wasm/]
        w2[\wasm/]
        w3[\wasm/]
    end

    IndexedDb[(IndexedDb)]
end

pd{remote pd}


idb === IndexedDb
classDef idbClass stroke:cyan,color:cyan,fill:blue
class idb,IndexedDb idbClass
BlockProcessor === idb 
view ----->|ctx| idb 
wbuild & wplan & wview -.-> IndexedDb
linkStyle 0,1,2,3,4,5 stroke:cyan;

view ==o|view.wallet-id| wallets
custody ==o|custody.authorize| pk 
custody ==o|custody.authorize| wallets
linkStyle 6,7,8 stroke:red,color:red

page <-.-> storage

init ===o queriers & idb & WalletServices
init -.-o grpcEndpoint & wallets & idbConstants 
linkStyle 10,11,12,13,14,15 stroke:lime,color:lime

Minifront <-.->|content scripts| listeners <--> SessionManager <-->|adapter| router
page <-.|pageClient|.-> router
linkStyle 16,17,18,19 stroke:magenta


fvk --o view & custody
esp --o custody
linkStyle 20,21,22 stroke:lime

view ==> offscreenClient & wplan & wbuild

offscreenClient <-.-> spawnWorker <==> w1 & w2 & w3  

BlockProcessor === wview
BlockProcessor ==== queriers 
fvk ==o wview & BlockProcessor


view =====>|ctx| custody -."ctx<br>popup".-> page
view ----->|ctx| queriers


queriers --> pd & pd & pd

wview --- queriers
proxy --> pd


classDef localExtStorage color:black,fill:yellow;
class wallets,idbConstants,grpcEndpoint localExtStorage
classDef sessionExtStorage color:black,fill:orange;
class pk sessionExtStorage

classDef offscreenBoundary fill:purple
class offscreen offscreenBoundary

classDef serviceInit stroke:lime
class init serviceInit

classDef wasm fill:teal,color:orange;
class wview,wbuild,wplan,w1,w2,w3 wasm

classDef transport color:magenta
class view,custody,proxy transport

view ~~~ custody
init ~~~ storage
```

tl;dr the colored lines crossing the purple are bad.

green, 'Services' key access, should just happen on init.

cyan, idb access, should be fine if we can access idb in the offscreen document. if we can't, ts can easily run queries through a chrome runtime connection, but wasm idb access may be problematic.

red, storage access, may be mostly replaceable in some cases by the green key access.

### i did not do any of that work yet

but this feels like good progress towards understanding what's necessary

### but i did achieve some reorganization

the simplified event/message flow for builds should turn out very nice - workers will launch straight from `optimisticBuild`, instead of through multiple layers of indirection

https://github.com/penumbra-zone/web/blob/040d18e6337d4ace853098d5ff966588db82392b/packages/router/src/grpc/view-protocol-server/util/build-tx.ts#L35-L46

and from the same file:

https://github.com/penumbra-zone/web/blob/040d18e6337d4ace853098d5ff966588db82392b/packages/router/src/grpc/view-protocol-server/util/build-tx.ts#L89-L114

## what are the obstacles to offscreen rpc?

1. offscreen may have limitations on things like idb access. unfortunately, documentation is limited. i'll investigate next week
2. if idb limitations exist, it may be difficult to work around them inside wasm
3. launching things like the popup might require one more step of message-passing

## what is not an obstacle to offscreen rpc?

1. offscreen cannot access extension storage. but the impls access it in very few places, and this needs to be changed anyway.
2. if idb limitations exist, we have already developed the tools to efficiently stream queries across the chrome runtime.
3. rpc messaging should be exactly identical. the same tools we use to cross the chrome runtime now will be suitable for use in offscreen.


apps/extension/src/entry/penumbra-rpc.ts
turbocrime, line 1: initial attempt at this included 'services' in the rpc offscreen. this probably isn't necessary but it was quick to try

turbocrime:
#488 would obviate wasm-idb concerns

turbocrime:
confirmed that offscreen has no problem accessing/writing to the same idb as the rest of the extension

Valentine1898:
Is this PR supposed to solve issue #488?

PR #713, opened 2024-03-11
Update deps to latest

grod220:
Comes with a number of linting updates


apps/extension/src/routes/popup/approval/origin.tsx
```diff
@@ -83,7 +83,7 @@ export const OriginApproval = () => {
                   </div>
                 </div>
                 <div className='z-30 flex h-11 w-full items-center overflow-x-scroll rounded-lg bg-background p-2 text-muted-foreground [scrollbar-color:red_red]'>
-                  <div className='mx-auto items-center p-2 text-center leading-[0.8em] [background-clip:content-box] [background-image:repeating-linear-gradient(45deg,red,black_15px)] first-line:[background-color:black]'>
```
grod220, line 86: The linter was complaining about `first-line:[background-color:black]` not being valid, however, it looks like most of these aren't used anway

apps/extension/src/shared/components/display-origin-url.tsx
```diff
@@ -5,13 +5,13 @@ export const DisplayOriginURL = ({ url: { protocol, hostname, port } }: { url: U
       <span className='tracking-[-0.21em]'>{':/'}</span>
       {'/'}
     </span>
-    <span className='tracking-[-0.05em] brightness-150 saturate-150'>{hostname}</span>
+    <span className='-tracking-wider brightness-150 saturate-150'>{hostname}</span>
```
Valentine1898, line 8: can we use `tracking-tighter` instead of `-tracking-wider`?

Valentine1898 (approved):
LGTM

apps/extension/src/shared/components/display-origin-url.tsx
grod220: ah true! Will fix.

TalDerei:
LGTM ~

apps/extension/src/routes/popup/approval/origin.tsx
turbocrime: this was actually an important style that is typically invisible but creates a visual indication if the line breaks

turbocrime: also, in this case the linter is incorrect

Issue #693, opened 2024-03-06
Add rust tests for tx.rs


Valentine1898:

grod220:
Closes as we will open more specific function issues

Issue #557, opened 2024-02-16
Change to dynamically-computed `TransactionView`s


hdevalence:
We should change the extension code to not save the `TransactionView` and instead compute it on-demand from the `Transaction` + `TransactionPerspective`.

When we add equivalent prices support #530 in we will want the views to be dynamic. If the view is not stored this is easy: retrieve a stored `TransactionPerspective` (without any `prices` set), add the prices to the in-memory `TransactionPerspective`, then use it to generate the `TransactionView` and return.

PR #714, opened 2024-03-11
557 change to dynamically computed transactionviews

Valentine1898:


packages/types/src/indexed-db.ts
```diff
@@ -122,9 +123,9 @@ export interface PenumbraDb extends DBSchema {
     key: 'params';
     value: Jsonified<FmdParameters>;
   };
-  TRANSACTION_INFO: {
-    key: Jsonified<Required<TransactionInfo>['id']['inner']>; // base64
-    value: Jsonified<TransactionInfo>;
+  TRANSACTIONS: {
+    key: string; // base64 TransactionInfo['id']['inner'];
+    value: Jsonified<TransactionInfo>; // TransactionInfo with undefined view and perspective
```
Valentine1898, line 128: I was originally thinking of using a custom type to save transactions, something like 
```typescript
export interface TransactionRecord {
  id: Jsonified<TransactionId>; 
  height: bigint; 
  transaction: Jsonified<Transaction>; 
}
``` 
But this led to not very convenient calls `toJson()` and `fromJson()` for `id` and `transaction` in many places in the router package. 
It is much more convenient for us to use proto message types to store in indexed-db, so I decided to use TransactionInfo in which `TransactionPersective` and `TtransactionView` will not be defined

packages/types/src/indexed-db.ts
```diff
@@ -122,9 +123,9 @@ export interface PenumbraDb extends DBSchema {
     key: 'params';
     value: Jsonified<FmdParameters>;
   };
-  TRANSACTION_INFO: {
-    key: Jsonified<Required<TransactionInfo>['id']['inner']>; // base64
-    value: Jsonified<TransactionInfo>;
+  TRANSACTIONS: {
```
Valentine1898, line 126: Table renamed to better display so that we no longer store the full TransactionInfo

packages/wasm/src/transaction.ts
```diff
@@ -6,7 +6,7 @@ import {
   TransactionView,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 
-export const transactionInfo = async (
+export const generateTransactionInfo = async (
```
Valentine1898, line 9: I renamed it because it caused conflicts with RPC `transactionInfo`

packages/storage/src/indexed-db/index.ts
```diff
@@ -208,27 +209,40 @@ export class IndexedDb implements IndexedDbInterface {
...
+    yield* new ReadableStream({
+      start: async cont => {
+        let cursor = await this.db.transaction('TRANSACTIONS').store.openCursor();
+        while (cursor) {
+          cont.enqueue(TransactionInfo.fromJson(cursor.value));
+          cursor = await cursor.continue();
+        }
+        cont.close();
+      },
+    });
```
grod220, line 222: Any reason for this? If not, let's change back. If so, please add a code comment.

packages/storage/src/indexed-db/index.ts
```diff
@@ -208,27 +209,40 @@ export class IndexedDb implements IndexedDbInterface {
...
+  async saveTransaction(
+    id: TransactionId,
+    height: bigint,
+    transaction: Transaction,
+  ): Promise<void> {
+    const tx = new TransactionInfo({
+      id,
+      height,
+      transaction,
+    });
```
grod220, line 234: A bit more readable on one line:

```
const tx = new TransactionInfo({ id, height, transaction });
```

grod220 (approved):

packages/wasm/src/transaction.ts
Valentine1898: We will need to split the `generateTransactionInfo()` function into two separate ones in the future 
1. Create `TransactionPersective` from `Transaction`
2. Generate `TransactionView` from `TransactionPersective` 
This will be needed to add prices to the `TransactionPersective` at an intermediate stage. see https://github.com/penumbra-zone/web/issues/557

I didn't do it in this PR because ideally we need to implement tests for wasm first https://github.com/penumbra-zone/web/issues/693

packages/storage/src/indexed-db/index.ts
Valentine1898: Yeah, I'll change it back

Issue #566, opened 2024-02-16
Change swap estimation to handle unfilled amounts


hdevalence:
A swap on Penumbra produces two output notes, one of _each_ asset, which allows the DEX engine to handle unfilled trades.

For instance, there's currently no liquidity on the UM:CUBE pair, so submitting an UM=>CUBE swap will return the original inputs. However, this isn't indicated to the user:

<img width="611" alt="Screenshot 2024-02-16 at 2 18 43 PM" src="https://github.com/penumbra-zone/web/assets/44879/b16c882a-3242-4183-b986-d0cb613c7003">

The response from the `SimulateTrade` endpoint contains a `SwapExecution`, documented [here](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1#penumbra.core.component.dex.v1.SwapExecution). It does not directly contain the unfilled input amount, though it could do this fairly easily to make things easier for its callers.

We'll need to either change the RPC to include the unfilled input amount, or compute it client side, and then show as part of the simulation that some of the input amount would be unfilled, showing two output `ValueView`s rather than one.

PR #715, opened 2024-03-11
Show unfilled in estimate

grod220:
Closes: https://github.com/penumbra-zone/web/issues/566

Here's a normal swap estimate:

Screenshot 2024-03-11 at 2 27 13 PM

With this PR, if a part of the estimate is expected to be unfilled, we'll show labels:

Screenshot 2024-03-11 at 2 26 55 PM


packages/types/src/index.ts
```diff
@@ -19,8 +19,6 @@ export * from './querier';
 export * from './amount';
 export * from './chain';
 export * from './jsonified';
-export * from './schemas';
-export * from './type-predicates';
```
grod220, line 23: Cleaning up a number of unused zod types that were before the Getter days

Valentine1898 (approved):
LGTM! 

But I think we will need to improve the UX for swaps in the future. 
Having unfilled amount leads to additional action in the transaction and thus higher transaction cost, instead it would be better to ask the user to change the amount to the maximum amount that can be filled.

hdevalence:
> Having unfilled amount leads to additional action in the transaction and thus higher transaction cost, instead it would be better to ask the user to change the amount to the maximum amount that can be filled.

Agree on needing to improve swap UX in the future, but this is not the case: every `SwapClaim` produces two output notes, there's no difference to the user whether the swap was completely filled or not.

Valentine1898:
> > Having unfilled amount leads to additional action in the transaction and thus higher transaction cost, instead it would be better to ask the user to change the amount to the maximum amount that can be filled.
> 
> Agree on needing to improve swap UX in the future, but this is not the case: every `SwapClaim` produces two output notes, there's no difference to the user whether the swap was completely filled or not.

Right, my bad

Issue #712, opened 2024-03-11
After onboarding service worker hangs


grod220:
Opening an issue for this so it's not forgotten. After this, we can deploy.

Reference: https://discord.com/channels/824484045370818580/1044457038886998137/1214934870040514570

PR #720, opened 2024-03-12
attempt to start, redirect to onboarding

turbocrime:
solves #712

### DIFF
--- a/apps/extension/public/manifest.json
+++ b/apps/extension/public/manifest.json
@@ -26,6 +26,10 @@
       "run_at": "document_start"
     }
   ],
+  "options_ui": {
+    "page": "page.html",
+    "open_in_tab": true
+  },
   "background": {
     "service_worker": "service-worker.js"
   },
--- a/apps/extension/src/routes/popup/home/index.tsx
+++ b/apps/extension/src/routes/popup/home/index.tsx
@@ -1,11 +1,10 @@
-import { redirect } from 'react-router-dom';
 import { SelectAccount } from '@penumbra-zone/ui';
-import { PopupPath } from '../paths';
 import { IndexHeader } from './index-header';
 import { useStore } from '../../../state';
 import { BlockSync } from './block-sync';
-import { localExtStorage, sessionExtStorage } from '@penumbra-zone/storage';
+import { localExtStorage } from '@penumbra-zone/storage';
 import { addrByIndexSelector } from '../../../state/wallets';
+import { needsLogin } from '../popup-needs';
 
 export interface PopupLoaderData {
   fullSyncHeight: number;
@@ -14,22 +13,8 @@ export interface PopupLoaderData {
 // Because Zustand initializes default empty (prior to persisted storage synced),
 // We need to manually check storage for accounts & password in the loader.
 // Will redirect to onboarding or password check if necessary.
-export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> => {
-  const wallets = await localExtStorage.get('wallets');
-
-  if (!wallets.length) {
-    await chrome.tabs.create({ url: chrome.runtime.getURL(`page.html`) });
-    window.close();
-  }
-
-  const password = await sessionExtStorage.get('passwordKey');
-
-  if (!password) return redirect(PopupPath.LOGIN);
-
-  const fullSyncHeight = await localExtStorage.get('fullSyncHeight');
-
-  return { fullSyncHeight };
-};
+export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> =>
+  (await needsLogin()) ?? { fullSyncHeight: await localExtStorage.get('fullSyncHeight') };
 
 export const PopupIndex = () => {
   const getAddrByIndex = useStore(addrByIndexSelector);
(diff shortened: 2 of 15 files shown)


apps/extension/public/manifest.json
```diff
@@ -26,6 +26,10 @@
       "run_at": "document_start"
     }
   ],
+  "options_ui": {
+    "page": "page.html",
+    "open_in_tab": true
+  },
```
grod220, line 32: what does this do?

apps/extension/src/listeners.ts
```diff
@@ -2,6 +2,8 @@ import { approveOrigin, originAlreadyApproved } from './approve-origin';
 import { Prax, PraxResponder } from './message/prax';
 import { JsonValue } from '@bufbuild/protobuf';
 
+console.log('listeners');
```
grod220, line 5: can remove

grod220 (commented):
I seem to be getting some strange loading artifacts (sometimes no UI, sometimes just loading bar).

Also with the exponential backoff, I suppose we are hoping it initializes after the wallet gets added with not too much wait? What about onboarding sending an initialization message (think we had that before?). 

https://github.com/penumbra-zone/web/assets/16624263/136b861c-044e-430b-9be2-e0d9730515fd

packages/router/src/grpc/view-protocol-server/app-parameters.ts
turbocrime, line 1: it seems like one of the root causes was that parameters were not yet present in idb - imo either minifront should handle that condition, or services init should ensure they're present, but this was a simpler fix

turbocrime:
> I seem to be getting some strange loading artifacts (sometimes no UI, sometimes just loading bar).

i wasn't able to achieve some of the UI states you showed in this video, but i did make the first load reliable.
 
> Also with the exponential backoff, I suppose we are hoping it initializes after the wallet gets added with not too much wait? What about onboarding sending an initialization message (think we had that before?).

now using both.

packages/services/src/index.ts
```diff
@@ -61,7 +83,7 @@ export class Services implements ServicesInterface {
   }
 
   private async initializeWalletServices(): Promise<WalletServices> {
-    const { walletId, fullViewingKey } = await this.config.getWallet();
+    const { walletId, fullViewingKey, idbVersion: dbVersion } = await this.config;
     const params = await this.querier.app.appParams();
     if (!params.sctParams?.epochDuration) throw new Error('Epoch duration unknown');
```
turbocrime, line 88: pairing with @jessepinho i think it's not appropriate to get epoch duration here but that should be a separate issue

packages/router/src/grpc/view-protocol-server/app-parameters.ts
```diff
@@ -1,10 +1,16 @@
...
   const services = ctx.values.get(servicesCtx);
   const { indexedDb } = await services.getWalletServices();
+
+  const subscription = indexedDb.subscribe('APP_PARAMETERS');
   const parameters = await indexedDb.getAppParams();
-  if (!parameters) throw new Error('App parameters not available');
-  return { parameters };
+  if (parameters) return { parameters };
+  for await (const update of subscription)
+    return { parameters: AppParameters.fromJson(update.value) };
```
jessepinho, line 13: Ah, nice — does this prevent the error that often shows up after clearing the cache that says something about not being able to find AppParameters? If so, this is a great solution!

apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -14,22 +13,8 @@ export interface PopupLoaderData {
 // Because Zustand initializes default empty (prior to persisted storage synced),
 // We need to manually check storage for accounts & password in the loader.
 // Will redirect to onboarding or password check if necessary.
-export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> => {
-  const wallets = await localExtStorage.get('wallets');
-
-  if (!wallets.length) {
-    await chrome.tabs.create({ url: chrome.runtime.getURL(`page.html`) });
-    window.close();
-  }
```
jessepinho, line 23: Are we OK with not checking for wallets anymore? Don't we want to check `await needsLogin()` _and_ `await needsOnboard()`?

jessepinho (approved):
LGTM! One small Q but feel free to merge if I'm missing something

grod220 (approved):

apps/extension/src/routes/popup/home/index.tsx
turbocrime: the login page cascades to needsOnboard

apps/extension/src/routes/popup/popup-needs.ts
```diff
@@ -0,0 +1,18 @@
...
+  return redirect(PopupPath.LOGIN);
+};
+
+export const needsOnboard = async () => {
+  const wallets = await localExtStorage.get('wallets');
+  if (wallets.length) return;
+
+  void chrome.runtime.openOptionsPage();
+  window.close();
+};
```
grod220, line 18: These need to return null right? Otherwise there will be that extension update error we saw while pairing.

PR #721, opened 2024-03-12
Prevent duplicate renders of delegations by aborting earlier requests when new ones start

jessepinho:
This fixes an issue where, if you switch quickly back and forth between account indexes on the staking page, multiple copies of the same delegations are sometimes rendered since the previous request is still loading.

### DIFF
--- a/apps/minifront/src/state/staking/index.ts
+++ b/apps/minifront/src/state/staking/index.ts
@@ -70,6 +70,7 @@ export interface StakingSlice {
    * `delegationsByAccount`. Should be called each time `account` is changed.
    */
   loadDelegationsForCurrentAccount: () => Promise<void>;
+  loadDelegationsForCurrentAccountAbortController?: AbortController;
   /**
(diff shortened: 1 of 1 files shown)


apps/minifront/src/state/staking/index.ts
```diff
@@ -70,6 +70,7 @@ export interface StakingSlice {
    * `delegationsByAccount`. Should be called each time `account` is changed.
    */
   loadDelegationsForCurrentAccount: () => Promise<void>;
+  loadDelegationsForCurrentAccountAbortController?: AbortController;
```
grod220, line 73: Good lord that's a long var name

grod220 (approved):

apps/minifront/src/state/staking/index.ts
jessepinho: Haha, yeah. I was thinking of calling it just `abortController`, but I wanted it to be clear what it's being used for.

PR #725, opened 2024-03-12
Extract common transaction tasks to a planBuildBroadcast helper

### LINKED ISSUE #717: DRY up Zustand action code for building + submitting transactions
We have a bunch of different places in the webapp's Zustand code where we build and submit transactions. The code in all these places is nearly identical, with the only parts that differ being the transaction type:

```ts
// apps/minifront/src/state/send.ts

export const createSendSlice = ()  => (set, get) => {
  return {
  // ...
    sendTx: async () => {
      set(state => {
        state.send.txInProgress = true;
      });

      const toast = new TransactionToast('send');
      toast.onStart();

      try {
        const transactionPlan = await plan(assembleRequest(get().send));
        const transaction = await authWitnessBuild({ transactionPlan }, status =>
          toast.onBuildStatus(status),
        );
        const txHash = await getTxHash(transaction);
        toast.txHash(txHash);
        const { detectionHeight } = await broadcast({ transaction, awaitDetection: true }, status =>
          toast.onBroadcastStatus(status),
        );
        toast.onSuccess(detectionHeight);

        // Reset form
        set(state => {
          state.send.amount = '';
        });
      } catch (e) {
        if (userDeniedTransaction(e)) {
          toast.onDenied();
        } else {
          toast.onFailure(e);
        }
      } finally {
        set(state => {
          state.send.txInProgress = false;
        });
      }
    },
  }
}
```

This code should be extracted to a well-tested helper function, with parameters for the parts that differ.

jessepinho:
`planBuildBroadcast` is back! 😆  

We've refactored our Zustand actions a number of times and we're back to the point where there's a ton of shared code between them. I extracted the bits that are general to pretty much any transaction: managing toasts, planning then building then broadcasting, etc. The rest gets passed in.

I manually tested:
- [x] A send transaction
- [x] A swap + swap claim transaction
- [x] A delegate transaction
- [x] An undelegate transaction
- [x] An undelegate claim transaction

And they all worked :)

Closes #717

### DIFF
--- a/apps/minifront/src/state/helpers.ts
+++ b/apps/minifront/src/state/helpers.ts
@@ -8,7 +8,6 @@ import {
   WitnessAndBuildResponse,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { viewClient } from '../clients';
-import { TransactionClassification, uint8ArrayToHex } from '@penumbra-zone/types';
 import { sha256Hash } from '@penumbra-zone/crypto-web';
 import {
   Transaction,
@@ -17,6 +16,8 @@ import {
 import { TransactionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/txhash/v1/txhash_pb';
 import { PartialMessage } from '@bufbuild/protobuf';
 import { TransactionToast } from '@penumbra-zone/ui';
+import { TransactionClassification } from '@penumbra-zone/types/src/transaction';
+import { uint8ArrayToHex } from '@penumbra-zone/types/src/hex';
 
 /**
  * Handles the common use case of planning, building, and broadcasting a
--- a/apps/minifront/src/state/helpers.ts
+++ b/apps/minifront/src/state/helpers.ts
@@ -8,6 +8,7 @@ import {
   WitnessAndBuildResponse,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { viewClient } from '../clients';
+import { TransactionClassification, uint8ArrayToHex } from '@penumbra-zone/types';
 import { sha256Hash } from '@penumbra-zone/crypto-web';
 import {
   Transaction,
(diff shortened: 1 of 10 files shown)


apps/minifront/src/state/helpers.ts
```diff
@@ -26,49 +79,32 @@ export const plan = async (
   return plan;
 };
 
-export const authWitnessBuild = async (
```
jessepinho, line 29: Combined `authWitnessBuild` and `witnessBuild`, since they were identical other than the `viewClient` method they called.

packages/ui/components/ui/tx/view/action-details.tsx
```diff
@@ -17,7 +17,7 @@ export const ActionDetails = ({ children }: { children: ReactNode }) => {
 };
 
 const Separator = () => (
-  <div className='mx-2 h-[1px] min-w-8 grow border-b-[1px] border-dotted border-light-brown' />
+  <div className='mx-2 h-px min-w-8 grow border-b-DEFAULT border-dotted border-light-brown' />
```
jessepinho, line 20: Linter complaints 🤷🏻‍♂️

apps/minifront/src/state/helpers.ts
```diff
@@ -26,49 +79,32 @@ export const plan = async (
...
-  req: PartialMessage<WitnessAndBuildRequest>,
+const build = async (
+  req: PartialMessage<AuthorizeAndBuildRequest> | PartialMessage<WitnessAndBuildRequest>,
+  skipAuth: boolean,
   onStatusUpdate: (
     status?: (AuthorizeAndBuildResponse | WitnessAndBuildResponse)['status'],
   ) => void,
 ) => {
-  for await (const { status } of viewClient.witnessAndBuild(req)) {
+  const buildFn = skipAuth ? 'witnessAndBuild' : 'authorizeAndBuild';
```
turbocrime, line 89: once again, we could use the classification rather than a boolean. also, would prefer to select the actual function instead of an index function.

maybe buildFn should be a parameter, since we have access to the classification in the caller. or the caller could actually use the function and pass the promised response as a parameter

turbocrime (approved):
had some suggestions but this is def an improvement

apps/minifront/src/state/helpers.ts
```diff
@@ -17,6 +17,59 @@ import {
...
+  req: PartialMessage<TransactionPlannerRequest>,
+  options?: {
+    /**
+     * If set to `true`, the `ViewService#witnessAndBuild` method will be used,
+     * which does not prompt the user to authorize the transaction. If `false`,
+     * the `ViewService#authorizeAndBuild` method will be used, which _does_
+     * prompt the user to authorize the transaction. (This is required in the
+     * case of most transactions.) Default: `false`
+     */
+    skipAuth?: boolean;
```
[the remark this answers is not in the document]
jessepinho: Well, the `transactionClassification` is mostly for the sake of toasts. There could be a scenario in the future where we have a mix of action types in the actions array, and some _would_ need auth.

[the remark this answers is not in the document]
jessepinho: (That said, if you feel strongly about this, I can change it in a follow-up PR.)

apps/minifront/src/state/helpers.ts
jessepinho: JK, I can just use the function's `.name` property. Will do that.

turbocrime:
ahhh that's how you got it past me

PR #726, opened 2024-03-12
restore overflow attention style

turbocrime:
looks like this was accidentally removed - not sure why there were so many changes in that deps upgrade PR

this style is useful because it visually indicates a line-break in the URL display field, which is necessary in order for the user to verify the origin

with the url `https://thisurlistryingtotrickyou-oooooooooooooooooooooooooooooooooooooooooooooooooooo-app.testnet.penumbra.zone-oooooooooooooooooooooooooooooooooooooooooooooooooooo-bymakingtheboxdisplaystrangely.verylongurl.ml:1234/`

without the style:

Screenshot 2024-03-12 at 15 44 03

with the style:

Screenshot 2024-03-12 at 15 44 32

edit: and a 'legit' dialog for comparison

Screenshot 2024-03-12 at 17 34 27

### DIFF
--- a/apps/extension/src/routes/popup/approval/origin.tsx
+++ b/apps/extension/src/routes/popup/approval/origin.tsx
@@ -82,8 +82,8 @@ export const OriginApproval = () => {
                     )}
                   </div>
                 </div>
-                <div className='z-30 flex h-11 w-full items-center overflow-x-scroll rounded-lg bg-background p-2 text-muted-foreground [scrollbar-color:red_red]'>
-                  <div className='mx-auto items-center p-2 text-center leading-[0.8em] [background-clip:content-box] [background-image:repeating-linear-gradient(45deg,red,black_15px)] first-line:[background-color:black]'>
+                <div className='z-30 flex min-h-11 w-full items-center overflow-x-scroll rounded-lg bg-background p-2 text-muted-foreground [scrollbar-color:red_red]'>
+                  <div className='mx-auto items-center p-2 text-center leading-[0.8em]'>
                     <DisplayOriginURL url={new URL(requestOrigin)} />
                   </div>
                 </div>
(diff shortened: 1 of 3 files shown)


turbocrime:
not sure why tailwind plugin didn't version bump in the fix pr

jessepinho (commented):
I'm kinda confused... both of those screenshots look broken, right?

turbocrime:
> I'm kinda confused... both of those screenshots look broken, right?

yes, but one is *obviously* broken, while the other is more subtle. we can't expect users to be very familiar with the approval dialog because they will see it very infrequently - some users may see a legitimate dialog only once, or not at all

Screenshot 2024-03-12 at 17 34 27

hdevalence:
What if we put a length limit on host names?

jessepinho (approved):
ah I get it now. it'd be great to have a way to show the whole thing, e.g., via a tooltip. but lgtm if this is restoring what it used to look like

hdevalence:
It seems like banning host names longer than 50 characters would prevent the problem. Its always possible to relax later and is relatively easy to reason about

hdevalence:
I think this is a good thing to be worried about but we could potentially avoid the problem entirely if we made a simpler choice here.

grod220:
If truncating is a security concern, shouldn't we just expand the box to always show the url no matter the length?

turbocrime:
> It seems like banning host names longer than 50 characters would prevent the problem. Its always possible to relax later and is relatively easy to reason about

i think it's very likely that node-hosted minifronts will have >50char origin names

apps/extension/src/routes/popup/approval/origin.tsx
```diff
@@ -82,8 +82,8 @@ export const OriginApproval = () => {
                     )}
                   </div>
                 </div>
-                <div className='z-30 flex h-11 w-full items-center overflow-x-scroll rounded-lg bg-background p-2 text-muted-foreground [scrollbar-color:red_red]'>
-                  <div className='mx-auto items-center p-2 text-center leading-[0.8em] [background-clip:content-box] [background-image:repeating-linear-gradient(45deg,red,black_15px)] first-line:[background-color:black]'>
+                <div className='z-30 flex min-h-11 w-full items-center overflow-x-scroll rounded-lg bg-background p-2 text-muted-foreground [scrollbar-color:red_red]'>
+                  <div className='mx-auto items-center p-2 text-center leading-[0.8em]'>
```
turbocrime, line 86: style removed, box just expands now. this could potentially conceal lower parts of the UI if the url is very long

PR #740, opened 2024-03-13
Fix how we detect that the user has denied a transaction

jessepinho:
@turbocrime it looks like the error we get when the user denies a transaction is now just a string, rather than an instance of `ConnectError`. Is that intentional? (If not, we probably shouldn't merge this PR.)


jessepinho:
Superseded by #742

PR #742, opened 2024-03-13
Fix how we handle various errors from the extension

### LINKED ISSUE #349: Web app logged in detection
If I am not logged into the extension and I attempt to send a transaction, it will issue an error:
Screenshot 2024-01-04 at 4 56 14 PM

This is an expected behavior as issuing a request should require a logged in state. However, the frontend should know if the user is logged in and, if not, show a message in the button `login to extension`.

Will require adding another `Std` method to query login state.

### LINKED ISSUE #517: Show more appropriate toast when a user denies a transaction authorization
Right now, when you start a transaction but then hit "Deny" in the approval popup, you get this:

Image

Not only is this not a particularly human-readable error, it also shouldn't be represented as an error at all.

A better option would be to show a neutral "info" toast that auto-closes after a few seconds, like this:

Image

### LINKED ISSUE #519: Show more appropriate toast when a user closes the approval popup without approving or denying
Similar to #517, this ticket is to improve the user experience after a user closes the approval popup without approving or denying the transaction.

Right now, we see this error when closing the popup:

Image

A better option would be to show a neutral "info" toast that auto-closes after a few seconds, like this:

Image

jessepinho:
https://github.com/penumbra-zone/web/assets/1121544/de6001fa-0d69-43bc-9619-f3d3566efe12

There were a number of issues with how we were handling errors coming from the extension. This PR fixes the following:

1. Show a nicer message when the user is not logged into the extension (closes #349).
2. Show a nicer message when the user closes the approval popup without approving or denying (closes #519).
3. Show a nicer message when the user denies authorization (re-fixes #517, which recently had a regression)

### DIFF
--- a/apps/extension/src/state/tx-approval.ts
+++ b/apps/extension/src/state/tx-approval.ts
@@ -102,6 +102,7 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set, 
       choice,
       transactionView: transactionViewString,
       authorizeRequest: authorizeRequestString,
+      handleCloseWindow,
     } = get().txApproval;
 
     if (!responder) throw new Error('No responder');
@@ -143,6 +144,8 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set, 
         state.txApproval.asPublic = undefined;
         state.txApproval.transactionClassification = undefined;
       });
+
+      if (window.onbeforeunload === handleCloseWindow) window.onbeforeunload = null;
     }
   },
 });
--- a/apps/extension/src/state/tx-approval.ts
+++ b/apps/extension/src/state/tx-approval.ts
@@ -37,6 +37,7 @@ export interface TxApprovalSlice {
     req: InternalRequest<TxApproval>,
     responder: (m: InternalResponse<TxApproval>) => void,
   ) => Promise<void>;
+  handleCloseWindow: () => void;
 
   setChoice: (choice: UserChoice) => void;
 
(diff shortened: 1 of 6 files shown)


jessepinho:
@grod220 you mentioned in #349 that login detection would require adding an `Std` method. Not sure what that means — did you mean an RPC method? If so, I went with a different approach here, simply throwing a `Code.Unauthenticated` error _from the extension_ if the user is not logged in, rather than proactively querying _from the webapp_. This seems better, since the right time to check if the user is logged in is at the moment the tx approval request is being sent to the extension. If we query earlier from the web app, it's possible that the user would log out from the extension between when we made the query and when we send the tx approval request.

grod220 (approved):
Nice! looks good.

> @grod220 you mentioned in https://github.com/penumbra-zone/web/issues/349 that login detection would require adding an Std method. Not sure what that means — did you mean an RPC method?

Think your flow makes sense to me!

apps/extension/src/state/tx-approval.ts
```diff
@@ -78,6 +79,15 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set,
...
     });
+
+    window.onbeforeunload = get().txApproval.handleCloseWindow;
+  },
+
+  handleCloseWindow: () => {
+    set(state => {
+      state.txApproval.choice = UserChoice.Ignored;
+    });
+    get().txApproval.sendResponse();
```
turbocrime, line 90: ah, this is incorrect. ignored is for use by origin approval to indicate the source of a request should be ignored, and not able to trigger any more requests.

in this context it's meaningless. without this empty sendResponse, would get a chrome.runtime error at the caller, indicating the window closed.

jessepinho: OK, I'll fix this

Issue #571, opened 2024-02-17
Show a more helpful UI when the RPC is unavailable


jessepinho:
Currently looks like this:

![Image](https://github.com/penumbra-zone/web/assets/1121544/b63920aa-4ab6-4e9c-abb8-7ac3dfcc0961)

Would be nice to show a splash page with a helpful error message.

PR #748, opened 2024-03-13
Make sure errors get surfaced when the RPC is unavailable

jessepinho:
Relates to #571


turbocrime (approved):

turbocrime:
should we just use that promise.withresolvers package everywhere?

jessepinho:
@turbocrime 
> should we just use that promise.withresolvers package everywhere?
Meh 🤷🏻‍♂️  we only need it in two packages so far, looks like

PR #752, opened 2024-03-13
Show a nicer UI when minifront can't connect to the extension, and for other errors

### LINKED ISSUE #339: Retry logic for stale extension state
Relates to: https://github.com/penumbra-zone/web/issues/51

It's a bit hard to replicate, but when a request hasn't come to the extension in a while it's common to get a `Receiving end does not exist` error. Upon refreshing the page it fixes itself. I believe the issue relates to the extension wake up time. In the previous content scripts, we used [exponential backoff](https://www.npmjs.com/package/exponential-backoff) to simply retry. This resolves the issue.

Screenshot 2024-01-03 at 9 04 48 AM

### LINKED ISSUE #571: Show a more helpful UI when the RPC is unavailable

jessepinho:
Previously, when the RPC node was unavailable, the `<ErrorBoundary />` component would catch the error, stringify it, and render it to the screen.

Now, we detect whether the given error is specific to the RPC being unavailable, and we show a message for that case.

## Before
image

## After
image

I also prettified the `<ErrorBoundary />` in both the minifront and node-status apps to show other types of errors inside a `<SplashPage />` — e.g.:

image

And lastly, I created a 404 page:
image

Closes #571 
Closes #339

### DIFF
--- a/apps/minifront/src/components/extension-unavailable.tsx
+++ b/apps/minifront/src/components/extension-unavailable.tsx
@@ -1,6 +1,9 @@
 import { SplashPage } from '@penumbra-zone/ui';
 import { HeadTag } from './metadata/head-tag';
 
+const NODE_STATUS_PAGE_URL =
+  window.location.hostname === 'localhost' ? 'http://localhost:5174' : '/';
+
 export const ExtensionUnavailable = () => {
   return (
     <>
@@ -8,7 +11,12 @@ export const ExtensionUnavailable = () => {
 
       <SplashPage title='Penumbra extension unavailble'>
         {`We can't currently connect to the Penumbra extension. This could be because the RPC node
-        that you're connected to is down. Please try reloading this page; and if that doesn't work,
+        that you're connected to is down. Please check `}
+        <a href={NODE_STATUS_PAGE_URL} className='underline'>
+          the node&apos;s status page
+        </a>
+        ;
+        {` and if that doesn't work,
         please consider using a different RPC node.`}
       </SplashPage>
     </>
--- a/apps/minifront/src/components/extension-unavailable.tsx
+++ b/apps/minifront/src/components/extension-unavailable.tsx
@@ -10,14 +10,19 @@ export const ExtensionUnavailable = () => {
       <HeadTag />
 
       <SplashPage title='Penumbra extension unavailble'>
-        {`We can't currently connect to the Penumbra extension. This could be because the RPC node
-        that you're connected to is down. Please check `}
-        <a href={NODE_STATUS_PAGE_URL} className='underline'>
-          the node&apos;s status page
-        </a>
-        ;
-        {` and if that doesn't work,
-        please consider using a different RPC node.`}
+        <p className='mb-2'>We can&apos;t currently connect to the Penumbra extension.</p>
+        <p className='mb-2'>
+          This page may have been left open for too long, causing a timeout. Please reload this page
+          and see if that fixes the issue.
+        </p>
+        <p>
+          If it doesn&apos;t, the RPC node that you&apos;re connected to could be down. Check{' '}
+          <a href={NODE_STATUS_PAGE_URL} className='underline'>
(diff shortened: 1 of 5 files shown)


apps/minifront/src/fetchers/chain-id.ts
```diff
@@ -1,8 +1,6 @@
 import { viewClient } from '../clients';
 
-export const getChainId = async (): Promise<string> => {
+export const getChainId = async (): Promise<string | undefined> => {
   const { parameters } = await viewClient.appParameters({});
-  if (!parameters?.chainId) throw new Error('No chainId in response');
-
-  return parameters.chainId;
+  return parameters?.chainId;
```
jessepinho, line 5: This is a little change @turbocrime suggested. We don't need to throw if the chain ID doesn't come back from the server.

apps/minifront/src/components/shared/error-boundary.tsx
```diff
@@ -1,17 +1,25 @@
...
-  console.error(error);
+  console.error('ErrorBoundary caught error:', error);
 
   return (
-    <div className='text-red'>
-      <h1 className='text-xl'>{String(error)}</h1>
-    </div>
+    <SplashPage title='Uh-oh!' description='Looks like there was an error.'>
+      {String(error)}
+    </SplashPage>
```
turbocrime, line 23: maybe this is inconsequential. but i (and a lot of users) find 'uh-oh' type error pages weirdly infuriating, and i would personally prefer to have a drier header phrase

apps/node-status/src/components/error-boundary.tsx
```diff
@@ -1,13 +1,14 @@
...
   const error = useRouteError();
 
-  console.error(error);
+  console.error('ErrorBoundary caught error:', error);
 
   return (
-    <div className='text-red'>
-      <h1 className='text-xl'>{String(error)}</h1>
-    </div>
+    <SplashPage title='Uh-oh!' description='Looks like there was an error.'>
```
turbocrime, line 10: same

apps/minifront/src/components/extension-unavailable.tsx
```diff
@@ -0,0 +1,29 @@
+import { SplashPage } from '@penumbra-zone/ui';
+import { HeadTag } from './metadata/head-tag';
+
+const NODE_STATUS_PAGE_URL =
+  window.location.hostname === 'localhost' ? 'http://localhost:5174' : '/';
```
turbocrime, line 5: this probably needs a .env and webpack define

turbocrime: actually, it seems to be the same as the grpc endpoint uri

turbocrime (approved):
perhaps correct node_status uri but otherwise merge away

apps/minifront/src/components/extension-unavailable.tsx
```diff
@@ -0,0 +1,29 @@
...
+        </p>
+        <p>
+          If it doesn&apos;t, the RPC node that you&apos;re connected to could be down. Check{' '}
+          <a href={NODE_STATUS_PAGE_URL} className='underline'>
+            the node&apos;s status page
+          </a>{' '}
+          and, if it is down, consider switching to a different RPC URL in the Penumbra
+          extension&apos;s settings.
+        </p>
+      </SplashPage>
```
turbocrime, line 26: these cases should be readily distinguishable by the error message

apps/minifront/src/components/extension-unavailable.tsx
jessepinho: this is tricky... we'd want to define the port number in .env, you mean, right? that would mean putting `5174` in `node-status`'s `.env` file, and then importing it from minifront? (Cause I assume we'd also want to use the `.env` value in `node-status`'s `package.json` `dev` script

PR #753, opened 2024-03-13
propagate listener errors and transport failure from streaming requests

turbocrime:


packages/transport-dom/src/create.ts
```diff
@@ -127,15 +127,19 @@ export const createChannelTransport = ({
...
-        trailer: new Headers(result.trailer),
-        message: method.O.fromJson(result.message, jsonOptions),
+        header: new Headers((await success).header),
+        trailer: new Headers((await success).trailer),
+        message: await success.then(({ message }) => {
+          const o = new method.O();
+          Any.fromJson(message, jsonOptions).unpackTo(o);
+          return o;
+        }),
       };
```
turbocrime, line 143: now using the `Any.fromJson` pattern, previously this was erroneously using the message class's method which had subtly different behavior regarding unknown fields

packages/transport-dom/src/create.ts
```diff
@@ -127,15 +127,19 @@ export const createChannelTransport = ({
...
         stream: false,
-        header: new Headers(result.header),
-        trailer: new Headers(result.trailer),
-        message: method.O.fromJson(result.message, jsonOptions),
+        header: new Headers((await success).header),
+        trailer: new Headers((await success).trailer),
+        message: await success.then(({ message }) => {
+          const o = new method.O();
+          Any.fromJson(message, jsonOptions).unpackTo(o);
+          return o;
```
jessepinho, line 141: What's this change do?

packages/transport-dom/src/create.ts
```diff
@@ -127,15 +127,19 @@ export const createChannelTransport = ({
       const message = Any.pack(new method.I(input)).toJson(jsonOptions);
       port.postMessage({ requestId, message, header });
 
-      const result = await Promise.race([response, listenerError.promise]);
+      const success = Promise.race([response, listenerError.promise]);
```
jessepinho, line 130: I feel like `success` sounds like a boolean, but 🤷🏻‍♂️

turbocrime:
oops this was reffing the wrong issue

PR #754, opened 2024-03-13
 propagate listener errors and transport failure from streaming requests

### LINKED ISSUE #746: Transport may not throw pending streaming requests
paired with @jessepinho to reliably throw unary requests pending against an errored channelTransport but streaming requests did not receive the same treatment. following on from #742 #748

turbocrime:
fixes #746

### DIFF
--- a/packages/transport-dom/src/create.ts
+++ b/packages/transport-dom/src/create.ts
@@ -127,15 +127,19 @@ export const createChannelTransport = ({
       const message = Any.pack(new method.I(input)).toJson(jsonOptions);
       port.postMessage({ requestId, message, header });
 
-      const result = await Promise.race([response, listenerError.promise]);
+      const success = Promise.race([response, listenerError.promise]);
 
       return {
         service,
         method,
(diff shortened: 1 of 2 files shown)


packages/transport-dom/src/create.ts
```diff
@@ -127,15 +127,19 @@ export const createChannelTransport = ({
...
-        header: new Headers(result.header),
-        trailer: new Headers(result.trailer),
-        message: method.O.fromJson(result.message, jsonOptions),
+        header: new Headers((await success).header),
+        trailer: new Headers((await success).trailer),
+        message: await success.then(({ message }) => {
+          const o = new method.O();
+          Any.fromJson(message, jsonOptions).unpackTo(o);
+          return o;
+        }),
```
turbocrime, line 142: now using the `Any.fromJson` pattern, previously this was erroneously using the message class's method which had subtly different behavior regarding unknown fields

jessepinho (approved):

Issue #751, opened 2024-03-13
Transaction fails to build with "Cannot use 'in' operator to search for 'error' in undefined"


turbocrime:
https://discord.com/channels/824484045370818580/915710988312002610/1217560532144492546

![image](https://github.com/penumbra-zone/web/assets/134443988/5330740e-ca7d-4d65-aa19-607b8660f7aa)

there are at least two issues here

1. approval popup is failing to stand up properly. may be a zustand or ser/de or getter issue
2. the actual failure's error is failing to propagate properly. much of that flow has changed since the last release but we can still tighten that up

grod220:
Think this is done?

TalDerei:
what's the status of this issue?

TalDerei:
closing as we haven't encountered this issue in a long time and it seems resolved

PR #755, opened 2024-03-14
cross-document error propagation

turbocrime:
should now convey serialized errors all the way to the client in every situation

no size diff

Screenshot 2024-03-13 at 17 17 59

should help with #751 but there will be another underlying cause


jessepinho (changes requested):
Hey, this seems to introduce regressions from https://github.com/penumbra-zone/web/pull/742 (which was merged earlier today). Can you please make sure those 3 cases are working first?

turbocrime:
okay i got confused by the broken HMR :[ i thought i wasn't seeing the issue, because it kinda works depending on where you edit things

wip pushed up i'll work on it tomorrow. annoyingly i think errors are propagating but they're getting detyped when they cross the dom message channel

apps/extension/src/message/popup.ts
turbocrime, line 1: popup responses may be absent

apps/extension/src/entry/offscreen-handler.ts
turbocrime, line 1: offscreen now uses ConnectError to serialize failures, and handles more failures

apps/extension/src/popup.ts
turbocrime, line 1: popup utility catches a closed window into an absent response

apps/extension/src/state/tx-approval.ts
turbocrime, line 1: slice logic no longer necessary

apps/minifront/src/state/helpers.ts
turbocrime, line 1: we don't have `ConnectError` in this scope, so still doing string detection. i think if we want to detect actual codes instead of the code annotated in the message, that should be a separate PR

packages/router/src/grpc/offscreen-client.ts
turbocrime, line 1: maybe unnecessary to change, but racing every build result with cancel was excessive and caused more throws than necessary

packages/transport-chrome/session-client.ts
turbocrime, line 1: now using ConnectError in content scripts

packages/transport-dom/src/create.ts
```diff
@@ -164,17 +174,15 @@ export const createChannelTransport = ({
...
-        const [{ value }, { done }] = [
-          (await iter.next()) as IteratorYieldResult<PartialMessage<I>>,
-          await iter.next(),
-        ];
-        if (!done)
+        const [{ value } = { value: null }, { done }] = [await iter.next(), await iter.next()];
+        if (done && typeof value === 'object' && value != null) {
+          const message = Any.pack(new method.I(value as object)).toJson(jsonOptions);
+          port.postMessage({ requestId, message, header } satisfies TransportMessage);
+        } else
```
turbocrime, line 181: this removes the need for a very specific cast by using a default value, and then using a safer cast. the expected behavior is now confirmed with tests

apps/extension/src/popup.ts
```diff
@@ -45,7 +53,9 @@ const spawnDetachedPopup = async (path: string) => {
 const throwIfAlreadyOpen = (path: string) =>
   chrome.runtime
     .getContexts({
-      documentUrls: [chrome.runtime.getURL(path)],
+      documentUrls: [
+        path.startsWith(chrome.runtime.getURL('')) ? path : chrome.runtime.getURL(path),
```
jessepinho, line 57: What's this do? / why this change?

packages/transport-chrome/session-client.ts
```diff
@@ -60,13 +61,10 @@ export class CRSessionClient {
...
-    this.clientPort.addEventListener('message', (ev: MessageEvent<unknown>) => {
-      if (isTransportEvent(ev.data)) {
-        const { requestId } = ev.data;
-        this.clientPort.postMessage({ requestId, error: 'Connection closed' });
-      }
+    this.clientPort.postMessage({
+      error: errorToJson(new ConnectError('Connection closed', Code.Unavailable), undefined),
     });
-    this.clientPort.postMessage({ error: { code: 'unavailable', message: 'Connection closed' } });
+    this.clientPort.close();
```
jessepinho, line 67: Why this change?

apps/extension/src/popup.ts
```diff
@@ -6,14 +6,22 @@ import type {
...
-  return chrome.runtime.sendMessage<InternalRequest<M>, InternalResponse<M>>(req);
+  await new Promise(resolve => setTimeout(resolve, 800));
+  return chrome.runtime.sendMessage<InternalRequest<M>, InternalResponse<M>>(req).catch(e => {
+    if (isChromeResponderDroppedError(e))
+      return {
+        type: req.type,
+        data: null,
+      };
+    else throw e;
+  });
```
jessepinho, line 24: I get what's happening here, but not how it fits into the PR as a whole

jessepinho (commented):
Hey so, again, there's some very specific niche knowledge in here about messaging that makes it impossible to really review this without a more thorough PR description or inline comments. I added a few questions inline — if you can explain those, then I can give it a final review. But _please_ do this before requesting a review in the future, because otherwise I have to do multiple rounds of reviews.

apps/extension/src/popup.ts
turbocrime: left a pr comment here that may have been concealed by gh - the previous logic was failing to match paths because it would unnecessarily append an extension origin to a path that already contained it

packages/transport-chrome/session-client.ts
turbocrime: using connecterror now in this script. the lines immediately above this are conveying a transport-level error that is equivalent to this

https://github.com/penumbra-zone/web/blob/0cc03146874e28e8fff14046e5ba04a084f1f533/packages/transport-chrome/session-client.ts#L62-L68

previously this also attached a handler that closed every subsequent request attempt. but, any existing transports are failed by the close message, and no subsequent transports can init.

apps/extension/src/popup.ts
turbocrime: this pr is addressing cross-document error propagation.

the closing of the popup document surfaces as an error, but it is not a failure, it is a meaningful response. handling the termination of a document from within the document is complex, and we shouldn't reimplement it for every kind of popup, so it's handled generally for all popups.

jessepinho (approved):
ship it!

PR #761, opened 2024-03-14
Packages release workflow [part 2]

grod220:
Follow up to: https://github.com/penumbra-zone/web/pull/760


grod220:
Nvm, seems to work

PR #763, opened 2024-03-14
Remove deprecated position state

Valentine1898:
This fixes the errors during synchronization that I discovered 
telegram-cloud-photo-size-2-5447309195226503859-y

Related to https://github.com/penumbra-zone/penumbra/pull/3883

### DIFF
--- a/packages/wasm/crate/src/view_server.rs
+++ b/packages/wasm/crate/src/view_server.rs
@@ -1,23 +1,20 @@
 use std::convert::TryInto;
 use std::{collections::BTreeMap, str::FromStr};
 
-use serde::{Deserialize, Serialize};
-use serde_wasm_bindgen::Error;
-use serde_wasm_bindgen::Serializer;
-use wasm_bindgen::prelude::wasm_bindgen;
-use wasm_bindgen::JsValue;
-
 use penumbra_asset::asset::{Id, Metadata};
 use penumbra_compact_block::{CompactBlock, StatePayload};
-use penumbra_dex::lp::position::Position;
-use penumbra_dex::lp::LpNft;
 use penumbra_keys::FullViewingKey;
 use penumbra_sct::Nullifier;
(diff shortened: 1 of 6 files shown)


packages/wasm/crate/src/view_server.rs
```diff
@@ -304,26 +301,6 @@ impl ViewServer {
...
-        position_state_value: JsValue,
-    ) -> Result<JsValue, Error> {
-        utils::set_panic_hook();
-
-        let position: Position = serde_wasm_bindgen::from_value(position_value)?;
-        let position_state = serde_wasm_bindgen::from_value(position_state_value)?;
-        let lp_nft = LpNft::new(position.id(), position_state);
-        let denom = lp_nft.denom();
-        serde_wasm_bindgen::to_value(&denom)
-    }
```
Valentine1898, line 326: This is not part of the view server actually, moved to dex.rs

packages/wasm/src/dex.ts
```diff
@@ -1,10 +1,21 @@
...
   const result = compute_position_id(position.toJson()) as unknown;
   return PositionId.fromJsonString(JSON.stringify(result));
 };
+
+export const getLpNftMetadata = (
+  positionId: PositionId,
+  positionState: PositionState,
+): Metadata => {
+  const result = get_lpnft_asset(positionId.toJson(), positionState.toJson()) as JsonValue;
+  return Metadata.fromJsonString(JSON.stringify(result));
```
grod220, line 20: Do we have to stringify? Can we not do:
```
return Metadata.fromJson(result);
```

grod220 (approved):
Nice catch! Just one comment.

packages/wasm/src/dex.ts
Valentine1898: You're right, it works without stringify

PR #764, opened 2024-03-14
556 add prices table to indexeddb

### LINKED ISSUE #556: Add prices table to IndexedDB
Replaces #526, in light of https://github.com/penumbra-zone/penumbra/issues/3838

I think the following columns would be sufficient:

- `priced_asset: AssetId`: the asset being priced
- `numeraire: AssetId`: the numeraire measuring the price
- `numeraire_per_unit: f64`: the amount of the numeraire per unit of the priced asset
- `expiry_height: u64`: the block height at which time the price is considered "stale" and no longer valid

Then a query to generate a `TransactionPerspective.Price` for a `priced_asset`, `numeraire` pair can (1) check for the pair in the table and (2) check that `expiry_height < sync_height`, and we never have to clean up stale prices.

This also lets us naturally handle different cases as in the original #526 and in #527 and in #528:

- [ ] For #526, we have a config setting (not necessarily user-exposed, just config data for now) for the preferred numeraire `N`. When we scan a block, we look for trades of asset `X` into `N`, and use the trade info to write into the prices table with an expiry height some (relatively short) number of blocks ahead, say 200 (~= 16 minutes).
- [ ] For #527, as we sync across an epoch boundary (probably we want to skip this while catching up), we can fetch the validator set using the [`ValidatorInfo`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.stake.v1#penumbra.core.component.stake.v1.QueryService.ValidatorInfo) request, compute prices of delegation tokens in terms of the staking token, and write those "chain-native" prices in with an expiry height set to the max epoch length
- [ ] For #528, (which we don't need to do until implementing LP support for web), we can have some other configurable polling interval to check on the state of the user's positions and use that as the expiry height.

Valentine1898:
partially close #556

### DIFF
--- a/packages/services/src/index.ts
+++ b/packages/services/src/index.ts
@@ -115,7 +115,7 @@ export class Services implements ServicesInterface {
       viewServer,
       querier: this.querier,
       indexedDb,
-      numeraireAssetId: numeraireAssetId,
+      numeraireAssetId,
     });
 
     return { viewServer, blockProcessor, indexedDb, querier: this.querier };


packages/constants/src/assets.ts
```diff
@@ -11,6 +11,11 @@ export const STAKING_TOKEN_METADATA = localAssets.find(
   metadata => metadata.display === STAKING_TOKEN,
 )!;
 
+export const NUMERAIRE_TOKEN = 'test_usd';
+export const NUMERAIRE_TOKEN_ID = localAssets.find(
+  metadata => metadata.display === NUMERAIRE_TOKEN,
+)!.penumbraAssetId!;
+
```
Valentine1898, line 18: In the future we can easily change the `NUMERAIRE_TOKEN` if necessary

packages/query/src/block-processor.ts
```diff
@@ -388,4 +409,46 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+
+  private async updatePrices(swapOutputs: BatchSwapOutputData[], height: bigint) {
+    for (const swapOutput of swapOutputs) {
+      const swapAsset1 = getSwapAsset1(swapOutput);
+      const swapAsset2 = getSwapAsset2(swapOutput);
+
+      let numerairePerUnit = 0;
+      let pricedAsset: AssetId | undefined = undefined;
+
+      // case for trading pair <pricedAsset,numéraire>
```
Valentine1898, line 421: We cannot know in advance the order of assets in a trading pair, so we must handle both cases

packages/query/src/block-processor.ts
```diff
@@ -388,4 +409,46 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+    }
+  }
 }
+
+const calculatePrice = (delta: Amount, unfilled: Amount, lambda: Amount): number => {
+  const filledAmount = subtractAmounts(delta, unfilled);
+  return isZero(delta) || isZero(lambda) || isZero(filledAmount)
+    ? 0
+    : divideAmounts(lambda, filledAmount).toNumber();
+};
```
Valentine1898, line 454: https://discord.com/channels/824484045370818580/938328380917559326/1217859679057870858 <- price calculation discussions here

packages/types/src/indexed-db.ts
```diff
@@ -171,6 +178,10 @@ export interface PenumbraDb extends DBSchema {
     key: string; // bech32-encoded validator identity key
     value: Jsonified<ValidatorInfo>;
   };
+  PRICES: {
+    key: string[]; // composite key [base64 EstimatedPrice['priced_asset'][inner'], base64 EstimatedPrice['numeraire']['inner']]
```
Valentine1898, line 182: Using the composite key ensures that we will not have two records for the combination `priced_asset:numeraire` and will always store only the most recent value

packages/constants/src/assets.ts
grod220: See `apps/extension/.env`. We should try to use env variables for this given mainnet/testnet will have a different numeraire. Looks like we started this already with `USDC_ASSET_ID`. We should update or replace that.

packages/query/src/block-processor.ts
```diff
@@ -232,6 +247,12 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.saveTransactions(compactBlock.height, relevantTx);
       }
 
+      // we can't use third-party price oracles for privacy reasons,
+      // so we have to get asset prices from swap results during block scans
+      // and store them locally in indexed-db.
+      if (compactBlock.swapOutputs.length)
+        await this.updatePrices(compactBlock.swapOutputs, compactBlock.height);
```
grod220, line 254: Let's add brackets here

packages/query/src/block-processor.ts
```diff
@@ -388,4 +409,46 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+          getUnfilled2Amount(swapOutput),
+          getLambda1Amount(swapOutput),
+        );
+      }
+
+      if (pricedAsset === undefined || numerairePerUnit === 0) continue;
+
+      await this.indexedDb.updatePrice(pricedAsset, NUMERAIRE_TOKEN_ID, numerairePerUnit, height);
+    }
+  }
```
grod220, line 446: I think it's worth adding some tests for this. Some ideas:

- We move updatePrices into its own file as a separate function (not on the class). This would require passing in the deps (like storage) into the function. When separated, we can write tests.

- If not, just start creating a test suite `block-processor.test.ts` passing in deps on the class constructor.

packages/storage/src/indexed-db/index.ts
```diff
@@ -538,6 +540,25 @@ export class IndexedDb implements IndexedDbInterface {
...
+  ) {
+    const estimatedPrice = new EstimatedPrice({
+      pricedAsset,
+      numeraire,
+      numerairePerUnit,
+      asOfHeight: height,
+    });
+
+    await this.u.update({
+      table: 'PRICES',
```
grod220, line 557: This is missing the `key` right? or is it automatic?

Valentine1898: This works automatically if the key is specified via `keypath`

packages/services/src/index.ts
```diff
@@ -110,6 +115,7 @@ export class Services implements ServicesInterface {
       viewServer,
       querier: this.querier,
       indexedDb,
+      numeraireAssetId: numeraireAssetId,
```
grod220, line 118: Think you can do:

```diff
- numeraireAssetId: numeraireAssetId,
+ numeraireAssetId,
```

grod220 (approved):
Nice job with those tests. Think this is fine for now, but we should have a follow up PR enabling multiple numeraires.

PR #765, opened 2024-03-14
Use denied instead of ignored when a user closes the tx approval dialog

jessepinho:
Addresses [this comment](https://github.com/penumbra-zone/web/pull/742#discussion_r1524064576) left by @turbocrime.


turbocrime:
handled in #755 since i was changing this behavior anyway

PR #768, opened 2024-03-15
packages: perspective and bech32

turbocrime:
split out a set of functions in types

1. tools for manipulating transaction views and perspectives go into 'perspective'
2. tools for manipulating bech32 strings go into 'bech32'

and configure them to build well

this kinda grew out of some other work i was doing in custody rpc


packages/bech32/src/decode.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const bech32ToAddress143 = (addr: string): { inner: Uint8Array } => {
+  const decodeAddress = bech32m.decode(addr, BECH32_ADDRESS_LENGTH);
+  return { inner: new Uint8Array(bech32m.fromWords(decodeAddress.words)) };
+};
+
+export const bech32ToDefault90 = (addr: string): Uint8Array => {
+  const decodeAddress = bech32m.decode(addr);
+  return new Uint8Array(bech32m.fromWords(decodeAddress.words));
+};
```
turbocrime, line 12: this was formerly called "bech32ToUint8Array" though its behavior was subtly different than one might expect, given that name.

packages/storage/src/indexed-db/index.ts
```diff
@@ -404,7 +402,7 @@ export class IndexedDb implements IndexedDbInterface {
         notesForVoting.push(
           new NotesForVotingResponse({
             noteRecord: note,
-            identityKey: new IdentityKey({ ik: bech32ToUint8Array(bech32IdentityKey) }),
+            identityKey: new IdentityKey({ ik: bech32ToDefault90(bech32IdentityKey) }),
```
turbocrime, line 405: is it appropriate for this deserialization to truncate at 90 chars? this is the existing behavior, but i don't think anyone noticed it's doing that.

packages/bech32/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "scripts": {
+    "build": "vite build",
+    "clean": "rm -rfv dist",
+    "lint": "eslint \"**/*.ts*\""
+  },
+  "dependencies": {
+    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.8.0-20240313211643-35db357c277d.1",
+    "@bufbuild/protobuf": "^1.8.0",
+    "bech32": "^2.0.0"
```
turbocrime, line 28: appropriately, this is now the only place we directly depend on `bech32`

packages/bech32/src/decode.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const bech32ToAddress143 = (addr: string): { inner: Uint8Array } => {
+  const decodeAddress = bech32m.decode(addr, BECH32_ADDRESS_LENGTH);
+  return { inner: new Uint8Array(bech32m.fromWords(decodeAddress.words)) };
+};
+
+export const bech32ToIdentityKey143 = (key: string): { ik: Uint8Array } => {
+  const decodeAddress = bech32m.decode(key, BECH32_ADDRESS_LENGTH);
+  return { ik: new Uint8Array(bech32m.fromWords(decodeAddress.words)) };
+};
```
turbocrime, line 12: these are now more specific. the previous 'bech32ToUint8Array' also truncated at an arbitrary length, but now this makes it hopefully more obvious. is 143 correct for the identity key length?

packages/bech32/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@penumbra-zone/bech32",
+  "version": "0.0.1",
```
grod220, line 3: I think it's fine to start it on 1.0.0

packages/bech32/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "type": "module",
+  "exports": {
+    ".": "./src/index.ts"
+  },
+  "publishConfig": {
+    "files": [
+      "./dist"
+    ],
+    "exports": {
+      ".*": {
```
grod220, line 14: is `.*` different than `.`?

packages/bech32/package.json
```diff
@@ -0,0 +1,30 @@
...
+      }
+    }
+  },
+  "scripts": {
+    "build": "vite build",
+    "clean": "rm -rfv dist",
+    "lint": "eslint \"**/*.ts*\""
+  },
+  "dependencies": {
+    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.8.0-20240313211643-35db357c277d.1",
```
grod220, line 26: This is in the global package right?

packages/bech32/src/decode.ts
grod220: Does it clarify things to have `143` in the name? What does that inform to the developer? I think we probably drop unless there's another `bech32ToIdentityKey` version that we need to distinguish this from.

packages/bech32/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineConfig({
+  build: {
+    lib: {
+      entry: ['./src/index.ts'],
+      formats: ['es'],
+    },
+    rollupOptions: {
+      external: Object.keys(pkg.dependencies),
+    },
```
grod220, line 14: What does this do?

packages/bech32/src/decode.ts
```diff
@@ -0,0 +1,12 @@
+import { BECH32_ADDRESS_LENGTH } from './penumbra-bech32';
+import { bech32m } from 'bech32';
+
+export const bech32ToAddress143 = (addr: string): { inner: Uint8Array } => {
+  const decodeAddress = bech32m.decode(addr, BECH32_ADDRESS_LENGTH);
+  return { inner: new Uint8Array(bech32m.fromWords(decodeAddress.words)) };
+};
+
+export const bech32ToIdentityKey143 = (key: string): { ik: Uint8Array } => {
```
grod220, line 9: Everywhere this is used, this is placed within a new IdentityKey type. Can we change this to return just a `IdentityKey`?

packages/perspective/translators/memo-view.test.ts
grod220, line 1: Maybe @jessepinho you can give some thoughts on the translators package you initially created? Do you think we can put them all in this new `perspective` package? Or are they supposed to be more generic?

packages/wasm/package.json
```diff
@@ -16,11 +16,11 @@
...
   },
   "dependencies": {
-    "bech32": "^2.0.0",
-    "zod": "^3.22.4",
-    "@penumbra-zone/types": "workspace:*"
+    "@penumbra-zone/types": "workspace:*",
+    "zod": "^3.22.4"
   },
   "devDependencies": {
+    "@penumbra-zone/bech32": "workspace:*",
```
grod220, line 23: bech32 is used in runtime, probably should go up to normal deps (unless we are only using the types)

packages/bech32/src/decode.ts
turbocrime: 143 characters is the expected address length. if the bech32 lib doesn't receive an arg about length, it truncates at 90, because this is the expected length of a bitcoin address. so any helper function we provide will always truncate at some specific length.

the previous 'bech32ToUint8Array' seemed named to indicate it could handle an input of arbitrary length, but it was written to truncate at 143, the length of a penumbra address.inner. but was also used to decode "identity keys", and i don't know if truncation at 143 is the appropriate behavior for that.

packages/bech32/package.json
turbocrime: yes, but we do refer to its types. if we publish this as a package we need the output package.json to contain a dep

packages/bech32/package.json
turbocrime: a typo

packages/bech32/vite.config.ts
turbocrime: it prevents bundling dependencies

packages/bech32/src/decode.ts
turbocrime: ok, i went the other direction. in the interest of minimizing dependency i removed the imported type and now just construct a type that satisfies but is not explicitly an Address or IdentityKey

packages/perspective/vite.config.ts
```diff
@@ -0,0 +1,25 @@
...
+      },
+      formats: ['es'],
+    },
+    rollupOptions: {
+      external: (source, importer) => {
+        if (!importer) return false;
+        const rel = path.relative(__dirname, source);
+        const isExternal = rel.startsWith('@') || rel.startsWith('..');
+        return isExternal;
+      },
```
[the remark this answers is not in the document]
turbocrime: this prevents bundling dependencies

packages/wasm/package.json
turbocrime: it's only used in a test, which is not runtime, it belongs in devdeps

turbocrime:
okay, minimized PR to reorganization, removed all behavior changes. so bech32 util behavior is a mess and definitely seems wrong in places but i can make that a separate PR

turbocrime:
#778 for bech32 behavior

PR #769, opened 2024-03-15
better type in rethrow wrapper

turbocrime:
instead of weirdly reaching into the cjs build of connectrpc we can just declare an equivalent type.

this change has zero behavioral effect but helps with packaging

### DIFF
--- a/packages/types/src/base64.test.ts
+++ b/packages/types/src/base64.test.ts
@@ -6,7 +6,7 @@ import {
   InnerBase64Schema,
   uint8ArrayToBase64,
 } from './base64';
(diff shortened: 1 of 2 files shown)


packages/types/src/base64.test.ts
```diff
@@ -6,7 +6,7 @@ import {
   InnerBase64Schema,
   uint8ArrayToBase64,
 } from './base64';
-import { SafeParseError } from 'zod/lib/types';
+import type { SafeParseError } from 'zod';
```
turbocrime, line 9: this weird import was also causing similar problems in packaging

grod220 (approved):

PR #772, opened 2024-03-15
Update login redirect logic

grod220:
`chrome.action.openPopup` does not appear to be a function on my machine (latest version of chrome). This prevents the connection popup from triggering if logged out.

Some ideas here on how to handle login re-directs. Works for connection approvals but for transaction approvals, seems to kill the popup redirect for some reason (frontend just gets an error). Need to debug.

### DIFF
--- a/apps/extension/src/popup.ts
+++ b/apps/extension/src/popup.ts
@@ -1,11 +1,9 @@
-import { sessionExtStorage } from '@penumbra-zone/storage';
 import { PopupMessage, PopupRequest, PopupResponse, PopupType } from './message/popup';
 import { PopupPath } from './routes/popup/paths';
 import type {
   InternalRequest,
   InternalResponse,
 } from '@penumbra-zone/types/src/internal-msg/shared';
-import { Code, ConnectError } from '@connectrpc/connect';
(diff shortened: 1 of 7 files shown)


turbocrime:
i think we should never open a login popup that isn't a native extension popup. it's impossible for a user to distinguish between a detached extension popup, and a fake popup created by any website.

grod220:
> i think we should never open a login popup that isn't a native extension popup.

A native extension popup one that is attached right? Wasn't there an issue with this that Chrome did not allow programmatically opening these? Also, we should keep in mind, _all_ crypto wallets make use of popups for connection/tx approvals. It's quite standard in the industry.

Also, does https://github.com/penumbra-zone/web/pull/770 supersede this PR? As in, solves the `chrome.action.openPopup` bug? If so, I'll close this.

turbocrime:
#770 was designed to allow the existing not-logged-in logic to activate, as we weren't actually hitting it before, in the specific case of attempting to authorize a transaction. it also did some work towards moving tx-view generation into zustand rather than the custody/authorize handle. it still has some useful parts but will not programmatically launch an extension popup.

Issue #424, opened 2024-01-29
Make tx details page for swap + swap claim nicer


grod220:

hdevalence:
We'll want to augment the protos to support this: https://github.com/penumbra-zone/penumbra/issues/3832

Proposal for rendering of swap actions:
```
[VARIANT A: output NoteViews are present]
Swap
(L) 123.45 UM --> (L) 38.23 USDC { + (L) 12.78 UM [if unfilled inputs]}
{ if claim_tx is set: Claimed in fefe...fefe (link) }

[VARIANT B: output NoteViews are not present]
Swap
(L) 123.45 UM --> (L) USDC
```
Here (L) represents a circular logo. Ideally this would be handled by a single component.

In the component code, we'd probably want to actually have two "rows", one for each direction, with each row displayed conditionally on having nonzero inputs. (At the protocol level, it's technically possible to create a swap that swaps both directions simultaneously, though no software we have would ever do this).

For the swap claim view, we can just show the output note(s).

hdevalence:
Proposed steps:

- [ ] Extend `TransactionPerspective` to include the data that is needed to fill in the fields added in https://github.com/penumbra-zone/penumbra/issues/3832
- [ ] Fill in support in Rust stack for the new fields in the `SwapView` and `TransactionPerspective`: the batch swap output data etc in the `TransactionPerspective` needs to be correctly inserted when using the perspective to generate a `TransactionView`
- [ ] Pull in updated Rust code as a wasm dep, rebuild
- [ ] Change extension typescript code to insert the new additional data into the `TransactionPerspective` it uses to generate the view
- [ ] (Optionally, but seems like a good idea) change extension code to not save the `TransactionView` and instead compute it on-demand from the `Transaction` + `TransactionPerspective`. When we add equivalent prices support we will want the views to be dynamic.

hdevalence:
Similar structure to https://github.com/penumbra-zone/penumbra/issues/3838

Not sure how we want to augment the perspective to have the output notes. Currently we have a general purpose "advice" set, but this only allows looking up notes by commitment and the original `Swap` doesn't have commitments to its eventual outputs. Transactions can have multiple swaps, so maybe we need a mapping from swap commitment => pairs of output notes?

jessepinho:
Moving to blocked for now, as there's quite a bit of work that needs to be done before this ticket can be completed.

PR #773, opened 2024-03-15
Link swaps and swap claims in the UI

jessepinho:
image

image

This PR is the first pass at connecting swap and swap claim transactions in the UI. To do that, we need `SwapView` to have a populated `claimTx` field, and we need `SwapClaimView` to have a populated `swapTx` field. Some of that work was done in https://github.com/penumbra-zone/penumbra/pull/4035; the rest is done in this PR.

## In this PR
- In WASM, use the new `nullification_transaction_ids_by_commitment` and `creation_transaction_ids_by_nullifier` fields to add swap and swap claim transaction IDs to `SwapClaimView` and `SwapView`.
- Display those transaction IDs as links in the components that render swaps and swap claims.
- Create a `<TransactionId />` component that uses `<Pill />`.

## Not in this PR
- Rendering the _outputs_ from the swap claim in the swap's view (per [this comment](https://github.com/penumbra-zone/web/issues/424#issuecomment-1947375109)). That will come in a later PR.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -84,7 +84,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
   async identifyTransactions(
     spentNullifiers: Set<Nullifier>,
-    commitmentRecords: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
+    commitmentRecordsByStateCommitment: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
     blockTx: Transaction[],
   ) {
     const relevantTx = new Map<TransactionId, Transaction>();
@@ -115,25 +115,25 @@ export class BlockProcessor implements BlockProcessorInterface {
         }
       });
 
-      for (const spentNf of spentNullifiers) {
-        if (txNullifiers.some(txNf => spentNf.equals(txNf))) {
+      for (const spentNullifier of spentNullifiers) {
+        if (txNullifiers.some(txNullifier => spentNullifier.equals(txNullifier))) {
           txId = new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
           relevantTx.set(txId, tx);
-          spentNullifiers.delete(spentNf);
+          spentNullifiers.delete(spentNullifier);
         }
       }
 
-      for (const [recCom, record] of commitmentRecords) {
-        if (txCommitments.some(txCom => recCom.equals(txCom))) {
+      for (const [stateCommitment, spendableNoteRecord] of commitmentRecordsByStateCommitment) {
+        if (txCommitments.some(txCommitment => stateCommitment.equals(txCommitment))) {
           txId ??= new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
           relevantTx.set(txId, tx);
-          if (blankTxSource.equals(record.source)) {
-            record.source = new CommitmentSource({
+          if (blankTxSource.equals(spendableNoteRecord.source)) {
+            spendableNoteRecord.source = new CommitmentSource({
               source: { case: 'transaction', value: { id: txId.inner } },
             });
-            recordsWithSources.push(record);
+            recordsWithSources.push(spendableNoteRecord);
           }
-          commitmentRecords.delete(recCom);
+          commitmentRecordsByStateCommitment.delete(stateCommitment);
         }
       }
     }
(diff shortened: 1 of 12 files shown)


packages/types/src/indexed-db.ts
```diff
@@ -199,4 +199,5 @@ export const IDB_TABLES: Tables = {
   gas_prices: 'GAS_PRICES',
   epochs: 'EPOCHS',
   validator_infos: 'VALIDATOR_INFOS',
+  transactions: 'TRANSACTIONS',
```
jessepinho, line 202: This was missing from constants, so WASM would throw when trying to access it.

packages/query/src/block-processor.ts
```diff
@@ -66,7 +66,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
   async identifyTransactions(
     spentNullifiers: Set<Nullifier>,
-    commitmentRecords: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
```
jessepinho, line 69: There are no meaningful changes in this file — I thought early on in this process that I'd need to modify the block processor to work on this ticket, so I renamed some of the variables to make them a bit clearer. Even after my changes were no longer needed, I figured I'd keep them.

packages/wasm/crate/src/tx.rs
```diff
@@ -273,7 +276,89 @@ pub async fn transaction_info_inner(
...
+                                        })
+                                    );
+
+                                    if should_break {
+                                        break;
+                                    }
+
+                                }
+
+                                None::<()>
```
jessepinho, line 338: This feels absurdly unidiomatic, but the linter complained until I returned an `Option` enum from this closure.

packages/wasm/crate/src/tx.rs
```diff
@@ -273,7 +276,89 @@ pub async fn transaction_info_inner(
                         .insert(nullifier, spendable_note_record.note.clone());
                 }
             }
+            Action::Swap(swap) => {
```
jessepinho, line 279: There may be a neater, less nest-y way of writing all this, but that is outside of my Rust knowledge 😆

packages/wasm/crate/src/tx.rs
```diff
@@ -273,7 +276,89 @@ pub async fn transaction_info_inner(
...
+                    .map(|source| {
+                        if let Some(Source::Transaction(transaction)) = source.source {
+                            let id: Result<[u8; 32], Vec<u8>> = transaction.id.try_into();
+                            if let Ok(id_inner) = id {
+                                txp.creation_transaction_ids_by_nullifier
+                                    .insert(nullifier, TransactionId(id_inner));
+                            }
+                        }
+
+                        None::<()>
```
jessepinho, line 359: Same thing here

packages/wasm/crate/src/tx.rs
grod220: What do you think of this?

```rust
Action::Swap(swap) => {
                let commitment = swap.body.payload.commitment;

                let swap_position_option = storage
                    .get_swap_by_commitment(commitment.into())
                    .await?
                    .map(|swap_record| Position::from(swap_record.position));

                if let Some(swap_position) = swap_position_option {
                    add_swap_info_to_perspective(&storage, &fvk, &mut txp, &commitment, swap_position).await?;
                }
            }
```

and then an external fn
```rust

async fn add_swap_info_to_perspective(storage: &IndexedDBStorage, fvk: &FullViewingKey, txp: &mut penumbra_transaction::TransactionPerspective, commitment: &StateCommitment, swap_position: Position) -> Result<(), WasmError> {
    let derived_nullifier_from_swap =
        Nullifier::derive(fvk.nullifier_key(), swap_position, commitment);

    let transaction_infos = storage.get_transaction_infos().await?;

    for transaction_info in transaction_infos {
        transaction_info
            .transaction
            .and_then(|transaction| transaction.body)
            .iter()
            .for_each(|body| {
                for action in body.actions.iter() {
                    let transaction_id = action.action.as_ref()
                        .and_then(|action| {
                            match action {
                                v1::action::Action::SwapClaim(swap_claim) => swap_claim.body.as_ref(),
                                _ => None,
                            }
                        })
                        .and_then(|body| body.nullifier.as_ref())
                        .filter(|&nullifier| nullifier == &derived_nullifier_from_swap.to_proto())
                        .and(transaction_info.id.as_ref())
                        .and_then(|id| TransactionId::try_from(id.clone()).ok());

                    if let Some(transaction_id) = transaction_id {
                        txp.nullification_transaction_ids_by_commitment.insert(*commitment, transaction_id);
                        break;
                    }
                }
            });
    }
    Ok(())
}

```

packages/wasm/crate/src/tx.rs
grod220: How about:

```rust
 storage
                    .get_swap_by_nullifier(&nullifier)
                    .await?
                    .and_then(|swap_record| swap_record.source)
                    .into_iter()
                    .try_for_each(|source| -> WasmResult<()> {
                        let commitment_source: CommitmentSource = source.try_into()?;
                        if let Some(id) = commitment_source.id() {
                            txp.creation_transaction_ids_by_nullifier
                                .insert(nullifier, TransactionId(id));
                        }
                        Ok(())
                    })?;
```

grod220 (approved):
Some Rust suggestions, but after that 👍

packages/wasm/crate/src/tx.rs
jessepinho: Much better — thanks!

Issue #771, opened 2024-03-15
[Bug] If you do not create a wallet, the extension page will automatically open every 2 minutes in your browser 


Valentine1898:
In the latest version of the extension, if the user has not created a wallet or imported it, the extension will periodically open in a new chrome tab 

**How to reproduce**
1. Install extenstion but do not create or import a wallet
2. About every two minutes when you try to create a new tab in chrome you will open the extension page instead of a new tab
![image](https://github.com/penumbra-zone/web/assets/25391690/e5b338bc-e1e7-4b72-9922-1b9f58bc5d69)

PR #774, opened 2024-03-15
don't attempt to onboard at services init

turbocrime:
#771


turbocrime:
hard to tell, because this is dependent on the service worker being killed after a timeout, and some other unknown condition that restarts it

jessepinho (approved):
LGTM, I think — does this mean we're no longer going to automatically open the options page? (Does Chrome open it automatically the first time the extension is installed?)

turbocrime:
> LGTM, I think — does this mean we're no longer going to automatically open the options page? (Does Chrome open it automatically the first time the extension is installed?)

it seems it is opened automatically and also the popup open would trigger it, in most cases.

PR #776, opened 2024-03-15
extension v5.0.1 release new tab overriding fix

turbocrime:
release fix for #771

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -37,19 +37,17 @@ import { approveTransaction } from './approve-transaction';
 import { rpcImpls } from './impls';
 import { backOff } from 'exponential-backoff';
 
-// prevent spamming the focus-stealing openOptionsPage
-let openOptionsOnce: undefined | Promise<void>;
 const startServices = async () => {
   const grpcEndpoint = await localExtStorage.get('grpcEndpoint');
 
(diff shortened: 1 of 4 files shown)


apps/extension/src/service-worker.ts
```diff
@@ -37,19 +37,17 @@ import { approveTransaction } from './approve-transaction';
...
 import { backOff } from 'exponential-backoff';
 
-// prevent spamming the focus-stealing openOptionsPage
-let openOptionsOnce: undefined | Promise<void>;
 const startServices = async () => {
   const grpcEndpoint = await localExtStorage.get('grpcEndpoint');
 
   const wallet0 = (await localExtStorage.get('wallets'))[0];
-  if (!wallet0) openOptionsOnce ??= chrome.runtime.openOptionsPage();
+  if (!wallet0) throw new Error('No wallet found');
```
jessepinho, line 44: I'm confused — this was already merged in https://github.com/penumbra-zone/web/pull/774/files

jessepinho:
can we hold off on this release until I've fixed the `Array.fromAsync` polyfill?

apps/extension/src/service-worker.ts
turbocrime: this is a release branch that's forking from the v5.0.0 tag so we're not making a larger change than necessary

PR #777, opened 2024-03-15
Use Array.fromAsync polyfill

jessepinho:
Previously, we'd assumed that Chrome supported Array.fromAsync, but only the last few versions of Chrome did (121+), and a lot of users are still on older versions. So this PR polyfills the _implementation_ (not just the _type_) of `Array.fromAsync` in browsers where it doesn't yet exist.


turbocrime:
add declarations like https://github.com/penumbra-zone/web/commit/b91d60921010f2d09b89bbcb446f93b76297e5b5

turbocrime (approved):

Issue #676, opened 2024-03-05
Proving keys should be handled by the package that uses them


turbocrime:
currently, proving keys are downloaded by a script in the extension's package.json. this should be handled by the package/build step of the wasm package that actually needs the keys.

turbocrime:
https://github.com/penumbra-zone/penumbra/issues/4130

PR #780, opened 2024-03-18
refactor(wasm): modify turborepo config

TalDerei:
references #676 

The responsibility of downloading proving keys is moved to the `wasm` package. I further simplified the turborepo configuration defined in `turbo.json`, while still adhering to the existing dependency chain, by utilizing the `precompile` script hook provided by npm. This hook is automatically executed before the compile / dev / build scripts, ensuring the proving keys are available before proceeding with the compilation of the WASM package. The top-level scripts in package.json will call this precompile step before continuing with their respective execution. As such, we remove the unnecessary explicit `download-key` call in favor of this `precompile` step.

Additionally, per https://github.com/penumbra-zone/web/pull/434#discussion_r1472812909, we maintain a `bin` directory in the wasm package, and copy the bin directory inside the temporary `dist` directory in the extension package.


TalDerei:
@turbocrime can you check why `lint:rust` is failing the CI? seems like `wasm-pack` doesn't seem to be available in the environment where the `@penumbra-zone/wasm` package's compile script is being executed?

.github/workflows/turbo-ci.yml
```diff
@@ -113,7 +113,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
-      - run: pnpm turbo download-keys --cache-dir=.turbo
+      - run: pnpm run compile
```
turbocrime, line 116: `--cache-dir=.turbo` generally should be specified in ci workflows, or it will fail to load/save anything

apps/extension/src/utils/download-proving-keys.ts
```diff
@@ -16,6 +16,13 @@ const main = () => {
   const downloads = Object.values(provingKeys).map(async name => {
     const file = `${name}_pk.bin`;
     const outputPath = path.join(binDir, file);
+
+    // Check if the file already exists
+    if (fs.existsSync(outputPath)) {
+      console.log(`${file} already downloaded.`);
+      return;
+    }
```
turbocrime, line 24: if we're going to do this in the script, we should have a cli flag to disable this check.

apps/extension/webpack.config.js
```diff
@@ -112,7 +112,15 @@ export default (env, argv) => {
...
-        patterns: ['public', { from: 'bin', to: 'bin' }],
+        patterns: [
+          { from: 'public', to: '.' },
+          {
+            from: 'bin',
+            to: 'bin',
+            context: path.resolve(__dirname, '../../packages/wasm'),
+            noErrorOnMissing: true,
+          },
+        ],
```
turbocrime, line 123: iirc there is a more 'conventional' way to bundle assets from a dependency in webpack. i don't have an example on hand but i will see what i can find later

packages/wasm/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
+    "precompile": "tsx ../../apps/extension/src/utils/download-proving-keys.ts",
```
turbocrime, line 8: excellent, i like how this leverages the pre/post auto-execute feature of npm. it might be more technically correct as a 'post' script.

we might look at using more of these pre/post scripts for dependency management outside of turbo, which should be aware of them and manage cache appropriately - `playwright-install` before certain tests is another good candidate for this

turbo.json
```diff
@@ -26,21 +22,21 @@
     "dev": {
       "dotEnv": [".env"],
       "env": ["PRAX", "IDB_VERSION", "USDC_ASSET_ID", "MINIFRONT_URL", "PENUMBRA_NODE_PD_URL"],
-      "dependsOn": ["download-keys", "^build"],
+      "dependsOn": ["compile", "^build"],
```
turbocrime, line 25: here, `compile` should be handled as part of the dependency on build

turbocrime: but maybe this is also another candidate for package.json pre/post script

turbocrime (changes requested):

apps/extension/src/utils/download-proving-keys.ts
turbocrime: turbo should be able to correctly understand when to cache key download, and thus handle this condition transparently, if we configure it correctly. the ci uses it successfully to restore downloaded keys.

.github/workflows/turbo-ci.yml
turbocrime: as an example, ci on this PR is currently failing

turbocrime:
Screenshot 2024-03-20 at 12 23 16

1) it's preferable to use turbo cache for keys, to avoid redownloading keys at every CI step
2) it's preferable to use turbo cache for compile output, to avoid setting up compile tools for every CI step

turbo.json
TalDerei: handling `compile` as part of the package.json pre-script would fall into an infinite recursive call loop

TalDerei:
@turbocrime is this in a mergeable state? feel free to make any changes

packages/wasm/package.json
```diff
@@ -5,6 +5,8 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
+    "download-keys": "tsx ../../apps/extension/src/utils/download-proving-keys.ts",
```
turbocrime, line 8: shouldn't be referring to a script in another package - this package should contain the script

turbocrime:
probably the wasm package should contain the key download script, and expose it to other packages via the 'bin' field.

or, in the interest of reuse, make a separate package just for the keys, so wasm and prax can both use it.

packages/wasm/package.json
TalDerei: moved script to `wasm` package

turbocrime:
https://github.com/penumbra-zone/web/pull/872

TalDerei:
closing in favor of https://github.com/penumbra-zone/web/pull/872

Issue #722, opened 2024-03-12
eliminate tsconfig package


turbocrime:
use a top-level workspace `tsconfig.json` and then extend that, instead of using turborepo's encouraged practice of maintaining a private `tsconfig` package

turbocrime:
completed by https://github.com/penumbra-zone/web/pull/1460

PR #781, opened 2024-03-18
refactor(build): tsconfig

TalDerei:
references #722 

Removes dedicated `tsconfig` package in favor of a top-level base `tsconfig.json` that's shared across all packages in the workspace, ensuring all packages in the workspace start with a consistent base TypeScript configuration. This promotes consistency and reduces duplication. Each package still maintains it's own `tsconfig.json`, but uses inheritance to extend the top-level `tsconfig.json` and overrides compiler settings as necessary for the specific package.


grod220:
This is another one, where the previous config seemed ok (and more standard to turborepo).

TalDerei:
What's the status of this and #782? Should it be made merge-ready or close? @grod220 @turbocrime

grod220:
I'm not seeing the benefit of these. Think we can close until we find a good reason later.

Issue #723, opened 2024-03-12
eliminate eslint-config-custom package


turbocrime:
use a top-level flat config, instead of using turborepo's encouraged practice of maintaining a local `eslint-config-custom` package

turbocrime:
closed by #1461

Issue #729, opened 2024-03-12
Tracking issue: NPM publishing


turbocrime:
- [x] https://github.com/penumbra-zone/web/issues/718
- [x] https://github.com/penumbra-zone/web/pull/672
- [x] https://github.com/penumbra-zone/web/issues/750
- [x] https://github.com/penumbra-zone/web/issues/724
- [x] #893
- [x] #892
- [x] https://github.com/penumbra-zone/web/issues/676
- [x] https://github.com/penumbra-zone/web/issues/684
- [x] https://github.com/penumbra-zone/web/issues/737
- [x] #894
- [x] https://github.com/penumbra-zone/web/issues/733
- [x] #741
- [x] https://github.com/penumbra-zone/web/issues/730
- [x] https://github.com/penumbra-zone/web/issues/738
- [x] https://github.com/penumbra-zone/web/issues/731
- [x] https://github.com/penumbra-zone/web/issues/728
- [x] https://github.com/penumbra-zone/web/issues/736
- [x] #680 
- [x] https://github.com/penumbra-zone/web/issues/749
- [x] https://github.com/penumbra-zone/web/issues/762

turbocrime:
https://github.com/penumbra-zone/web/issues/722 may be relevant if we don't bundle when publishing.

currently tsconfigs all extend a parent config in another package. this won't be included in pack, so tsconfig in the package will just break

TalDerei:
what's the current status of our NPM publishing workflow?

TalDerei:
closing as complete

PR #782, opened 2024-03-18
refactor(eslint): top-level eslint

TalDerei:
references #723 

Removes `eslint-config-custom` package and implements top-level flat ESLint configuration, serving as the the primary linter configuration for all packages. package specific `.eslintrc.cjs` files are instructed to use the configuration defined in the top-level, and it more explicitly specifies the package's `tsconfig.json` for TypeScript-related linting.


grod220:
Curious, do you think this solves a problem we have? The current config we have seems to be a turborepo standard thing. Moving it to the top level seems to be a net-neutral effect.

TalDerei:
I think by removing dedicated configuration packages and leveraging inheritance, it has the nice property of reducing potential duplication. Since it extends the top-level configuration, it's also clearer when there are package-specific overrides, which can make it easier to read. This applies to the `tsconfig.json` as well. 

I was sequentially going down the [NPM publishing](https://github.com/penumbra-zone/web/issues/729) tracking issue to reduce our backlog.

grod220:
I suppose my observation is that it doesn't remove the need for dedicated configs, it was using inheritance before, and there does not seem to be any de-duplication. At best, it seems like a net-neutral change. And in that case, I think it's best to stick with the default turborepo recommendation.

Issue #724, opened 2024-03-12
stop using barrel files


turbocrime:
barrel import/export files are preventing our bundlers from identifying dead code. i've complained about this before but it's rather critical now that we're trying to publish consumables

package.json 'exports' field supports wildcards, for the purpose of easily modularizing packages. tsconfig and vite both understand exports and we can take advantage of this to reduce bundle sizes.

Issue #736, opened 2024-03-13
Improve external deps import paths


grod220:
_(Given https://github.com/penumbra-zone/web/pull/672 is merged)_

Removing barrel files makes imports look like this:

```typescript
import { shorten } from '@penumbra-zone/types/src/string';
```

External apps that import our package from NPM could do this:

```typescript
import { shorten } from '@penumbra-zone/types/string';

// package.json
 "exports": {
    "./string": "./dist/string.js"
  },
```

Internal packages that would want the same thing would have to do this:

```typescript
import { shorten } from '@penumbra-zone/types/string';

// package.json
 "exports": {
    "./string": "./src/string.ts"
  },
```

We cannot do them for both. In the current best case scenario would be doing this:
```typescript
 "exports": {
    "./string": "./dist/string.js" // for external packages
    "./src/string": "./dist/string.js" // for internal packages
  },
```

However, given our effort to remove barrel imports, having two export paths for every export gets unwieldy very fast (see types package).

This issue is for doing research on how to improve this situation that is the most maintainable possible. Idea: perhaps at build time for external packages, the package.json fields can be overwritten programatically.

turbocrime:
the conventional way to achieve this is an `exports` block in `publishConfig` https://docs.npmjs.com/cli/v10/configuring-npm/package-json#publishconfig

TalDerei:
we've had a bunch of changes with package exports and publishing, is this still relevant?

PR #783, opened 2024-03-18
refactor(imports): remove barrel imports

TalDerei:
references #724

eliminates barrel imports to facilitate better tree-shaking (dead code elimination) by bundlers.


TalDerei:
https://github.com/penumbra-zone/web/issues/736 still unresolved

apps/extension/src/routes/popup/settings/settings-full-viewing-key.tsx
```diff
@@ -45,7 +48,7 @@ export const SettingsFullViewingKey = () => {
                 label={
                   <p className='font-headline font-semibold text-muted-foreground'>Password</p>
                 }
-                onChange={e => {
+                onChange={(e: { target: { value: SetStateAction<string> } }) => {
```
grod220, line 51: Don't think we need this as it's an implicit type. Or do we?

apps/minifront/src/components/send/ibc/ibc-form.tsx
grod220, line 1: This feels like it shouldn't be an added file, think it's from my other PR. Think that means you need to rebase against latest `main`. I'd suggest squashing all your commits to one first.

packages/perspective/package.json
```diff
@@ -4,16 +4,39 @@
...
+    "./action-view": "./translators/action-view.ts",
+    "./address-view": "./translators/address-view.ts",
+    "./memo-view": "./translators/memo-view.ts",
+    "./output-view": "./translators/output-view.ts",
+    "./spend-view": "./translators/spend-view.ts",
+    "./transaction-view": "./translators/transaction-view.ts",
+    "./types": "./translators/types.ts",
+    "./get-address-view": "./plan/get-address-view.ts",
+    "./index": "./plan/index.ts",
+    "./view-action-plan": "./plan/view-action-plan.ts"
```
grod220, line 19: Think we can keep the wild card no?

TalDerei: modified imports to keep the wild card

PR #784, opened 2024-03-18
ui package experiment

turbocrime:
experimenting with reorg and build of ui package.

not sure if we actually want to publish this, but it had a lot of features that made it good for figuring out how to configure vite.


apps/minifront/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "vite --port 5173",
     "clean": "rm -rfv dist",
-    "build": "tsc && vite build",
+    "build": "tsc --noEmit && vite build",
```
grod220, line 9: Why is it better as a flag than on tsconfig? Seems like a better place before.

packages/transport-dom/src/stream.test.ts
```diff
@@ -111,7 +111,10 @@ describe('Stream Transformers', () => {
   describe('MessageToJson', () => {
     test('transforms Message to JSON value', async () => {
       const jsonOptions = { typeRegistry };
-      const messageToJson = new MessageToJson(jsonOptions);
+      const messageToJson = new MessageToJson(jsonOptions) as unknown as ReadableWritablePair<
+        JsonValue,
+        SayRequest
+      >;
```
grod220, line 117: A type error caught?

packages/tsconfig/base.json
```diff
@@ -2,18 +2,21 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "display": "Default",
   "compilerOptions": {
+    "paths": {
+      "@penumbra-zone/*": ["../packages/*"]
+    },
```
grod220, line 7: What's this do?

packages/tsconfig/base.json
```diff
@@ -24,8 +27,9 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "resolveJsonModule": true,
-    "target": "ES2020",
-    "module": "ES2022"
+    "target": "esnext",
```
grod220, line 30: This feels too current. Think ES2022 covers like 90% of our users. We're not sure esnext won't release a new feature that could be too new.

packages/ui/package.json
```diff
@@ -4,7 +4,44 @@
...
+    "./styles/*": "./styles/*",
+    "./fonts/*": "./fonts/*"
+  },
+  "publishConfig": {
+    "files": [
+      "dist",
+      "styles",
+      "fonts",
+      "postcss.config.js",
+      "tailwind.config.js"
```
grod220, line 22: Doesn't `dist` folder contain these others ones?

packages/ui/src/components/identicon.tsx
```diff
@@ -60,3 +66,44 @@ const IdenticonSolid = ({ uniqueIdentifier, size = 120, className }: IdenticonPr
...
+    text: color
+      .mostReadable(c, ['white', 'black'], {
+        includeFallbackColors: true,
+        level: 'AAA',
+        size: 'small',
+      })
+      .saturate()
+      .darken(20)
+      .toHexString(),
+  };
```
grod220, line 108: Think this was better in separate files before

packages/ui/vite.config.ts
```diff
@@ -0,0 +1,54 @@
...
+import dts from 'vite-plugin-dts';
+import { externalizeDeps } from 'vite-plugin-externalize-deps';
+import tsconfigPaths from 'vite-tsconfig-paths';
+import libAssetsPlugin from '@laynezh/vite-plugin-lib-assets';
+
+// eslint-disable-next-line import/no-relative-packages
+import workspacePkg from '../../package.json';
+
+const components = Object.fromEntries(
+  globbySync([
```
grod220, line 13: What does globby do?

packages/ui/vite.config.ts
```diff
@@ -0,0 +1,54 @@
...
+    lib: {
+      entry: {
+        utils: 'src/utils/index.ts',
+        ...components,
+      },
+      formats: ['es'],
+    },
+  },
+  resolve: {
+    preserveSymlinks: true,
```
grod220, line 35: What's this do?

packages/ui/vite.config.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+  plugins: [
+    libAssetsPlugin({}),
+    tsconfigPaths(),
+    externalizeDeps({
```
grod220, line 40: What do each of these do?

Issue #775, opened 2024-03-15
not allowed to open popup programmatically


turbocrime:
<img width="674" alt="Screenshot 2024-03-15 at 11 38 36" src="https://github.com/penumbra-zone/web/assets/134443988/102da305-e5cf-4ebc-b7ba-3eb1c1e311dc">

tried a couple different methods for this. some of them work in the extension during testing and then don't work later? there are known differences between unpacked loaded extension and published extensions but some may be undocumented

grod220:
Looks like this relates to: https://github.com/penumbra-zone/web/pull/772

grod220:
Also, this seemed to always be an issue on my machine no matter the environment

PR #787, opened 2024-03-18
Support `NotLoggedIn` status on connection page

grod220:
Closes https://github.com/penumbra-zone/web/issues/775
Update to #772

We now handle the not-logged-in state and surface the error to the user on the connection page

Screenshot 2024-03-18 at 9 00 32 PM


TalDerei:
might something like this, similar to when when a user tries to initiate a transaction when the wallet is locked, in red be cleaner? 

Screenshot 2024-03-18 at 3 00 37 PM

grod220:
Now using toasts:

Screenshot 2024-03-18 at 10 54 55 PM
Screenshot 2024-03-18 at 10 54 30 PM

turbocrime:
have some suggestions about message types i will pr to this branch

packages/client/src/global.ts
```diff
@@ -1,3 +1,8 @@
+export enum PenumbraRequestFailure {
```
turbocrime, line 1: failure reasons not prax-specific

apps/extension/src/popup.ts
```diff
@@ -1,37 +1,30 @@
-import { sessionExtStorage } from '@penumbra-zone/storage';
-import { PopupMessage, PopupRequest, PopupResponse, PopupType } from './message/popup';
+import { PopupMessage, PopupRequest, PopupType } from './message/popup';
```
turbocrime, line 1: non-response handled in popup launcher

apps/extension/src/state/origin-approval.ts
```diff
@@ -61,6 +61,7 @@ export const createOriginApprovalSlice = (): SliceCreator<OriginApprovalSlice> =
         data: {
           choice,
           origin: requestOrigin,
+          date: Date.now(),
```
turbocrime, line 64: slice just returns an actual complete record now

apps/extension/src/popup.ts
```diff
@@ -61,22 +54,21 @@ const throwIfAlreadyOpen = (path: string) =>
...
 
-  const loggedIn = Boolean(await sessionExtStorage.get('passwordKey'));
-
-  if (!loggedIn) {
-    popUrl.hash = PopupPath.LOGIN;
-    void spawnExtensionPopup(popUrl.href);
-    throw new ConnectError('User must login to extension', Code.Unauthenticated);
-  }
-
   switch (pop) {
```
turbocrime, line 65: needs login composable per popup type (future popups may not need it?)

turbocrime (approved):
cc @grod220 for approval

some of this diff is a bit larger than needs be

improved popup etc

importantly: failure reasons no longer prax-specific, part of  client package

Issue #747, opened 2024-03-13
Zustand race condition


grod220:
When the popup loads, it pulls from local storage and populates state with the values:

https://github.com/penumbra-zone/web/blob/26503d77f5e54496fcb8ba12b992d9debdd20270/apps/extension/src/state/persist.ts#L30-L49

The issue is that it the store is loaded synchronously. This means the async calls are a fire-and-forget. This poses a race condition.

If loading storage isn't as fast as react rendering (20% of the time?), we get this:
<img width="383" alt="Screenshot 2024-03-13 at 10 38 21 PM" src="https://github.com/penumbra-zone/web/assets/16624263/19eecccb-ad95-44f5-bdc0-6637701815bf">

We should remove this throw here:
https://github.com/penumbra-zone/web/blob/26503d77f5e54496fcb8ba12b992d9debdd20270/apps/extension/src/state/wallets.ts#L82

And handle the undefined case gracefully.

PR #788, opened 2024-03-18
Fix zustand race condition

grod220:
Closes https://github.com/penumbra-zone/web/issues/747

Checks first if a wallet is available before rendering UI. At the moment, zustand data loading from storage is sometimes slower than react rendering.

### DIFF
--- a/packages/ui/components/ui/block-sync-status/large.tsx
+++ b/packages/ui/components/ui/block-sync-status/large.tsx
@@ -104,7 +104,7 @@ const FullySyncedState = ({ latestKnownBlockHeight, fullSyncHeight }: SyncingSta
             width='50'
             color='var(--teal)'
             wrapperClass={cn(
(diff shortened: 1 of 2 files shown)


packages/ui/components/ui/block-sync-status/large.tsx
```diff
@@ -104,7 +104,7 @@ const FullySyncedState = ({ latestKnownBlockHeight, fullSyncHeight }: SyncingSta
             width='50'
             color='var(--teal)'
             wrapperClass={cn(
-              'transition-all duration-300 absolute bottom-0 right-3',
+              'transition-all duration-300 absolute bottom-0 -left-7',
```
grod220, line 107: This guy has been annoying me as well. This way, as the number grows, it will always be spaced out.

Screenshot 2024-03-18 at 9 16 06 PM

TalDerei (approved):

PR #790, opened 2024-03-18
Animate validators as they render

jessepinho:
Well, that was easy.

https://github.com/penumbra-zone/web/assets/1121544/fb21c0b5-5edc-4192-9f99-7736c669a845

(Note that I had to install `framer-motion` inside `minifront`. This shouldn't actually add to the bundle size, though, since it's also installed in `packages/ui`, which is used by `minifront`. That said, we could probably save a bit of bundle size by not using framer-motion at all and rolling our own animations, or using the [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API).)

### DIFF
--- a/apps/minifront/src/components/staking/account/delegations.tsx
+++ b/apps/minifront/src/components/staking/account/delegations.tsx
@@ -1,3 +1,4 @@
+import { AnimatePresence, motion } from 'framer-motion';
 import { AllSlices } from '../../../state';
 import { DelegationValueView } from './delegation-value-view';
 import { ValueView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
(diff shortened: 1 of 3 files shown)


apps/minifront/src/components/staking/account/delegations.tsx
```diff
@@ -23,17 +24,24 @@ export const Delegations = () => {
...
-            delegation,
-          )}
-        />
-      ))}
+      <AnimatePresence>
+        {delegations.map(delegation => (
+          <motion.div
+            key={bech32IdentityKey(getIdentityKeyFromValueView(delegation))}
+            layout
+            className='bg-charcoal'
```
jessepinho, line 32: Gave it a background so that text doesn't overlap other text during the animations

Valentine1898 (approved):
LGTM 

I'd prefer an animated skeleton (but I'm not a designer)

grod220 (approved):
I like it!

PR #795, opened 2024-03-19
New Tab: IBC 

### LINKED ISSUE #248: Ics20Withdraw part 2
Complete remainder of IBC out work (first issue: https://github.com/penumbra-zone/web/issues/132). 

Further work is currently blocked by a new testnet release. This will provide the new `revisionHeight` serialization bug fixes.

Things left to do:
- Re-enable IBC tab
- Add form validation
- Test if it works

grod220:
Closes #248

Does not address design, just functional elements. Design to come next. Updates:

- Replaces governance for new IBC tab
- Adds logic about what assets can be ibc-withdrawn
- Add validation checks (via new bech32 functions)
- Add tx details page for ics20 withdraws


packages/constants/src/chains.ts
```diff
@@ -1,17 +1,32 @@
-// Canonical data source: https://github.com/cosmos/chain-registry/tree/master
+import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+
 export interface Chain {
   displayName: string;
   chainId: string;
   ibcChannel: string;
   iconUrl: string;
+  nativeAssets: AssetId[];
```
grod220, line 8: ibc token ids will be added here when we figure out what those assets are TBD

packages/ui/components/ui/tx/view/isc20-withdrawal.tsx
```diff
@@ -0,0 +1,51 @@
+import { Ics20Withdrawal } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/ibc/v1/ibc_pb';
+import { ViewBox } from './viewbox';
+import { ActionDetails } from './action-details';
+import { joinLoHiAmount } from '@penumbra-zone/types/src/amount';
+import { bech32Address } from '@penumbra-zone/bech32';
+
+export const Ics20WithdrawalComponent = ({ value }: { value: Ics20Withdrawal }) => {
```
grod220, line 7: Ics20Withdraw's do not have a special private/public case fyi

Screenshot 2024-03-19 at 3 52 15 PM

apps/minifront/src/state/ibc.ts
```diff
@@ -141,3 +149,39 @@ const getPlanRequest = async ({
...
+      ? false
+      : !validateAddress(state.ibc.chain, state.ibc.destinationChainAddress),
+    amountErr: !state.ibc.selection ? false : validateAmount(state.ibc.selection, state.ibc.amount),
+  };
+};
+
+const validateAddress = (chain: Chain | undefined, address: string): boolean => {
+  if (!chain || address === '') return false;
+  return bech32IsValid(address, chain.addressPrefix);
+};
```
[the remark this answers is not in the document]
hdevalence: Can we allow any address if the chain is undefined?

packages/constants/src/chains.ts
hdevalence: We should not have an explicit list of native assets per chain, we already have an asset registry and we can figure out the chain from the channel in the denom

grod220: Can you provide a few examples of what these denoms look like?

grod220: Ah, wait, is it just a normal ibc asset like:

```
ibc/14F9BC3E44B8A9C1BE1FB08980FAB87034C9905EF17CF2F5008FC085218811CC
```

If so, that's a hash of: `transfer/channel-141/uosmo`. So that means we'd need to loop over our asset registry, doing the hashing of each one until we find a match in order to see if that ibc asset belongs to the particular channel the user selected. Does that sound right to you?

packages/ui/components/ui/tx/view/isc20-withdrawal.tsx
hdevalence: Not immediately but we probably do want to have a view for the Ics20Withdrawal so that we can insert:

- token metadata
- address view (of return address)

Valentine1898:
I get this error when trying to enter amount greater than 2, while I have a note c amount 2. 
It looks like I can't spend more than one note with IBC withdrawal 
 
telegram-cloud-photo-size-2-5462977910232177415-x

packages/constants/src/chains.ts
hdevalence: Nope — what I was trying to flag is that unlike ibc-go, we don't do the ibc/HASH thing. So the denom of that IBC asset is just

transfer/channel-141/uosmo

and its asset ID is the asset ID for that denom.

The way to think about this is: in the cosmos sdk, they started with coin denoms, then later realized that there should be asset IDs, and retrofitted an asset ID mechanism _inside_ the denom string. But in Penumbra, we had to have asset IDs from the start to record values in the shielded pool, so we made the asset ID be a hash of the denom from the start. And then we had to build a system for getting metadata based on asset ID as the base case.

The upshot is that handling IBC assets _inside_ Penumbra is way simpler, it's just like metadata of any other asset, IBC or not, and we can inspect the path directly in the denom.

apps/minifront/src/state/ibc.ts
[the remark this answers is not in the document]
grod220: This particular check where `chain` can be undefined relates to zustand booting up empty and then populating with chain data from our hard-coded registry. We can't transfer without a channel-id being available first.

grod220:
@Valentine1898 let's pair on that error you're seeing

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -33,8 +33,8 @@ export const TransactionApproval = () => {
 
   return (
     <div className='flex h-screen flex-col justify-between'>
-      <div className='grow overflow-auto p-[30px] pt-10'>
-        <p className='bg-text-linear bg-clip-text font-headline text-2xl font-bold text-transparent'>
+      <div className='grow overflow-auto p-[30px] pb-44 pt-10'>
+        <p className='bg-text-linear bg-clip-text pb-2 font-headline text-2xl font-bold text-transparent'>
```
grod220, line 37: Fixes bug where tx approval window does not scroll all the way down given approve/deny buttons overlap it with a fixed position

packages/ui/components/ui/tx/view/memo-view.tsx
```diff
@@ -10,8 +10,12 @@ export const MemoViewComponent = ({ memo: { memoView } }: { memo: MemoView }) =>
...
             label='Memo Text'
             visibleContent={
-              // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-              <div>{memoView.value.plaintext?.text || '&nbsp;'}</div>
+              <div>
+                {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
+                {memoView.value.plaintext?.text || (
+                  <span className='italic text-gray-600'>== none ==</span>
+                )}
+              </div>
```
grod220, line 18: Was confused by seeing `&nbsp;` in the text box. So added a bit of styling instead:

Screenshot 2024-03-20 at 12 29 07 PM

Valentine1898 (approved):
Good job 👍, all LGTM

PR #797, opened 2024-03-19
actually run tests

turbocrime:
some tests weren't running in CI, and turbo was caching some tests too aggressively

now all tests should run, and turbo should not cache test results. notably, it became obvious that 'services' and 'client' contain no tests.

we should consider not using turbo for tests at all, since `vitest run` from the workspace root is equivalent and has no cache logic to fight against.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -14,7 +14,7 @@
       "outputs": ["bin/**"]
     },
     "compile": {
-      "inputs": ["crate/src/**"],
+      "inputs": ["crate/src/**", "crate/Cargo.toml", "crate/Cargo.lock"],
       "outputs": ["wasm/**"]
     },
     "host": {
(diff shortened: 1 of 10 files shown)


turbo.json
```diff
@@ -31,7 +31,7 @@
       "persistent": true,
       "cache": false
     },
-    "test": { "dependsOn": ["compile", "//#playwright-install"] },
+    "test": { "dependsOn": ["compile", "//#playwright-install"], "cache": false },
     "test:rust": { "dependsOn": ["download-keys", "compile"] },
```
TalDerei, line 35: `cache": false` for `test:rust` necessary? otherwise LGTM

TalDerei (approved):

turbo.json
turbocrime: the rust tests are very different. also they take an eternity in CI so i would very much like to cache them when possible.

they seem reliable so far, but i added some config to encourage turbo to run on crate change.

hdevalence: is this repo using the faster runners like the monorepo? that could help

turbocrime: buildjet 16vcpu on everything that uses cargo

PR #801, opened 2024-03-20
New asset pattern helper

grod220:
A proposal for new class that assists in typing so we don't have to do the `groups as unknown as GroupType` pattern.

### DIFF
--- a/packages/constants/src/assets.test.ts
+++ b/packages/constants/src/assets.test.ts
@@ -1,25 +1,25 @@
 import { describe, expect, it } from 'vitest';
-import { assetPatterns } from './assets';
+import { assetPatterns, RegexMatcher } from './assets';
 
 describe('assetPatterns', () => {
   describe('lpNftPattern', () => {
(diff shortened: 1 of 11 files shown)


packages/constants/src/assets.test.ts
```diff
@@ -56,34 +56,79 @@ describe('assetPatterns', () => {
...
+      const matcher = new RegexMatcher(regex);
+      expect(matcher.matches('123')).toBe(false);
+    });
+  });
+
+  it('should return undefined if no groups are present', () => {
+    const regex = /hello/;
+    const matcher = new RegexMatcher(regex);
+    // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
+    expect(matcher.capture('hello')).toBeUndefined();
```
grod220, line 110: Kinda nice the type system knows it shouldn't be calling something without capture groups

Valentine1898 (approved):

PR #802, opened 2024-03-20
Validate zod only in dev

grod220:
Given we do this on each block save and block sync is performance critical, we should only validate in dev mode. I don't notice a difference (at least on my machine), but still best to do. 

See discord for context: https://discord.com/channels/824484045370818580/1044457038886998137/1220036483612807298


TalDerei (approved):

grod220:
Yeah, I think the chrome profiler is misattributing to zod. But regardless, I suppose it's best to do this anyway. I believe it's really the json serialization stuff we are seeing there.

PR #804, opened 2024-03-20
generate tx view in slice

turbocrime:
pulled out of work refactoring custody/authorize.

tx view generation tools are now in their own package. we use these tools in the tx approval slice now, and query the view rpc for metadata when necessary, instead of serializing all metadata and posting it with the request message. the slice accesses the fvk directly, meaning the custody rpc no longer requires fvk at all.


jessepinho (approved):
Nice change! I remember there being some discussion about whether it's OK for the FVK to be in places that are outside of the `grpc/custody` directory, but I think we decided that's fine as long as it's still in the extension, right? If so, 🚢 it!

Issue #798, opened 2024-03-20
[Bug] Planner throws an error in cases where we need to spend more than one note


Valentine1898:
![image](https://github.com/penumbra-zone/web/assets/25391690/a778fbc6-f5a1-409f-bc2f-99c1ad1bfbb3)

If the balance on the account is split into several notes, the planner will give an error when trying to spend more than the smallest of the notes contains

Valentine1898:
We're getting an error here 
https://github.com/penumbra-zone/web/blob/00fc1340ce0ed27dac643e056a779440638aa0d6/packages/wasm/crate/src/planner.rs#L412-L424

Valentine1898:
Current results of the bug investigation:
1. Most likely planner is working correctly 
2. Part of my balances are on ephemeral addresses, and part on the main address. 
3. It seems that the display of balances is also broken because of this. Two Account 0's are displayed, with the same address, and the entire UM balance is on one address
<img width="657" alt="image" src="https://github.com/penumbra-zone/web/assets/25391690/cb9ebd14-69f2-4677-9698-7a542b55664a">
4. When sending a `TransactionPlannerRequest` we pass source - the account from which we want to spend the funds. 
But source is `AddressIndex`, a normal address and an ephemeral address will have a different `AddressIndex`. more details here https://discord.com/channels/824484045370818580/1044457038886998137/1220143628404392006

The reason for the bug is that minifront thinks that the entire UM balance is on one account, but it sends to planner `source` with ephemeral `AddressIndex`, which has only one note with a small balance. Planner get notes by source, and obviously can't spend it to cover the entire balance of the transaction.

PR #807, opened 2024-03-20
 #798  [Bug] planner throws an error in cases where we need to spend more than one note

Valentine1898:
Bug description in issue https://github.com/penumbra-zone/web/issues/798

### DIFF
--- a/apps/minifront/src/fetchers/balances/by-account.ts
+++ b/apps/minifront/src/fetchers/balances/by-account.ts
@@ -1,26 +1,25 @@
 import {
   Address,
-  AddressIndex,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { getBalances } from '.';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { getAddress, getAddressIndex } from '@penumbra-zone/getters/src/address-view';
 
 export interface BalancesByAccount {
-  index: AddressIndex;
+  index: number;
   address: Address;
   balances: BalancesResponse[];
 }
 
 const groupByAccount = (acc: BalancesByAccount[], curr: BalancesResponse): BalancesByAccount[] => {
   const index = getAddressIndex(curr.accountAddress);
-  const grouping = acc.find(a => a.index.equals(index));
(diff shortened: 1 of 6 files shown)


apps/minifront/src/fetchers/balances/by-account.ts
```diff
@@ -1,26 +1,25 @@
...
   Address,
-  AddressIndex,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { getBalances } from '.';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { getAddress, getAddressIndex } from '@penumbra-zone/getters/src/address-view';
 
 export interface BalancesByAccount {
-  index: AddressIndex;
+  index: number;
```
Valentine1898, line 9: Before:

image

After: 

image

apps/minifront/src/fetchers/balances/by-account.ts
```diff
@@ -1,26 +1,25 @@
...
   Address,
-  AddressIndex,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { getBalances } from '.';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { getAddress, getAddressIndex } from '@penumbra-zone/getters/src/address-view';
 
 export interface BalancesByAccount {
-  index: AddressIndex;
+  account: number;
```
Valentine1898, line 9: We should not use `AddressIndex` when grouping by account, because one account can have many ephemeral addresses, each of which will have a different `AddressIndex`

Before:

image

After: 

image

packages/wasm/crate/src/storage.rs
```diff
@@ -140,8 +140,11 @@ impl IndexedDBStorage {
                 continue;
             }
 
-            if Some(record.address_index) != address_index {
-                continue;
+            // Planner should omit the address index randomizer and compare only the account index
+            if let Some(ref ai) = address_index {
```
grod220, line 144: Do we need `ref`? Looks like this works the same:

```
if let Some(ai) = address_index {
```

grod220 (approved):
Nice catch 👍

PR #813, opened 2024-03-21
too fancy json view

turbocrime:
im sorry im like this.

but try inspecting a tx with this it's a much nicer experience

### DIFF
--- a/packages/ui/components/ui/json-viewer.tsx
+++ b/packages/ui/components/ui/json-viewer.tsx
@@ -1,12 +1,42 @@
 import type { JsonObject, JsonValue } from '@bufbuild/protobuf';
 import { JsonView } from 'react-json-view-lite';
 
+const collapseLevel = 3;
+
+const objectDepth = (o: JsonValue): number =>
+  o && typeof o === 'object' ? 1 + Math.max(-1, ...Object.values(o).map(objectDepth)) : 0;
+
+const objectLength = (o: JsonValue): number =>
+  o && typeof o === 'object' ? Object.entries(o).length : 0;
+
+const shouldExpandNode = (level: number, value: JsonValue, field?: string) => {
+  if (
+    // exmpand empty, so they can be collapsed by css
+    objectLength(value) === 0 ||
+    objectDepth(value) === 0 ||
+    // expand arrays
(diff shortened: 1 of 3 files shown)


packages/ui/components/ui/json-viewer.tsx
```diff
@@ -1,30 +1,129 @@
...
+  o && typeof o === 'object' ? Object.entries(o).length : 0;
 
 export const JsonViewer = ({ jsonObj }: { jsonObj: JsonObject | JsonValue[] }) => {
+  const [expandAll, setExpandAll] = useState(false);
+
+  const shouldExpandNode = useCallback(
+    (level: number, value: JsonValue, field?: string) => {
+      if (expandAll) return true;
+      if (
+        // expand empty, so they can be collapsed by css
```
jessepinho, line 24: Not sure I understand this — collapsed by CSS?

jessepinho (approved):
SUCH a nice improvement 👏🏻 had a couple Qs/suggestions but feel free to merge after addressing those.

packages/ui/components/ui/json-viewer.tsx
turbocrime: instead of ever rendering an empty object as `field: { ••• }` and causing the user to click it and see nothing, all empty objects are expanded. then css can render it compactly as `field: { }`

when i wrote the comment i think i didn't have 'collapsed' showing as inline yet, so a collapsed empty would take up multiple lines.

packages/ui/components/ui/json-viewer.tsx
```diff
@@ -1,30 +1,129 @@
 import type { JsonObject, JsonValue } from '@bufbuild/protobuf';
 import { JsonView } from 'react-json-view-lite';
+import { cn } from '../../lib/utils';
+import { CopyToClipboardIconButton } from './copy-to-clipboard-icon-button';
+import { DoubleArrowDownIcon, DoubleArrowUpIcon } from '@radix-ui/react-icons';
+import { Button } from './button';
+import { useCallback, useState } from 'react';
+
+const collapseLevel = 3;
```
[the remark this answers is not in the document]
turbocrime: i just moved it into the function

PR #815, opened 2024-03-21
Barrel imports (changeset)

TalDerei:
forgot to add changeset to https://github.com/penumbra-zone/web/pull/783


grod220 (approved):
Does every single package/app deserve a major bump? These feels like they don’t need one, but will leave it up to you!

- eslint-config-custom
- @penumbra-zone/tailwind-config

Also keep in mind that you don’t need to give a version bump to things that have dependencies on packages that got bumped. Later `changeset version` will automatically give those the appropriate bumps it needs.

PR #816, opened 2024-03-21
Price delegation tokens

### LINKED ISSUE #527: Price delegation tokens
For delegation tokens, populate ValueView's equivalentValues array with pricing data:

- On asset balances request, if a delegation token is found, query database for exchange rate
- If exchange rate is absent or old (they change each epoch), query full node for latest exchange rates and save to db
- Multiply delegation token by exchange rate for UM token amount to put in equivalentValues array

### LINKED ISSUE #683: Show the UM-equivalent value of delegation tokens below the delegation token balance
Rough mockup:
image

could just be 0 UM until the actual data exists
and we might want to have a color code for the background of the value rendering so that there's a visual cue that it's an alternate representation

jessepinho:
image

We want to show the equivalent values of delegation tokens, using the staking token as the numeraire (i.e., the standard by which the delegation tokens' value is measured).

## In this PR
- Show delegation tokens' equivalents on the staking page.
- Remove rendering of equivalent values from `ValueViewComponent`, per Henry's comment [here](https://discord.com/channels/824484045370818580/1044457038886998137/1220609960828731472).
- Add a `pricedAsset` index to the IndexedDB `PRICES` table, so that we can grab all prices for a given asset.
- Create a new `<Pill />` component; use it for `ValueViewComponent`.

Closes #527 
Closes #683


packages/query/src/price-indexer.ts
```diff
@@ -46,8 +46,9 @@ export const calculatePrice = (delta: Amount, unfilled: Amount, lambda: Amount):
  * This function processes only (1) price and ignores (2) price
  * We can get a BSOD with zero deltas(inputs), and we shouldn't save the price in that case
  */
-export const updatePrices = async (
+export const updatePricesFromSwaps = async (
```
jessepinho, line 49: Wanted to make it clear that this helper is specifically for the swaps use case, not for e.g., delegation tokens.

packages/query/src/block-processor.ts
```diff
@@ -276,15 +295,22 @@ export class BlockProcessor implements BlockProcessorInterface {
...
 
-      if (metadataFromNode) {
-        await this.indexedDb.saveAssetsMetadata(customizeSymbol(metadataFromNode));
-      }
+    if (metadataFromNode) {
+      await this.indexedDb.saveAssetsMetadata(customizeSymbol(metadataFromNode));
+      return metadataFromNode;
     }
+
+    return undefined;
```
jessepinho, line 313: Just extracting this logic out into a new `saveAndReturnMetadata` method so that it can be reused below.

packages/query/src/block-processor.ts
```diff
@@ -35,6 +52,14 @@ const blankTxSource = new CommitmentSource({
   source: { case: 'transaction', value: { id: new Uint8Array() } },
 });
 
+const getExchangeRateFromValidatorInfoResponse = getValidatorInfo
+  .pipe(getRateData)
+  .pipe(getValidatorExchangeRate);
+
+const getIdentityKeyFromValidatorInfoResponse = getValidatorInfo.pipe(
+  getIdentityKeyFromValidatorInfo,
+);
```
jessepinho, line 61: These seem like such specific use cases that I figured they'd be better here, rather than in the `getters` package. Thoughts?

packages/query/src/queriers/shielded-pool.ts
```diff
@@ -14,7 +14,7 @@ export class ShieldedPoolQuerier implements ShieldedPoolQuerierInterface {
     this.client = createClient(grpcEndpoint, QueryService);
   }
 
-  async assetMetadata(assetId: AssetId): Promise<Metadata | undefined> {
+  async assetMetadataById(assetId: AssetId): Promise<Metadata | undefined> {
```
jessepinho, line 17: Not strictly necessary for this PR, but I figured it made sense to align the method name with the RPC method.

packages/ui/components/ui/conditional-wrap.tsx
```diff
@@ -0,0 +1,32 @@
...
+ *     </Tooltip>
+ *   )}
+ * >
+ *   Here is the content that may or may not need a tooltip.
+ * </ConditionalWrap>
+ * ```
+ *
+ * @see https://stackoverflow.com/a/56870316/974981
+ */
+export const ConditionalWrap = ({
```
jessepinho, line 24: I created this because I was going to conditionally render a tooltip around the value view. But per Henry's comment [here](https://discord.com/channels/824484045370818580/1044457038886998137/1220609960828731472), we're only going to show equivalent values in circumstances where the parent component knows which values it's interested in and renders those specifically. So this component isn't needed for this PR anymore, but I figured I'd leave it in anyway in case we need it in the future. Thoughts?

packages/query/src/block-processor.ts
grod220: I'd be in favor of moving them to getters. If we are successful, lots of folks will be wanting to use & contribute to getters.

packages/query/src/block-processor.ts
```diff
@@ -391,17 +423,46 @@ export class BlockProcessor implements BlockProcessorInterface {
...
       if (!validatorInfoResponse.validatorInfo) continue;
 
       // Await the upsert. This makes it possible for users of this method to
       // await the entire method, if they want to block all other code until all
       // validator infos have been upserted.
       await this.indexedDb.upsertValidatorInfo(validatorInfoResponse.validatorInfo);
+
+      // Don't await this, though -- price equivalents for delegation tokens are
+      // non-critical, and shouldn't block the rest of the block processor.
+      void this.updatePriceForValidatorDelegationToken(validatorInfoResponse, nextEpochStartHeight);
```
grod220, line 440: In reality though, this will simply run when the call stack is empty right? In between loops of the `compactBlockRange()` iterator? If so, it probably doesn't make a difference?

This is one of those moments where I wish it were easier to just throw this in another thread and forget it.

packages/query/src/block-processor.ts
```diff
@@ -2,6 +2,13 @@ import { RootQuerier } from './root-querier';
 import { sha256Hash } from '@penumbra-zone/crypto-web/src/sha256';
 import { computePositionId, getLpNftMetadata } from '@penumbra-zone/wasm/src/dex';
 import { decodeSctRoot } from '@penumbra-zone/wasm/src/sct';
+
+import Array from '@penumbra-zone/polyfills/src/Array.fromAsync';
```
grod220, line 6: What do we need the polyfill for?

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -56,8 +63,12 @@ type AccountMap = Record<AddressIndex['account'], BalancesMap>;
 
 class BalancesAggregator {
   readonly accounts: AccountMap = {};
+  private readonly estimatedPriceByPricedAsset: Record<string, EstimatedPrice[]> = {};
```
grod220, line 66: How about:
```
Record<Stringified<AssetId['inner']>, EstimatedPrice[]>
```

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -109,6 +120,39 @@ class BalancesAggregator {
     );
     currentAmount.lo = newAmount.lo;
     currentAmount.hi = newAmount.hi;
+
+    await this.aggregateEquivalentValues(valueView, toAdd);
```
grod220, line 124: How about we pull this out of this function so it sits on the parent alongside:

```typescript
await this.aggregateAmount(valueView, n);
await this.aggregateEquivalentValues(valueView, toAdd);
```

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -109,6 +120,39 @@ class BalancesAggregator {
...
+  }
+
+  private async aggregateEquivalentValues(valueView: ValueView, toAdd: SpendableNoteRecord) {
```
grod220, line 127: Can we get a brief doc string about how this is aggregating?

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -109,6 +120,39 @@ class BalancesAggregator {
...
+    const assetId = getAssetIdFromRecord.optional()(toAdd);
+    if (!assetId?.inner) return;
+
+    const amount = getAmount(valueView);
+
+    const equivalentValues: EquivalentValue[] = [];
+
+    for (const price of this.estimatedPriceByPricedAsset[uint8ArrayToBase64(assetId.inner)] ?? []) {
```
grod220, line 135: Not strictly necessary, but may be a nicer api to have a small external class to manage this object so you could do something like:
```
this.estimatedPriceByPricedAsset.add(assetId)
```
and the class would handle the object keys and getting as well.

It may help also moving the initialization logic out as well to this helper class.

packages/ui/components/ui/conditional-wrap.tsx
grod220: Sure, think we can keep this

grod220 (approved):
Great work here! A general design comment:

Screenshot 2024-03-25 at 12 38 41 PM

Are we distinguishing visually enough that the user does not have _both_ of these, but the second is an estimated value of the first? Perhaps something like this would help?
Screenshot 2024-03-25 at 12 40 15 PM (1)

packages/query/src/block-processor.ts
jessepinho: Sort of... `.updatePriceForValidatorDelegationToken()`has a few `await`s of its own, so I guess those would get threaded between following frames of JS work.

packages/query/src/block-processor.ts
jessepinho: Whoops — guessing I had been using `Array.fromAsync` before but removed it, but the linter didn't catch it since we're using `Array` in this file. 

I just made another commit to remove this import. Good catch.

packages/router/src/grpc/view-protocol-server/balances.ts
jessepinho: done

packages/router/src/grpc/view-protocol-server/balances.ts
jessepinho: Good call, done

packages/router/src/grpc/view-protocol-server/balances.ts
jessepinho: Lemme do this as a quick-follow up PR in case there are disagreements about how it's implemented.

Issue #796, opened 2024-03-19
Create tools for measuring performance


turbocrime:
- [ ] create tool to measure chain download time
- [ ] performance marks in block processor
- [ ] performance marks in build
- [ ] performance marks in idb queries

PR #820, opened 2024-03-23
trial decrypt from ts

turbocrime:
i couldn't avoid playing with it.

this is messy but it seems to be able to filter out uninteresting blocks in approx 4.5ms, whereas jsonifying a compact block takes about 45ms. so i expected a 10x speedup

- with these changes: sync to 10k takes ~12sec
- with out these changes sync to 10k takes ~12sec ???
- a no-op block processor: download to 10k takes ~1sec

there's large variation in all times :/

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -68,64 +68,6 @@ export class BlockProcessor implements BlockProcessorInterface {
     this.abortController.abort(`Sync abort ${r}`);
   }
 
-  private async identifyTransactions(
-    spentNullifiers: Set<Nullifier>,
-    commitmentRecords: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
-    blockTx: Transaction[],
-  ) {
-    const relevantTx = new Map<TransactionId, Transaction>();
-    const recordsWithSources = new Array<SpendableNoteRecord | SwapRecord>();
-    for (const tx of blockTx) {
-      let txId: TransactionId | undefined;
-
-      const txCommitments = (tx.body?.actions ?? []).flatMap(({ action }) => {
-        switch (action.case) {
-          case 'output':
-            return action.value.body?.notePayload?.noteCommitment;
-          case 'swap':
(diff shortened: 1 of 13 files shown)


packages/query/src/block-processor.ts
```diff
@@ -146,6 +146,65 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+            }
+            default: {
+              throw new Error('unknown state payload');
+            }
+          }
+        }
+
+        // didn't find anything. time to forget all of that
+        forgettable.map(commitment => this.viewServer.forgetCommitment(commitment));
+        this.viewServer.dontScanBlock(compactBlock.height);
```
turbocrime, line 205: i'm only batching this because if we had found something, we would pass the compact block to wasm, which would progressively forget as it scans. so i have to avoid mutating the sct until i know the block is totally uninteresting

packages/query/src/block-processor.ts
```diff
@@ -146,6 +146,65 @@ export class BlockProcessor implements BlockProcessorInterface {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
+      let trial: undefined | string;
+      let epoch = false;
+
+      trialDecrypt: {
```
turbocrime, line 152: i know a labelled block is unusual, this was just to try it out

turbocrime (commented):
if we exposed nullifier::derive, tools for the swap outputs, and sct, i think we could easily move the block processor logic entirely into typescript, and get rid of the logic split.

hdevalence:
I don't think it's desirable to move the sync logic into typescript. The profiling shows that the wasm code we have is so fast it barely shows up on the profile, while the typescript logic spends most of its time doing serialization. This seems like a reason to not do this, and instead go in the opposite direction, moving logic into Rust/wasm, which is fast, rather than the typescript code, which is slow.

turbocrime:
agreed

turbocrime:
this didn't pan out but maybe the timing code can serve https://github.com/penumbra-zone/web/issues/796

Issue #533, opened 2024-02-15
Should not be able to select same asset in/out


grod220:
User's should not be able to select the same asset for the in/out box. Will involve filtering and a smarter way to initialize the component.

<img width="305" alt="Screenshot 2024-02-15 at 3 00 53 PM" src="https://github.com/penumbra-zone/web/assets/16624263/45621453-ac57-45fb-9568-fe3909c109ea">

This could be part of `assertValidSwaps` in the custody service's `authorize` RPC method. See https://github.com/penumbra-zone/web/pull/821#discussion_r1537044452

hdevalence:
I think we should restructure this UI to have an arrow signifying the swap and only selecting the output token (why do we show the wallet balance of what we're swapping _into_?)

grod220:
Design of the asset out wallet balance basically came from looking as Sushi, but we can remove it if you find it not helpful:

<img width="538" alt="Screenshot 2024-02-16 at 10 23 37 AM" src="https://github.com/penumbra-zone/web/assets/16624263/6b4568fb-0859-453f-a913-2305e566f04c">

They also have a helpful thing where clicking the arrow swaps the asset in/out denoms

TalDerei:
@jessepinho adding comment from https://github.com/penumbra-zone/web/pull/821 to this ticket.

"when trying to swap the same pair, the extension throws **"failed to deliver transaction: check_stateless failed: Trading pair must be distinct"**.

maybe we should have an additional `assertDistinctSwapAssets` assertion check that handles that error with a dedicated toast?"

grod220:
> maybe we should have an additional assertDistinctSwapAssets assertion check that handles that error with a dedicated toast?

In theory, after this issue is addressed, this invalid tx should not be possible to perform right?

jessepinho:
@grod220:

> In theory, after this issue is addressed, this invalid tx should not be possible to perform right?

This assumes people will be using minifront with our extension. But in fact, they could be using a different webapp with our extension, so IMO, both the extension and the UI should check this.

PR #821, opened 2024-03-23
Reject `Swap` approvals with foreign claim addresses

### LINKED ISSUE #548: Reject `Swap` approvals with foreign claim addresses
The claim address is specified in the `SwapPlan`. It should always be an address the user controls.

The extension's approval code should check that any submitted `SwapPlan` has a claim address controlled by the user, and automatically reject it if not.

Adding this check means that we don't need to display the claim address to the user in the approval view.

jessepinho:
From #548:

> The claim address is specified in the SwapPlan. It should always be an address the user controls.
>
> **The extension's approval code should check that any submitted SwapPlan has a claim address controlled by the user, and automatically reject it if not.**
>
> Adding this check means that we don't need to display the claim address to the user in the approval view.

## In this PR
- In the `CustodyService`'s `authorize` implementation, assert that transaction swaps (if there are any) contain claim addresses controlled by the current user.
  - To keep things clean, and because we may want to add more assertions as we think of other cases like this, I put the assertion code and its tests in a separate file from the `authorize` implementation.
- Rename the poorly named `isControlledAddress` helper to `getAddressIndexByAddress` to better represent what it does.
- Create a new `isControlledAddress` helper that actually returns a boolean. 
- Remove the claim address from `<SwapViewComponent />`.

## Manual testing

The code:

```diff
diff --git a/apps/minifront/src/state/swap.ts b/apps/minifront/src/state/swap.ts
index 49ca99fe..52a867f4 100644
--- a/apps/minifront/src/state/swap.ts
+++ b/apps/minifront/src/state/swap.ts
@@ -24,6 +24,7 @@ import { getAssetId } from '@penumbra-zone/getters/src/metadata';
 import { getSwapCommitmentFromTx } from '@penumbra-zone/getters/src/transaction';
 import { getAddressIndex } from '@penumbra-zone/getters/src/address-view';
 import { toBaseUnit } from '@penumbra-zone/types/src/lo-hi';
+import { bech32ToAddress } from '@penumbra-zone/bech32/src/address';

 export interface SimulateSwapResult {
   output: ValueView;
@@ -163,7 +164,13 @@ const assembleSwapRequest = async ({ assetIn, amount, assetOut }: SwapSlice) =>
           ),
           assetId: getAssetIdFromValueView(assetIn.balanceView),
         },
-        claimAddress: await getAddressByIndex(addressIndex.account),
+        // claimAddress: await getAddressByIndex(addressIndex.account),
+        claimAddress: {
+          inner: bech32ToAddress(
+            'penumbra1tj9s0dh8ymphnw2qgrva57wpcmpr38e9a8w05t5c8dyynr7dt58hgnare7tunqyuryc7yag084rqxrsrj9yx8uk9yhs8lmg64hl993wn3rphlvan3paxdd6suyhwnpkd6cwzes',
+          ),
+        },
+
         // TODO: Calculate this properly in subsequent PR
         //       Asset Id should almost certainly be upenumbra,
         //       may need to indicate native denom in registry
```

The result:

https://github.com/penumbra-zone/web/assets/1121544/1a4caf95-9a73-4f11-a11f-2ee58be57574

Note that it just says "Transaction canceled," rather than showing the error message about the website stealing your funds. That's because it throws a `ConnectError` with `Code.PermissionDenied`, which minifront catches and displays as "Transaction canceled." 

Closes #548

### DIFF
--- a/packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
+++ b/packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
@@ -1,0 +1,19 @@
+import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { TransactionPlan } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import { Code, ConnectError } from '@connectrpc/connect';
+
+export const assertValidSwaps = (
+  plan: TransactionPlan,
+  isControlledAddress: (address?: Address) => boolean,
+): void => {
+  plan.actions.forEach(action => {
+    if (action.action.case !== 'swap') return;
+
+    if (!isControlledAddress(action.action.value.swapPlaintext?.claimAddress)) {
+      throw new ConnectError(
+        "Tried to initiate a swap with a claim address belonging to a different user. This means that, when the swap is claimed, the funds would go to someone else's address, not yours. This should never happen. The website you are using may be trying to steal your funds!",
+        Code.PermissionDenied,
+      );
+    }
+  });
+};
--- a/packages/router/src/grpc/custody/authorize/index.ts
+++ b/packages/router/src/grpc/custody/authorize/index.ts
@@ -7,7 +7,7 @@ import { Key } from '@penumbra-zone/crypto-web/src/encryption';
 import { Code, ConnectError } from '@connectrpc/connect';
 import { Box } from '@penumbra-zone/types/src/box';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
-import { assertValidSwaps } from './assert-valid-swaps';
+import { assertSwapClaimAddressesBelongToCurrentUser } from './assert-swap-claim-addresses-belong-to-current-user';
 import { isControlledAddress } from '@penumbra-zone/wasm/src/address';
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
@@ -19,7 +19,9 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   const walletServices = await ctx.values.get(servicesCtx).getWalletServices();
 
   const { fullViewingKey } = walletServices.viewServer;
-  assertValidSwaps(req.plan, address => isControlledAddress(fullViewingKey, address));
+  assertSwapClaimAddressesBelongToCurrentUser(req.plan, address =>
+    isControlledAddress(fullViewingKey, address),
+  );
 
   if (!approveReq) throw new ConnectError('Approver not found', Code.Unavailable);
 
--- a/packages/router/src/grpc/custody/authorize/index.ts
+++ b/packages/router/src/grpc/custody/authorize/index.ts
@@ -1,12 +1,15 @@
-import type { Impl } from '.';
-import { extLocalCtx, extSessionCtx } from '../../ctx/prax';
-import { approverCtx } from '../../ctx/approver';
+import type { Impl } from '..';
+import { extLocalCtx, extSessionCtx, servicesCtx } from '../../../ctx/prax';
+import { approverCtx } from '../../../ctx/approver';
 import { generateSpendKey } from '@penumbra-zone/wasm/src/keys';
 import { authorizePlan } from '@penumbra-zone/wasm/src/build';
 import { Key } from '@penumbra-zone/crypto-web/src/encryption';
 import { Code, ConnectError } from '@connectrpc/connect';
 import { Box } from '@penumbra-zone/types/src/box';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
+import { assertValidSwaps } from './assert-valid-swaps';
+import { isControlledAddress } from '@penumbra-zone/wasm/src/address';
+import { bech32Address } from '@penumbra-zone/bech32/src/address';
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!req.plan) throw new ConnectError('No plan included in request', Code.InvalidArgument);
@@ -14,6 +17,12 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   const approveReq = ctx.values.get(approverCtx);
   const sess = ctx.values.get(extSessionCtx);
   const local = ctx.values.get(extLocalCtx);
+  const services = ctx.values.get(servicesCtx);
+
+  const { fullViewingKey } = (await services.getWalletServices()).viewServer;
+  assertValidSwaps(req.plan, address =>
+    address ? Boolean(isControlledAddress(fullViewingKey, bech32Address(address))) : false,
+  );
 
   if (!approveReq) throw new ConnectError('Approver not found', Code.Unavailable);
 
--- a/packages/router/src/grpc/custody/authorize/index.ts
+++ b/packages/router/src/grpc/custody/authorize/index.ts
@@ -8,7 +8,7 @@ import { Code, ConnectError } from '@connectrpc/connect';
 import { Box } from '@penumbra-zone/types/src/box';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
 import { assertValidSwaps } from './assert-valid-swaps';
-import { getAddressIndexByAddress } from '@penumbra-zone/wasm/src/address';
+import { isControlledAddress } from '@penumbra-zone/wasm/src/address';
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!req.plan) throw new ConnectError('No plan included in request', Code.InvalidArgument);
@@ -19,9 +19,7 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
   const services = ctx.values.get(servicesCtx);
 
   const { fullViewingKey } = (await services.getWalletServices()).viewServer;
-  assertValidSwaps(req.plan, address =>
-    address ? Boolean(getAddressIndexByAddress(fullViewingKey, address)) : false,
-  );
+  assertValidSwaps(req.plan, address => isControlledAddress(fullViewingKey, address));
 
   if (!approveReq) throw new ConnectError('Approver not found', Code.Unavailable);
 
--- a/packages/router/src/grpc/custody/authorize/index.ts
+++ b/packages/router/src/grpc/custody/authorize/index.ts
@@ -9,7 +9,6 @@ import { Box } from '@penumbra-zone/types/src/box';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
 import { assertValidSwaps } from './assert-valid-swaps';
 import { getAddressIndexByAddress } from '@penumbra-zone/wasm/src/address';
-import { bech32Address } from '@penumbra-zone/bech32/src/address';
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!req.plan) throw new ConnectError('No plan included in request', Code.InvalidArgument);
@@ -21,7 +20,7 @@ export const authorize: Impl['authorize'] = async (req, ctx) => {
 
   const { fullViewingKey } = (await services.getWalletServices()).viewServer;
   assertValidSwaps(req.plan, address =>
-    address ? Boolean(getAddressIndexByAddress(fullViewingKey, bech32Address(address))) : false,
+    address ? Boolean(getAddressIndexByAddress(fullViewingKey, address)) : false,
   );
 
   if (!approveReq) throw new ConnectError('Approver not found', Code.Unavailable);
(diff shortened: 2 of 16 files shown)


packages/ui/components/ui/tx/view/swap.tsx
```diff
@@ -3,17 +3,10 @@ import { SwapView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/
 import { fromBaseUnitAmount, joinLoHiAmount } from '@penumbra-zone/types/src/amount';
 import { uint8ArrayToBase64 } from '@penumbra-zone/types/src/base64';
 import { ActionDetails } from './action-details';
-import { AddressView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-import { AddressViewComponent } from './address-view';
 
 export const SwapViewComponent = ({ value }: { value: SwapView }) => {
```
jessepinho, line 7: We no longer need to show the claim address. Here's what that looks like in the approval dialog (it's the same in the Swap transaction view of a successful transaction, too):

image

packages/router/src/grpc/test-utils.ts
```diff
@@ -35,7 +35,7 @@ export interface ShieldedPoolMock {
 }
 
 export interface ViewServerMock {
-  fullViewingKey?: Mock;
+  fullViewingKey?: string;
```
jessepinho, line 38: This should have always been a string

packages/wasm/src/address.ts
```diff
@@ -1,14 +1,21 @@
...
+import { bech32Address } from '@penumbra-zone/bech32/src/address';
 
 export const getShortAddressByIndex = (fullViewingKey: string, index: number) =>
   get_short_address_by_index(fullViewingKey, index) as string;
 
-export const isControlledAddress = (
+export const getAddressIndexByAddress = (
   fullViewingKey: string,
-  address: string,
+  address: Address,
```
jessepinho, line 14: Instead of expecting a bech32-encoded address, we just accept an `Address`, and do the encoding ourselves inside this function.

TalDerei:
minor nit; to maintain consistency, can we deactivate the swap button until the user enters a non-zero amount, similar to sends?

TalDerei:
when trying to swap the same pair, the extension throws `"failed to deliver transaction: check_stateless failed: Trading pair must be distinct"`. We should handle that error with a dedicated toast.

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
```diff
@@ -0,0 +1,19 @@
+import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { TransactionPlan } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import { Code, ConnectError } from '@connectrpc/connect';
+
+export const assertValidSwaps = (
```
TalDerei, line 5: when trying to swap the same pair, the extension throws "failed to deliver transaction: check_stateless failed: Trading pair must be distinct".

maybe we should have a separate `assertDistinctAssets` assertion check that handles that error with a dedicated toast?

TalDerei:
> That's because it throws a `ConnectError` with `Code.PermissionDenied`, which minifront catches and displays as "Transaction canceled."

we might want to consider turning this into a more stateful error message describing why the transaction was canceled down the line.

TalDerei:
@jessepinho this is out of scope for this PR, but maybe we should separately consider adding a minor delay between swap and swap claim toasts, or slight offset in the UI? I'm not sure this is the most intuitive UX, because the view changes so quickly between transactions that I find myself missing the display of the first transaction. Curious what you think. 

Screenshot 2024-03-24 at 9 56 33 PM

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
```diff
@@ -0,0 +1,19 @@
...
+  isControlledAddress: (address?: Address) => boolean,
+): void => {
+  plan.actions.forEach(action => {
+    if (action.action.case !== 'swap') return;
+
+    if (!isControlledAddress(action.action.value.swapPlaintext?.claimAddress)) {
+      throw new ConnectError(
+        "Tried to initiate a swap with a claim address belonging to a different user. This means that, when the swap is claimed, the funds would go to someone else's address, not yours. This should never happen. The website you are using may be trying to steal your funds!",
+        Code.PermissionDenied,
+      );
```
grod220, line 16: This seems like quite a helpful error message. Can we show this in our tooltip instead of it being swallowed?

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
grod220: This is an interesting question. Should the extension only forcibly not allow harmful tx's or should we do some additional help? In this case, it feels like the outcome would be the same and that really the frontend needs to implement: https://github.com/penumbra-zone/web/issues/533

packages/router/src/grpc/custody/authorize/index.ts
```diff
@@ -1,19 +1,25 @@
...
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!req.plan) throw new ConnectError('No plan included in request', Code.InvalidArgument);
 
   const approveReq = ctx.values.get(approverCtx);
   const sess = ctx.values.get(extSessionCtx);
   const local = ctx.values.get(extLocalCtx);
+  const services = ctx.values.get(servicesCtx);
+
+  const { fullViewingKey } = (await services.getWalletServices()).viewServer;
```
grod220, line 21: I personally find await's nested in things hard to read. What do you think about?

```typescript
  const services = await ctx.values.get(servicesCtx).getWalletServices();

  assertValidSwaps(req.plan, address =>
    isControlledAddress(services.viewServer.fullViewingKey, address),
  );
```

packages/ui/components/ui/tx/view/swap.tsx
grod220: Hmm. I'm not sure we should hide it. The claiming part of a swap seems so integral to penumbra. I don't think it hurts to show it.

packages/wasm/src/address.ts
```diff
@@ -1,14 +1,21 @@
...
+  AddressIndex,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { JsonValue } from '@bufbuild/protobuf';
+import { bech32Address } from '@penumbra-zone/bech32/src/address';
 
 export const getShortAddressByIndex = (fullViewingKey: string, index: number) =>
   get_short_address_by_index(fullViewingKey, index) as string;
 
-export const isControlledAddress = (
+export const getAddressIndexByAddress = (
```
grod220, line 12: If a consumer is doing `getAddressIndexByAddress()` they are assuming it's controlled right? If so, think we should consider just throwing an error if not instead of returning an undefined.

packages/wasm/src/address.ts
```diff
@@ -1,14 +1,21 @@
...
   fullViewingKey: string,
-  address: string,
+  address: Address,
 ): AddressIndex | undefined => {
-  const res = is_controlled_address(fullViewingKey, address) as JsonValue;
+  const res = is_controlled_address(fullViewingKey, bech32Address(address)) as JsonValue;
```
grod220, line 16: Think we should consider changing the rust api as well. 

`is_controlled_address()` should return a boolean

`get_address_index_by_address()` should return Result<AddressIndex> (throws if not able to)

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
jessepinho: Marking as resolved, as that's already captured in a separate ticket

packages/ui/components/ui/tx/view/swap.tsx
jessepinho: Hm OK. I was basing it on Henry's comment in #548:

> Adding this check means that we don't need to display the claim address to the user in the approval view.

That said, I think the main point of that ticket was to ensure the claim address belonged to the current user, so I'll add it back.

grod220: Then up to you! 🤷

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
turbocrime, line 5: can we name this function to specifically indicate it checks that the user controls the claim address of the swap, since that's the purpose of it?

packages/router/src/grpc/custody/authorize/index.ts
turbocrime:
```suggestion
  const { viewServer: { fullViewingKey } } = await services.getWalletServices();
```

but more generally, we should probably just have a context item for fvk

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
jessepinho: This is related to @TalDerei 's comment above:

> we might want to consider displaying a more stateful error message describing why the swap transaction was canceled down the line.

Here's the issue: minifront will never try to steal the user's funds, so we never need to show this message to minifront users.

We _should_ show this message to users on other sites who try to steal the user's funds, but obviously, a malicious website isn't going to proactively show this error to people. But if they're trying to steal people's funds, they'll presumably test the stealing-people's-funds functionality first, and they'll realize that this error shows up, at which point they'll either A) give up, or B) at least suppress this error.

So I'm not actually sure how useful this error is?

The only way it could be useful is if we could show this error _in the transaction approval popup_, which the malicious website has no control over. Not sure if this is possible, though, since this error is thrown inside an RPC method implementation. What if we merge this as-is for now, and then revisit this question in a separate ticket?

packages/router/src/grpc/custody/authorize/index.ts
jessepinho: Good call, done

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
jessepinho: Ah yeah, good call. I was thinking that we could eventually extend `assertValidSwaps` to do all kinds of assertions on swaps, but right now it's only doing one kind of assertion. [Fixed](https://github.com/penumbra-zone/web/pull/821/commits/f5a720d0ed3bd7ecca3a3ff85c3f0b8b44221f38).

jessepinho:
@TalDerei:

> minor nit; to maintain consistency, can we deactivate the swap button until the user enters a non-zero amount, similar to how we handle send transactions?

> This is out of scope for this PR, but maybe we should separately consider adding a minor delay between swap and swap claim toasts, or slight offset in the UI?

Could you please create separate tickets for these?

packages/wasm/src/address.ts
jessepinho: Hm, not necessarily. The `indexByAddress` RPC method uses this helper, and that method can take any address and simply return an empty object if the address isn't controlled.

grod220 (approved):

packages/ui/components/ui/tx/view/swap.tsx
hdevalence: Yeah, I'd prefer to hide it. The user shouldn't ever have to know about the claim address, it should just do the right thing automatically.

hdevalence: As a follow-up to this PR, we should also change the swap rendering to show formatted values, so that as a user I see something like "Swap 1234 UM => TestUsd"

packages/router/src/grpc/custody/authorize/assert-valid-swaps.ts
TalDerei: created separate ticket to limit scope

packages/router/src/grpc/custody/authorize/index.ts
```diff
@@ -1,19 +1,27 @@
...
 
   const approveReq = ctx.values.get(approverCtx);
   const sess = ctx.values.get(extSessionCtx);
   const local = ctx.values.get(extLocalCtx);
+  const walletServices = await ctx.values.get(servicesCtx).getWalletServices();
+
+  const { fullViewingKey } = walletServices.viewServer;
+  assertSwapClaimAddressesBelongToCurrentUser(req.plan, address =>
+    isControlledAddress(fullViewingKey, address),
+  );
```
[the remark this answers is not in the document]
hdevalence: I think it makes sense to think of these as policies, and have one function per policy that approves or denies, and call each function in sequence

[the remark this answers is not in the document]
jessepinho: I can dig it 👌🏻

Issue #735, opened 2024-03-13
Show slippage when estimating a swap


hdevalence:
The swap dialog should show the estimated slippage along with the expected output.

All the information needed is already in the RPC response: the first trace in the `SwapExecution` is the best (starting) execution, and the last trace is the final (worst) execution. The end-to-end price of a trace is the amount of the first asset in the chain divided by the amount of the last asset in the chain (or vice versa, depending on which asset we're pricing in terms of which).

So the slippage can be computed by comparing the best price to the worst price, and displayed to the user.

PR #828, opened 2024-03-25
Add price impact to swap estimate

grod220:
Closes https://github.com/penumbra-zone/web/issues/735 <--- turns out this isn't a slippage thing, but a price impact one
Closes https://github.com/penumbra-zone/web/issues/829

When a user makes a swap, the size of their trade could impact the price of the asset in the market. This `price impact` percentage is something we want to display to the user when they are estimating the swap they are about to take. This is especially important in low-liquidity pools where users may not want to take that kind of hit.

Zero price impact:
Screenshot 2024-03-25 at 9 57 46 PM

Price impact present:
Screenshot 2024-03-25 at 9 57 54 PM

Price impact worse than 10%
Screenshot 2024-03-25 at 9 58 06 PM

### DIFF
--- a/apps/minifront/src/components/swap/asset-out-box.tsx
+++ b/apps/minifront/src/components/swap/asset-out-box.tsx
@@ -18,7 +18,7 @@ import { cn } from '@penumbra-zone/ui/lib/utils';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { Amount } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/num/v1/num_pb';
 
-import { isZero } from '@penumbra-zone/types/src/amount';
+import { formatNumber, isZero } from '@penumbra-zone/types/src/amount';
 import { getAmount } from '@penumbra-zone/getters/src/value-view';
 
 const findMatchingBalance = (
@@ -56,7 +56,7 @@ export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
       <div className='mb-2 flex items-center justify-between gap-1 md:gap-2'>
         <p className='text-sm font-bold md:text-base'>Swap into</p>
       </div>
-      <div className='flex flex-col gap-4'>
+      <div className='flex justify-between gap-4'>
         <div className='flex items-start justify-start'>
           {simulateOutResult ? (
             <Result result={simulateOutResult} />
@@ -64,12 +64,10 @@ export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
             <EstimateButton simulateFn={simulateSwap} loading={simulateOutLoading} />
           )}
         </div>
-        <div>
-          <AssetOutSelector
-            balances={balances}
-            assetOut={matchingBalance}
-            setAssetOut={setAssetOut}
-          />
+        <div className='flex flex-col'>
+          <div className='ml-auto w-auto shrink-0'>
+            <AssetOutSelector assetOut={matchingBalance} setAssetOut={setAssetOut} />
+          </div>
           <div className='mt-[6px] flex items-start justify-between'>
             <div />
             <div className='flex items-start gap-1'>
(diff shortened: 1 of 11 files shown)


apps/minifront/src/components/swap/asset-out-box.tsx
```diff
@@ -56,33 +56,55 @@ export const AssetOutBox = ({ balances }: AssetOutBoxProps) => {
...
+          <div className='ml-auto w-auto shrink-0'>
+            <AssetOutSelector assetOut={matchingBalance} setAssetOut={setAssetOut} />
+          </div>
+          <div className='mt-[6px] flex items-start justify-between'>
+            <div />
+            <div className='flex items-start gap-1'>
+              <img src='./wallet.svg' alt='Wallet' className='size-5' />
+              <ValueViewComponent view={matchingBalance} showIcon={false} />
+            </div>
+          </div>
```
grod220, line 77: The CSS in this section was kinda wonky. This revision separates the content on the left vs right better.

apps/minifront/src/components/swap/asset-out-selector.tsx
```diff
@@ -11,20 +11,18 @@ import {
...
   assetOut: ValueView | undefined;
   setAssetOut: (metadata: Metadata) => void;
 }
 
 /** @todo Refactor to use `SelectTokenModal` */
-export const AssetOutSelector = ({ balances, setAssetOut, assetOut }: AssetOutSelectorProps) => {
+export const AssetOutSelector = ({ setAssetOut, assetOut }: AssetOutSelectorProps) => {
   return (
     <Dialog>
-      <DialogTrigger disabled={!balances.length}>
```
grod220, line 27: The assets out should not be dependent on the user's balances. This was a mistake.

apps/minifront/src/components/swap/swap-form.tsx
```diff
@@ -2,14 +2,15 @@ import { Button } from '@penumbra-zone/ui/components/ui/button';
...
-import { swapSelector } from '../../state/swap';
+import { swapSelector, swapValidationErrors } from '../../state/swap';
 import { AssetOutBox } from './asset-out-box';
 import { SwapLoaderResponse } from './swap-loader';
 
 export const SwapForm = () => {
   const { assetBalances } = useLoaderData() as SwapLoaderResponse;
   const { assetIn, setAssetIn, amount, setAmount, initiateSwapTx, txInProgress } =
     useStore(swapSelector);
+  const validationErrs = useStore(swapValidationErrors);
```
grod220, line 13: Validation on the swap form wasn't in place, so added it

apps/minifront/src/state/swap.ts
```diff
@@ -187,4 +200,48 @@ export const issueSwapClaim = async (swapCommitment: StateCommitment) => {
...
+  const bestOutputAmount = getMatchingAmount(bestTrace.value, outputAssetId);
+  const bestTraceEstimatedPrice = divideAmounts(bestOutputAmount, bestInputAmount);
+
+  // Difference = (priceB - priceA) / priceA
+  const percentDifference = swapEstimatePrice
+    .minus(bestTraceEstimatedPrice)
+    .div(bestTraceEstimatedPrice);
+
+  return percentDifference.toNumber();
+};
```
grod220, line 230: @erwanor mind taking a look at this code section?

apps/minifront/src/state/swap.ts
```diff
@@ -187,4 +200,48 @@ export const issueSwapClaim = async (swapCommitment: StateCommitment) => {
   await planBuildBroadcast('swapClaim', req, { skipAuth: true });
 };
 
+/*
+  Price impact is the change in price as a consequence of the trade's size. In SwapExecution, the \
+  first trace in the array is the best execution for the swap. To calculate price impact, take
+  the price of the trade and see the % diff off the best execution trace.
+ */
```
grod220, line 207: @hdevalence note that this is a different method than what you proposed in the original issue. After pairing with @erwanor, this felt more accurate.

packages/types/src/amount.ts
```diff
@@ -48,47 +48,26 @@ export const multiplyAmountByNumber = (amount: Amount, multiplier: number): Amou
   return new Amount(loHi);
 };
 
-export const divideAmounts = (dividend: Amount, divider: Amount): BigNumber => {
-  if (isZero(divider)) throw new Error('Division by zero');
+export const divideAmounts = (dividend: Amount, divisor: Amount): BigNumber => {
```
grod220, line 51: Think `divisor` was a more correct term here

apps/minifront/src/state/swap.ts
erwanor: This LGTM!

Valentine1898 (approved):
LGTM! 
Should we allow swap without estimate? And should we then show a warning in the popup about high price impact?

grod220:
> Should we allow swap without estimate? 

Indeed. Requesting an estimate is quite normal in every other dex protocol. However, in our case, it reveals specific intentions to the rpc you are connected to. So for privacy reasons, it's opt in.

> And should we then show a warning in the popup about high price impact?

Given above, we couldn't pro-actively do this on behalf of the user.

Issue #810, opened 2024-03-21
header titles jump around because of inconsistent positioning


hdevalence:
we need to make sure that they're always in boxes and that only the box coloring changes

https://github.com/penumbra-zone/web/assets/44879/b2e67e38-6060-4d36-af88-1c50d4b137f2

PR #832, opened 2024-03-26
Prevent jumping headers

grod220:
Closes https://github.com/penumbra-zone/web/issues/810

Making the padding not conditional prevents them from resizing if active


Valentine1898 (approved):
LGTM

PR #833, opened 2024-03-26
Clarify wasm func naming

grod220:
Follow up from: https://github.com/penumbra-zone/web/pull/821/files#r1538165300

A small PR to improve the naming of the wasm function getting index from an address.

### DIFF
--- a/packages/wasm/crate/src/keys.rs
+++ b/packages/wasm/crate/src/keys.rs
@@ -1,9 +1,6 @@
 use std::str::FromStr;
 
 use anyhow;
-use rand_core::OsRng;
-use wasm_bindgen::prelude::*;
-
 use penumbra_keys::keys::{Bip44Path, SeedPhrase, SpendKey};
 use penumbra_keys::{Address, FullViewingKey};
 use penumbra_proof_params::{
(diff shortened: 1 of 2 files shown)


packages/wasm/crate/src/keys.rs
```diff
@@ -132,13 +131,14 @@ pub fn get_ephemeral_address(full_viewing_key: &str, index: u32) -> WasmResult<J
...
-/// Check if the address is FVK controlled
+/// Returns the AddressIndex of an address.
+/// If it is not controlled by the FVK, it returns a `None`
 /// Arguments:
 ///     full_viewing_key: `bech32 String`
 ///     address: `bech32 String`
 /// Returns: `Option<pb::AddressIndex>`
 #[wasm_bindgen]
-pub fn is_controlled_address(full_viewing_key: &str, address: &str) -> WasmResult<JsValue> {
+pub fn get_index_by_address(full_viewing_key: &str, address: &str) -> WasmResult<JsValue> {
```
Valentine1898, line 141: We pass `full_viewing_key` and `address` here as bech32 strings, although we can use the proto types `FullViewingKey` and `Address`. 
In addition, we manually convert `Address` to bech32
```typescript
get_index_by_address(fullViewingKey, bech32Address(address)) as JsonValue;
``` 
But at the same time, fvk always exists in extension as a bech32 string 

I don't think we should fix it in this PR since it doesn't only affect the `get_index_by_address` function but the whole wasm, but I wonder what you think about it .

Valentine1898 (approved):
LGTM

packages/wasm/crate/src/keys.rs
grod220: > we manually convert Address to bech32

Ah, I should fix this in this PR 👍 

> fvk always exists in extension as a bech32 string

Indeed, it is a bit strange we pass fvk around as a string versus use the bufbuild type for this. We should definitely update that before mainnet, can you make a ticket for this?

PR #834, opened 2024-03-26
Improve wasm signatures

grod220:
Closes https://github.com/penumbra-zone/web/issues/845

Based on feedback in retro to update wasm signatures to ensure we do not use `JsValue` instead of the [supported types](https://rustwasm.github.io/wasm-bindgen/reference/types.html).

Along the way, I also noticed we were serializing through JSON.stringify() unnecessarily (I think? No documentation on why for these.).


packages/wasm/crate/src/tx.rs
```diff
@@ -36,34 +36,6 @@ impl TxInfoResponse {
...
-pub fn decode_tx(tx_bytes: &str) -> WasmResult<JsValue> {
-    utils::set_panic_hook();
-
-    let tx_vec: Vec<u8> =
-        base64::Engine::decode(&base64::engine::general_purpose::STANDARD, tx_bytes)?;
-    let transaction: Transaction = Transaction::try_from(tx_vec)?;
-    let result = serde_wasm_bindgen::to_value(&transaction)?;
-    Ok(result)
-}
-
```
grod220, line 66: Both unused

packages/wasm/src/view-server.ts
```diff
@@ -73,22 +80,13 @@ export class ViewServer implements ViewServerInterface {
...
-      new_swaps: (new_swaps ?? []).map(s => JSON.stringify(s)),
-    };
+    const result = this.wasmViewServer.flush_updates() as FlushResult;
+    const { height, sct_updates, new_notes, new_swaps } = result;
     return {
-      height: BigInt(height ?? false),
+      height: BigInt(height ?? 0),
       sctUpdates: validateSchema(SctUpdatesSchema, sct_updates),
-      newNotes: wasmJson.new_notes.map(n => SpendableNoteRecord.fromJsonString(n)),
-      newSwaps: wasmJson.new_swaps.map(s => SwapRecord.fromJsonString(s)),
```
grod220, line 91: The deserialization here felt a bit odd. Given it's so different, it feels like it was intentional, but it's hard to tell without any kind of documentation alongside.

Regardless, it feels unnecessary to do `json -> string -> class` when we can do `json -> class`.

packages/wasm/crate/src/tx.rs
turbocrime: these are equivalent to bufbuild's toBinary/fromBinary anyway

packages/wasm/src/view-server.ts
```diff
@@ -73,22 +80,13 @@ export class ViewServer implements ViewServerInterface {
...
-    const wasmJson = {
-      new_notes: (new_notes ?? []).map(n => JSON.stringify(n)),
-      new_swaps: (new_swaps ?? []).map(s => JSON.stringify(s)),
-    };
+    const result = this.wasmViewServer.flush_updates() as FlushResult;
+    const { height, sct_updates, new_notes, new_swaps } = result;
     return {
-      height: BigInt(height ?? false),
+      height: BigInt(height ?? 0),
       sctUpdates: validateSchema(SctUpdatesSchema, sct_updates),
```
turbocrime, line 87: is it possible to avoid this schema and emit a type for sctUpdates from wasm, or use a pb type

turbocrime (commented):
there are some convertable parameter types as well that i noted the other day. will suggest in a moment

packages/wasm/src/view-server.ts
grod220: It would be nice, but there aren't protobufs for any of these. Added an issue: https://github.com/penumbra-zone/penumbra/issues/4118

packages/wasm/src/view-server.ts
turbocrime: i remember this happened for a reason, but i don't remember what that reason was. worth being careful about this.

notably this is one of the few places where a pb type is emitted to js, which apparently works for some reason.

turbocrime:
examples how to do these in https://github.com/penumbra-zone/web/pull/844

grod220:
Should wait to merge until after v0.70 release + extensive testing

packages/wasm/src/view-server.ts
turbocrime: ah, iirc something like the `case` statement was different, and perhaps breaking here. but that might have been due to a direct cast

turbocrime:
as mentioned in #844 it might be good for some of these return types to have a structure that satisfies pb types, such as `{ inner: Uint8Array }`. but i'm not sure how to achieve that in wasm-bindgen without exporting a name. might need a 'typescript custom section'

grod220:
> it might be good for some of these return types to have a structure that satisfies pb types, such as { inner: Uint8Array }

I don't think this is a desirable trait for a few reasons:

- A number of these structs have more than just an inner field. For instance, `PositionId` gets deseralized to `{ inner: Uint8Array,  altBech32m: string}`.
- The motivation behind these changes is to essentially make things on the typescript side more typesafe given JsonValue is an `any`. We achieve with Rust's `encode_to_vec()` + typescript's `fromBinary`. It has just as much simplicity as JsonValue but now have explicit `Uint8Array` types to pass around. Specifying the inner part, does not really add anything in terms of type-safety.
- Serialization is an implementation detail of connectrpc/@buf. In theory, they should be able to change it (say make it like `{ data: Uint8Array }` and it would not affect our application. In general, with two different systems communicating, one should not try to guess the best shape for the other's but rely on the serialization lib to handle that stuff.

packages/wasm/src/view-server.ts
grod220: I've not run into any errors with the simplified way. We should proceed and re-visit if we later discover if there is an issue.

packages/wasm/crate/src/dex.rs
```diff
@@ -1,36 +1,37 @@
...
 ///     position_state_value: `lp::position::State`
-/// Returns: `DenomMetadata`
+/// Returns: `Uint8Array representing a DenomMetadata`
 #[wasm_bindgen]
 pub fn get_lpnft_asset(
-    position_id_value: JsValue,
-    position_state_value: JsValue,
-) -> Result<JsValue, Error> {
+    position_id_value: &[u8],
+    position_state_value: &[u8],
```
Valentine1898, line 29: It makes sense to rename `position_id_value` -> `position_id` 
I added `_value` to mark `JsValue` parameters, but it didn't make sense either.

Valentine1898 (approved):
LGTM! 
A few suggestions for improving the docs

grod220:
Discussed offline. Seems like there a more JsValue's that can be squashed. Updating now!

PR #835, opened 2024-03-26
Fix account switcher bug

grod220:
1. Makes the hover color more visible (especially helpful in the extension, the web app has a higher contrast background anyway)
2. Increase width when right/left arrow are clicked so the number doesn't get cut off

Think this perhaps relates to some changes to this earlier in the month @jessepinho

Before:

https://github.com/penumbra-zone/web/assets/16624263/fbeb23dd-802e-47b8-b7b6-8f0e7f09ded8

After:

https://github.com/penumbra-zone/web/assets/16624263/3d54a982-ea9d-4f54-a307-71cdce5ffe76


Valentine1898 (approved):
LGTM!

PR #837, opened 2024-03-26
v0.70

TalDerei:
bump dep versions


TalDerei:
Changes will be consumed by https://github.com/penumbra-zone/web/pull/837 or version release

Issue #678, opened 2024-03-06
Use error codes from core to show more helpful UIs when transactions fail


jessepinho:
🚨 Blocked until https://github.com/penumbra-zone/penumbra/issues/3961 is addressed

Instead of [parsing](https://github.com/penumbra-zone/web/blob/v4.2.0/packages/query/src/queriers/tendermint.ts#L8-L13) and showing raw errors from Tendermint, we should show custom error messages depending on an error code we get from core.

i.e., instead of `ConnectError: [unknown] Tendermint: failed to deliver transaction: check_stateful failed: undelegation was prepared for next epoch 16 but the next epoch is 23`, we should show `Sorry, that transaction is now stale. Please try again.` (based on the error code)

hdevalence:
This will probably have to wait until post-mainnet, since we don't report error codes and changing that will require a lot of refactoring of the core code.

PR #838, opened 2024-03-26
Account for error where delegation was prepared for different epoch

jessepinho:
This should almost never happen. Basically, this should only happen when the user opens the web app in one epoch, then the epoch changes, then the user tries to delegate to a validator. In that case, we should at least show the error to the user rather than hanging indefinitely on "waiting for confirmation."

### DIFF
--- a/packages/query/src/queriers/tendermint.ts
+++ b/packages/query/src/queriers/tendermint.ts
@@ -5,16 +5,6 @@ import { TransactionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/
 import { Transaction } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 import type { TendermintQuerierInterface } from '@penumbra-zone/types/src/querier';
 
-// Should add more errors as we discover them
-const knownTendermintErrors = [
-  'proof did not verify',
-  'is not a valid field element',
-  'is not a valid SCT root',
-  'cannot claim unbonding tokens before the end epoch',
-  'undelegation was prepared for next epoch',
-  'delegation was prepared for epoch',
-];
-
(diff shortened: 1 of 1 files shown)


grod220 (dismissed):
At some point, we need to loop back and re-think about how we are catching these tendermint errors. In fact, I think the API is likely flawed if we continue having to parse our own errors like this.

grod220:
We've definitely been doing this incorrectly 😅 . Here's how it's handled in Rust:
https://github.com/penumbra-zone/penumbra/blob/aed1f8f607bfb4f9635d88f43b5ef8c5813036fd/crates/view/src/service.rs#L208-L218

Strange API, but if a non-zero response code is passed back, that indicates an error is present and viewable in the log.

```typescript
  async broadcastTx(tx: Transaction) {
    const params = tx.toBinary();
    // Note that "synchronous" here means "wait for the tx to be accepted by
    // the fullnode", not "wait for the tx to be included on chain.
    const { hash, log, code } = await this.client.broadcastTxSync({ params });

    // Non-zero response code indicates an error
    if (code !== 0n) {
      throw new Error(`Tendermint error ${code.toString()}: ${log}`);
    }
    return new TransactionId({ inner: hash });
  }
```

grod220 (approved):
Feel free to implement the change ⬆️  and merge

TalDerei (approved):

turbocrime (approved):

jessepinho:
@grod220 good catch! I've updated this with your suggestion and I'll merge when it's green.

Heads-up that I also created tickets for [core](https://github.com/penumbra-zone/penumbra/issues/3961) and [minifront](https://github.com/penumbra-zone/web/issues/678) to use error codes for showing more helpful error messages when a transaction fails. Henry was saying that should come after mainnet, though.

PR #839, opened 2024-03-26
Update core dependencies to v0.70.3

jessepinho:
## In this PR
- Update Cargo and NPM dependencies on core packages.
- Fix various places where we were previously using epochs, but are now using block heights.
  - This includes unbonding tokens. Unbonding tokens used to be named like `uunbonding_epoch_N_penumbravalid1abc123`. Now, unbonding tokens refer to a start _block height_, rather than a start _epoch_, so they're of the form `unbonding_start_at_N_penumbravalidabc123`. Also, they now have multiple denom units, so the display denom just starts with `unbonding_` rather than `uunbonding`.
- Fix a bug with how we stored epochs.
  - Previously, we stored new epochs only at epoch transitions. However, this meant that the first epoch (index 0) was missing! So, I updated the block processor to create an initial epoch if we're starting sync at block height 0.

### DIFF
--- a/apps/extension/.env
+++ b/apps/extension/.env
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=30
+IDB_VERSION=31
 USDC_ASSET_ID="reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
 MINIFRONT_URL=https://app.testnet.penumbra.zone/
 PENUMBRA_NODE_PD_URL=https://grpc.testnet.penumbra.zone/
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -145,13 +145,18 @@ export class BlockProcessor implements BlockProcessorInterface {
     const startHeight = fullSyncHeight ? fullSyncHeight + 1n : 0n;
     let latestKnownBlockHeight = await this.querier.tendermint.latestBlockHeight();
 
-    // In the `for` loop below, we only update validator infos once we've
-    // reached the latest known epoch. This means that, if a user is syncing for
-    // the first time, they could experience a broken UI until the latest known
-    // epoch is reached, since they may have delegation tokens but no validator
-    // info to go with them. So we'll update validator infos at the beginning of
-    // sync as well, and force the rest of sync to wait until it's done.
-    if (startHeight === 0n) await this.updateValidatorInfos(0n);
+    if (startHeight === 0n) {
+      // In the `for` loop below, we only update validator infos once we've
+      // reached the latest known epoch. This means that, if a user is syncing
+      // for the first time, they could experience a broken UI until the latest
+      // known epoch is reached, since they may have delegation tokens but no
+      // validator info to go with them. So we'll update validator infos at the
+      // beginning of sync as well, and force the rest of sync to wait until
+      // it's done.
+      await this.updateValidatorInfos(0n);
+
+      await this.indexedDb.addEpoch(0n);
+    }
 
     // this is an indefinite stream of the (compact) chain from the network
     // intended to run continuously
@@ -273,7 +278,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
       const isLastBlockOfEpoch = !!compactBlock.epochRoot;
       if (isLastBlockOfEpoch) {
-        void this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
+        await this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
       }
     }
   }
--- a/packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
@@ -75,11 +75,10 @@ describe('AssetMetadataById request handler', () => {
     expect(metadataByIdResponse.denomMetadata?.equals(metadataFromNode)).toBeTruthy();
   });
 
-  test('should return an empty response when metadata not found in idb and node', async () => {
+  test('should fail to get metadata when metadata not found in idb and node', async () => {
     mockIndexedDb.getAssetsMetadata?.mockResolvedValue(undefined);
     mockShieldedPool.assetMetadataById.mockResolvedValueOnce(undefined);
-    const response = new AssetMetadataByIdResponse(await assetMetadataById(request, mockCtx));
-    expect(response.equals(new AssetMetadataByIdResponse())).toBeTruthy();
+    await expect(assetMetadataById(request, mockCtx)).rejects.toThrow();
   });
 
   test('should fail if assetId is missing in request', async () => {
--- a/packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
+++ b/packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
@@ -75,10 +75,11 @@ describe('AssetMetadataById request handler', () => {
     expect(metadataByIdResponse.denomMetadata?.equals(metadataFromNode)).toBeTruthy();
   });
 
-  test('should fail to get metadata when metadata not found in idb and node', async () => {
+  test('should return an empty response when metadata not found in idb and node', async () => {
     mockIndexedDb.getAssetsMetadata?.mockResolvedValue(undefined);
     mockShieldedPool.assetMetadataById.mockResolvedValueOnce(undefined);
-    await expect(assetMetadataById(request, mockCtx)).rejects.toThrow();
+    const response = new AssetMetadataByIdResponse(await assetMetadataById(request, mockCtx));
+    expect(response.equals(new AssetMetadataByIdResponse())).toBeTruthy();
   });
 
   test('should fail if assetId is missing in request', async () => {
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -458,12 +458,14 @@ export class IndexedDb implements IndexedDbInterface {
     });
   }
 
-  async addEpoch(startHeight: bigint, index?: bigint): Promise<void> {
-    if (index === undefined) {
-      const cursor = await this.db.transaction('EPOCHS', 'readonly').store.openCursor(null, 'prev');
-      const previousEpoch = cursor?.value ? Epoch.fromJson(cursor.value) : undefined;
-      index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
-    }
+  /**
+   * Adds a new epoch with the given start height. Automatically sets the epoch
+   * index by finding the previous epoch index, and adding 1n.
+   */
+  async addEpoch(startHeight: bigint): Promise<void> {
+    const cursor = await this.db.transaction('EPOCHS', 'readonly').store.openCursor(null, 'prev');
+    const previousEpoch = cursor?.value ? Epoch.fromJson(cursor.value) : undefined;
+    const index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
 
     const newEpoch = {
       startHeight: startHeight.toString(),
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -489,8 +489,6 @@ export class IndexedDb implements IndexedDbInterface {
      */
     height: bigint,
   ): Promise<Epoch | undefined> {
-    let cursor = await this.db.transaction('EPOCHS', 'readonly').store.openCursor();
-
     let epoch: Epoch | undefined;
 
     /**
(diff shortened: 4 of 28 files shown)


packages/wasm/crate/Cargo.toml
```diff
@@ -15,21 +15,21 @@ default = ["console_error_panic_hook"]
...
-penumbra-tct           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-tct" }
-penumbra-transaction   = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-transaction", default-features = false }
+penumbra-asset         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-asset" }
+penumbra-compact-block = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-compact-block", default-features = false }
+penumbra-dex           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-dex", default-features = false }
+penumbra-fee           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-fee", default-features = false }
+penumbra-governance    = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-governance", default-features = false }
+penumbra-ibc           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.0", package = "penumbra-ibc", default-features = false }
```
grod220, line 23: These two packages do not seem to be used in our code, we can scrap

TalDerei: removed

apps/extension/.env
```diff
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=30
+IDB_VERSION=31
```
jessepinho, line 3: Needed for the update to unbonding tokens, which now have a different symbol

packages/wasm/crate/Cargo.toml
```diff
@@ -15,21 +15,19 @@ default = ["console_error_panic_hook"]
...
-penumbra-proto         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-proto", default-features = false }
-penumbra-sct           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-sct", default-features = false }
-penumbra-shielded-pool = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-shielded-pool", default-features = false }
-penumbra-stake         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-stake", default-features = false }
-penumbra-tct           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-tct" }
-penumbra-transaction   = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.68.3", package = "penumbra-transaction", default-features = false }
+penumbra-asset         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.70.2", package = "penumbra-asset" }
```
jessepinho, line 18: Just updating `v0.68.3` to `v0.70.2` on all of these

packages/query/src/block-processor.ts
```diff
@@ -273,7 +278,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
       const isLastBlockOfEpoch = !!compactBlock.epochRoot;
       if (isLastBlockOfEpoch) {
-        void this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
+        await this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
```
jessepinho, line 281: Needed to `await` due to an unlikely, but possible, race condition. When we save a new epoch to the database during an epoch transition, we look in the database for the _previous_ latest known epoch. We grab its epoch index, increment it by 1, then use that index for the new epoch we're saving to the database.

If that took a long time, for some reason, and in the meantime the block processor had already gotten to the next epoch, we could have two epochs with the same epoch index.

This change makes sure that epoch indexes increment in the correct order.

packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
```diff
@@ -75,10 +75,11 @@ describe('AssetMetadataById request handler', () => {
     expect(metadataByIdResponse.denomMetadata?.equals(metadataFromNode)).toBeTruthy();
   });
 
-  test('should fail to get metadata when metadata not found in idb and node', async () => {
+  test('should return an empty response when metadata not found in idb and node', async () => {
```
jessepinho, line 78: There's really no need to throw, since it's acceptable to return an empty response from this RPC method. So I changed the behavior.

packages/storage/src/indexed-db/index.ts
```diff
@@ -458,12 +458,14 @@ export class IndexedDb implements IndexedDbInterface {
     });
   }
 
-  async addEpoch(startHeight: bigint, index?: bigint): Promise<void> {
```
jessepinho, line 461: We're never actually using the `index` argument anywhere, so I just got rid of it.

packages/ui/components/ui/tx/view/undelegate.tsx
```diff
@@ -13,10 +13,6 @@ export const UndelegateComponent = ({ value }: { value: Undelegate }) => {
       label='Undelegate'
       visibleContent={
         <ActionDetails>
-          <ActionDetails.Row label='Epoch index'>
-            {value.startEpochIndex.toString()}
-          </ActionDetails.Row>
```
jessepinho, line 18: This is a little strange to me: the `startEpochIndex` property is deprecated, so we should no longer show it. Ideally, we'd show the unbonding start height instead – but that property doesn't exist on the `Undelegate` proto! So for now, we'll just hide this, and revisit this if/when core adds that property to the `Undelegate` action.

packages/ui/components/ui/tx/view/value/index.tsx
```diff
@@ -33,7 +33,7 @@ export const ValueViewComponent = ({
     const value = view.valueView.value;
     const metadata = view.valueView.value.metadata;
     const amount = value.amount ?? new Amount();
-    const exponent = getDisplayDenomExponent(metadata);
```
jessepinho, line 36: Reviewers: I could use some help here.

When a user clicks the Undelegate button, the transaction approval dialog opens. We try to fetch the metadata for the undelegation tokens they'll be getting, but that will always come back empty, since the unbonding token type is now specific to the block height at which unbonding starts. That is, the unbonding token name is `unbonding_start_at_N_penumbravalid1...`, where `N` is the block height that unbonding started at.

There will _never_ be token metadata for an unbonding token until _after_ the transaction has been committed to the chain, since the token type is created when the transaction is committed.

But that means that the undelegation tokens are shown as "Unknown asset" in the transaction approval dialog:
image

Perhaps we should generate the unbonding token metadata on the fly here? If so, could someone point me in the right direction for how to do that? (That said, I'd love to get this PR merged before handling that, since _all_ delegation is blocked until this PR is merged.)

jessepinho: Another wrinkle to all this: this issue, where the metadata request always returns an empty response for an undelegation token that doesn't exist yet, wasn't happening before when undelegation tokens were tied to epochs rather than block heights.

I don't know how that's possible. We should have had the same problem back then, right? That is:

1. We transition to a new epoch.
2. Someone goes to undelegate tokens from a validator.
3. Since no one else has undelegated tokens from that validator in this epoch yet, the metadata for `uunbonding_epoch_N_...` doesn't exist yet, so the metadata response is empty.
4. `getDisplayDenomExponent` throws, because it's being passed `undefined` metadata.

And yet, this wasn't happening before — no error was being thrown. It only started with this change, where unbonding tokens are tied to block height rather than epoch index.

I'm investigating this, but it's a bit concerning since I don't understand why it worked before!

packages/wasm/crate/src/planner.rs
```diff
@@ -300,30 +300,37 @@ pub async fn plan_transaction(
...
     for tpr::Delegate { amount, rate_data } in request.delegations {
+        let epoch = storage.get_latest_known_epoch().await?.unwrap();
         let amount = amount
             .ok_or_else(|| anyhow!("missing amount in delegation"))?
             .try_into()?;
         let rate_data: RateData = rate_data
             .ok_or_else(|| anyhow!("missing rate data in delegation"))?
             .try_into()?;
-        actions.push(rate_data.build_delegate(amount).into());
+        actions.push(rate_data.build_delegate(epoch.into(), amount).into());
```
jessepinho, line 310: The epoch in the rate data has been deprecated, so we have to pass it in now.

packages/query/src/block-processor.ts
```diff
@@ -145,13 +145,18 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      // In the `for` loop below, we only update validator infos once we've
+      // reached the latest known epoch. This means that, if a user is syncing
+      // for the first time, they could experience a broken UI until the latest
+      // known epoch is reached, since they may have delegation tokens but no
+      // validator info to go with them. So we'll update validator infos at the
+      // beginning of sync as well, and force the rest of sync to wait until
+      // it's done.
+      await this.updateValidatorInfos(0n);
+
+      await this.indexedDb.addEpoch(0n);
```
jessepinho, line 158: Previously, we were skipping saving the first epoch to the database. This fixes that bug.

packages/router/src/grpc/view-protocol-server/asset-metadata-by-id.test.ts
grod220: Response on other comment

packages/storage/src/indexed-db/index.ts
```diff
@@ -458,12 +458,14 @@ export class IndexedDb implements IndexedDbInterface {
...
-      index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
-    }
+  /**
+   * Adds a new epoch with the given start height. Automatically sets the epoch
+   * index by finding the previous epoch index, and adding 1n.
+   */
+  async addEpoch(startHeight: bigint): Promise<void> {
+    const cursor = await this.db.transaction('EPOCHS', 'readonly').store.openCursor(null, 'prev');
+    const previousEpoch = cursor?.value ? Epoch.fromJson(cursor.value) : undefined;
+    const index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
```
grod220, line 468: Can we use an async iterator here?

```typescript
  async addEpoch(startHeight: bigint): Promise<void> {
    for await (const cursor of this.db.transaction('EPOCHS').store) {
      const previousEpoch = Epoch.fromJson(cursor.value);
      const index = previousEpoch.index + 1n;
      const newEpoch = {
        startHeight: startHeight.toString(),
        index: index.toString(),
      };
      await this.u.update({
        table: 'EPOCHS',
        value: newEpoch,
      });
    }
  }
```

grod220: think getEpochByHeight() can be converted to one too

packages/ui/components/ui/tx/view/value/index.tsx
grod220: > Perhaps we should generate the unbonding token metadata on the fly here?

The `assetMetadataById` rpc method feels like a good place to generate metadata for this edge case so web apps have meaningful display data here.

> And yet, this wasn't happening before

Umm, no idea! 😅

packages/wasm/crate/src/storage.rs
```diff
@@ -305,4 +306,20 @@ impl IndexedDBStorage {
...
+        if let Some(cursor) = store
+            .open_cursor_with_direction(web_sys::IdbCursorDirection::Prev)?
+            .await?
+        {
+            // Return the first epoch index, since we're cursoring from the end.
+            let epoch: Epoch = serde_wasm_bindgen::from_value(cursor.value())?;
+            return Ok(Some(epoch));
+        }
+
+        Ok(None)
```
grod220, line 323: Maybe we can combine these?

```rust
    Ok(store
        .open_cursor_with_direction(web_sys::IdbCursorDirection::Prev)?
        .await?
        .map(|cursor| serde_wasm_bindgen::from_value(cursor.value()).ok())
        .flatten())
}
```

grod220 (approved):
Wow, some really domain specific knowledge here 😅 . Good job! Left some comments.

packages/ui/components/ui/tx/view/undelegate.tsx
turbocrime: can you leave the code here and just style it hidden or comment it out?

TalDerei:
Are we waiting for the 0.71.0 point release later today?

TalDerei (approved):

packages/storage/src/indexed-db/index.ts
jessepinho: Hm, why do it that way? I only want to grab the very last epoch from the table, not iterate over all of them. I guess I could just add a `break` or `return` statement at the end of the first run through the loop, but that feels a bit unidiomatic — people reading this code in the future will initially see it as a loop, and then realize the code is only running once.

jessepinho: (I did update `getEpochByHeight` to use this syntax though, per your suggestion.)

packages/wasm/crate/src/storage.rs
jessepinho: Ahh, good call. Done.

packages/storage/src/indexed-db/index.ts
grod220: Ah, didn't realize that was happening. Can you add a code comment? :)

grod220: Oh wait there is one already lol whoops

PR #841, opened 2024-03-27
Update all non-crate deps

grod220:
A follow up to https://github.com/penumbra-zone/web/pull/839 in preparation for release


apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -87,7 +87,7 @@ const postRequest = () => {
       window.origin,
     );
     request.promise
-      .catch(e => connection.reject(e))
+      .catch((e: unknown) => connection.reject(e))
```
grod220, line 90: New eslint rule: https://typescript-eslint.io/rules/use-unknown-in-catch-callback-variable/

Catch needs to specify that it's an unknown

packages/ui/package.json
```diff
@@ -33,37 +33,37 @@
...
-    "framer-motion": "^11.0.8",
+    "framer-motion": "^11.0.22",
     "humanize-duration": "^3.31.0",
-    "lucide-react": "^0.354.0",
+    "lucide-react": "^0.363.0",
     "react-dom": "^18.2.0",
     "react-json-view-lite": "^1.3.0",
     "react-loader-spinner": "^6.1.6",
     "react-router-dom": "^6.22.3",
     "sonner": "1.4.3",
```
grod220, line 43: Sonner has an update, but the latest version has a bug: https://github.com/emilkowalski/sonner/issues/378. We need to wait a bit one this one.

packages/client/src/prax.ts
```diff
@@ -11,7 +11,7 @@ import { createChannelTransport } from '@penumbra-zone/transport-dom/src/create'
 import { PenumbraSymbol } from './global';
 import { jsonOptions } from '@penumbra-zone/types/src/registry';
 
-const prax_id = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe' as const;
+const prax_id = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe';
```
grod220, line 14: Eslint rule that is was an unnecessary const

packages/eslint-config-custom/index.js
```diff
@@ -59,6 +59,7 @@ module.exports = {
         ],
       },
     ],
+    '@typescript-eslint/restrict-template-expressions': ['error', { allowNumber: true }],
```
grod220, line 62: New eslint rule: https://typescript-eslint.io/rules/restrict-template-expressions/

I think this makes sense for everything except for numbers. I think we all know that putting a number into a string template stringifies it and there isn't any crazy things like `[object Object]` to worry about.

turbo.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "globalDependencies": ["**/.env.*local"],
-  "globalEnv": ["NODE_ENV"],
+  "globalEnv": ["NODE_ENV", "MODE"],
```
grod220, line 4: New turbo eslint rule that requires us to put env variables here

TalDerei:
LGTM

TalDerei (approved):

PR #844, opened 2024-03-27
suggested wasm updates

turbocrime:
many signatures like this could be updated

edit: okay, now this contains fixes and fn rename `decodeSctRoot` -> `validSctRoot`

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -384,9 +384,7 @@ export class BlockProcessor implements BlockProcessorInterface {
   // This is expensive to do every block, so should only be done in development.
   // @ts-expect-error Only used ad-hoc in dev
   private async assertRootValid(blockHeight: bigint): Promise<void> {
-    const sourceOfTruth: Uint8Array = await this.querier.cnidarium.keyValue(
-      `sct/anchor/${blockHeight}`,
-    );
+    const sourceOfTruth = await this.querier.cnidarium.keyValue(`sct/anchor/${blockHeight}`);
     const inMemoryRoot = this.viewServer.getSctRoot();
     if (!validSctRoot({ inner: sourceOfTruth }))
       throw new Error(`Block height: ${blockHeight}. Remote cnidarium provided invalid SCT root.`);
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -1,7 +1,7 @@
 import { RootQuerier } from './root-querier';
 import { sha256Hash } from '@penumbra-zone/crypto-web/src/sha256';
 import { computePositionId, getLpNftMetadata } from '@penumbra-zone/wasm/src/dex';
-import { decodeSctRoot } from '@penumbra-zone/wasm/src/sct';
+import { validSctRoot as validSctRoot } from '@penumbra-zone/wasm/src/sct';
 
 import {
   getExchangeRateFromValidatorInfoResponse,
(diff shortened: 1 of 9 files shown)


grod220 (approved):
makes sense!

turbocrime:
it would be nice to directly emit an object that satisfies `{ inner: UInt8Array }` without explicitly naming a type in several places, but i'm not sure how to achieve that

turbocrime:
what is the expected use/behavior of decodeSctRoot? it seems to return the same byte string, and i suspect it isn't actually a transaction id despite being documented as such

turbocrime:
corrected parameter type, renamed function.

packages/wasm/crate/src/sct.rs
```diff
@@ -0,0 +1,12 @@
...
+use penumbra_proto::DomainType;
+
+/// check validity of SCT root by attempting to parse it
+/// Arguments:
+///     root: `Uint8Array representing sct root hash`
+/// Returns: boolean
+#[wasm_bindgen]
+pub fn valid_sct_root(root: Vec<u8>) -> bool {
+    return penumbra_tct::Root::decode(root.as_slice()).is_ok();
+}
```
turbocrime, line 12: omitted panic hook for this fn because everything should be caught by `.is_ok()`

turbocrime:
this is targeting your pr so feel free to merge when you're moving forward with it

packages/query/src/block-processor.ts
```diff
@@ -386,12 +386,12 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      throw new Error(
-        `Block height: ${blockHeight}. Wasm root does not match remote source of truth. Programmer error.`,
-      );
-    }
+    if (!validSctRoot({ inner: sourceOfTruth }))
+      throw new Error(`Block height: ${blockHeight}. Remote cnidarium provided invalid SCT root.`);
+    if (!validSctRoot(inMemoryRoot))
+      throw new Error(`Block height: ${blockHeight}. Invalid local SCT root.`);
+    if (!inMemoryRoot.equals({ inner: sourceOfTruth }))
+      throw new Error(`Block height: ${blockHeight}. Local SCT root does not match remote.`);
```
grod220, line 394: We don't so much want to validate the root, but ensure our locally stored one is the same as the one the remote rpc says. There's also been some breaking changes since we've used this. Going to make some updates here.

packages/wasm/crate/src/keys.rs
```diff
@@ -21,10 +20,7 @@ use crate::utils;
 /// function will additionally require downloading the proving key parameter `.bin`
 /// file for each key type.
 #[wasm_bindgen]
-pub fn load_proving_key(parameters: JsValue, key_type: &str) -> WasmResult<()> {
-    // Deserialize JsValue into Vec<u8>.
-    let parameters_bytes: Vec<u8> = Uint8Array::new(&parameters).to_vec();
-
+pub fn load_proving_key(key: Vec<u8>, key_type: &str) -> WasmResult<()> {
```
grod220, line 23: This is definitely a good example of where bytes are needed anyway. It doesn't make sense to serialize through JsValue. However, I don't think this fits in all cases. Will comment on https://github.com/penumbra-zone/web/issues/845 as I make my way through it.

grod220 (commented):
Copying stuff over to my branch as it's ahead locally and can't merge cleanly anymore.  Will close this one.

PR #846, opened 2024-03-28
842 use bufbuild type to storetransmit fvk instead of bech32 string

### LINKED ISSUE #842: Use bufbuild type to store/transmit FVK instead of bech32 string
Instead of using bech32 string we should use bufbuild type https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.keys.v1#penumbra.core.keys.v1.FullViewingKey

Valentine1898:
close #842

### DIFF
--- a/apps/extension/src/state/connected-sites.test.ts
+++ b/apps/extension/src/state/connected-sites.test.ts
@@ -19,6 +19,7 @@ describe('Connected Sites Slice', () => {
 
   beforeEach(() => {
     localStorage = mockLocalExtStorage();
+    //
     useStore = create<AllSlices>()(initializeStore(mockSessionExtStorage(), localStorage));
   });
 
--- a/apps/extension/src/state/connected-sites.test.ts
+++ b/apps/extension/src/state/connected-sites.test.ts
@@ -1,17 +1,14 @@
 import { beforeEach, describe, expect, test } from 'vitest';
 import { create, StoreApi, UseBoundStore } from 'zustand';
 import { AllSlices, initializeStore } from '.';
-import {
-  localDefaults,
-  LocalStorageState,
-  OriginRecord,
-} from '@penumbra-zone/storage/src/chrome/local';
+import { localDefaults } from '@penumbra-zone/storage/src/chrome/local';
 import { ExtensionStorage } from '@penumbra-zone/storage/src/chrome/base';
 import {
   mockLocalExtStorage,
   mockSessionExtStorage,
 } from '@penumbra-zone/storage/src/chrome/test-utils/mock';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
+import { LocalStorageState, OriginRecord } from '@penumbra-zone/types/src/local-storage';
 
 describe('Connected Sites Slice', () => {
   let useStore: UseBoundStore<StoreApi<AllSlices>>;
--- a/apps/extension/src/state/connected-sites.test.ts
+++ b/apps/extension/src/state/connected-sites.test.ts
@@ -16,7 +16,6 @@ describe('Connected Sites Slice', () => {
 
   beforeEach(() => {
     localStorage = mockLocalExtStorage();
-    //
     useStore = create<AllSlices>()(initializeStore(mockSessionExtStorage(), localStorage));
   });
 
--- a/apps/extension/src/state/connected-sites.test.ts
+++ b/apps/extension/src/state/connected-sites.test.ts
@@ -2,13 +2,13 @@ import { beforeEach, describe, expect, test } from 'vitest';
 import { create, StoreApi, UseBoundStore } from 'zustand';
 import { AllSlices, initializeStore } from '.';
 import { localDefaults } from '@penumbra-zone/storage/src/chrome/local';
+import { LocalStorageState, OriginRecord } from '@penumbra-zone/storage/src/chrome/types';
 import { ExtensionStorage } from '@penumbra-zone/storage/src/chrome/base';
 import {
   mockLocalExtStorage,
   mockSessionExtStorage,
 } from '@penumbra-zone/storage/src/chrome/test-utils/mock';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
-import { LocalStorageState, OriginRecord } from '@penumbra-zone/types/src/local-storage';
 
 describe('Connected Sites Slice', () => {
   let useStore: UseBoundStore<StoreApi<AllSlices>>;
--- a/apps/extension/src/state/wallets.ts
+++ b/apps/extension/src/state/wallets.ts
@@ -91,7 +91,7 @@ export const addrByIndexSelector =
     const active = getActiveWallet(state);
     if (!active) throw new Error('No active wallet');
 
-    let fullViewingKey = FullViewingKey.fromJsonString(active.fullViewingKey);
+    const fullViewingKey = FullViewingKey.fromJsonString(active.fullViewingKey);
     return ephemeral
       ? getEphemeralByIndex(fullViewingKey, index)
       : getAddressByIndex(fullViewingKey, index);
--- a/apps/extension/src/state/wallets.ts
+++ b/apps/extension/src/state/wallets.ts
@@ -8,13 +8,13 @@ import {
 } from '@penumbra-zone/wasm/src/keys';
 import { Key } from '@penumbra-zone/crypto-web/src/encryption';
 import { ExtensionStorage } from '@penumbra-zone/storage/src/chrome/base';
-import { LocalStorageState } from '@penumbra-zone/storage/src/chrome/local';
 import {
   Address,
   FullViewingKey,
   SpendKey,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { Wallet, WalletCreate } from '@penumbra-zone/types/src/wallet';
+import { LocalStorageState } from '@penumbra-zone/types/src/local-storage';
 
 export interface WalletsSlice {
   all: Wallet[];
--- a/apps/extension/src/state/wallets.ts
+++ b/apps/extension/src/state/wallets.ts
@@ -8,13 +8,13 @@ import {
 } from '@penumbra-zone/wasm/src/keys';
 import { Key } from '@penumbra-zone/crypto-web/src/encryption';
 import { ExtensionStorage } from '@penumbra-zone/storage/src/chrome/base';
+import { LocalStorageState } from '@penumbra-zone/storage/src/chrome/types';
 import {
   Address,
   FullViewingKey,
   SpendKey,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { Wallet, WalletCreate } from '@penumbra-zone/types/src/wallet';
-import { LocalStorageState } from '@penumbra-zone/types/src/local-storage';
 
 export interface WalletsSlice {
   all: Wallet[];
--- a/apps/extension/src/state/wallets.ts
+++ b/apps/extension/src/state/wallets.ts
@@ -40,7 +40,12 @@ export const createWalletsSlice =
         const key = await Key.fromJson(passwordKey);
         const encryptedSeedPhrase = await key.seal(seedPhraseStr);
         const walletId = getWalletId(fullViewingKey);
-        const newWallet = new Wallet(label, walletId, fullViewingKey, { encryptedSeedPhrase });
+        const newWallet = new Wallet(
+          label,
+          walletId.toJsonString(),
+          fullViewingKey.toJsonString(),
+          { encryptedSeedPhrase },
+        );
 
         set(state => {
           state.wallets.all.unshift(newWallet);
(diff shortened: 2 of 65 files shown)


packages/types/src/wallet.ts
```diff
@@ -8,8 +13,8 @@ export interface WalletCreate {
 // Stored in chrome.local.storage
 export interface WalletJson {
   label: string;
-  id: string;
-  fullViewingKey: string;
+  id: Stringified<WalletId>;
+  fullViewingKey: Stringified<FullViewingKey>;
   custody: { encryptedSeedPhrase: BoxJson };
 }
```
Valentine1898, line 19: `WalletJson` is stored in **chrome.local.storage** and `fullViewingKey` and `id` fields must be strings because we can only store strings in local storage. 
If before it was just bech32 strings, now it should be `Stringified<FullViewingKey>` and `Stringified<WalletId>`

packages/types/src/wallet.ts
```diff
@@ -23,22 +28,27 @@ export type Custody = HotWallet; // Later on, could have different types (like l
 export class Wallet {
   constructor(
     readonly label: string,
-    readonly id: string,
-    readonly fullViewingKey: string,
+    readonly id: WalletId,
+    readonly fullViewingKey: FullViewingKey,
```
Valentine1898, line 32: Wallet is stored in zustand memory, we can use here bufbuild types `FullViewingKey` and `WalletId`

packages/types/src/wallet.ts
```diff
@@ -23,22 +28,27 @@ export type Custody = HotWallet; // Later on, could have different types (like l
...
-      encryptedSeedPhrase: Box.fromJson(obj.custody.encryptedSeedPhrase),
-    });
+    return new Wallet(
+      obj.label,
+      WalletId.fromJsonString(obj.id),
+      FullViewingKey.fromJsonString(obj.fullViewingKey),
+      {
+        encryptedSeedPhrase: Box.fromJson(obj.custody.encryptedSeedPhrase),
+      },
+    );
```
Valentine1898, line 44: When converting `WalletJson` -> `Wallet`, we must use `fromJsonString()`

apps/extension/src/state/connected-sites.test.ts
```diff
@@ -19,6 +19,7 @@ describe('Connected Sites Slice', () => {
 
   beforeEach(() => {
     localStorage = mockLocalExtStorage();
+    //
     useStore = create<AllSlices>()(initializeStore(mockSessionExtStorage(), localStorage));
```
Valentine1898, line 23: Here and in a number of other tests I get an error 
```typescript
TS2322: Type 'UseBoundStore<WithImmer<StoreApi<AllSlices>>>' is not assignable to type 'UseBoundStore<StoreApi<AllSlices>>'.
``` 

image

apps/extension/src/state/wallets.ts
```diff
@@ -9,15 +9,19 @@ import {
...
 import { Wallet, WalletCreate } from '@penumbra-zone/types/src/wallet';
 
 export interface WalletsSlice {
   all: Wallet[];
   addWallet: (toAdd: WalletCreate) => Promise<Wallet>;
   getSeedPhrase: () => Promise<string[]>;
-  getSpendKey: () => Promise<string>;
-  getFullViewingKey: () => Promise<string>;
+  getSpendKey: () => Promise<SpendKey>;
+  getFullViewingKey: () => Promise<FullViewingKey>;
```
grod220, line 24: Given we are doing this pattern a bunch of times:

```typescript
bech32FullViewingKey(await getFullViewingKey())
```

I think we can store them as strings in the slice

```typescript
getFullViewingKey: () => Promise<FullViewingKey>;
```

apps/extension/src/state/connected-sites.test.ts
grod220: So this is a downside of zustand+immer 😞 . It's not allowed to store bufbuild classes directly. It seems to be a limitation because immer cannot properly do a safe copy of it. So we likely need to be using `WalletJson` here. But does this invalidate this whole task? Maybe not, we still should be storing FullViewingKey.toJson() in storage instead of a string.

packages/bech32/src/full-viewing-key.ts
```diff
@@ -0,0 +1,10 @@
+import { bech32m } from 'bech32';
+import { PENUMBRA_BECH32_FVK_LENGTH, PENUMBRA_BECH32_FVK_PREFIX } from './penumbra-bech32';
+
+export const bech32FullViewingKey = (fvk: { inner: Uint8Array }): string =>
```
grod220, line 4: Think this would be good:

```typescript
const bech32FullViewingKey = (fvk: FullViewingKey): string {}
```

packages/bech32/src/full-viewing-key.ts
```diff
@@ -0,0 +1,10 @@
...
+  bech32m.encode(PENUMBRA_BECH32_FVK_PREFIX, bech32m.toWords(fvk.inner));
+
+export const bech32ToFullViewingKey = (fvk: string) => {
+  const decodeAddress = bech32m.decode(fvk, PENUMBRA_BECH32_FVK_LENGTH);
+  return new Uint8Array(bech32m.fromWords(decodeAddress.words));
+};
```
grod220, line 10: Think this should return a `new FullViewingKey()`

packages/storage/src/chrome/local.ts
```diff
@@ -35,10 +42,36 @@ export const localDefaults: LocalStorageState = {
...
+        const walletId = new WalletId({
+          inner: bech32ToWalletId(id),
+        });
+        return {
+          fullViewingKey: fvk.toJsonString(),
+          id: walletId.toJsonString(),
+          label,
+          custody,
+        };
+      }),
```
grod220, line 67: Can you move this to a new v1-migration.ts file?

```typescript
const migrations: Migration = {
  wallets: {
    [LocalStorageVersion.V1]: v1WalletMigration
  },
};
```

Also, can you write tests for this specific migration as well?

packages/wasm/crate/src/build.rs
```diff
@@ -1,28 +1,27 @@
...
-///     full_viewing_key: `bech32m String`,
+///     full_viewing_key: `Full`,
 ///     witness_data: `WitnessData``
 /// Returns: `Action`
 #[wasm_bindgen]
 pub fn build_action(
     transaction_plan: JsValue,
     action_plan: JsValue,
-    full_viewing_key: &str,
+    full_viewing_key: JsValue,
```
grod220, line 24: Somewhat related to https://github.com/penumbra-zone/web/pull/844, but we can do something like this so we get types:

```rust
pub fn build_action(
    transaction_plan: JsValue,
    action_plan: JsValue,
    full_viewing_key: &[u8],
    witness_data: JsValue,
) -> WasmResult<JsValue> {
    let full_viewing_key: FullViewingKey = FullViewingKey::decode(full_viewing_key)?;
```

and then in typescript we pass `fullViewingKey.inner`

Valentine1898: Why is it better? Do we want to get rid of `toJson()` serialization and `serde_wasm_bindgen`? 
Should we do this in all cases when handling `{inner: Uint8Array}` types?

grod220: I'm working through these cases in https://github.com/penumbra-zone/web/pull/834. I go back and forth honestly. Perhaps it's not better than JsValue. We still pass an opaque type that needs deserializing on the other side. Maybe ignore this for now.

Valentine1898:
The V1 -> V2 migration was successful, but now that we have V2 in local storage state we get an error 
```
Cannot read properties of undefined (reading 'V1')
``` 
image

grod220:
Wherever that bug is, sounds like a good candidate for testing 😅

packages/wasm/crate/src/build.rs
grod220: Ok, here's the idea with this:
- wasm: https://github.com/penumbra-zone/web/pull/834/files#diff-8379276b019dd49da4c1b561332a55996d838efd25a8e02609f6ed40625e94d1R103-R108
- typescript: https://github.com/penumbra-zone/web/pull/834/files#diff-be70ea133ec02a72fd2fc4cd0b3c2782d9ce8d074a426f5490e9abf78e49c29bR14-R17

`JsValue` can be _anything_. And it gets typed to an `any`. However, in some cases, we can encode/decode through binary. If we pass an Uint8Array straight to wasm and receive one back, the typescript signature will be strongly enforced. It increases typesafety. It won't work in every case, but for pb types, it often can.

apps/extension/src/state/wallets.ts
Valentine1898: > Given we are doing this pattern a bunch of times:
Only once, so I left it as is 
```typescript
getFullViewingKey: () => Promise<FullViewingKey>;
```

packages/wasm/crate/src/keys.rs
```diff
@@ -44,122 +45,85 @@ pub fn load_proving_key(parameters: JsValue, key_type: &str) -> WasmResult<()> {
...
 #[wasm_bindgen]
-pub fn get_wallet_id(full_viewing_key: &str) -> WasmResult<String> {
+pub fn get_wallet_id(full_viewing_key: &[u8]) -> WasmResult<JsValue> {
     utils::set_panic_hook();
 
-    let fvk = FullViewingKey::from_str(full_viewing_key)?;
-    Ok(fvk.wallet_id().to_string())
+    let fvk: FullViewingKey = FullViewingKey::decode(full_viewing_key)?;
+    let result = serde_wasm_bindgen::to_value(&fvk.wallet_id())?;
+    Ok(result)
```
Valentine1898, line 82: @grod220 
`WalletId` has no method `.encode_to_vec()`

packages/storage/src/chrome/local.ts
```diff
@@ -1,25 +1,7 @@
...
-  date: number;
-}
-
-export interface LocalStorageState {
-  wallets: WalletJson[];
-  grpcEndpoint: string;
-  passwordKeyPrint?: KeyPrintJson;
-  fullSyncHeight: number;
-  knownSites: OriginRecord[];
-}
```
Valentine1898, line 22: I moved the types that were defined `@penumbra-zone/storage/src/chrome/local` to `@penumbra-zone/types/src/local-storage`. 
This was to avoid a cyclic import which created a bug with the error `Cannot read properties of undefined (reading 'V1')` .   
This also created a lot of import changes in this PR.

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -94,7 +99,9 @@ describe('IndexedDb', () => {
         chainId: 'test',
         accountAddr: 'penumbra123xyz',
         dbVersion: 2,
-        walletId: `walletid${Math.random()}`,
+        walletId: new WalletId({
+          inner: Uint8Array.from({ length: 8 }, () => Math.floor(Math.random() * 256)),
+        }),
```
Valentine1898, line 104: This test is actually incorrect, if we want to test idb version update, we should not generate a random dbName, only increase the version.
But the thing is, if we fix this test, it causes all storage tests to hang and timeout error. I will open a separate PR for this fix (since this PR is big without it)

packages/wasm/crate/src/keys.rs
```diff
@@ -44,122 +45,85 @@ pub fn load_proving_key(parameters: JsValue, key_type: &str) -> WasmResult<()> {
...
-/// Returns: `String`
-#[wasm_bindgen]
-pub fn get_short_address_by_index(full_viewing_key: &str, index: u32) -> WasmResult<JsValue> {
-    utils::set_panic_hook();
-
-    let fvk = FullViewingKey::from_str(full_viewing_key)?;
-    let (address, _dtk) = fvk.incoming().payment_address(index.into());
-    let short_address = address.display_short_form();
-    Ok(JsValue::from_str(&short_address))
-}
```
Valentine1898, line 165: Not used

packages/wasm/crate/src/keys.rs
grod220: Looks like WalletId is missing a trait in the core repo. I'll open a PR to fix there. In the meantime, let's do:

```rust
pub fn get_wallet_id(full_viewing_key: JsValue) -> WasmResult<Vec<u8>> {
    utils::set_panic_hook();

    let fvk: FullViewingKey = serde_wasm_bindgen::from_value(full_viewing_key)?;
    // Can do `fvk.wallet_id().encode_to_vec()` when Domain impl added to WalletId in core
    let wallet_id_proto = WalletId::from(fvk.wallet_id());
    Ok(wallet_id_proto.encode_to_vec())
}
```

Valentine1898: `penumbra_proto::core::keys::v1::WalletId` is also missing a trait

grod220 (approved):
Really well done! This wasn't easy 😅 . Before you merge, please do some extensive testing with all user flows (migrating from v1/swapping/staking/send/etc) to ensure there aren't any bugs we missed.

Valentine1898:
I'll just point out 
Migration works only in case of version increase (V1-->V2). 
If you try to rollback extension with V2 version of local storage to V1 version you will just break everything. 

I don't think users will have a case when a version is decreased, and developers should just consider that this can happen when switching between branches.

grod220:
> If you try to rollback extension with V2 version of local storage to V1 version you will just break everything.

Ah, good point. After you merge, please make a note in #browser-extension in discord.

PR #848, opened 2024-03-28
fix account switcher and address component

turbocrime:
this was causing rendering errors and strange behavior in parent components. it should now render more reliably

for some reason it's showing henry as 'collaborating', but that's not quite true. this came out of some work to use this component and make its style configurable for light theme, in my ibc style branch

### DIFF
--- a/apps/minifront/src/components/dashboard/assets-table.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table.tsx
@@ -41,7 +41,7 @@ export default function AssetsTable() {
                 </h2>
               </div>
 
-              <div className='max-w-72 truncate'>
+              <div className='max-w-72 text-muted-foreground'>
                 <AddressComponent address={a.address} />
               </div>
             </div>
--- a/packages/ui/components/ui/address-component.tsx
+++ b/packages/ui/components/ui/address-component.tsx
@@ -1,23 +1,27 @@
 import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { bech32Address } from '@penumbra-zone/bech32/src/address';
+import { cn } from '../../lib/utils';
 
 interface AddressComponentProps {
   address: Address;
   ephemeral?: boolean;
+  ephemeralClassName?: string;
 }
 
 /**
  * Displays an address. The address is truncated to the prefix plus 24
  * characters, and rendered in color when it is ephemeral.
  */
-export const AddressComponent = ({ address, ephemeral }: AddressComponentProps) => {
+export const AddressComponent = ({
+  address,
+  ephemeral,
+  ephemeralClassName = 'text-rust',
+}: AddressComponentProps) => {
   const bech32Addr = bech32Address(address);
 
   return (
-    <span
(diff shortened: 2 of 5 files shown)


packages/ui/components/ui/address-component.tsx
```diff
@@ -1,23 +1,27 @@
...
+  ephemeral,
+  ephemeralClassName = 'text-rust',
+}: AddressComponentProps) => {
   const bech32Addr = bech32Address(address);
 
   return (
-    <span
-      className={'font-mono' + (ephemeral ? ' text-[#8D5728]' : ' text-muted-foreground truncate')}
-    >
+    <div className={cn('font-mono', 'truncate w-0 min-w-full', ephemeral && ephemeralClassName)}>
```
turbocrime, line 23: now inherits text color from parent, and 'alternate' color is configurable

packages/ui/components/ui/select-account.tsx
```diff
@@ -33,52 +33,43 @@ export const SelectAccount = ({ getAddrByIndex }: SelectAccountProps) => {
...
-              </div>
-
-              <p className='truncate text-sm'>
-                <AddressComponent address={address} ephemeral={ephemeral} />
-              </p>
-            </div>
-            <CopyToClipboardIconButton text={bech32Address(address)} />
+        <div className='mt-4 flex items-center justify-between gap-1 rounded-lg border bg-background px-3 py-4'>
+          <AddressIcon address={address} size={24} />
+          <div className='w-full text-muted-foreground'>
```
turbocrime, line 42: the old default text color `text-muted-foreground` is applied at existing use locations

apps/minifront/src/components/dashboard/assets-table.tsx
```diff
@@ -41,7 +41,7 @@ export default function AssetsTable() {
                 </h2>
               </div>
 
-              <div className='max-w-72 truncate'>
+              <div className='max-w-72 text-muted-foreground'>
                 <AddressComponent address={a.address} />
               </div>
             </div>
```
grod220, line 47: We appear to have lost the address part, before:

Screenshot 2024-03-28 at 10 28 55 AM

After:
Screenshot 2024-03-28 at 10 28 42 AM

packages/ui/components/ui/select-account.tsx
```diff
@@ -33,52 +33,43 @@ export const SelectAccount = ({ getAddrByIndex }: SelectAccountProps) => {
...
-                <AddressIcon address={address} size={24} />
-              </div>
-
-              <p className='truncate text-sm'>
-                <AddressComponent address={address} ephemeral={ephemeral} />
-              </p>
-            </div>
-            <CopyToClipboardIconButton text={bech32Address(address)} />
+        <div className='mt-4 flex items-center justify-between gap-1 rounded-lg border bg-background px-3 py-4'>
+          <AddressIcon address={address} size={24} />
```
grod220, line 41: The identicon looks to have lost some roundness here: 
Screenshot 2024-03-28 at 10 31 08 AM

turbocrime: this is a failure in identicon, have a fix that isn't PR'd yet

PR #850, opened 2024-03-28
ibc page

turbocrime:
ibc page design depending on layout changes

some common components now configurable/styleable

Screenshot 2024-03-28 at 00 15 50


grod220:
Fixing these conflicts and going to get this merged in so we have a base to work off of

Issue #432, opened 2024-01-31
Move toward a "black box" component library model for `packages/ui`


jessepinho:
At the moment, our `@penumbra-zone/ui` package contains components whose appearance can be highly customized by the devs using them, particularly through the use of the `className` prop.

Ideally, a component library should be much less customizable. In the same way that an API exposes specific endpoints rather than allow any client to make SQL queries directly against its database, a component library should have a specific set of "endpoints" (in the form of component props) that allow them to be customized in _pre-defined_, rather than _ad-hoc_, ways.

The idea is that **developers shouldn't generally be making design decisions** when working on a given UI ticket, as that leads to poor UX decision-making. Designers are the ones responsible for making sure that an app has visual unity, and that each user flow is optimized to make it as painless and intuitive as possible. Developers, meanwhile, are generally focused on making a feature work, even if it comes at the expense of those two design concerns. Thus, if the design of an app is led by developers, its visual unity will gradually disintegrate over time, as each developer tweaks designs to their own tastes. And, common workflows will often suffer without a designer's research and expertise.

Thus, I am suggesting the following:
1. Develop a component library + design system in e.g., Figma.
    - This should include enough components that nearly all UI tickets can be completed by combining existing components, rather than creating new ones.
    - The system should be extremely specific. For example:
      - It should specify base layout and spacing units (such as 4px for spacing and e.g., a 12-column grid).
      - Each component should have its _internal_ spatial relationships defined. (External spatial relationships are always determined by the container, not be the component itself. e.g., a button should never have external margins, as those should be set by their containing elements to avoid CSS override hell; but it is permitted to have internal padding.)
    - The system should specify shared values like border radii, brand colors, typefaces, etc. along with what each is used for.
2. Implement this component library in `@penumbra-zone/ui`.
    - Customization of these components, beyond what is specified in the Figma designs, should be disabled. For example, if Figma calls for a button with several "variants" like `primary`, `secondary`, and `icon` — each of which have different background colors and sizes — these can be specified via the `<Button />` component's `variant` prop, _NOT_ via a `className` prop that allows customization of colors and sizes.
[...]

PR #852, opened 2024-03-28
styleable valueview

turbocrime:
removed pill element because it was just a div with round corners, and only used by valueview

valueview accepts a className prop


grod220 (commented):
Accepting `className` as a prop feels like it's the tailwind/shadcnui direction, but isn't so compatible with: https://github.com/penumbra-zone/web/issues/432. Maybe @jessepinho best if you weigh in.

Side note, would be great to fix the text alignment while we are here:
Screenshot 2024-03-28 at 11 10 49 AM

turbocrime:
yes it's incorrect, imo, should be using cva, but i would rather take this flexible approach for minimal change as my objective is to complete ibc style

jessepinho:
The `Pill` is also being used in #858 so I'd prefer we don't get rid of it. It's sort of a foundational component we can use in a lot of contexts.

Also to Gabe's point re #432, I'm strongly opposed to having `className` props on components that are used in a bunch of places. It makes maintainability a nightmare. Instead, if there are several ways you want to be able to represent `ValueView`s via `ValueViewComponent`, add e.g., a `variant` prop so that `ValueViewComponent` _itself_ can decide how to render for various use cases, rather than a parent component reaching in and changing its internal appearance.

Issue #830, opened 2024-03-25
Swap Claim UX Improvements 


TalDerei:
Maybe we should separately consider adding a minor delay between swap and swap claim toasts, or slight offset in the UI? I'm not sure this is the most intuitive UX, because the view changes so quickly between transactions that I find myself missing the display of the first transaction.

<img width="442" alt="316391165-a0f5ff09-c7f7-45c4-a2fd-6360d579f4fe" src="https://github.com/penumbra-zone/web/assets/70081547/8c299f1e-0dc0-434d-aacd-7bc5f8381b4d">

PR #858, opened 2024-03-28
Show nicer view of swap outputs 

### LINKED ISSUE #424: Make tx details page for swap + swap claim nicer

jessepinho:
🚨 ~ 331 line changes are from tests

Per #424, we wanted to show a more user-friendly view of swaps and swap claims. This has taken a lot of work in past PRs on the core and WASM side of things, so that we have the necessary data to render everything.

This PR ties it all together by rendering a swap's outputs (and unfilled amount) on a swap page, and rendering the claimed amounts on a swap claim page.

## Swap page
The swap page is a little complex, because it has to handle the following cases:

### One swap input exists but no swap outputs exist.
This happens we're viewing the swap inside the transaction approval dialog, so the transaction hasn't yet been built and thus has no outputs yet. In this case, we show e.g., "100UM -> TestUSD" — that is, we just show the _symbol_ for test_usd, but not the amount, since we don't know the amount yet.

image

### One swap input exists and one swap output exists.
This is the typical case — there's enough liquidity to cover the entire swap, so the user does a one-way swap from asset 1 to asset 2.

image

### One swap input exists and two swap outputs exist.
This happens when there isn't enough liquidity to cover the entire swap (or, even, any of it). In this case, a user does a one-way swap from asset 1 to asset 2 and gets two outputs: the amount of asset 2 that the liquidity was able to cover, and the amount of asset 1 that the liquidity was not able to cover (sort of like change). This is shown as the "Unfilled" amount.

image

### Two swap inputs exist and two swap outputs exist.
This is a two-way swap and it... basically will never happen. This means the user is swapping from asset 1 to asset 2 AND from asset 2 back to asset 1 in the same Swap action. There would never be a good reason to do this — especially since you're just burning fees. That said, I did account for this possibility in the view code; but since no Penumbra code makes this type of transaction possible, I didn't test it. However, this is what the two-way swap component looks like (ignore the fact that amount of each currency doesn't make sense — I was just rendering it with test data):

image

## Swap claim page
This is a bit simpler. We just have to account for when:

### One output exists.
This is the typical one-way swap case where there was enough liquidity to cover the entire swap.

image

### Two outputs exist.
This happens when either A) there was an unfilled amount, so the user can claim some of asset 1 AND some of asset 2; or B) the user did a two-way swap, so there is some of both assets to claim (again, this should never happen).

image

Closes #424


packages/ui/components/ui/tx/view/value-with-address.tsx
```diff
@@ -0,0 +1,30 @@
+import { AddressView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { ReactNode } from 'react';
+import { AddressViewComponent } from './address-view';
+
+/**
+ * Displays "from" or "to" and an address. Useful for spends, outputs, swaps,
+ * etc. where there's a value paired with an address.
+ */
+export const ValueWithAddress = ({
```
jessepinho, line 9: image

packages/ui/components/ui/tx/view/output.tsx
```diff
@@ -1,22 +1,22 @@
...
-    const note = value.outputView.value.note!;
+    const note = getNote(value);
+    const address = getAddress(note);
+
     return (
       <ViewBox
         label='Output'
         visibleContent={
-          <div className='flex flex-col justify-between gap-2 sm:flex-row sm:gap-0'>
+          <ValueWithAddress addressView={address} label='to'>
```
jessepinho, line 17: Extracted this markup to `ValueWithAddress` — see the screenshot in my comment on the `ValueWithAddress` code below

packages/ui/components/ui/tx/view/swap.tsx
```diff
@@ -1,78 +0,0 @@
-import { ViewBox } from './viewbox';
```
jessepinho, line 1: Moved to `...view/swap/index.tsx`

TalDerei:
@jessepinho tagging #830 as potential relevant scope

jessepinho:
@TalDerei that'd be separate scope from this ticket — this is just about how swap transactions are displayed in A) the txn approval dialog and B) the transaction view. I agree with you on the need for better UX when submitting a swap + swap claim, though!

packages/wasm/crate/src/tx.rs
```diff
@@ -278,13 +279,25 @@ pub async fn transaction_info_inner(
...
+                swap_record
+                    .clone()
+                    .and_then(|swap_record| swap_record.output_data)
+                    .map(|output_data| {
+                        if let Ok(bsod) = BatchSwapOutputData::try_from(output_data) {
+                            txp.batch_swap_output_data.push(bsod);
+                        }
+
+                        Some(())
+                    });
```
grod220, line 297: Think this should be moved up above `let swap_position_option =...`. Also, can we get a code comment on what this is doing?

grod220: What do you think about this?

```rust
       Action::Swap(swap) => {
                let commitment = swap.body.payload.commitment;
                if let Some(swap_record) = storage.get_swap_by_commitment(commitment.into()).await? {
                    // Add swap output to perspective
                    if let Some(output_data) = swap_record.output_data {
                        let bsod = BatchSwapOutputData::try_from(output_data)?;
                        txp.batch_swap_output_data.push(bsod)
                    }

                    // Add swap claim to perspective
                    let swap_position = Position::from(swap_record.position);
                    add_swap_claim_txn_to_perspective(
                        &storage,
                        &fvk,
                        &mut txp,
                        &commitment,
                        swap_position,
                    ).await?;
                }
            }

```

grod220 (approved):
Ah this looks so much better 💯

packages/wasm/crate/src/tx.rs
jessepinho: Ah, much cleaner. Will update.

Issue #739, opened 2024-03-13
Use `ViewService#unbondingTokens` to determine (and indicate to the user) which unbonding tokens are claimable vs. not


jessepinho:

grod220:
This is done right? @jessepinho

jessepinho:
Yep!

PR #862, opened 2024-03-30
Implement `ViewService#unbondingTokensByAddressIndex`; update to v0.71.0

jessepinho:
🚨 Most line changes are lockfiles and tests

This is the first part of #739 — creating an implementation of `ViewService#unbondingTokensByAddressIndex` that will, in my next PR, be used to show claimable vs not-yet-claimable unbonding tokens on the staking page.

## In this PR
- Update to v0.71.0 of core
- Create `ViewService#unbondingTokensByAddressIndex` impl

Relates to #739


packages/router/src/grpc/view-protocol-server/unbonding-tokens-by-address-index/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  async function* (req, ctx) {
+    for await (const balancesResponse of balances(
+      new BalancesRequest({ accountFilter: req.addressIndex }),
+      ctx,
+    )) {
+      if (!isUnbondingTokenBalance(balancesResponse)) continue;
+      const claimable = await getIsClaimable(balancesResponse, ctx);
+
+      // See https://github.com/typescript-eslint/typescript-eslint/issues/7114
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
grod220, line 20: I wonder if we should disable this rule globally with this comment until the bug is fixed

grod220 (approved):
LGTM

PR #863, opened 2024-03-31
Split out claimable vs. not-yet-claimable unbonding tokens in the UI

### LINKED ISSUE #739: Use `ViewService#unbondingTokens` to determine (and indicate to the user) which unbonding tokens are claimable vs. not

jessepinho:
We've been showing a user's unbonding tokens in a single group. When they click the `Claim` button, we create a single transaction with undelegate claims for every unbonding token that user holds. This means that the entire transaction (and all its undelegate claims) will fail if _any_ undelegate claim fails, since they're all batched into the same transaction.

With this PR, we use the new `ViewService#unbondingTokensByAddressIndex` endpoint to determine which unbonding tokens are claimable and which are not yet claimable. We only show the `Claim` button for claimable tokens, and we show the claimable and not-yet-claimable tokens separately.

image

image

## In this PR
- Use the new `ViewService#unbondingTokensByAddressIndex` to load unbonding tokens into state.
- Fix a bug where some accounts were incorrectly filtered out of the account switcher.
  - It used to be that the account switcher was being filtered via the `accountSelector`, which looked at a few different state properties to determine which accounts were eligible to be included on the staking page.

    The problem was, it only took staking and unbonding tokens into account. Which means that, if you had only delegation tokens in a given account, that account would not be accessible via the account switcher.
    
    So I created a method, `loadAndReduceBalances`, which a) puts the user's staking token balances into state, and b) determines which of the user's accounts are relevant to staking (i.e., because they contain staking, delegation, or unbonding tokens).
- Extract an `<UnbondingTokens />` component that shows the user's balance of unbonding tokens for either claimable or not-yet-claimable tokens.

Closes #739

### DIFF
--- a/apps/minifront/src/components/staking/layout.tsx
+++ b/apps/minifront/src/components/staking/layout.tsx
@@ -7,7 +7,7 @@ import { Account } from './account';
 export const StakingLoader = async () => {
   await throwIfPraxNotConnectedTimeout();
   // Await to avoid screen flicker.
-  await useStore.getState().staking.loadUnstakedTokensByAccount();
+  await useStore.getState().staking.loadAndReduceBalances();
 
   return null;
 };
--- a/apps/minifront/src/components/staking/layout.tsx
+++ b/apps/minifront/src/components/staking/layout.tsx
@@ -1,8 +1,8 @@
 import { useEffect } from 'react';
-import { useStore } from '../../state';
-import { stakingSelector } from '../../state/staking';
+import { AllSlices, useStore } from '../../state';
 import { throwIfPraxNotConnectedTimeout } from '@penumbra-zone/client';
 import { Account } from './account';
+import { useStoreShallow } from '../../utils/use-store-shallow';
 
 export const StakingLoader = async () => {
   await throwIfPraxNotConnectedTimeout();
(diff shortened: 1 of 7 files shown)


apps/minifront/src/components/staking/layout.tsx
```diff
@@ -1,25 +1,32 @@
...
+    useStoreShallow(stakingLayoutSelector);
 
   /** Load delegations every time the account changes. */
-  useEffect(
-    () => void loadDelegationsForCurrentAccount(),
-    [account, loadDelegationsForCurrentAccount],
-  );
+  useEffect(() => {
+    void loadDelegationsForCurrentAccount();
+    void loadUnbondingTokensForCurrentAccount();
```
grod220, line 28: These seem to always be together in a number of places right? Should we just combine them into one `loadStakingPositions()` function?

apps/minifront/src/state/staking/index.ts
```diff
@@ -292,7 +362,8 @@ export const createStakingSlice = (): SliceCreator<StakingSlice> => (set, get) =
       // Reload delegation tokens and unstaked tokens to reflect their updated
       // balances.
       void get().staking.loadDelegationsForCurrentAccount();
-      void get().staking.loadUnstakedAndUnbondingTokensByAccount();
+      void get().staking.loadAndReduceBalances();
+      void get().staking.loadUnbondingTokensForCurrentAccount();
```
grod220, line 366: Perhaps all three can just be combined into one? Does it help being separate?

grod220 (approved):

apps/minifront/src/components/staking/layout.tsx
jessepinho: They don't all need to get called at the same time, so unfortunately, they need to be split this way. For example:
- `loadDelegationsForCurrentAccount` and `loadUnbondingTokensForCurrentAccount` get called here in the effect
- `loadAndReduceBalances` gets called in the loader
- `loadDelegationsForCurrentAccount` and `loadAndReduceBalances` get (re-)called after delegating (but there's no need to reload unbonding tokens at that point)
- All three get called when undelegating
- `loadAndReduceBalances` and `loadUnbondingTokensForCurrentAccount` get called after undelegate claim, since those pieces of data need to be reloaded

apps/minifront/src/state/staking/index.ts
jessepinho: see above

PR #864, opened 2024-03-31
Check for an existing start epoch before creating one

### LINKED ISSUE #865: Delegations failing on testnet
Image

Image

When submitting a delegate transaction on testnet (`https://grpc.testnet.penumbra.zone`), I get the error in the first screenshot. The second screenshot is the latest epoch in my extension's IndexedDB database.

This ticket is to investigate and fix.

jessepinho:
This fixes a bug we had before where, for some reason (probably related to exponential back-off), the start epoch with index 0 was being saved multiple times. This threw off the count for epochs, making delegate actions fail (since they use epoch indexes). To fix this, we create epoch 0 outside of the syncAndStore method, at DB initialization.

Closes #865

### DIFF
--- a/apps/extension/.env
+++ b/apps/extension/.env
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=31
+IDB_VERSION=32
 USDC_ASSET_ID="reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
 MINIFRONT_URL=https://app.testnet.penumbra.zone/
 PENUMBRA_NODE_PD_URL=https://grpc.testnet.penumbra.zone/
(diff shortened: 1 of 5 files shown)


apps/extension/.env
```diff
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=31
+IDB_VERSION=32
```
jessepinho, line 3: Anyone who has duplicate epochs in their database already will need to regenerate their data from scratch — hence the version bump.

TalDerei:
LGTM. I'm not seeing any duplicate nullifiers in indexDB, but what does different epochs with the same `startHeight` mean in this context? sync delay?

Screenshot 2024-04-01 at 7 45 51 AM

TalDerei (approved):

jessepinho:
@TalDerei different epochs with the same start height _is_ a duplicate epoch — it means we saved more than one epoch (with different indexes) at the same start height.

And yeah, I think it's due to a sync delay. We use a `backOff()` function to repeatedly call the block processor's `syncAndStore` method if it fails at first. I think it must have been creating the 0th epoch multiple times as a result.

PR #868, opened 2024-04-01
Add RPC selection to onboarding flow

### LINKED ISSUE #605: Add RPC selection to onboarding flow
Currently, Penumbra Labs runs an RPC at `grpc.testnet{,-preview}.penumbra.zone`. However, we won't be doing this for a production mainnet. We want to ensure users have choice in their RPC provider and that there isn't a single entity acting as a chokepoint.

For this reason, the extension needs to have the user select their RPC endpoint while they are onboarding onto the extension. This should happen after they input their seed phrase and configure a passphrase.  To do this, we'll need:

- [ ] JSON configuration data in the extension source with a list of endpoints. Presumably each list item should include:
  - [ ] The RPC URL itself (eg `https://grpc.testnet.penumbra.zone`)
  - [ ] A human-readable name (eg `Penumbra Labs Testnet RPC`)
  - [ ] Optionally a logo image for the RPC provider (eg the Penumbra Labs logo)
- [ ] A dialog in the user flow after they set their seed phrase that prompts them to select from the list. The order of the list should be randomized on every appearance
- [ ] The list should have a special entry at the end labeled "Custom RPC" with free-form user input for the RPC URL.
- [ ] The list should have the first item (which is random, because of order randomization) preselected, so that a user who doesn't care about the RPC selection can just immediately click "Next" without having to think about it if they don't want to.

This would also help me with flows around local devnets, switching to preview, etc; it's easier to reset the whole wallet state and be able to set the target chain without having to edit internal constants.

jessepinho:
https://github.com/penumbra-zone/web/assets/1121544/344a0ffa-dac1-4620-99e6-bf1b1f2a8a3c

## In this PR
- Add an RPC selection step to the onboarding flow
- Rework how RPC impls are defined, so that we don't define them until we have a gRPC endpoint set in local extension storage.
- Add `addListener`/`removeListener` methods to `ExtensionStorage`, so that we can wait to make requests until the `grpcEndpoint` property exists in extension storage.
- Use the above two points to wait to start up the extension services until an RPC endpoint has been chosen.
  - Note that this means we no longer get errors on extension install! Yay!

## Next PR
- Add RPC selection to the Settings page of the extension

Relates to #605


apps/extension/src/routes/page/onboarding/set-password.tsx
```diff
@@ -16,16 +16,16 @@ import { PasswordInput } from '../../../shared/components/password-input';
...
   const [confirmation, setConfirmation] = useState('');
 
   const handleSubmit = (event: FormEvent | MouseEvent) => {
     event.preventDefault();
 
     void (async () => {
-      await finalOnboardingSave(password);
-      navigate(PagePath.ONBOARDING_SUCCESS);
+      await onboardingSave(password);
+      navigate(PagePath.SET_RPC_ENDPOINT);
```
jessepinho, line 28: Note that I didn't make this change in the `RestorePassword` component. It doesn't seem to be part of the onboarding flow, and I actually am unclear on if that component is actually used anywhere?

apps/extension/src/get-rpc-impls.ts
```diff
@@ -24,36 +24,40 @@ import { viewImpl } from '@penumbra-zone/router/src/grpc/view-protocol-server';
 import { localExtStorage } from '@penumbra-zone/storage/src/chrome/local';
 import { ServiceType } from '@bufbuild/protobuf';
 
-const grpcEndpoint = await localExtStorage.get('grpcEndpoint');
+export const getRpcImpls = async () => {
```
jessepinho, line 27: Wrapped `rpcImpls` in a `getRpcImpls` function so that we can wait to call it until we know the gRPC endpoint exists in local storage. (View this file in "hide whitespace" mode — there are almost no changes.)

apps/extension/src/utils/use-store-shallow.ts
```diff
@@ -0,0 +1,27 @@
+import { useShallow } from 'zustand/react/shallow';
```
jessepinho, line 1: This is copied exactly from `minifront`.

packages/storage/src/chrome/local.ts
```diff
@@ -7,12 +7,10 @@ import { LocalStorageState, LocalStorageVersion } from './types';
 // declaration in `apps/extension/prax.d.ts` because we are in an independent
 // package. we should probably move the localExtStorage declaration into the
 // extension app.
-declare const DEFAULT_GRPC_URL: string;
 declare const MINIFRONT_URL: string;
 
 export const localDefaults: LocalStorageState = {
   wallets: [],
-  grpcEndpoint: DEFAULT_GRPC_URL,
```
jessepinho, line 15: We no longer want to define a default gRPC URL; instead, we want to wait for the user to choose the URL.

packages/ui/components/ui/select-list.tsx
```diff
@@ -0,0 +1,103 @@
...
+ *     onSelect={() => {
+ *       customOptionInput.current?.focus();
+ *     }}
+ *     isSelected={option === customOption}
+ *     image={<CustomIcon className='size-full' />}
+ *   />
+ * </SelectList>
+ * ```
+ */
+export const SelectList = ({ children }: { children: ReactNode }) => (
```
jessepinho, line 49: image

packages/ui/components/ui/select-list.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import type { Meta, StoryObj } from '@storybook/react';
```
jessepinho, line 1: image

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const randomlySortedEndpoints = useMemo(() => [...RPC_ENDPOINTS].sort(randomSort), []);
+  const [grpcEndpoint, setGrpcEndpoint] = useState(randomlySortedEndpoints[0]!.url);
+  const setGrpcEndpointInZustand = useStore(state => state.network.setGRPCEndpoint);
+  const customRpcEndpointInput = useRef<HTMLInputElement | null>(null);
+  const isCustomRpcEndpoint = !RPC_ENDPOINTS.some(({ url }) => url === grpcEndpoint);
+
+  const handleSubmit = async (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    await setGrpcEndpointInZustand(grpcEndpoint);
+    void chrome.runtime.sendMessage(ServicesMessage.OnboardComplete);
```
jessepinho, line 26: @turbocrime please confirm that this is correct (moving `OnboardComplete`) to this step, since this is the new last step of onboarding.

apps/extension/src/routes/page/onboarding/set-password.tsx
grod220: Feel free to delete it. It's unused for like 8 months.

apps/extension/src/shared/rpc-endpoints.ts
```diff
@@ -0,0 +1,20 @@
...
+    name: 'Penumbra Labs Testnet RPC',
+    url: 'https://grpc.testnet.penumbra.zone',
+    imageUrl: STAKING_TOKEN_METADATA.images[0]?.svg,
+  },
+  {
+    name: 'Penumbra Labs Testnet Preview RPC',
+    url: 'https://grpc.testnet-preview.penumbra.zone',
+    imageUrl: STAKING_TOKEN_METADATA.images[0]?.svg,
+  },
+];
```
grod220, line 20: Let's move this to `packages/constants`, it probably should be included in the migration of our registry to another repo

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Network } from 'lucide-react';
+
+const randomSort = () => (Math.random() >= 0.5 ? 1 : -1);
+
+export const SetRpcEndpoint = () => {
+  const navigate = usePageNav();
+  const randomlySortedEndpoints = useMemo(() => [...RPC_ENDPOINTS].sort(randomSort), []);
+  const [grpcEndpoint, setGrpcEndpoint] = useState(randomlySortedEndpoints[0]!.url);
+  const setGrpcEndpointInZustand = useStore(state => state.network.setGRPCEndpoint);
+  const customRpcEndpointInput = useRef<HTMLInputElement | null>(null);
```
grod220, line 20: What do we need the ref for?

grod220, line 20: What do you think about moving all this logic into zustand with an RPC slice? My hunch is perhaps we'd want to re-use this within our settings tab in the extension and moving the logic out likely makes this easier.

grod220: It also reduces the redundancies with `setGrpcEndpointInZustand` & `setGrpcEndpoint`

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  return (
+    <FadeTransition>
+      <Card className='w-[400px]' gradient>
+        <CardHeader>
+          <CardTitle>Select your preferred RPC endpoint</CardTitle>
+          <CardDescription>
+            If you&apos;re unsure which one to choose, leave this option set to the default.
+          </CardDescription>
```
grod220, line 37: Can we add some text about the importance of trusting this RPC as it reveals intentions by what you request from it?

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
```diff
@@ -0,0 +1,94 @@
...
+                <input
+                  type='url'
+                  ref={customRpcEndpointInput}
+                  value={isCustomRpcEndpoint && !!grpcEndpoint ? grpcEndpoint : ''}
+                  onChange={e => setGrpcEndpoint(e.target.value)}
+                  className='w-full bg-transparent'
+                />
+              }
+              onSelect={() => {
+                if (!isCustomRpcEndpoint) setGrpcEndpoint('');
```
grod220, line 70: Can we do some validation like the way we do in settings right now? If it cannot connect to the grpc url, we do not allow it to be set.

packages/storage/src/chrome/base.ts
```diff
@@ -1,9 +1,15 @@
...
+
 export interface IStorage {
   get(key: string): Promise<Record<string, unknown>>;
   set(items: Record<string, unknown>): Promise<void>;
   remove(key: string): Promise<void>;
+  onChanged: {
+    addListener(listener: Listener): void;
+    removeListener(listener: Listener): void;
+  };
 }
```
grod220, line 13: I feel like there is a missed opportunity to use a generic here. For instance, `ExtensionStorage` can pass `T` to it right? And then we can use that to improve these methods. Or maybe not, up to you!

grod220 (approved):
Left some comments, think there are some areas it would be helpful to add to

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
jessepinho: It allows us to call `.focus()` on it when the custom RPC endpoint option is selected

apps/extension/src/service-worker.ts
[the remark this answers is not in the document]
jessepinho: Ah, good call. Done.

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
jessepinho: done

packages/storage/src/chrome/base.ts
jessepinho: Just tried this, and it looks like that's not possible because `chrome.storage.local` doesn't support generics :( so we can't actually get type-safety here — we have to treat the changed keys as just `string` rather than `keyof T`.

apps/extension/src/routes/page/onboarding/set-rpc-endpoint.tsx
jessepinho: Going to take the liberty of doing this (and addressing your comment above about moving logic into Zustand) in a separate PR, since it looks like that'll require some (necessary) refactoring that I was anyway going to do when adding a similar RPC selection form to the extension's settings page.

PR #869, opened 2024-04-01
Move transaction validity assertions out of authorize and into transaction-planner

jessepinho:
@Valentine1898 pointed out in [this comment](https://github.com/penumbra-zone/web/pull/861#discussion_r1546765377) that the assertions about the validity of a transaction should be run in the planning stage, rather than the authorization stage.

So I moved the assertions over. They now expect a `TransactionPlannerRequest`, rather than a `TransactionPlan`.

However, I realized after doing this that this won't block bad actors from e.g., submitting a swap transaction with a claim address that doesn't belong to the current user. That's because web frontends that interact with our extension don't actually _have_ to submit a transaction planner request to build a transaction. They can simply build the transaction plan themselves, then submit it to the extension, thereby bypassing these assertions.

Thus, I think this PR should NOT be merged. However, I'm opening it anyway so that reviewers can comment, in case I'm missing something.


Valentine1898:
I suggested moving only the `assertSwapAssetsAreNotTheSame() `check since the previous PR only addressed it. 

We should definitely not move the `assertSwapClaimAddressesBelongToCurrentUser()` check to the planner layer

Valentine1898:
@jessepinho but overall, you made a good point that other developers may not use the planner but create `TransactionPlan` manually. In that case, it makes sense to leave all the checks to the authorization step

hdevalence:
Probably it makes sense to have checks in both places but if there's a distinction to be drawn here it might be between "invalid plans" and "bad plans".

Checking that a swap isn't for an asset to itself is an "invalid plan" check, if the tx is built and submitted the chain will reject it and the user is no worse off. Checking that a swap doesn't have a foreign claim address is a "bad plan" check, if the tx is built and submitted the user could lose funds.

jessepinho:
Ah, sorry @Valentine1898 , I misunderstood you.

@hdevalence — to your point, I'll leave the same-asset-swaps check in the planner, and the swap-claim-address-belongs-to-current-user check in the authorize step.

jessepinho:
OK @Valentine1898 would you mind re-reviewing? I put the assertion with security implications back inside `CustodyService#authorize`, and left the assertion re: same asset in/out in the transaction planner impl.

Valentine1898 (approved):
LGTM!

PR #870, opened 2024-04-02
validate root

turbocrime:


grod220:
Think we can proceed with https://github.com/penumbra-zone/web/pull/834 and it covers this

PR #872, opened 2024-04-02
handle proving keys in dedicated package

turbocrime:
* use curl instead of writing a downloader
* keys are built into a dedicated package. key users simply import
* keys package provides bin for manipulating keys
* keys package provides action-key mapping

edit:

okay, webpack ruins everything. keys are downloaded by the package, and exported, but webpack needs manual copying and explicit urls.

package exports nothing but the keys, as files. user must provide mapping


turbo.json
turbocrime, line 1: this file is entirely wrong

packages/keys/download-keys.ts
```diff
@@ -0,0 +1,119 @@
...
+  await downloadKeys();
+  const checks = check();
+  await Promise.allSettled(checks);
+  await Promise.all(checks).then(
+    () => {
+      retry = 0;
+      stdout.log(outDir);
+    },
+    () => {
+      retry--;
```
turbocrime, line 115: loop logic wrong

apps/extension/webpack.config.js
```diff
@@ -115,10 +115,8 @@ export default (env, argv) => {
         patterns: [
           { from: 'public', to: '.' },
           {
-            from: 'bin',
-            to: 'bin',
-            context: path.resolve(__dirname, '../../packages/wasm'),
-            noErrorOnMissing: true,
+            from: process.env.KEY_PATH,
+            to: 'keys',
```
turbocrime, line 119: perhaps import.meta.resolve instead of env

apps/extension/package.json
```diff
@@ -7,7 +7,8 @@
   "scripts": {
     "dev": "webpack --watch --mode=development -d inline-source-map",
     "clean": "rm -rfv dist bin",
-    "build": "webpack",
+    "prebuild": "download-keys",
+    "build": "KEY_PATH=$(download-keys) webpack",
```
turbocrime, line 11: don't need both of these

packages/keys/download-keys.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+do {
+  await downloadKeys();
+  const checks = check();
+  await Promise.allSettled(checks);
+  await Promise.all(checks).then(
+    () => {
+      retry = 0;
+      stdout.log(outDir);
```
turbocrime, line 112: the only thing this entire script echoes to stdout path to the dir containing a successful key download. if it fails, nothing to stdout. all other output is status on stderr

packages/keys/download-keys.ts
```diff
@@ -0,0 +1,119 @@
...
+      ? // clobber will overwrite existing files
+        ['--clobber']
+      : // continue-at will resume or skip existing files
+        ['--continue-at', '-']),
+    '--location',
+    '--remote-name',
+    // curl supports url globbing
+    `${remoteSource}/{${fileNames.join(',')}}`,
+  ]);
+}
```
turbocrime, line 86: just use curl instead of implementing all of these things

packages/keys/download-keys.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    if (expectSha !== checkSha) {
+      console.log(`${name}: FAILED`);
+      process.exitCode ??= 0;
+      process.exitCode++;
+      if (rmFailedOption || (retryOption && !clobberOption)) fs.rm(filePath);
+      throw new Error(`${name}: ${checkSha}`);
+    } else console.log(`${name}: OK`);
+  });
+}
```
turbocrime, line 103: looked at using shasum/sha256sum but it's more complex, often not available, and node already includes the necessary crypto tools

packages/keys/download-keys.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+let workspace: string | undefined;
+workspace =
+  (await import('workspace-root').then(
+    ({ workspaceRoot }) => workspaceRoot(),
+    () => null, // package not present, or no root
+  )) || undefined;
```
turbocrime, line 35: if a tool for detecting workspace-root is present, it downloads to a dir in root. this is not necessary but may be convenient for anyone else consuming this package in their own monorepo

apps/extension/package.json
TalDerei:
```
  "scripts": {
    "dev": "webpack --watch --mode=development -d inline-source-map",
    "clean": "rm -rfv dist bin",
    "build": "webpack",
    "test": "vitest run"
  }
```

why wouldn't we keep the build script `"build": "webpack"`?

apps/extension/webpack.config.js
TalDerei: .

packages/wasm/package.json
```diff
@@ -5,8 +5,6 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
-    "download-keys": "tsx ./utils/download-proving-keys.ts",
-    "precompile": "npm run download-keys",
```
TalDerei, line 9: I rather liked the use of a `precompile` script. I'm partial to using more of that in the codebase

`download-keys` is also used in the context of the ci pipeline since it was missing keys: https://github.com/penumbra-zone/web/blob/config-pk/.github/workflows/turbo-ci.yml#L116

turbo.json
```diff
@@ -10,8 +10,7 @@
       "outputs": ["dist/**"]
     },
     "download-keys": {
-      "inputs": ["./utils/download-proving-keys.ts"],
-      "outputs": ["bin/**"]
```
TalDerei, line 14: similar to previous comment think we should keep an explicit `download-keys` script for CI purposes. I couldn't make some of the tests pass without it.

packages/keys/download-keys.ts
TalDerei: nice approach!

TalDerei:
I'll default to you on this since it's your issue. I'm impartial to the exact approach

turbocrime:
> I'll default to you on this since it's your issue. I'm impartial to the exact approach

retargeting PR

TalDerei:
we should find a way to incorporate this from the other PR

Screenshot 2024-04-14 at 5 27 14 PM

apps/extension/webpack.config.js
turbocrime, line 1: webpack doesn't support `import.meta.resolve` in the processed files, and produced broken builds when i exported URLs from the keys package. so i have to copy and generate paths in config

apps/extension/src/wasm-build-action.ts
turbocrime, line 1: i don't like this but i don't want to spend more time on this

packages/wasm/crate/src/keys.rs
turbocrime, line 1: just using the camelCase names we expect so we don't have to keep a second mapping around

packages/wasm/crate/tests/build.rs
turbocrime, line 1: maybe could use a local copy but then i'd worry about deduplication. this is fine

packages/wasm/src/build.ts
turbocrime, line 1: instead of guessing, wasm expects the caller to provide a key path

packages/keys/download-keys
turbocrime, line 1: simple

packages/keys/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "type": "module",
+  "files": [
+    "download-keys",
+    "keys"
+  ],
+  "exports": {
+    ".": "./keys"
+  },
+  "scripts": {
+    "postinstall": "./download-keys ./keys"
```
turbocrime, line 14: postinstall 😎

TalDerei (approved):

TalDerei:
LGTM

TalDerei:
@turbocrime can we add a `changeset` for this?

Issue #642, opened 2024-02-29
Fix misleading rendering in TransactionViewComponent


hdevalence:
The current transaction view component rendering shows unknown actions as "hidden". This is extremely problematic since it misleads users about what information is actually published by a transaction.

We need to change all of the actions we haven't implemented yet to show an "unimplemented action" text rather than an incognito view.

Later, as we fill in the action views, we should have code comments justifying why information that we don't show is omitted from the view.

TalDerei:
@grod220 is this done? If not, we should mark all bug-related tasks with `mainnet` tag

grod220:
Labeled 👍

PR #874, opened 2024-04-03
Make unimplemented views clear

grod220:
Closes https://github.com/penumbra-zone/web/issues/642

At the moment, unimplemented views are seen as incognito. This is a mistaken default. This creates a view that communicates to the user that we haven't got to creating a view for it.

Screenshot 2024-04-03 at 3 26 48 PM

Up next, will create an issue to audit our action views (especially the opaque ones).


Valentine1898 (approved):
LGTM

PR #877, opened 2024-04-03
Set gRPC endpoint during onboarding, part 2

### LINKED ISSUE #605: Add RPC selection to onboarding flow

jessepinho:
This PR takes the UI from #868 and uses it in the RPC settings page of the extension. It also does a bit of refactoring to better share the functionality between the RPC settings page and the RPC onboarding page. And lastly, it shows the selected chain ID (or error message) at the bottom of the form.

https://github.com/penumbra-zone/web/assets/1121544/63f05064-a426-4e67-82f7-11dc15e65e39

https://github.com/penumbra-zone/web/assets/1121544/7192fd2b-bc64-4804-8f2a-f664eb50151d

## In this PR
- Extract a `<GrpcEndpointForm />` component that's shared between the onboarding screen and the settings screen of the extension.
- Create a `<PaddingWrapper />` component to ensure consistent padding in the various component popup screens.

Closes #605


apps/extension/src/routes/page/onboarding/routes.tsx
```diff
@@ -6,7 +6,7 @@ import { ImportSeedPhrase } from './import';
 import { OnboardingSuccess } from './success';
 import { SetPassword } from './set-password';
 import { pageIndexLoader } from '..';
-import { SetRpcEndpoint } from './set-rpc-endpoint';
+import { SetGrpcEndpoint } from './set-grpc-endpoint';
```
jessepinho, line 9: Consolidating the naming

apps/extension/src/routes/page/onboarding/set-grpc-endpoint.tsx
```diff
@@ -0,0 +1,34 @@
+import { Card, CardDescription, CardHeader, CardTitle } from '@penumbra-zone/ui/components/ui/card';
+import { FadeTransition } from '@penumbra-zone/ui/components/ui/fade-transition';
+import { usePageNav } from '../../../utils/navigate';
+import { PagePath } from '../paths';
+import { ServicesMessage } from '@penumbra-zone/types/src/services';
+import { GrpcEndpointForm } from '../../../shared/components/grpc-endpoint-form';
+
+export const SetGrpcEndpoint = () => {
```
jessepinho, line 8: This file used to be `set-rpc-endpoint.tsx`. When I renamed it to `set-grpc-endpoint.tsx`, _and_ extracted the `GrpcEndpointForm` component, the file was so different that Git thought of it as an entirely different file.

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
```diff
@@ -0,0 +1,117 @@
...
+const isValidUrl = (url: string) => {
+  try {
+    new URL(url);
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+export const useGrpcEndpointForm = () => {
```
jessepinho, line 27: This hook basically extracts all the business logic out of the `SettingsRPC` component

apps/extension/src/routes/popup/padding-wrapper.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react';
+
+/**
+ * Wrap this around any top-level components that should have horizontal and
+ * bottom padding in the extension popup.
+ */
+export const PaddingWrapper = ({ children }: { children: ReactNode }) => {
```
jessepinho, line 7: For now, this is only used in one component (the RPC settings page). Once this PR is merged, I'll make a follow-up PR to use this throughout the extension popup screens.

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  } = useGrpcEndpointForm();
+  const customGrpcEndpointInput = useRef<HTMLInputElement | null>(null);
+
+  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (isSubmitButtonEnabled) void onSubmit(onSuccess);
+  };
+
+  return (
+    <div className='flex flex-col gap-2'>
```
turbocrime, line 37: minor but i'd make this a bare form element instead of wrapping it in a div

apps/extension/src/routes/popup/padding-wrapper.tsx
turbocrime: it seems like a custom component is the incorrect way to do this, it should be part of a layout or other existing parent component

apps/extension/package.json
```diff
@@ -32,6 +32,7 @@
     "exponential-backoff": "^3.1.1",
     "framer-motion": "^11.0.22",
     "immer": "^10.0.4",
+    "lodash": "^4.17.21",
```
turbocrime, line 35: tree shaking still sucks, and lodash is about about 1.5mb. i went through some effort to remove it as a dep a while back - if we're just going to use `debounce`, i would prefer to just write a debounce fn

apps/extension/src/routes/page/onboarding/routes.tsx
turbocrime: would prefer to align on "rpc", it's a more generic term

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
```diff
@@ -0,0 +1,117 @@
...
+    [handleChangeGrpcEndpointInput, grpcEndpointInput],
+  );
+
+  const onSubmit = async (
+    /** Callback to run when the RPC endpoint successfully saves */
+    onSuccess: () => void | Promise<void>,
+  ) => {
+    setIsSubmitButtonEnabled(false);
+    await setGrpcEndpoint(grpcEndpointInput);
+    // If the chain id has changed, our cache is invalid
```
turbocrime, line 95: on mainnet, this is a more serious condition than is typical in our experience as developers constantly flipping between testnet/preview. there should be a confirmation or other significant ui hurdle at this point.

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
jessepinho: Yeah I was going to, but I actually have a single element _outside_ the form at the bottom of this component that needs to be spaced apart from the form with `gap-2.

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
jessepinho: On it

jessepinho: OK, I added a confirmation dialog in [this commit](https://github.com/penumbra-zone/web/pull/877/commits/4251dd970f81c78a9e347f51d1519f867b1e6954?diff=split&w=1).

Was the rest of this PR a ✅ ? If so, could you please approve if that latest commit looks good?

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            />
+          </SelectList>
+
+          <a
+            href='https://github.com/penumbra-zone/web/blob/main/packages/constants/src/grpc-endpoints.ts'
+            target='_blank'
+            rel='noreferrer'
+            className='block text-right text-xs text-muted-foreground'
+          >
+            Add to this list
```
TalDerei, line 87: clicking on this link in the RPC settings page of the extension takes you to this outdated link: https://github.com/penumbra-zone/web/blob/main/packages/constants/src/grpc-endpoints.ts

turbocrime (approved):

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
jessepinho: @TalDerei that's actually the new link!

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
jessepinho: Yeah, you're right — I'll do a follow-up PR for that

apps/extension/src/routes/page/onboarding/routes.tsx
jessepinho: Ugh, same. I'll merge this and then discuss in Discord. The issue was, the local storage key is `grpcEndpoint`, so I figured we should standardize on that since it's already in a bunch of users' local storage.

PR #880, opened 2024-04-04
Fix id increasing version test

Valentine1898:
The indexed-db version increase test was not correct. 
In fact, instead of increasing the version of the existing db, we created a new db with a different random name

### DIFF
--- a/packages/storage/src/indexed-db/indexed-db.test.ts
+++ b/packages/storage/src/indexed-db/indexed-db.test.ts
@@ -100,7 +100,7 @@ describe('IndexedDb', () => {
         chainId: 'test',
         accountAddr: 'penumbra123xyz',
         dbVersion: 2,
-        walletId: version1Props.walletId
+        walletId: version1Props.walletId,
       };
       const dbB = await IndexedDb.initialize(version2Props);
       expect((await dbB.getAssetsMetadata(metadataA.penumbraAssetId!))?.name).toBeUndefined();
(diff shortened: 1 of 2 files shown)


packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -94,16 +94,14 @@ describe('IndexedDb', () => {
...
+      dbA.close();
 
       const version2Props = {
         chainId: 'test',
         accountAddr: 'penumbra123xyz',
         dbVersion: 2,
-        walletId: new WalletId({
-          inner: Uint8Array.from({ length: 8 }, () => Math.floor(Math.random() * 256)),
-        }),
+        walletId: version1Props.walletId
```
Valentine1898, line 103: Now the same walletId as for version 1

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -94,16 +94,14 @@ describe('IndexedDb', () => {
       const version1Props = generateInitialProps();
       const dbA = await IndexedDb.initialize(version1Props);
       await dbA.saveAssetsMetadata(metadataA);
+      dbA.close();
```
Valentine1898, line 97: The thing is that in order to upgrade idb to a new version you have to wait for it to close. 
If idb is not closed, the test will last forever.  
This did not lead to a bug only because when updating the version extension restarts and closes indexed-db

grod220 (approved):

PR #881, opened 2024-04-04
DRY up extension settings code; extract a `<SettingsScreen />` component for all settings screens in the extension

jessepinho:
🚨 I'd recommend viewing this PR with the "Hide whitespace" setting turned on.

In #877, I created a `<PaddingWrapper />` component for use throughout the various settings screens. But then I realized that all the settings screens had a lot more in common than just the padding, so in this PR I deleted `<PaddingWrapper />` and extracted a `<SettingsScreen />` component to take care of the common elements of settings screens (header, back button, padding, icon, etc.).

The only visual difference in this PR is that "Lock wallet" is now listed along with other menu options, rather than being pushed all the way to the bottom:
image

### DIFF
--- a/apps/extension/src/routes/popup/padding-wrapper.tsx
+++ b/apps/extension/src/routes/popup/padding-wrapper.tsx
@@ -1,9 +1,0 @@
-import { ReactNode } from 'react';
-
-/**
- * Wrap this around any top-level components that should have horizontal and
- * bottom padding in the extension popup.
- */
-export const PaddingWrapper = ({ children }: { children: ReactNode }) => {
-  return <div className='px-[30px] pb-[30px]'>{children}</div>;
-};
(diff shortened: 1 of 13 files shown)


grod220 (approved):

TalDerei (approved):

turbocrime (approved):
this is an improvement, but my intended meaning was that the layout is already positioned to directly handle everything that SettingsScreen is doing

https://github.com/penumbra-zone/web/blob/1e1f1e61a6b8da880fd5039aa2c2016f48687d40/apps/extension/src/routes/popup/settings/index.tsx#L1-L10

and things like title/icon are effectively route information

i think it is worth keeping the ‘lock wallet’ link pushed to the bottom.

jessepinho:
@turbocrime OK, I pushed the lock button to the bottom.

Going to merge this for now since it's an improvement. TBH, not sure I'm crazy about moving part of the UI for settings screens into the `<Settings />` layout component, since that means that the UI — and the knowledge about a particular route — will be divided between the route's component and the layout component containing it. i.e., the RPC selection screen will render the title "RPC" and the network icon, but when you go to the RPC selection screen's code, you won't see either of those in its code. Meanwhile, the parent layout component has to know which title and icon go with each route, when that information should probably be localized to the route component.

Not sure if I'm explaining this well, and I'm open to discussion on it.

Issue #878, opened 2024-04-03
Runtime configuration of SCT divergence checks in web extension


hdevalence:
When using 6.0.0 from the Chrome store, I cannot make any transactions because of a sync bug causing an SCT divergence:

> Error: [unknown] Tendermint error 1: failed to deliver transaction: check_stateful failed: provided anchor 602a5492caad089c2443c8e264ce87b08a652979f5e93d944293f4adb08a6908 is not a valid SCT root

I cannot debug this because there's no way to enable the SCT divergence checks at runtime. To fix this, there should be a field I can manually set in the storage inspector that configures whether SCT divergence checks are enabled. We don't need a UI for this, other than the Chrome devtools, but in the future, if someone else has an SCT divergence, there should be a way for them to produce the divergence height without having to reinstall the software.

PR #882, opened 2024-04-04
Add console checks for sct divergence

grod220:
Closes https://github.com/penumbra-zone/web/issues/878

https://github.com/penumbra-zone/web/assets/16624263/c0596976-bc1f-45f9-8bc6-a84e7aaec7fa

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -36,6 +36,13 @@ import { getAssetId } from '@penumbra-zone/getters/src/metadata';
 import { STAKING_TOKEN_METADATA } from '@penumbra-zone/constants/src/assets';
 import { toDecimalExchangeRate } from '@penumbra-zone/types/src/amount';
 import { ValidatorInfoResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/stake/v1/stake_pb';
+import { uint8ArrayToHex } from '@penumbra-zone/types/src/hex';
+
+declare global {
+  // `var` required for global declaration (as let/const are block scoped)
+  // eslint-disable-next-line no-var
+  var ASSERT_ROOT_VALID: boolean | undefined;
+}
 
 interface QueryClientProps {
   querier: RootQuerier;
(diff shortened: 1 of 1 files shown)


packages/query/src/block-processor.ts
```diff
@@ -141,7 +148,7 @@ export class BlockProcessor implements BlockProcessorInterface {
 
   private async syncAndStore() {
     const fullSyncHeight = await this.indexedDb.getFullSyncHeight();
-    const startHeight = fullSyncHeight ? fullSyncHeight + 1n : 0n;
+    const startHeight = fullSyncHeight !== undefined ? fullSyncHeight + 1n : 0n; // Must compare to undefined as 0n is falsy
```
grod220, line 151: Found an interesting bug. We sync block 0. However, 0 is a falsy value. That means if state was reset between blocks 0 and 1000, we'd have a sct root divergence.

Valentine1898: Nice catch 👍 
I wonder if this fixes the divergence that occurs on Henry's wallet (I suspect Henry didn't stop synchronization specifically between the 0 and 1000 block)

Valentine1898 (approved):

packages/query/src/block-processor.ts
hdevalence: Hmm, I _did_ make transactions in the first 1000 blocks, so if it paused to fetch full tx details, that could have been an opportunity for interruption...

packages/query/src/block-processor.ts
```diff
@@ -36,6 +36,13 @@ import { getAssetId } from '@penumbra-zone/getters/src/metadata';
 import { STAKING_TOKEN_METADATA } from '@penumbra-zone/constants/src/assets';
 import { toDecimalExchangeRate } from '@penumbra-zone/types/src/amount';
 import { ValidatorInfoResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/stake/v1/stake_pb';
+import { uint8ArrayToHex } from '@penumbra-zone/types/src/hex';
+
+declare global {
+  // `var` required for global declaration (as let/const are block scoped)
+  // eslint-disable-next-line no-var
+  var ASSERT_ROOT_VALID: boolean | undefined;
+}
```
hdevalence, line 45: @grod220 can we get some docs in the dev docs about the existence of this global and how you would set it and why?

grod220: Sounds good 👍 
https://github.com/penumbra-zone/web/pull/891

Issue #824, opened 2024-03-25
Add block sync states


grod220:
For our block sync status indicator in the extension, it waits until its loaded until it fades in.

Loading & error state:

<img width="368" alt="Screenshot 2024-03-25 at 1 31 54 PM" src="https://github.com/penumbra-zone/web/assets/16624263/cfa3d299-fb5d-4aad-8742-827ae0e09d99">

Syncing state:

<img width="372" alt="Screenshot 2024-03-25 at 1 32 19 PM" src="https://github.com/penumbra-zone/web/assets/16624263/06a24bee-64cb-4e10-b177-8423745563b9">

This is not the case with our minifront block sync UI component where it has states for:
- Loading
- In progress
- Completed
- Error

We should add these states to the extension UI component as well. This is particularly important given the genesis block is _quite_ large and could possibly take a long time. So we should add a 5th state (loading genesis block) if we see in storage no syncs have taken place. If so, the user should see a special kind of loader for the genesis block so it doesn't look stuck at say `0/1242521123` for a long time.

PR #883, opened 2024-04-04
Clear cache fix & added sync states

grod220:
Closes https://github.com/penumbra-zone/web/issues/824

Does two main things:
- On cache clearing, waits for db to clear before transitioning back to index. Sometimes takes a bit, so added a loading state for the button.
- Added genesis and normal loading states for block sync bar. Gives users an indication something is happening. At the moment, it simply doesn't render when its in the loading state.

https://github.com/penumbra-zone/web/assets/16624263/ef6dbfb7-bf07-4a80-a147-c83e7a77eed0

### DIFF
--- a/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
+++ b/apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
@@ -6,9 +6,30 @@ import { TrashGradientIcon } from '../../../icons/trash-gradient';
 import { ServicesMessage } from '@penumbra-zone/types/src/services';
 import { usePopupNav } from '../../../utils/navigate';
 import { PopupPath } from '../paths';
+import { useStore } from '../../../state';
+import { useState } from 'react';
+
+const useCacheClear = () => {
+  const navigate = usePopupNav();
+  const [loading, setLoading] = useState(false);
+
+  const handleCacheClear = () => {
+    setLoading(true);
+
+    void (async function () {
+      await chrome.runtime.sendMessage(ServicesMessage.ClearCache);
+      useStore.setState(state => {
+        state.network.fullSyncHeight = undefined;
+      });
+      navigate(PopupPath.INDEX);
+    })();
+  };
+
+  return { handleCacheClear, loading };
+};
 
 export const SettingsClearCache = () => {
-  const navigate = usePopupNav();
+  const { handleCacheClear, loading } = useCacheClear();
   return (
     <FadeTransition>
       <div className='flex min-h-screen w-screen flex-col gap-6'>
@@ -28,15 +49,13 @@ export const SettingsClearCache = () => {
             </p>
           </div>
           <Button
+            disabled={loading}
             variant='gradient'
             size='lg'
             className='w-full'
-            onClick={() => {
-              void chrome.runtime.sendMessage(ServicesMessage.ClearCache);
-              navigate(PopupPath.INDEX);
-            }}
+            onClick={handleCacheClear}
           >
-            Confirm
+            {loading ? 'Clearing cache...' : 'Confirm'}
           </Button>
         </div>
       </div>
--- a/packages/storage/src/chrome/local.ts
+++ b/packages/storage/src/chrome/local.ts
@@ -11,7 +11,7 @@ declare const MINIFRONT_URL: string;
 
 export const localDefaults: LocalStorageState = {
   wallets: [],
-  fullSyncHeight: 0,
+  fullSyncHeight: undefined,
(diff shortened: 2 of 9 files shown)


packages/storage/src/chrome/local.ts
```diff
@@ -11,7 +11,7 @@ declare const MINIFRONT_URL: string;
 
 export const localDefaults: LocalStorageState = {
   wallets: [],
-  fullSyncHeight: 0,
+  fullSyncHeight: undefined,
```
grod220, line 14: This is sort of a bug somewhat relating to: https://github.com/penumbra-zone/web/pull/882.

0 sync height indicates that we've synced the genesis block. So `undefined` should be used as the proper default indicating no sync has been done. Lots of changes in components flow from this one change.

apps/extension/src/routes/popup/settings/settings-clear-cache.tsx
```diff
@@ -6,9 +6,30 @@ import { TrashGradientIcon } from '../../../icons/trash-gradient';
...
+  const [loading, setLoading] = useState(false);
+
+  const handleCacheClear = () => {
+    setLoading(true);
+
+    void (async function () {
+      await chrome.runtime.sendMessage(ServicesMessage.ClearCache);
+      useStore.setState(state => {
+        state.network.fullSyncHeight = undefined;
+      });
```
grod220, line 23: Two things happening here:
- Awaiting cache clearing to finish
- resetting zustand sync height

Only after this do we navigate. Sometimes this can be slower, so added a loading state.

packages/ui/components/ui/block-sync-status/large.tsx
```diff
@@ -47,6 +49,38 @@ const BlockSyncErrorState = ({ error }: { error: unknown }) => {
...
 };
 
+// Not having a fullSyncHeight indicates we are doing a fresh sync.
+// The genesis block is particularly large, so we have special text to indicate that.
+const AwaitingState = ({ genesisSyncing }: { genesisSyncing: boolean }) => {
+  return (
+    <motion.div
+      className='flex w-full flex-col gap-1'
+      initial={{ opacity: 0 }}
+      animate={{ opacity: 1, transition: { duration: 0.3, ease: 'easeOut', delay: 0.75 } }}
```
grod220, line 59: Set a short delay. If not, flickers the screen.

packages/ui/components/ui/block-sync-status/large.tsx
```diff
@@ -56,7 +90,7 @@ const SyncingState = ({ fullSyncHeight, latestKnownBlockHeight }: SyncingStatePr
   return (
     <motion.div
       className='flex w-full flex-col items-center gap-1'
-      initial={{ opacity: 1 }}
+      initial={{ opacity: 0 }}
```
grod220, line 93: bug fix

packages/ui/components/ui/block-sync-status/large.tsx
```diff
@@ -12,7 +12,9 @@ export const LargeBlockSyncStatus = ({
   error,
 }: BlockSyncProps) => {
   if (error) return <BlockSyncErrorState error={error} />;
-  if (!latestKnownBlockHeight || !fullSyncHeight) return <div />;
+  if (!latestKnownBlockHeight || !fullSyncHeight) {
```
jessepinho, line 15: Shouldn't this be checking for `!== undefined`? Since `0n` now means that the first block has synced?

packages/ui/components/ui/block-sync-status/large.tsx
jessepinho: Not sure I understand what flickers, but setting a hard-coded time delay to avoid flicker feels a little hacky to me, since whether it works depends on the speed of the user's computer. Is there not a deterministic way to prevent flicker?

jessepinho (approved):
Couple small Qs, but feel free to merge after addressing those. This is a big UX win!

packages/ui/components/ui/block-sync-status/large.tsx
grod220: Aha! This edge case is the gift that keeps on giving.

packages/ui/components/ui/block-sync-status/large.tsx
grod220: Here's without it. Essentially, we need a sort of UI debouncer to prevent it. I found a slight animation delay is sufficient, but perhaps there is a better way 🤔 

https://github.com/penumbra-zone/web/assets/16624263/5af95a28-a00d-42ef-a295-4a7822f5768d

PR #886, opened 2024-04-04
Show equivalent values in the assets table

jessepinho:
Pretty straightforward. I also added a tooltip for equivalent values to make it clear what they represent:

image


grod220 (approved):
Design-wise, I do wonder if we should bring back the table layout:

| Asset         | Equivalent Values    |
|---------------|----------------------|
| `10um`        | `332.2usdc`          |
| `53uxyz` | `66um`, `324234usdc` |

TalDerei (approved):

jessepinho:
@grod220 I'll merge this and do a quick spike on making it use a table

PR #890, opened 2024-04-05
Improve memo rendering with rows

hdevalence:
Uses the rows component to render the entire memo as one object (since it is one object). Also fixes formatting for empty memos.

Screenshot 2024-04-04 at 7 28 05 PM
Screenshot 2024-04-04 at 7 27 39 PM
Screenshot 2024-04-04 at 7 26 20 PM
Screenshot 2024-04-04 at 7 26 13 PM

As a side effect to doing this, I added configuration to mark individual rows as visible or not. I originally used this to make a version of the "opaque" memo that showed that individual fields were opaque, and then concluded that it wasn't a good design.

However, I left them in the code since they'll be useful for making rendered views of actions that can show that some fields are present while others are opaque.

### DIFF
--- a/packages/ui/components/ui/tx/view/memo-view.tsx
+++ b/packages/ui/components/ui/tx/view/memo-view.tsx
@@ -1,36 +1,32 @@
 import { MemoView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 import { AddressViewComponent } from './address-view';
-import { ViewBox, ViewSection } from './viewbox';
+import { ViewBox } from './viewbox';
+import { ActionDetails } from './action-details';
 
 export const MemoViewComponent = ({ memo: { memoView } }: { memo: MemoView }) => {
   switch (memoView.case) {
     case 'visible':
       return (
-        <ViewSection heading='Memo'>
-          <ViewBox
-            label='Memo Text'
-            visibleContent={
-              <div>
-                {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
-                {memoView.value.plaintext?.text || (
-                  <span className='italic text-gray-600'>== none ==</span>
-                )}
-              </div>
-            }
-          />
-          <ViewBox
-            label='Sender Return Address'
-            visibleContent={<AddressViewComponent view={memoView.value.plaintext!.returnAddress} />}
-          />
-        </ViewSection>
+        <ViewBox
+          label='Memo'
+          visibleContent={
+            <div className='flex flex-col gap-4'>
+              <ActionDetails>
+                <ActionDetails.Row label='Sender Return Address'>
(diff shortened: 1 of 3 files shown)


packages/ui/components/ui/tx/view/viewbox.tsx
```diff
@@ -7,22 +7,25 @@ import { IncognitoIcon } from '../../icons/incognito';
...
   label: string;
   visibleContent?: React.ReactElement;
+  isOpaque?: boolean;
 }
 
 const Label = ({ label }: { label: string }) => <span className='text-lg'>{label}</span>;
 
-export const ViewBox = ({ label, visibleContent }: ViewBoxProps) => {
+export const ViewBox = ({ label, visibleContent, isOpaque }: ViewBoxProps) => {
+  // if isOpaque is undefined, set it to !visibleContent
```
jessepinho, line 16: Hm, couldn't this conflate private content with undefined or empty content, like an empty string?

packages/ui/components/ui/tx/view/memo-view.tsx
```diff
@@ -1,36 +1,32 @@
...
+        <ViewBox
+          label='Memo'
+          visibleContent={
+            <div className='flex flex-col gap-4'>
+              <ActionDetails>
+                <ActionDetails.Row label='Sender Return Address'>
+                  <AddressViewComponent view={memoView.value.plaintext!.returnAddress} />
+                </ActionDetails.Row>
+                <ActionDetails.Row label='Memo Text'>
+                  <span className='italic' style={{ wordBreak: 'normal' }}>
```
jessepinho, line 19: Could we get rid of the italic?

jessepinho (approved):
Two minor things. Feel free to merge after addressing, or if they're not a concern.

packages/ui/components/ui/tx/view/viewbox.tsx
hdevalence: Yeah, but the current ViewBox component doesn't handle children in a "real" way, it has some stuff passed in and it was using that to determine if it should show the incognito icon. The new parameter I added is an explicit override, the fallback check that could be error-prone is the existing behavior.

I think it would be better if the viewbox took child elements and had explicitly controlled visibility display, but I didn't want to rewrite the whole component and all its uses along the way to improving the memo rendering.

packages/ui/components/ui/tx/view/memo-view.tsx
hdevalence: I did the italics on purpose.  I felt that it was a reasonably unobtrusive way to signal that the text was different from the other text on the page — all of the other words and labels are generated by the frontend, but the memo text is supplied by the counterparty (who is maybe adversarial, eg not escaping it would be bad since they could then inject scripts into the user's page, but even if we're not XSSable it seems good to signal that the text comes from the transaction, not from the UI).

jessepinho: Makes sense!

PR #896, opened 2024-04-05
Generate unbonding token metadata on the fly for the approval dialog

### LINKED ISSUE #856: Generate unbonding token metadata on the fly for the approval dialog
Image

Right now, the unbonding tokens in the transaction approval dialog are rendered as "unknown asset". This is because the token metadata for this unbonding token doesn't yet exist, since it's specific to the block that unbonding starts at.

To fix this, we'll need to generate token metadata for unbonding tokens on the fly, using logic similar to [this](https://github.com/penumbra-zone/penumbra/blob/b429bc7/crates/core/asset/src/asset/registry.rs#L355-L380).

jessepinho:
When a user undelegates from a validator — meaning, they want to unbond their delegation tokens so that they can reclaim those tokens as UM — Penumbra issues "unbonding" tokens. These unbonding tokens are tied to the specific validator that the user is undelegating from, as well as to the block height at which the user made the undelegate transaction. (It just so happens that Penumbra's current implementation clamps the block height to the start of the current epoch; but this could change in the future, so we treat it as though it's tied to a block height, rather than tied to an epoch.)

Because unbonding tokens are so specific, there often is no metadata for them in the database when a user wishes to undelegate from a validator. (The only way there _would_ be metadata for them in the database is if the user had already undelegated from that same validator at that same block height at least once before. Since the block height is clamped to the start of the epoch, this is technically possible; but the user would still face this issue the first time they undelegate from that validator during that epoch.) As a result, the transaction approval dialog would show "Unknown asset" for the "Output" action that outputs unbonding tokens. This is a confusing experience for users, who would have no way of knowing that the "Unknown asset" they're getting is the unbonding token.

This PR fixes that issue by generating and saving asset metadata to the database _while planning the transaction_. This means that, by the time the transaction approval dialog displays the transaction plan, the metadata will exist in the database, and thus the dialog will correctly show the asset metadata.

The (very slight) downside to this approach is that, if the user cancels the transaction and never attempts another undelegation at that block height, IndexedDB will contain metadata for an asset that doesn't actually exist for the user on-chain. i.e., if the user clicks "Undelegate" for a validator — which pops open the transaction approval dialog and saves the generated asset metadata for the unbonding tokens — and then cancels the transaction, the generated metadata will still exist in the database even if it never gets "used" for any unbonding tokens. This seems like an OK trade-off, however.

Closes #856

### DIFF
--- a/apps/extension/.env
+++ b/apps/extension/.env
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=32
+IDB_VERSION=33
 USDC_ASSET_ID="reum7wQmk/owgvGMWMZn/6RFPV24zIKq3W6In/WwZgg="
 MINIFRONT_URL=https://app.testnet.penumbra.zone/
 PENUMBRA_NODE_PD_URL=https://grpc.testnet.penumbra.zone/
--- a/apps/extension/src/state/tx-approval.ts
+++ b/apps/extension/src/state/tx-approval.ts
@@ -66,9 +66,9 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set, 
     const getMetadata = async (assetId: AssetId) => {
       try {
         const { denomMetadata } = await viewClient.assetMetadataById({ assetId });
-        return denomMetadata ?? new Metadata();
+        return denomMetadata ?? new Metadata({ penumbraAssetId: assetId });
       } catch {
-        return new Metadata();
+        return new Metadata({ penumbraAssetId: assetId });
       }
     };
 
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -118,7 +118,10 @@ export class IndexedDb implements IndexedDbInterface {
 
     const instance = new this(db, new IbdUpdater(db), constants, chainId);
     await instance.saveLocalAssetsMetadata(); // Pre-load asset metadata
-    await instance.addEpoch(0n); // Create first epoch
+
+    const existing0thEpoch = await instance.getEpochByHeight(0n);
+    if (!existing0thEpoch) await instance.addEpoch(0n); // Create first epoch
+
     return instance;
   }
   close(): void {
--- a/packages/wasm/crate/src/metadata.rs
+++ b/packages/wasm/crate/src/metadata.rs
@@ -1,42 +1,63 @@
-use penumbra_proto::core::asset::v1::Metadata;
+use anyhow::anyhow;
+use penumbra_asset::asset::Metadata as MetadataDomainType;
+use penumbra_proto::{core::asset::v1::Metadata, DomainType};
 use regex::Regex;
+use wasm_bindgen::prelude::wasm_bindgen;
+
+use crate::{error::WasmResult, utils};
 
 pub static UNBONDING_TOKEN_REGEX: &str = "^uunbonding_(?P<data>start_at_(?P<start>[0-9]+)_(?P<validator>penumbravalid1(?P<id>[a-zA-HJ-NP-Z0-9]+)))$";
 pub static DELEGATION_TOKEN_REGEX: &str =
     "^udelegation_(?P<data>penumbravalid1[a-zA-HJ-NP-Z0-9]+)$";
 pub static SHORTENED_ID_LENGTH: usize = 8;
 
-pub fn customize_symbol(metadata: Metadata) -> Metadata {
-    if let Some(unbonding_match) = Regex::new(UNBONDING_TOKEN_REGEX)
-        .expect("regex is valid")
-        .captures(&metadata.base)
-    {
-        let id_match = unbonding_match.name("id").unwrap().as_str();
-        let shortened_id = id_match
-            .chars()
-            .take(SHORTENED_ID_LENGTH)
-            .collect::<String>();
-        let start_match = unbonding_match.name("start").unwrap().as_str();
+#[wasm_bindgen]
+pub fn customize_symbol(metadata_bytes: &[u8]) -> WasmResult<Vec<u8>> {
+    utils::set_panic_hook();
 
-        Metadata {
-            symbol: format!("unbondUMat{start_match}({shortened_id}…)"),
-            ..metadata
-        }
-    } else if let Some(delegation_match) = Regex::new(DELEGATION_TOKEN_REGEX)
-        .expect("regex is valid")
-        .captures(&metadata.base)
-    {
-        let id_match = delegation_match.name("id").unwrap().as_str();
-        let shortened_id = id_match
-            .chars()
-            .take(SHORTENED_ID_LENGTH)
-            .collect::<String>();
+    let metadata_domain_type = MetadataDomainType::decode(metadata_bytes)?;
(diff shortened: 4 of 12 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -118,7 +118,10 @@ export class IndexedDb implements IndexedDbInterface {
 
     const instance = new this(db, new IbdUpdater(db), constants, chainId);
     await instance.saveLocalAssetsMetadata(); // Pre-load asset metadata
-    await instance.addEpoch(0n); // Create first epoch
+
+    const existing0thEpoch = await instance.getEpochByHeight(0n);
+    if (!existing0thEpoch) await instance.addEpoch(0n); // Create first epoch
```
jessepinho, line 123: I hadn't realized that this code was run every time the extension was restarted, so I added a check to make sure the 0th epoch doesn't exist before creating it.

apps/extension/.env
```diff
@@ -1,6 +1,6 @@
 
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
-IDB_VERSION=32
+IDB_VERSION=33
```
jessepinho, line 3: Needed for [this fix](https://github.com/penumbra-zone/web/pull/896/files#diff-8298d111bc94531ec943d5f024a9ab57657ad28670caea51e3aa9eda88bf3954)

apps/extension/src/state/tx-approval.ts
```diff
@@ -66,9 +66,9 @@ export const createTxApprovalSlice = (): SliceCreator<TxApprovalSlice> => (set,
     const getMetadata = async (assetId: AssetId) => {
       try {
         const { denomMetadata } = await viewClient.assetMetadataById({ assetId });
-        return denomMetadata ?? new Metadata();
+        return denomMetadata ?? new Metadata({ penumbraAssetId: assetId });
       } catch {
-        return new Metadata();
+        return new Metadata({ penumbraAssetId: assetId });
```
jessepinho, line 71: Not strictly needed for this PR, but it should have been like this already.

packages/wasm/crate/src/metadata.rs
```diff
@@ -0,0 +1,42 @@
+use penumbra_proto::core::asset::v1::Metadata;
+use regex::Regex;
+
+pub static UNBONDING_TOKEN_REGEX: &str = "^uunbonding_(?P<data>start_at_(?P<start>[0-9]+)_(?P<validator>penumbravalid1(?P<id>[a-zA-HJ-NP-Z0-9]+)))$";
+pub static DELEGATION_TOKEN_REGEX: &str =
+    "^udelegation_(?P<data>penumbravalid1[a-zA-HJ-NP-Z0-9]+)$";
```
jessepinho, line 6: This is a copy of core's [delegation token regex](https://github.com/penumbra-zone/penumbra/blob/de135e8/crates/core/component/stake/src/delegation_token.rs#L67) and [unbonding token regex](https://github.com/penumbra-zone/penumbra/blob/de135e8/crates/core/component/stake/src/unbonding_token.rs#L71), although the unbonding token regex has been modified from the original to be able to extract the ID.

I'm not crazy about having duplicate Regex, so if this PR is merged, I'll make a PR to core to extract the Regex for those tokens so they can be imported here. (I'll also have to modify the unbonding token regex to include the ID.)

packages/wasm/crate/src/metadata.rs
```diff
@@ -0,0 +1,42 @@
...
+pub static SHORTENED_ID_LENGTH: usize = 8;
+
+pub fn customize_symbol(metadata: Metadata) -> Metadata {
```
jessepinho, line 9: This is the exact same functionality as our [TypeScript `customizeSymbol` function](https://github.com/penumbra-zone/web/blob/main/packages/types/src/customize-symbol.ts), ported to Rust.

I briefly spiked on just calling out to this Rust function from our TypeScript code, rather than having one copy of this function on each of the Rust and TypeScript sides. However, that would mean modifying this function to take a `Vec<u8>` encoding of the metadata object, which would then make this function a bit messier to use from the Rust planner code that calls it, because we'd have to encode and decode metadata to achieve this.

Not sure what the best way forward for this is — reviewers, thoughts?

grod220: Two things:
- In Rust, save for tests, we should try to avoid panic'ing. A panic is basically like a hard crash where we couldn't even catch on the javascript side. Which means`.expect()`, should be replaced with `?` and `unwrap` (unless guaranteed to never panic) should be handled.
- To expose this function to the frontend, perhaps we can have an _inner function? Whatever naming, we can have two functions, one to handle TS and one for internal wasm usage.


What do you think of this?


```rust
// along with other errors in error.rs...
#[derive(Error, Debug)]
pub enum WasmError {
    // ...
    #[error("{0}")]
    RegexError(#[from] regex::Error),
}

#[wasm_bindgen]
pub fn customize_symbol(metadata_bytes: &[u8]) -> WasmResult<Vec<u8>> {
    utils::set_panic_hook();

    let metadata = Metadata::decode(metadata_bytes)?;
    customize_symbol_inner(metadata)?;
    Ok(metadata.encode_to_vec())
}


pub fn customize_symbol_inner(mut metadata: Metadata) -> WasmResult<Metadata> {
    let unbonding_re = Regex::new(UNBONDING_TOKEN_REGEX)?;
    let delegation_re = Regex::new(DELEGATION_TOKEN_REGEX)?;

    if let Some(captures) = unbonding_re.captures(&metadata.base) {
        let shortened_id = shorten_id(&captures)?;
        let start_match = captures.name("start")
            .ok_or_else(|| anyhow!("<start> not matched in unbonding token regex"))?
            .as_str();
        metadata.symbol = format!("unbondUMat{start_match}({shortened_id}...)");
    } else if let Some(captures) = delegation_re.captures(&metadata.base) {
        let shortened_id = shorten_id(&captures)?;
        metadata.symbol = format!("delUM({shortened_id}...)");
    }

    Ok(metadata)
}

fn shorten_id(captures: &regex::Captures) -> WasmResult<String> {
    let id_match = captures.name("id").ok_or_else(|| anyhow!("<id> not matched in staking token regex"))?;
    Ok(id_match.as_str()
        .chars()
        .take(SHORTENED_ID_LENGTH)
        .collect())
}

```

If you don't like mutating the input, this can be converted to just cloning the input and mutating that.

packages/wasm/crate/src/planner.rs
```diff
@@ -480,3 +483,25 @@ pub async fn plan_transaction(
...
+async fn save_unbonding_token_metadata_if_needed(
+    undelegate: &Undelegate,
+    storage: &IndexedDBStorage,
+) -> WasmResult<()> {
+    let metadata = undelegate.unbonding_token().denom();
+
+    if storage.get_asset(&metadata.id()).await?.is_none() {
+        let metadata_proto = metadata.to_proto();
+        let customized_metadata_proto = customize_symbol(metadata_proto);
+        let customized_metadata = Metadata::try_from(customized_metadata_proto)?;
```
grod220, line 502: It seems we can't access metadata fields so we convert to proto and back to do so? Hmm. Is this a pattern in the core repo too?

grod220 (approved):
Suggestions to improve `customize_symbol`. After that 👍

PR #897, opened 2024-04-05
Render the assets table as an actual table

jessepinho:
🚨 I'd recommend reviewing this [with the "hide whitespace" feature turned on](https://github.com/penumbra-zone/web/pull/897/files?diff=split&w=1).

In response to @grod220 's [comment](https://github.com/penumbra-zone/web/pull/886#pullrequestreview-1981798384), I did a quick spike on rendering the assets table using an actual `<table>`:

image

Here's how it'd look with multiple equivalent values:

image


grod220 (approved):
I do think this is clearer than relying on the tooltip. That said, our whole table layout can almost certainly will be improved once we get a designer. It feels off, but I don't really have notes to fix it for now.

PR #906, opened 2024-04-08
Get rid of horizontal scrollbar on dashboard page

### LINKED ISSUE #899: Remove horizontal scrollbar at bottom of dashboard page
Image

There's currently a scrollbar at the bottom of the dashboard page. This should be removed.

jessepinho:
See inline code comments.

## Before
Note the horizontal scrollbar at the bottom.

image

## After
image

Closes #899

### DIFF
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -43,7 +43,7 @@ export default function AssetsTable() {
           <TableHeader key={account.account} className='group'>
             <TableRow>
               <TableHead colSpan={2}>
-                <div className='flex max-w-full flex-col justify-center gap-2 md:flex-row pt-8 group-[:first-of-type]:pt-0'>
+                <div className='flex max-w-full flex-col justify-center gap-2 pt-8 group-[:first-of-type]:pt-0 md:flex-row'>
                   <div className='flex items-center justify-center gap-2'>
                     <AddressIcon address={account.address} size={20} />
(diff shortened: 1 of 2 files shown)


apps/minifront/src/components/dashboard/assets-table/index.tsx
```diff
@@ -43,7 +43,7 @@ export default function AssetsTable() {
           <TableHeader key={account.account} className='group'>
             <TableRow>
               <TableHead colSpan={2}>
-                <div className='flex max-w-full flex-col justify-center gap-2 md:flex-row pt-8 group-[:first-of-type]:pt-0'>
+                <div className='flex max-w-full flex-col justify-center gap-2 pt-8 group-[:first-of-type]:pt-0 md:flex-row'>
```
jessepinho, line 46: This was just one of those tailwind class-order linting things. Figured I'd fix it while I was investigating this.

apps/minifront/src/components/layout.tsx
```diff
@@ -35,7 +35,7 @@ export const Layout = () => {
   return (
     <>
       <HeadTag />
-      <div className='flex min-h-screen w-screen flex-col'>
+      <div className='flex min-h-screen w-full flex-col'>
```
jessepinho, line 38: This is the actual fix. `w-screen` was setting the div's width to `100vw`, which causes overflow when there's a _vertical_ scrollbar. `w-full` sets it to `100%`, which accounts for the vertical scrollbar.

The reason we only saw this problem on the dashboard, is that the dashboard was the only page with a vertical scrollbar!

Valentine1898 (approved):
LGTM!

PR #909, opened 2024-04-09
fix epoch saving

### LINKED ISSUE #908: Incorrect current epoch when sending a delegatе transaction
I was getting an error when I sent the delegation transaction   
telegram-cloud-photo-size-2-5233619977951632205-x

This is because the epoch index is thrown off when the sync is restarted 
telegram-cloud-photo-size-2-5233619977951632207-y

Valentine1898:
close #908

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -423,7 +423,7 @@ export class BlockProcessor implements BlockProcessorInterface {
     endHeightOfPreviousEpoch: bigint,
     latestKnownBlockHeight: bigint,
   ): Promise<void> {
-    const { sctParams } = await this.querier.app.appParams();
+    const { sctParams } = (await this.indexedDb.getAppParams()) ?? {};
     const nextEpochStartHeight = endHeightOfPreviousEpoch + 1n;
 
     await this.indexedDb.addEpoch(nextEpochStartHeight);
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -423,11 +423,10 @@ export class BlockProcessor implements BlockProcessorInterface {
     endHeightOfPreviousEpoch: bigint,
     latestKnownBlockHeight: bigint,
   ): Promise<void> {
-    const { sctParams } = (await this.indexedDb.getAppParams()) ?? {};
     const nextEpochStartHeight = endHeightOfPreviousEpoch + 1n;
-
     await this.indexedDb.addEpoch(nextEpochStartHeight);
 
+    const { sctParams } = (await this.indexedDb.getAppParams()) ?? {};
     const nextEpochIsLatestKnownEpoch =
       sctParams && latestKnownBlockHeight - nextEpochStartHeight < sctParams.epochDuration;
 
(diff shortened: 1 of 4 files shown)


packages/query/src/block-processor.ts
```diff
@@ -423,7 +423,7 @@ export class BlockProcessor implements BlockProcessorInterface {
     endHeightOfPreviousEpoch: bigint,
     latestKnownBlockHeight: bigint,
   ): Promise<void> {
-    const { sctParams } = await this.querier.app.appParams();
+    const { sctParams } = (await this.indexedDb.getAppParams()) ?? {};
```
Valentine1898, line 426: I also fixed this, we don't need to request `sctParams` from the full node since we store them in indexed-db

packages/storage/src/indexed-db/index.ts
```diff
@@ -477,6 +477,9 @@ export class IndexedDb implements IndexedDbInterface {
     const previousEpoch = cursor?.value ? Epoch.fromJson(cursor.value) : undefined;
     const index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
 
+    // avoid saving the same epoch twice
+    if (previousEpoch?.startHeight === startHeight) return;
+
```
Valentine1898, line 482: At first I thought it would be best to link to `epochRoot`. 
But it seems that binding to a unique `startHeight` is just as good, plus I'm not sure that `epochRoot` is guaranteed to be unique for every epoch

grod220 (commented):
Will let Jesse review this one, but good catch!

packages/query/src/block-processor.ts
jessepinho: @grod220 why's that?

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -545,6 +545,13 @@ describe('IndexedDb', () => {
         expect(result2?.index).toBe(2n);
         expect(result3?.index).toBe(3n);
       });
+
+      it('should not save the epoch with the same startHeight twice', async () => {
+        await db.addEpoch(epoch3.startHeight);
+
+        const result3 = await db.getEpochByHeight(350n);
+        expect(result3?.index).toBe(3n);
```
jessepinho, line 553:
```suggestion
        const result = await db.getEpochByHeight(350n);
        expect(result?.index).toBe(3n);
```

jessepinho (approved):
Great catch — minor comment re: a variable name, then ship it!

packages/query/src/block-processor.ts
grod220: The params are used further down. There seems to be code in between that doesn't use the params.

jessepinho:
oh sorry @Valentine1898 , before you merge, could you please bump the IndexedDB version number so that people's DBs re-sync from scratch without the problematic epochs?

Valentine1898:
> oh sorry @Valentine1898 , before you merge, could you please bump the IndexedDB version number so that people's DBs re-sync from scratch without the problematic epochs?

Yes, it makes sense even though the indexed-db structure has not changed

PR #910, opened 2024-04-09
Publish storage package to NPM

grod220:
Closes https://github.com/penumbra-zone/web/issues/892

Removes `/src/ from imports too

Inspired by Kent Dodds https://www.briefs.fm/3-minutes-with-kent/221, this PR does not bundle before publishing. Given all of our consumers will be TS+ESM users, we can simply just expose our src files.


Valentine1898 (approved):
LGTM!

PR #914, opened 2024-04-09
Remove two-way swap code

### LINKED ISSUE #895: Delete "one-way swap"/"two-way swap" code (?)
The swap view component has code to classify swaps as "one way" or "two way" and render both. But there is no tooling (now or in the future) that should ever produce "two way" swaps.

So we could optimize for simplicity, by having the swap view only render the case we expect and fall back to an "unsupported swap" in the extremely unlikely event that a user is viewing a weird swap that they crafted with their own custom tooling. Otherwise we potentially carry forward maintenance burden of code users will never see.

jessepinho:
Henry asked that we remove the code related to two-way swaps, so as to reduce the future burden of code maintenance for a use case that should never happen.

Closes #895


Valentine1898 (approved):
LGTM

Issue #866, opened 2024-04-01
app.testnet pre-approved origin is not detected properly


turbocrime:
<img width="1217" alt="Screenshot 2024-04-01 at 10 58 14" src="https://github.com/penumbra-zone/web/assets/134443988/f6433cc7-2bda-4b14-b14e-5e1f0dad7c21">

<img width="470" alt="Screenshot 2024-04-01 at 10 58 37" src="https://github.com/penumbra-zone/web/assets/134443988/23f76668-7c8b-47e8-9306-93a9f506db4d">

hdevalence:
hmm, do we want to have pre-approved origins? might be worth discussing at a product level

hdevalence:
hmm, do we want to have pre-approved origins? might be worth discussing at a product level

PR #915, opened 2024-04-09
Add link to frontend in popup

### LINKED ISSUE #628: Add link to frontend in popup
Users want to be able to easily access a link to a frontend from the extension popup.

A link to the RPC's frontend should be available on the popup index page.

jessepinho:
Per #628, we want to show a link to the Penumbra frontend of the user's choice. By default, the Penumbra frontend is `app.testnet.penumbra.zone` (AKA the first pre-filled "known site"). Any 

## In this PR
- Rename `NetworksPopover` to `Network` (seems that the previous name was a holdover from previous functionality it had), and give it the capability of rendering a link to the frontend.
- Refactor `<SettingsConnectedSites />` to break it up into smaller components.
- Significantly refactor how the `ConnectedSites` slice works in the extension. Previously, Zustand actions were saving data in denormalized ways, presumably to cache their results. This may have been to avoid unnecessary re-renders. However, this is no longer necessary now that we're using Zustand selectors properly (via, e.g., [`useStoreShallow()`](https://github.com/penumbra-zone/web/blob/main/apps/extension/src/utils/use-store-shallow.ts)). Now, all derived data is calculated via selectors and helper functions.
- Convert Tailwind config to TypeScript so it can be used in JS via the new `RESOLVED_TAILWIND_CONFIG` const.

## Punting for now
- At the moment, despite `app.testnet.penumbra.zone` being a prefilled "known site", navigating to https://app.testnet.penumbra.zone/ still shows the "Connect" UI:
    image
    
    Not sure why this is, but it's a pre-existing bug, so I didn't address it in this PR.
    
    UPDATE: Looks like there's already a ticket for this: https://github.com/penumbra-zone/web/issues/866

Closes #628


packages/ui/components/ui/network.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+const meta: Meta<typeof Network> = {
+  component: Network,
+  title: 'Network',
+  tags: ['autodocs'],
+};
+export default meta;
+
+type Story = StoryObj<typeof Network>;
+
+export const Basic: Story = {
```
jessepinho, line 13: image

apps/extension/src/routes/popup/home/index-header.tsx
```diff
@@ -1,27 +1,30 @@
...
+  const frontendUrl = useStore(state => state.connectedSites.frontendUrl);
 
   return (
     <header className='top-0 z-40 w-full'>
-      <div className='flex items-center justify-between pt-4'>
+      <div className='flex items-center justify-between gap-4 pt-4'>
         <HamburgerMenuIcon
           onClick={() => navigate(PopupPath.SETTINGS)}
-          className='size-6 cursor-pointer hover:opacity-50'
+          className='size-6 shrink-0 cursor-pointer hover:opacity-50'
```
jessepinho, line 19: Accounting for various spacing issues here to make sure that the `<Network />` component fits even when it has the external link icon.

packages/ui/components/ui/network.tsx
```diff
@@ -0,0 +1,38 @@
...
+  name: string;
+  connectIndicator?: boolean;
+  href?: string;
+}
+
+/**
+ * Displays the network (chain ID) we're connected to, as well as an optional
+ * connection indicator and a link to the frontend.
+ */
+export const Network = ({ name, href, connectIndicator = true }: NetworkProps) => {
```
jessepinho, line 14: This was previously `NetworksPopover`

apps/extension/src/routes/popup/settings/settings-connected-sites/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+const getGroupedSites = (knownSites: OriginRecord[], filter?: string) => {
+  const groupedSites = Map.groupBy(knownSites, ({ choice }) => choice);
+  const filterFn = (site: OriginRecord) => !filter || site.origin.includes(filter);
+
+  return {
+    approvedSites: groupedSites.get(UserChoice.Approved)?.filter(filterFn) ?? [],
+    deniedSites: groupedSites.get(UserChoice.Denied)?.filter(filterFn) ?? [],
+    ignoredSites: groupedSites.get(UserChoice.Ignored)?.filter(filterFn) ?? [],
+  };
+};
```
jessepinho, line 28: This is a helper function rather than a selector because, as a selector, it would force a re-render every time the state changes. That's because these arrays get redefined every time the state changes, so even `useStoreShallow` would detect changes to state and re-render.

apps/extension/src/state/connected-sites.ts
```diff
@@ -2,74 +2,41 @@ import { ExtensionStorage } from '@penumbra-zone/storage/src/chrome/base';
...
       set(state => {
-        state.connectedSites.knownSites = knownSites;
-        state.connectedSites.approvedSites = groupedSites.get(UserChoice.Approved) ?? [];
-        state.connectedSites.deniedSites = groupedSites.get(UserChoice.Denied) ?? [];
-        state.connectedSites.ignoredSites = groupedSites.get(UserChoice.Ignored) ?? [];
+        state.connectedSites.filter = search;
       });
     },
 
-    setFilter: (search?: string) => {
```
jessepinho, line 42: `setFilter` was previously calculating a bunch of derived values and setting _them_ to state, which should really be done in selectors. So I simplified `setFilter` to just update the filter property.

apps/extension/src/state/persist.ts
```diff
@@ -70,6 +74,39 @@ function syncLocal(changes: Record<string, chrome.storage.StorageChange>, set: S
...
+  if (changes['frontendUrl']) {
+    const stored = changes['frontendUrl'].newValue as
+      | StorageItem<LocalStorageState['frontendUrl']>
+      | undefined;
+    set(
+      produce((state: AllSlices) => {
+        state.connectedSites.frontendUrl = stored?.value ?? state.connectedSites.frontendUrl;
+      }),
+    );
+  }
```
jessepinho, line 109: Adding a few missing properties to what gets synced on storage changes.

We really need to refactor this persist implementation to auto-sync between state and local storage. Otherwise, we have sort of spaghetti code. For example, [`setFrontendUrl`](https://github.com/penumbra-zone/web/pull/915/files#diff-7ec170797ae75c84fd4d1e8acd3cdfbb74fe61bd02d734c2199446b68b34ef74R26-R28) only sets a key in local storage; it doesn't actually update the Zustand state, because this persist implementation syncs from local storage to state. But that's hard for devs to understand.

IMO, ideally, our persistence implementation should have a mapping of keys in local storage to their location in state, and automatically handle the syncing whenever store values are updated — rather than store methods having to call `local.set()`. But that refactor seemed out of scope for this PR.

packages/ui/components/ui/network.tsx
Valentine1898: I ran the extension first and only then read the PR. 
Actually, it was not very obvious to me that clicking on network should lead to minifront (I actively use other wallet extensions).
I think the closest analogy for minifront is the metamask portfolio or keplr portfolio, and it would probably be familiar to me if the transition to minifront was similar. 

At the same time, it would be predictable for me if clicking on network would lead to the network status page. 
https://grpc.testnet.penumbra.zone

image
image

jessepinho: @Valentine1898 Great feedback — I'll whip something together for this and re-request review

Valentine1898:
If a user deletes a site marked as the primary Penumbra frontend, they still continue to go to that site when they click on the network. 
This sounds like a bug, but at the same time the solution options seem bad
- block deletion of a site that is marked as primary 
- automatically set a new primary from the list of sites that remain

packages/ui/components/ui/network.tsx
jessepinho: @Valentine1898 I've redesigned it like this:

image

jessepinho:
@Valentine1898 

> If a user deletes a site marked as the primary Penumbra frontend, they still continue to go to that site when they click on the network.

Great catch — didn't think of that. I just pushed up a commit to automatically pick the first remaining known site as the primary frontend if the currently selected one gets deleted. I actually think that's a fine way to handle that — not a bad solution.

Valentine1898 (approved):
image
I like where the button to go to the primary frontend is located. 
Perhaps we should think about calling it something other than Portfolio. Because the user can choose anything as primary frontend, not just minifront. But I have no idea what to call it, so we can rename this button later in a separate PR. 

I'll approve this PR so I don't block it until tomorrow, but note that some CI checks were not successful

PR #916, opened 2024-04-09
794 improve pricing method

### LINKED ISSUE #794: Improve pricing method
- [ ] need to add support for multiple numeraires  
- [ ] add filtering of outdated prices

Valentine1898:
This PR contains some improvements on asset pricing 
- added support for multiple NUMERAIRES, allowing customers to determine for themselves which price estimate to use for display 
- added price relevance threshold depending on asset type. Crypto assets are very volatile and we can't use too old prices to display estimated values 

Close #794

### DIFF
--- a/packages/constants/src/assets.ts
+++ b/packages/constants/src/assets.ts
@@ -9,6 +9,11 @@ export const localAssets: Metadata[] = LocalAssetRegistry.map(a =>
 export const NUMERAIRE_DENOMS: string[] = ['test_usd', 'usdc'];
 export const NUMERAIRES: Metadata[] = localAssets.filter(m => NUMERAIRE_DENOMS.includes(m.display));
 
+export const PRICE_RELEVANCE_THRESHOLDS = {
+  delegationToken: 719,
+  default: 200,
+  // Додайте інші типи токенів та їх пороги тут
+};
 export const STAKING_TOKEN = 'penumbra';
 export const STAKING_TOKEN_METADATA = localAssets.find(
   metadata => metadata.display === STAKING_TOKEN,
--- a/packages/constants/src/assets.ts
+++ b/packages/constants/src/assets.ts
@@ -9,10 +9,13 @@ export const localAssets: Metadata[] = LocalAssetRegistry.map(a =>
 export const NUMERAIRE_DENOMS: string[] = ['test_usd', 'usdc'];
 export const NUMERAIRES: Metadata[] = localAssets.filter(m => NUMERAIRE_DENOMS.includes(m.display));
 
+// PRICE_RELEVANCE_THRESHOLDS defines how long prices for different asset types remain relevant (in blocks)
+// 1 block = 5 seconds, 200 blocks approximately equals 17 minutes
 export const PRICE_RELEVANCE_THRESHOLDS = {
   delegationToken: 719,
   default: 200,
 };
+
 export const STAKING_TOKEN = 'penumbra';
 export const STAKING_TOKEN_METADATA = localAssets.find(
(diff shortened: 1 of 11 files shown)


packages/constants/src/assets.ts
```diff
@@ -6,6 +6,14 @@ export const localAssets: Metadata[] = LocalAssetRegistry.map(a =>
   Metadata.fromJson(a as JsonValue),
 );
 
+export const NUMERAIRE_DENOMS: string[] = ['test_usd', 'usdc'];
+export const NUMERAIRES: Metadata[] = localAssets.filter(m => NUMERAIRE_DENOMS.includes(m.display));
```
Valentine1898, line 10: I assume we will have two different local assets registy for testnet and mainnet. And we will have to revise the definition of `NUMERAIRES`

packages/router/src/grpc/view-protocol-server/balances.ts
```diff
@@ -28,16 +28,17 @@ import { Amount } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/nu
...
 import { Stringified } from '@penumbra-zone/types/src/jsonified';
 
 // Handles aggregating amounts and filtering by account number/asset id
 export const balances: Impl['balances'] = async function* (req, ctx) {
   const services = ctx.values.get(servicesCtx);
-  const { indexedDb } = await services.getWalletServices();
+  const { indexedDb, querier } = await services.getWalletServices();
 
-  const aggregator = new BalancesAggregator(ctx, indexedDb);
+  const latestBlockHeight = await querier.tendermint.latestBlockHeight();
```
Valentine1898, line 39: I think it's better to use `latestBlockHeight` rather than `fullSyncHeight` to avoid showing old prices during the sync process

packages/storage/src/indexed-db/index.ts
```diff
@@ -580,11 +585,31 @@ export class IndexedDb implements IndexedDbInterface {
...
     const results = await this.db.getAllFromIndex('PRICES', 'pricedAsset', base64AssetId);
 
-    return results.map(price => EstimatedPrice.fromJson(price));
+    const priceRelevanceThreshold = this.determinePriceRelevanceThresholdForAsset(assetMetadata);
+    const minHeight = latestBlockHeight - BigInt(priceRelevanceThreshold);
+
+    return results
+      .map(price => EstimatedPrice.fromJson(price))
+      .filter(price => price.asOfHeight >= minHeight);
+  }
```
Valentine1898, line 606: I decided to do this in-memory filtering because we won't have a lot of price records for each asset. And using a cursor or double index would make the code and its support much more complicated

grod220 (approved):
Nice job! One more thing I noticed. It looks like a good number of prices in the db have `1` for the `numerairePerUnit` field. The remaining all appear to be between 1 and ~1.2.

Screenshot 2024-04-10 at 8 57 39 AM

Valentine1898:
> Nice job! One more thing I noticed. It looks like a good number of prices in the db have `1` for the `numerairePerUnit` field. The remaining all appear to be between 1 and ~1.2.
> 
> Screenshot 2024-04-10 at 8 57 39 AM
> Feel free to merge and follow up with another PR if it requires a good deal more of investigation.

These are all prices for delegation tokens, which are calculated based on the validator's exchange rate. 
The fact that all these prices are in the range of 1.01 - 1.4 is exactly normal. As for price 1, I assume that if the validator is not in the active validator set, its ExchangeRate is equal to one. Going to check this out, and will open a new PR if necessary

Issue #779, opened 2024-03-17
Popup should handle resizing


turbocrime:
<img width="1728" alt="2024-03-16_14 34 28" src="https://github.com/penumbra-zone/web/assets/134443988/58b149d9-fb5a-4c91-b248-d18d8f06c3a0">

https://discord.com/channels/824484045370818580/1044457038886998137/1218538710736437348

PR #917, opened 2024-04-10
Use accessible design for approve/deny buttons

### LINKED ISSUE #711: Connection approval accessible colors
See thread from Discord: https://discord.com/channels/824484045370818580/979785365437165658/1216610553418612767

Screenshot_2024-03-11_at_12 55 24_AM

We should revise the connection approval designs to be more accessible for colorblind folks. All suggestions should be validated by some accessibility tool that measures contrast levels.

Suggestions:
- Remove gradient from "Approval", make solid color
- Add approve + deny icons next to the text
- Brighten sentence "Approval will allow this host..."
- Consider outline to highlight difference

jessepinho:
image

## In this PR
- Add a `destructiveSecondary` variant to the `<Button />` component.
- Create a Storybook story for the `<Button />` component.
- Use the new variant in `<ApproveDeny />`, and simplify that component to always use the same design.

Closes #711

### DIFF
--- a/apps/extension/src/routes/popup/approval/origin.tsx
+++ b/apps/extension/src/routes/popup/approval/origin.tsx
@@ -3,7 +3,6 @@ import { useStore } from '../../../state';
 import { originApprovalSelector } from '../../../state/origin-approval';
 import { ApproveDeny } from './approve-deny';
 import { LinkGradientIcon } from '../../../icons/link-gradient';
-import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 import { DisplayOriginURL } from '../../../shared/components/display-origin-url';
 import { cn } from '@penumbra-zone/ui/lib/utils';
 import { UserChoice } from '@penumbra-zone/types/src/user-choice';
@@ -89,14 +88,9 @@ export const OriginApproval = () => {
                 </div>
               </div>
             </div>
-            <div className='mt-3 flex flex-col gap-3'>
-              <div className='text-center text-muted-foreground'>
-                This host wants to connect to your wallet.
-              </div>
-              <div className='flex items-center gap-2 text-rust'>
-                <ExclamationTriangleIcon />
-                Approval will allow this host to see your balance and transaction history.
-              </div>
+            <div className='mt-3 flex flex-col gap-3 text-muted-foreground'>
+              <p>This host wants to connect to your wallet.</p>
+              <p>Approval will allow this host to see your balance and transaction history.</p>
             </div>
           </div>
         </div>
--- a/apps/extension/src/routes/popup/approval/origin.tsx
+++ b/apps/extension/src/routes/popup/approval/origin.tsx
@@ -100,13 +100,7 @@ export const OriginApproval = () => {
             </div>
           </div>
         </div>
-        <ApproveDeny
-          variants={['gradient']}
-          approve={approve}
-          deny={deny}
-          ignore={lastRequest && ignore}
-          wait={3}
-        />
+        <ApproveDeny approve={approve} deny={deny} ignore={lastRequest && ignore} wait={3} />
       </div>
     </FadeTransition>
   );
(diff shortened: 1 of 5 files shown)


packages/ui/components/ui/button.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+  component: Button,
+  title: 'Button',
+  tags: ['autodocs'],
+  argTypes: {},
+};
+export default meta;
+
+type Story = StoryObj<typeof Button>;
+
+export const Basic: Story = {
```
jessepinho, line 15: image

packages/ui/.storybook/main.js
```diff
@@ -25,5 +25,8 @@ const config = {
   docs: {
     autodocs: 'tag',
   },
+  typescript: {
+    reactDocgen: 'react-docgen-typescript',
```
jessepinho, line 29: Auto-generates Storybook docs based on TypeScript typings of props, e.g.:
image

packages/ui/components/ui/button.tsx
```diff
@@ -4,7 +4,7 @@ import { cva, type VariantProps } from 'class-variance-authority';
 import { cn } from '../../lib/utils';
 
 const buttonVariants = cva(
-  'inline-flex items-center justify-center rounded-lg font-bold ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50',
+  'inline-flex items-center justify-center rounded-lg px-4 font-bold ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50',
```
jessepinho, line 7: Finally, we no longer need to add padding to every button, lol

grod220: Does that mean we have to remove padding from all our buttons to coincide with this change?

grod220 (approved):
Looks good!

TalDerei:
Screenshot 2024-04-11 at 11 26 44 AM
Screenshot 2024-04-11 at 11 29 36 AM

comparing these two, I rather prefer keeping the orange approval dialogue and warning symbol, but we should increase the brightness and shade of the color. Additionally, currently the padding creates dead black space on the right side in the approval dialogue. We should extend it to fill that space or center it.

apps/extension/src/routes/popup/approval/origin.tsx
```diff
@@ -89,24 +88,13 @@ export const OriginApproval = () => {
...
-              <div className='text-center text-muted-foreground'>
-                This host wants to connect to your wallet.
-              </div>
-              <div className='flex items-center gap-2 text-rust'>
-                <ExclamationTriangleIcon />
-                Approval will allow this host to see your balance and transaction history.
-              </div>
+            <div className='mt-3 flex flex-col gap-3 text-muted-foreground'>
+              <p>This host wants to connect to your wallet.</p>
+              <p>Approval will allow this host to see your balance and transaction history.</p>
```
TalDerei, line 93: I rather prefer keeping the orange approval dialogue and warning symbol, but we should increase the brightness and use darker color / tint

TalDerei (changes requested):

turbocrime (approved):

turbocrime:
> The resizing is also broken.

#779 known for a while

packages/ui/components/ui/button.tsx
jessepinho: I'd like to eventually, but it doesn't harm anything to leave them as-is for now, since they'll override the default padding.

jessepinho:
Addressed @TalDerei 's comments, but I'm not able to reproduce the issue of the black bar at the right. I'm going to merge this for now and follow up with @TalDerei re: the black bar separately.

TalDerei (approved):

PR #920, opened 2024-04-11
Align symbol with amount

grod220:
At the moment, amounts in the value view are not really lined up with the token symbol. The amount is too "high" comparatively. 
Screenshot 2024-04-11 at 4 33 37 PM

This PR drops it by a few pixels so they share the same baseline
Screenshot 2024-04-11 at 4 35 22 PM


Valentine1898 (approved):

Valentine1898:
Not related to this PR, but noticed a new warning in logs
image

PR #928, opened 2024-04-11
capture unhandled promise rejections in offscreen

turbocrime:
offscreen is rejecting a promise somewhere, but there is no mechanism to capture the failure

### DIFF
--- a/apps/extension/src/entry/offscreen-handler.ts
+++ b/apps/extension/src/entry/offscreen-handler.ts
@@ -7,13 +7,31 @@ import {
   isOffscreenRequest,
 } from '@penumbra-zone/types/src/internal-msg/offscreen';
 
+// propagates errors that occur in unawaited promises
+// see: https://stackoverflow.com/questions/39992417/how-to-bubble-a-web-worker-error-in-a-promise-via-worker-onerror
+const unhandledRejection = Promise.withResolvers<never>();
+self.addEventListener(
+  'unhandledrejection',
+  event => {
+    // the event object has two special properties:
+    // event.promise - the promise that generated the error
+    // event.reason  - the unhandled error object
+    unhandledRejection.reject(event.reason);
+  },
+  { once: true },
+);
+
 chrome.runtime.onMessage.addListener((req, _sender, respond) => {
   if (!isOffscreenRequest(req)) return false;
   const { type, request } = req;
   if (isActionBuildRequest(request)) {
     void (async () => {
       try {
-        const data = await spawnActionBuildWorker(request);
+        const data = await Promise.race([
+          spawnActionBuildWorker(request),
+          // this error might not correspond to the specific request it responds to
+          unhandledRejection.promise,
+        ]);
         respond({ type, data });
       } catch (e) {
         respond({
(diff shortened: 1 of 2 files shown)


packages/router/src/grpc/offscreen-client.ts
```diff
@@ -48,7 +48,10 @@ const releaseOffscreen = async () => {
 const sendOffscreenMessage = async <T extends OffscreenMessage>(
   req: InternalRequest<T>,
 ): Promise<InternalResponse<T>> =>
-  chrome.runtime.sendMessage<InternalRequest<T>, InternalResponse<T>>(req);
+  chrome.runtime.sendMessage<InternalRequest<T>, InternalResponse<T>>(req).catch(() => ({
+    type: req.type,
+    data: null,
+  }));
```
turbocrime, line 54: this doesn't suppress the serialized errors, which are thrown at https://github.com/penumbra-zone/web/blob/da3a476fc14c0e7c09c0b7bede07973841741029/packages/router/src/grpc/offscreen-client.ts#L80

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -61,29 +62,35 @@ const buildActions = (
...
       type: 'BUILD_ACTION',
       request: {
-        transactionPlan: transactionPlan.toJson() as Jsonified<TransactionPlan>,
-        witness: witness.toJson() as Jsonified<WitnessData>,
-        fullViewingKey: fullViewingKey.toJson() as Jsonified<FullViewingKey>,
+        ...partialRequest,
         actionPlanIndex,
-      } satisfies ActionBuildRequest,
-    });
-    if ('error' in buildRes) throw ConnectError.from(buildRes.error);
```
turbocrime, line 77: `throw ConnectError.from(buildRes.error);` was technically wrong. needed to use `errorFromJson`

apps/extension/src/entry/offscreen-handler.ts
```diff
@@ -28,24 +41,28 @@ chrome.runtime.onMessage.addListener((req, _sender, respond) => {
 });
 
 const spawnActionBuildWorker = (req: ActionBuildRequest) => {
```
jessepinho, line 43: The changes to this function don't change any functionality, right? They're just a refactor?

apps/extension/src/entry/offscreen-handler.ts
```diff
@@ -13,13 +13,26 @@ chrome.runtime.onMessage.addListener((req, _sender, respond) => {
...
+        // propagate errors that occur in unawaited promises
+        const unhandled = Promise.withResolvers<never>();
+        self.addEventListener('unhandledrejection', unhandled.reject, {
+          once: true,
+        });
+
+        const data = await Promise.race([
+          spawnActionBuildWorker(request),
+          unhandled.promise,
+        ]).finally(() => self.removeEventListener('unhandledrejection', unhandled.reject));
```
jessepinho, line 25: Not sure I understand this. If you're already `await`ing `spawnActionBuildWorker`, won't the rejections throw an error that will then move use down into the `catch` block below? or are they "detached" somehow?

If the latter, this is a genius solution 🤯 

However, isn't there a risk that, while waiting for the action to build, an unrelated promise rejection could happen that this would catch, resulting in this action build getting killed?

packages/router/src/grpc/offscreen-client.ts
```diff
@@ -61,29 +62,35 @@ const buildActions = (
   fullViewingKey: FullViewingKey,
   cancel: PromiseLike<never>,
 ): Promise<Action>[] => {
-  const active = activateOffscreen();
-  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  const activation = activateOffscreen();
+
+  // this is a lot of binary -> base64 serialization, so just do it once and reuse
```
jessepinho, line 67:
```suggestion
  // this is a lot of JSON serialization, so just do it once and reuse
```

(Right? We're not doing binary -> base64 here, right?)

jessepinho (commented):
Mostly LGTM, but I'm not sure I understand how the unhandled rejections part works. See my comment on `offscreen-handler`

apps/extension/src/entry/offscreen-handler.ts
turbocrime: correct, it's just a switch to the `Promise.withResolvers` syntax, could be reverted.

packages/router/src/grpc/offscreen-client.ts
turbocrime: these pb types include many binary fields, and pb json involves binary -> base64. binary -> base64 is notably slow

jessepinho: Gotcha — but we're still encoding to JSON, not base64, right?

apps/extension/src/entry/offscreen-handler.ts
jessepinho: Ah no need to revert, I just wanted to make sure I was understanding it correctly

apps/extension/src/entry/offscreen-handler.ts
jessepinho: OK, thanks, this makes sense. FWIW, it's also possible that a completely unrelated promise rejection from some other JS process could trigger this failure too. But A) that seems like it'd be uncommon, and B) the worst case scenario is that the transaction fails to build, and must be re-submitted. That also seems acceptable.

turbocrime: i believe unhandled promise rejections are document-scoped so this shouldn't be an issue

packages/router/src/grpc/offscreen-client.ts
turbocrime: the mentioned base64 is inside json yes. i'll update the comment

PR #929, opened 2024-04-12
Fix React complaint about missing key prop

jessepinho:


grod220:
Think this is an old branch, because I merged a fix for this already

jessepinho:
Whoops! My mistake.

PR #931, opened 2024-04-12
Create a script to zip minifront/node-status dist directories for deploying to core

### LINKED ISSUE #857: Create script to build minifront and node-status for core
Penumbra core stores the minifront and node-status apps as ZIP files via Git LFA [here](https://github.com/penumbra-zone/penumbra/tree/main/assets).

This means that, when we want to deploy a new version of minifront or node-status, we need to:
1. Run `pnpm build` on web.
1. `cd` into the `dist` directories in each of `minifront` and `node-status`. Run `zip -r minifront.zip *` from inside the `apps/minifront/dist` directory (and do the same inside `node-status`) to zip the app.
1. PR those zip files to core.

Let's make a script that takes care of steps 1-2. If step 3 turns out to be especially painful in the future, we can automate that too.

As part of this PR, also update the [deployment issue template](https://github.com/penumbra-zone/web/blob/cd7b5d25fd2fd20d737c62f84419f6c9d217aba5/.github/ISSUE_TEMPLATE/deployment.md) to mention using this script rather than manually zipping the `dist` directories.

jessepinho:
This PR creates a script to build and ZIP the dist directories of minifront and node-status, so as to prevent human error and ease the deployment process.

Closes #857


TalDerei (approved):
LGTM

PR #932, opened 2024-04-12
Fix: Syncing genesis state causes unresponsive minifront

### LINKED ISSUE #921: Syncing genesis state causes unresponsive minifront
We likely need to implement timeouts

https://github.com/penumbra-zone/web/assets/16624263/9efcd8e2-50d1-4edf-8df0-bee0e5820676

jessepinho:
Per #921, the entire minifront layout is blank while syncing genesis state. Since genesis state is large and may take a long time to sync, the user experience while syncing genesis state is pretty poor.

Fortunately, the fix was easy. It turns out that the root layout loader was querying for the chain ID. Since the chain ID query wouldn't respond until the extension was done syncing genesis state, the query was blocking the layout component from rendering. So I moved the chain ID query to the header (which is where it's used), so that the chain ID no longer blocks the layout.

(I wonder if we should stop putting queries into our loaders, as the loaders block their routes from rendering until loading is complete — which creates UX problems like this.)

## Manual testing
I added a 1,000-second timeout to the block processor before it syncs the first block by adding this line...
```ts
    await new Promise(resolve => setTimeout(resolve, 1_000_000));
```

...just before the `for await (const compactBlock...` line here:

https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L165

Then I opened the extension and cleared the cache. This got it to stay stuck in the "Syncing genesis state" state for 1,000 seconds, allowing me to test that the minifront UI still loaded.

Closes #921

### DIFF
--- a/apps/minifront/src/components/header/top-row.tsx
+++ b/apps/minifront/src/components/header/top-row.tsx
@@ -11,6 +11,8 @@ import {
 import { Network } from '@penumbra-zone/ui/components/ui/network';
 import { PagePath } from '../metadata/paths';
 import { TabletNavMenu } from './tablet-nav-menu';
+import { useEffect, useState } from 'react';
+import { getChainId } from '../../fetchers/chain-id';
 import { LayoutLoaderResult } from '../layout';
 
 // Infinite-expiry invite link to the #web-ext-feedback channel. Provided by
@@ -18,8 +20,19 @@ import { LayoutLoaderResult } from '../layout';
 // if there are any problems with this link.
 const WEB_EXT_FEEDBACK_DISCORD_CHANNEL = 'https://discord.gg/XDNcrhKVwV';
 
+const useChainId = (): string | undefined => {
+  const { isInstalled, isConnected } = useLoaderData() as LayoutLoaderResult;
+  const [chainId, setChainId] = useState<string | undefined>();
+
+  useEffect(() => {
+    if (isInstalled && isConnected) void getChainId().then(setChainId);
+  }, [isInstalled, isConnected]);
+
+  return chainId;
+};
+
 export const TopRow = () => {
-  const result = useLoaderData() as LayoutLoaderResult;
+  const chainId = useChainId();
 
   return (
     <div className='flex w-full flex-col items-center justify-between px-6 md:h-[82px] md:flex-row md:gap-12 md:px-12'>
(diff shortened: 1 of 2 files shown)


turbocrime (approved):

apps/minifront/src/components/header/top-row.tsx
```diff
@@ -11,15 +11,28 @@ import {
...
 // if there are any problems with this link.
 const WEB_EXT_FEEDBACK_DISCORD_CHANNEL = 'https://discord.gg/XDNcrhKVwV';
 
+const useChainId = (): string | undefined => {
+  const { isInstalled, isConnected } = useLoaderData() as LayoutLoaderResult;
+  const [chainId, setChainId] = useState<string | undefined>();
+
+  useEffect(() => {
+    if (isInstalled && isConnected) void getChainId().then(setChainId);
+  }, [isInstalled, isConnected]);
```
grod220, line 29: We should be using react-query for async data loading when it's in a component

grod220:
I think this solution is fine for the time being, however, it doesn't address the core issue. Our extension messaging seems to queue up requests for an infinite amount of time. This solution sort of works around it. However, in the future, we'll have many frontends (hopefully) that should expect a more typical request/response + timeout behavior.

> I wonder if we should stop putting queries into our loaders, as the loaders block their routes from rendering until loading is complete — which creates UX problems like this.

Async data loading patterns within react-land are kinda rough. They create the classic react waterfall rendering+loading issue that Remixjs was created to address. Loaders are preferred given data loading happens in parallel for all nested routes. However, for long running requests, like say the transactions page, we should probably move away from loaders (valentine opened an issue about this).

apps/minifront/src/components/header/top-row.tsx
jessepinho: Ah, my bad — I hadn't seen it being used in any recent code, so I assumed it was effectively being deprecated

jessepinho:
Ah, I wasn't aware that loaders work in parallel for nested routes. That's a nice benefit.

That said, `useQuery` would fire in parallel for all components that are loaded in nested routes, right? i.e., if component A is the parent and B is the child, and both use `useQuery`, nothing would stop those queries from firing once those components mount, right? (As long as A isn't waiting to render `<Outlet />` until its query resolves.)

PR #933, opened 2024-04-13
Fix layout and resizing issues in the extension popup

### LINKED ISSUE #779: Popup should handle resizing

jessepinho:
Explanation for changes are inline. In short:
- I made the popup root div have a _min_ width/height instead of a hard-coded width/height, thus allowing it to scale with the popup window size.
- I fixed the issue where the scrollbars were hidden behind the fixed-position approve/deny buttons, by removing the fixed positioning and using flexbox instead.
- You may notice in the screencap that there is still some strange scrolling behavior when you shrink the window _below_ 400px wide or 600px tall. That's due to the min widths that we apparently need for the attached window height. I believe that, to work around this, we'd need to make a separate root for detached vs. attached popup windows. We could do that, but I'd suggest we make that a separate PR (if reviewers even think it's necessary), since that'll require quite a bit of extra work.

https://github.com/penumbra-zone/web/assets/1121544/f30622fc-3798-4e04-bca1-6c440c31509c

Closes #779

### DIFF
--- a/apps/extension/src/routes/popup/approval/approve-deny.tsx
+++ b/apps/extension/src/routes/popup/approval/approve-deny.tsx
@@ -17,7 +17,7 @@ export const ApproveDeny = ({
   useEffect(startCountdown, [startCountdown]);
 
   return (
-    <div className='fixed inset-x-0 bottom-0 flex flex-row flex-wrap justify-center gap-4 bg-black p-4 shadow-lg'>
+    <div className='flex flex-row flex-wrap justify-center gap-4 bg-black p-4 shadow-lg'>
       <Button variant='gradient' className='w-full' size='lg' onClick={approve} disabled={!!count}>
         Approve {count !== 0 && `(${count})`}
       </Button>
--- a/apps/extension/src/routes/popup/approval/origin.tsx
+++ b/apps/extension/src/routes/popup/approval/origin.tsx
@@ -98,7 +98,9 @@ export const OriginApproval = () => {
             </div>
           </div>
         </div>
-        <ApproveDeny approve={approve} deny={deny} ignore={lastRequest && ignore} wait={3} />
+        <div className='flex grow flex-col justify-end'>
+          <ApproveDeny approve={approve} deny={deny} ignore={lastRequest && ignore} wait={3} />
+        </div>
       </div>
     </FadeTransition>
   );
--- a/apps/extension/src/routes/popup/approval/transaction/index.tsx
+++ b/apps/extension/src/routes/popup/approval/transaction/index.tsx
@@ -32,8 +32,8 @@ export const TransactionApproval = () => {
   };
 
   return (
-    <div className='flex h-screen flex-col justify-between'>
-      <div className='grow overflow-auto p-[30px] pb-44 pt-10'>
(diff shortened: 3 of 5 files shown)


apps/extension/src/routes/popup/approval/origin.tsx
```diff
@@ -100,13 +100,16 @@ export const OriginApproval = () => {
...
         </div>
-        <ApproveDeny
-          variants={['gradient']}
-          approve={approve}
-          deny={deny}
-          ignore={lastRequest && ignore}
-          wait={3}
-        />
+
+        <div className='flex grow flex-col justify-end'>
```
jessepinho, line 104: Pushes the Approve/Deny buttons all the way to the bottom of the window without having to use `position: fixed`

apps/extension/src/routes/popup/approval/approve-deny.tsx
```diff
@@ -19,7 +19,7 @@ export const ApproveDeny = ({
   useEffect(startCountdown, [startCountdown]);
 
   return (
-    <div className='fixed inset-x-0 bottom-0 flex flex-row flex-wrap justify-center gap-4 bg-black p-4 shadow-lg'>
```
jessepinho, line 22: Now, this div will only go up to where the approve/deny buttons are, so the scroll bar no longer goes behind them.

apps/extension/src/routes/popup/approval/transaction/index.tsx
```diff
@@ -32,8 +32,8 @@ export const TransactionApproval = () => {
   };
 
   return (
-    <div className='flex h-screen flex-col justify-between'>
-      <div className='grow overflow-auto p-[30px] pb-44 pt-10'>
+    <div className='flex h-screen flex-col'>
+      <div className='grow overflow-auto p-[30px] pt-10'>
```
jessepinho, line 36: This pushes the approve/deny buttons to the button using flex box

apps/extension/src/routes/popup/popup-layout.tsx
```diff
@@ -12,7 +12,7 @@ import { Outlet } from 'react-router-dom';
  * then removing the hard-coded width from `globals.css`.
  */
 export const PopupLayout = () => (
-  <div className='flex min-h-full flex-col bg-card-radial'>
+  <div className='flex grow flex-col bg-card-radial'>
```
jessepinho, line 15: Not sure why `height: 100%` doesn't fill a flexbox, but apparently it doesn't. So I had to use `grow` instead.

packages/ui/styles/globals.css
```diff
@@ -59,8 +59,11 @@
   }
 
   #popup-root {
-    width: 400px;
-    height: 600px;
+    min-width: 400px;
+    min-height: 600px;
+    height: 100%;
+    display: flex;
+    flex-direction: column;
```
jessepinho, line 66: `min-height` makes the attached popup the correct size, but doesn't restrict the root div from growing larger in the detached popup.

TalDerei (approved):

PR #934, opened 2024-04-13
Expand toasts by default

### LINKED ISSUE #830: Swap Claim UX Improvements

jessepinho:
Instead of using Sonner's default UI, which hides old toasts behind new ones, this PR expands all toasts by default. That way, when doing a swap + swap claim, you can clearly see both notifications:

image

I actually think this is a better UI choice anyway. There isn't really a compelling need to hide old toasts behind new ones.

Closes #830


TalDerei (approved):
Looks great!

PR #935, opened 2024-04-13
Stream transactions in transactions table to avoid blocking loading

### LINKED ISSUE #911: Do not wait for all transactions to load when going to the Transacions page
https://github.com/penumbra-zone/web/blob/716e14d15892e4e079be64b7741d3ed9afa97237/apps/minifront/src/fetchers/transactions.ts#L12-L23
We wait for all transactions to load when we open the Transactions page. 
I noticed a small freeze when going to the transactions page on my test wallet. If you have about a hundred transactions, this will become a problem. 

I think we should use the same approach used on the Staking page, dynamically update the transaction list as it loads. 

https://github.com/penumbra-zone/web/assets/25391690/472f6400-e12d-424d-9430-6a740a48b1a6

I tested on m3 pro, probably it can be even slower on weaker CPUs

 You may find it useful to use a script that will help you automatically generate many transactions (use swap to generate two transactions at a time)
```shell
#!/bin/bash

while true; do
  cargo run --release --bin pcli -- --home spam tx swap --into gm 0.1penumbra
  sleep 10
done

```

jessepinho:
https://github.com/penumbra-zone/web/assets/1121544/3b7d8252-a23d-4418-a931-6458b578fc0c

Reviewers, what do you think of this?

@Valentine1898 pointed out in #911 that there's a delay when switching to the transactions table if you have a lot of transactions. So in this PR, I switched it to stream results, so that it's more responsive.

This results in a similar issue to what we were dealing with on the staking page, though: the page might actually take longer to load all the data, even though it _feels_ faster, because of rendering costs.

That said, I generated nearly 200 transactions and then loaded the page, and it finished loading in under a second. So I think we're OK.

Closes #911

### DIFF
--- a/apps/minifront/src/components/dashboard/transaction-table.tsx
+++ b/apps/minifront/src/components/dashboard/transaction-table.tsx
@@ -10,7 +10,8 @@ import { Link, LoaderFunction } from 'react-router-dom';
 import { throwIfPraxNotConnectedTimeout } from '@penumbra-zone/client';
 import { shorten } from '@penumbra-zone/types/src/string';
 import { useStore } from '../../state';
-import { useEffect } from 'react';
+import { memo, useEffect } from 'react';
+import { TransactionSummary } from '../../state/transactions';
 
 export const TxsLoader: LoaderFunction = async (): Promise<null> => {
   await throwIfPraxNotConnectedTimeout();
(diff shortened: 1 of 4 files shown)


apps/minifront/src/components/dashboard/transaction-table.tsx
```diff
@@ -6,67 +6,65 @@ import {
...
+import { useStore } from '../../state';
+import { memo, useEffect } from 'react';
+import { TransactionSummary } from '../../state/transactions';
 
-export const TxsLoader: LoaderFunction = async (): Promise<TransactionSummary[]> => {
+export const TxsLoader: LoaderFunction = async (): Promise<null> => {
   await throwIfPraxNotConnectedTimeout();
-  return await getAllTransactions();
+  return null;
 };
```
turbocrime, line 19: notConnected is already handled by the top-level layoutloader. if that's all that remains here, the loader can be deleted.

turbocrime (approved):

apps/minifront/src/components/dashboard/transaction-table.tsx
jessepinho: Good catch! Fixed

apps/minifront/src/state/transactions.ts
```diff
@@ -0,0 +1,41 @@
...
+      const summaries = [...get().transactions.summaries, summary].sort(
+        (a, b) => b.height - a.height,
+      );
+
+      set(state => {
+        state.transactions.summaries = summaries;
+      });
+    }
+  },
+});
```
grod220, line 41: I wonder if this is better than using [useStream](https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/fetchers/stream.ts)? 🤔 

It does feel more generic and we could make `useStream` accept a transformer or create our own custom hook that wraps around that. Not sure, what do you think?

jessepinho: Yeah, I'd considered that. The only reason I didn't is that I assumed we'd want this data in Zustand state. But, it could go either way 🤷🏻‍♂️

PR #939, opened 2024-04-15
684 rename penumbra zonerouter package + publish to npm

### LINKED ISSUE #730: disaggregate `@penumbra-zone/services-context` package into `HandlerContext`
the name is meaningless, overloaded, and causes confusion. most of the things residing here are only provided to rpc handler context

edit: expanded detail

relevant to #341 

the services package and `ServicesInterface` is essentially a vestige of the old rpc messaging system, and serves a similar conceptual purpose to `HandlerContext` - if you notice, most `HandlerContext` use is just acquiring the 'Services` object.

the context interface should become more well-specified and implementation-agnostic set of interfaces, and `Services` should become a prax-specific set of objects and functions fulfilling those context interfaces.

Valentine1898:
In this PR: 
- renamed `services` -> `services-context` (most likely temporarily, until resolve #730)
- renamed `router` -> `services`
- remove the `/grpc/` nesting since these implementations can actually use any transport, not just grpc (they don't actually use grpc now) 
- `services-context` и `services` publish to npm


packages/services-context/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@penumbra-zone/services-context",
```
grod220, line 2: If this kinda depends on https://github.com/penumbra-zone/web/issues/730, think we can hold off publishing until that issue is picked up right?

grod220 (approved):

packages/services-context/package.json
Valentine1898: I guess so. Do you think I should revert all changes for `services-context` related to publish?

grod220: I suppose my comment was that if `services-context` is not really finalized as a name, let's hold off on publishing until that work is picked up. Not sure if anything needs reverted though.

PR #945, opened 2024-04-16
revert #939, rename to 'rpc'

turbocrime:
use name `rpc` instead of `services` to avoid confusion and conflict

restore services-context to original name, for no change. needed changes are more extensive than a rename


Valentine1898:
> use name `rpc` instead of `services` to avoid confusion and conflict

The rpc will cause less confusion, I agree. But at the same time it seems that rpc does not represent well the contents of a package that is not related to rpc (not only not related to remote calls, but being completely isolated from the transport). The package contains service imlementations. 
Similar functionality in rust is also named as service 
https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/service.rs

> restore services-context to original name, for no change. needed changes are more extensive than a rename

Exactly, I temporarily renamed it, realizing that there would be a refactoring of this package as part of https://github.com/penumbra-zone/web/issues/730.

PR #946, opened 2024-04-17
Use new registry [part 1]

grod220:
The new chain/asset registry is up and running: https://github.com/prax-wallet/registry with a helper library to fetch the data `@penumbra-labs/registry`.

This PR takes the first step in migrating away from the hardcoded grpc urls in constants to using the public ones in the new repo.

Also converted the webpack config to typescript for some better typesafety there.


apps/extension/package.json
```diff
@@ -5,13 +5,15 @@
   "description": "chrome-extension",
   "type": "module",
   "scripts": {
-    "dev": "webpack --watch --mode=development -d inline-source-map",
+    "dev": "NODE_ENV=testnet pnpm bundle --watch --mode=development -d inline-source-map",
+    "build": "NODE_ENV=mainnet pnpm bundle",
     "clean": "rm -rfv dist bin",
-    "build": "webpack",
+    "bundle": "NODE_OPTIONS=\"--import=./src/utils/webpack-register.js\" webpack",
```
grod220, line 11: Now that webpack config is typescript, we need to use a special kind of register so it can the cli can load it

apps/extension/src/hooks/chain-id.ts
```diff
@@ -11,8 +26,8 @@ export const getChainId = async (): Promise<string> => {
 export const useChainIdQuery = () => {
   const { data, refetch } = useQuery({
     queryKey: ['chain-id'],
-    queryFn: getChainId,
-    refetchInterval: false,
```
grod220, line 15: Given chain ids don't change on refetch, the `refetchInterval` was intended to instruct react-query not to refetch. But that was a misreading of the api, `staleTime: Infinity` is the correct one.

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -41,22 +43,36 @@ export const GrpcEndpointForm = ({
     <>
       <div className='flex flex-col gap-2'>
         <form className='flex flex-col gap-4' onSubmit={handleSubmit}>
+          {rpcsQuery.isLoading && <LoadingIndicator />}
```
grod220, line 46: Screenshot 2024-04-17 at 1 15 48 PM

apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -41,22 +43,36 @@ export const GrpcEndpointForm = ({
...
+          {rpcsQuery.error && (
+            <div className='text-red-700'>
+              Error loading chain registry: {String(rpcsQuery.error)}
+            </div>
+          )}
```
grod220, line 51: Screenshot 2024-04-17 at 1 18 47 PM

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
```diff
@@ -26,20 +26,24 @@ const isValidUrl = (url: string) => {
 };
 
 export const useGrpcEndpointForm = () => {
```
grod220, line 28: This component is pretty complex. When we have more time, we should think of an alternative design/paradigm to manage this state. At the moment, it's react-hook hell.

apps/extension/webpack.config.ts
```diff
@@ -0,0 +1,153 @@
+import * as dotenv from 'dotenv';
+import path from 'path';
+import CopyPlugin from 'copy-webpack-plugin';
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+import url from 'url';
+// eslint-disable-next-line import/no-named-as-default
+import webpack from 'webpack';
```
grod220, line 7: sadly `import * as webpack from 'webpack';` does not work for this like most cases

apps/extension/webpack.config.ts
```diff
@@ -0,0 +1,153 @@
...
+
+// Loads default vars from `.env.testnet` & `.env.mainnet` file in this directory.
+// Reference package.json build script. TODO: add `.env.mainnet` when ready.
+const envPath = process.env['NODE_ENV'] === 'mainnet' ? '.env.testnet' : '.env.testnet';
+dotenv.config({ path: envPath });
```
grod220, line 12: Allows us to pass in different envs per chain (quite relevant when we go to mainnet)

apps/extension/webpack.config.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+const definitions = {
+  CHAIN_ID: JSON.stringify(process.env['CHAIN_ID']),
+  PRAX: JSON.stringify(process.env['PRAX']),
+  PRAX_ORIGIN: JSON.stringify(`chrome-extension://${process.env['PRAX']}`),
+  IDB_VERSION: JSON.stringify(Number(process.env['IDB_VERSION'])),
+  MINIFRONT_URL: JSON.stringify(process.env['MINIFRONT_URL']),
+};
+
+const __dirname = new URL('.', import.meta.url).pathname;
```
grod220, line 31: Everything below is the same

apps/extension/webpack.config.ts
```diff
@@ -0,0 +1,153 @@
...
+import path from 'path';
+import CopyPlugin from 'copy-webpack-plugin';
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+import url from 'url';
+// eslint-disable-next-line import/no-named-as-default
+import webpack from 'webpack';
+
+// Loads default vars from `.env.testnet` & `.env.mainnet` file in this directory.
+// Reference package.json build script. TODO: add `.env.mainnet` when ready.
+const envPath = process.env['NODE_ENV'] === 'mainnet' ? '.env.testnet' : '.env.testnet';
```
Valentine1898, line 11: typo here (twice .env.testnet)

apps/extension/src/shared/components/grpc-endpoint-form/use-grpc-endpoint-form.ts
grod220: Yes, I have it on my sticky note to do so. I have an inclination that a finite state machine pattern would be better. However, there's always the pesky "reactive" thing that needs to hook into the react lifecycle.

apps/extension/webpack.config.ts
grod220: that was intentional 😅 , I'll make this clearer though

Valentine1898: a, I saw the TODO, but thought it was only related to creating an `.env.mainnet` file

Valentine1898 (approved):

PR #948, opened 2024-04-17
ibc slices

turbocrime:
cut down cosmos-kit PR to hand it off and focus on other issues

none of this works but hopefully it's a starting point.

contains mostly zustand slices. cosmos-kit is present in deps but doesn't build into the output, as only a type is referenced in a slice.

some cosmos-kit present, but all use but types is commented out, so there should be minimal build effect.


apps/minifront/src/components/ibc/ibc-chain-provider.tsx
turbocrime, line 1: this contains a no-op component

apps/minifront/vite.config.ts
turbocrime, line 1: node polyfills necessary for cosmos-kit and juno-network (if we use juno-network)

apps/minifront/src/state/cosmos-kit.ts
turbocrime, line 1: chainContext can't actually go into zustand, and none of its functions can be called outside of a component anyway. so this stores static items

apps/minifront/src/components/header/constants.tsx
turbocrime, line 1: oh, this slightly changes the header still

PR #950, opened 2024-04-17
context key for full viewing key

### LINKED ISSUE #826: context key for fvk
currently, method impls that need fvk do something like this

```
  const { fullViewingKey } = (await services.getWalletServices()).viewServer;
```

we should just have a context key that provides fvk to impls, and fvk should not be available from another location.

relevant to #341

Valentine1898:
> currently, method impls that need fvk do something like this
> 
> ```
>   const { fullViewingKey } = (await services.getWalletServices()).viewServer;
> ```
> 
> we should just have a context key that provides fvk to impls, and fvk should not be available from another location.
> 
> relevant to #341

Close #826

### DIFF
--- a/packages/services/src/view-service/address-by-index.test.ts
+++ b/packages/services/src/view-service/address-by-index.test.ts
@@ -5,33 +5,22 @@ import {
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { createContextValues, createHandlerContext, HandlerContext } from '@connectrpc/connect';
 import { ViewService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/view/v1/view_connect';
-import { servicesCtx } from '../ctx/prax';
 import { addressByIndex } from './address-by-index';
 import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-import type { ServicesInterface } from '@penumbra-zone/types/src/services';
 import { testFullViewingKey } from '../test-utils';
+import { fvkCtx } from '../ctx/full-viewing-key';
 
 describe('AddressByIndex request handler', () => {
-  let mockServices: ServicesInterface;
   let mockCtx: HandlerContext;
 
   beforeEach(() => {
-    mockServices = {
-      getWalletServices: () =>
-        Promise.resolve({
-          viewServer: {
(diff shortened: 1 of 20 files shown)


packages/types/src/servers.ts
```diff
@@ -1,10 +1,8 @@
 import { ScanBlockResult } from './state-commitment-tree';
 import { CompactBlock } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/compact_block/v1/compact_block_pb';
 import { MerkleRoot } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/crypto/tct/v1/tct_pb';
-import { FullViewingKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 
 export interface ViewServerInterface {
-  fullViewingKey: FullViewingKey;
```
Valentine1898, line 7: There's no need for it anymore

packages/services/src/view-service/address-by-index.test.ts
```diff
@@ -5,33 +5,22 @@ import {
...
   beforeEach(() => {
-    mockServices = {
-      getWalletServices: () =>
-        Promise.resolve({
-          viewServer: {
-            fullViewingKey: testFullViewingKey,
-          },
-        }),
-    } as ServicesInterface;
-
```
Valentine1898, line 27: This has led to a simplification of some of rpc tests

packages/services/src/view-service/address-by-index.ts
```diff
@@ -1,14 +1,13 @@
...
-export const addressByIndex: Impl['addressByIndex'] = async (req, ctx) => {
-  const services = ctx.values.get(servicesCtx);
-  const {
-    viewServer: { fullViewingKey },
-  } = await services.getWalletServices();
-
+export const addressByIndex: Impl['addressByIndex'] = (req, ctx) => {
+  const fullViewingKey = ctx.values.get(fvkCtx);
+  if (!fullViewingKey) {
+    throw new Error('Cannot access full viewing key');
```
turbocrime, line 9:
```suggestion
    throw new ConnectError('Cannot access full viewing key', Code.Unauthenticated);
```

and the same at many similar locations

turbocrime (approved):
nice! a suggestion, use `ConnectError` and `Code.Unauthenticated` instead of generic `Error`

merge when ready

PR #951, opened 2024-04-18
Use new registry [part 2]

grod220:
This time for IBC connections information. Removes the hard-coded ibc connections in constants package.


Valentine1898 (approved):
LGTM 👍

PR #954, opened 2024-04-18
rm unused webpack rule

turbocrime:
empty rule is no-op

### DIFF
--- a/apps/extension/webpack.config.ts
+++ b/apps/extension/webpack.config.ts
@@ -67,7 +67,6 @@ const config: webpack.Configuration = {
   },
   module: {
     rules: [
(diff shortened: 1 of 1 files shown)


apps/extension/webpack.config.ts
```diff
@@ -67,7 +67,6 @@ const config: webpack.Configuration = {
   },
   module: {
     rules: [
-      { test: /\.wasm/ },
```
grod220, line 70: When you say `empty rule is no-op`, I think that means without specifying a loader it doesn't do anything? I guess the `asyncWebAssembly` setting is handling everything for us already?

grod220 (approved):

Issue #737, opened 2024-03-13
Publish wasm package to npm


grod220:

turbocrime:
there is currently a package available at npm `@penumbra-zone/wasm` that does not represent the current package

turbocrime:
completed in #955

PR #955, opened 2024-04-18
`@penumbra-zone/wasm`

turbocrime:
work on #737

the wasm package on npm is outdated, and we have consumers trying to use, so i took a shot at this

## keys package
* another pr #958

## wasm package
* removed /src/ path segment
* publishes without build as per discussion
* keys are now an optional dependency

verified working on an example nextjs app with the output of `pnpm pack` which you can view at https://github.com/turbocrime/nextjs-penumbra-wasm-example/


turbocrime:
@grod220 the 'keys' package should not be versioned with changeset, as it corresponds directly to the penumbra core repo version. but i'm not sure how publishing works when we're not using changeset to manage. could use some attention.

packages/wasm/package.json
```diff
@@ -1,27 +1,32 @@
...
-    "@penumbra-zone/keys": "workspace:*",
     "@penumbra-zone/types": "workspace:*",
     "zod": "^3.22.4"
   },
   "devDependencies": {
     "@penumbra-zone/bech32": "workspace:*",
     "fake-indexeddb": "^5.0.2"
+  },
+  "optionalDependencies": {
+    "@penumbra-zone/keys": "workspace:*"
```
grod220, line 30: Interesting, never heard of this category of deps

grod220: Is it required to be a dynamic import like so? https://github.com/turbocrime/nextjs-penumbra-wasm-example/blob/360201378c47cc217a704ceaf8b32bce72e120a9/app/components/ComponentWithWasm.tsx#L6

packages/wasm/package.json
```diff
@@ -1,27 +1,32 @@
...
   "version": "3.0.0",
-  "private": true,
   "license": "MIT",
   "type": "module",
   "scripts": {
     "clean": "rm -rfv dist wasm ; cargo clean --manifest-path ./crate/Cargo.toml",
     "dev": "cargo watch -C ./crate --postpone -- pnpm turbo run compile",
     "compile": "cd crate ; wasm-pack build --no-pack --target bundler --out-name index --out-dir ../wasm",
-    "build": "tsc",
+    "prepack": "touch ./wasm/.npmignore",
```
grod220, line 10: What does this do/why is it required?

grod220 (approved):
Nice job! After this is merged, the bot will update this PR: https://github.com/penumbra-zone/web/pull/940. Then if you merge that, this will publish to npm for the consumer needing this.

Also, could you add documentation to a README.md on how to use this package? Think it would be helpful to give instructions on:
- No keys usage
- Usage w/ keys (and downloading that optional package)
- Possible required webpack/next configs

packages/wasm/package.json
turbocrime: this is the way i could get it to work with next. will try more today

turbocrime: note that isn't the keys package import that's a submodule that happens to be named keys

packages/wasm/package.json
turbocrime: this prevents ignoring the compiled wasm when packaging by overriding the wildcard gitignore

turbocrime: prepack is called during publishing

PR #956, opened 2024-04-18
IBC out: round up to nearest 10min

### LINKED ISSUE #812: Implement rounding on IBC-timeouts
Reference https://github.com/penumbra-zone/penumbra/issues/4064

At the moment, the web code implements milliseconds. Turns out, it actually needs nanoseconds to work properly. However, this is a potential privacy leak. After core code implements rounding, implement a mirror implementation to `pcli withdraw` in the web code: 

https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/state/ibc.ts#L97-L98

grod220:
Closes #812

Reference implementation: https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pcli/src/command/tx.rs#L1055-L1067

Also fixed our method to getting the timeout blocks

### DIFF
--- a/apps/minifront/src/state/ibc.ts
+++ b/apps/minifront/src/state/ibc.ts
@@ -6,13 +6,12 @@ import {
 import { BigNumber } from 'bignumber.js';
 import { ClientState } from '@buf/cosmos_ibc.bufbuild_es/ibc/lightclients/tendermint/v1/tendermint_pb';
 import { Height } from '@buf/cosmos_ibc.bufbuild_es/ibc/core/client/v1/client_pb';
-import { ibcClient, viewClient } from '../clients';
+import { ibcChannelClient, ibcClient, ibcConnectionClient, viewClient } from '../clients';
 import {
   getDisplayDenomExponentFromValueView,
   getMetadata,
 } from '@penumbra-zone/getters/src/value-view';
 import { getAddressIndex } from '@penumbra-zone/getters/src/address-view';
-import { typeRegistry } from '@penumbra-zone/types/src/registry';
 import { toBaseUnit } from '@penumbra-zone/types/src/lo-hi';
 import { planBuildBroadcast } from './helpers';
 import { amountMoreThanBalance } from './send';
@@ -25,6 +24,7 @@ import {
 import { bech32IsValid } from '@penumbra-zone/bech32/src/validate';
 import { errorToast } from '@penumbra-zone/ui/lib/toast/presets';
 import { Chain } from '@penumbra-labs/registry';
+import { typeRegistry } from '@penumbra-zone/types/src/registry';
 
 export interface IbcSendSlice {
   selection: BalancesResponse | undefined;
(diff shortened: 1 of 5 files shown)


apps/minifront/src/state/ibc.ts
```diff
@@ -90,29 +90,57 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
...
+  const tenMinutes = BigInt(10 * 60 * 1000 * 1000000);
+
+  // Round up to nearest 10 minutes
+  return nowPlusTwoDays + tenMinutes - (nowPlusTwoDays % tenMinutes);
+};
+
+// Reference in core: https://github.com/penumbra-zone/penumbra/blob/1376d4b4f47f44bcc82e8bbdf18262942edf461e/crates/bin/pcli/src/command/tx.rs#L998-L1050
 const getTimeout = async (
   chain: Chain,
 ): Promise<{ timeoutTime: bigint; timeoutHeight: Height }> => {
```
grod220, line 112: Turns out there can be many client states with the same chain id. This new method reflects the way core does it: https://github.com/penumbra-zone/penumbra/blob/1376d4b4f47f44bcc82e8bbdf18262942edf461e/crates/bin/pcli/src/command/tx.rs#L998-L1050. It precisely gets the client state we need.

apps/minifront/src/state/ibc.ts
```diff
@@ -90,29 +90,57 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
...
+  const nowInNanoseconds = BigInt(Date.now()) * BigInt(1_000_000);
+
+  // New timestamp after adding 2 days
+  const twoDays = BigInt(2 * 24 * 60 * 60 * 1000 * 1000000);
+  const nowPlusTwoDays = nowInNanoseconds + twoDays;
+  const tenMinutes = BigInt(10 * 60 * 1000 * 1000000);
+
+  // Round up to nearest 10 minutes
+  return nowPlusTwoDays + tenMinutes - (nowPlusTwoDays % tenMinutes);
+};
```
turbocrime, line 107: this is clearer for me to read.

* constants at the top, calculated with decreasing significance so it lines up
* name every variable with Ms or Ns depending on what it contains
* 'round up' confused me, i think 'next ten-minute interval' is clearer
* only *one* millis-to-nanos conversion, at the end
* use `_` separator on large integer

```suggestion
export const currentTimePlusTwoDaysRounded = (): bigint => {
  const tenMinsMs = BigInt(1000 * 60 * 10);
  const twoDaysMs = BigInt(1000 * 60 * 60 * 24 * 2);
  
  // Date.now returns milliseconds
  const nowMs = BigInt(Date.now());
  const twoDaysFromNowMs = nowMs + twoDaysMs;
  
  // round to next ten-minute interval
  const roundedTimeoutMs = twoDaysFromNowMs + tenMinsMs - (twoDaysFromNowMs % tenMinsMs);
  
  // 1 million nanoseconds per millisecond
  const roundedTimeoutNs = roundedTimeoutMs * BigInt(1_000_000);
  return roundedTimeoutNs;
};
```

turbocrime: i would also suggest accepting an input parameter, and having the caller provide Date.now()

apps/minifront/src/state/ibc.ts
```diff
@@ -90,29 +90,57 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
...
+    throw new Error('no clientId ConnectionEnd returned from ibcConnectionClient request');
+  }
+
+  const { clientState: clientStateUnpacked } = await ibcClient.clientState({ clientId: clientId });
+  if (!clientStateUnpacked) {
+    throw new Error('no clientState returned from ibcClient request');
+  }
+
+  const clientState = clientStateUnpacked.unpack(typeRegistry) as ClientState | undefined;
   if (!clientState) throw new Error('Could not find chain id client state');
```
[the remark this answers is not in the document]
grod220: Yeah, unpacked was the wrong name here. 

Changed to `unpackTo`, but mutating the input is kinda a strange api imo

apps/minifront/src/state/ibc.ts
grod220: Updated to your format. However, changed so that bigint conversion happens at the end.

TalDerei (approved):
the difference between rounding at the millisecond level versus the nanosecond level (as performed in core) is negligible. LGTM

apps/minifront/src/state/ibc.ts
```diff
@@ -90,29 +89,62 @@ export const createIbcSendSlice = (): SliceCreator<IbcSendSlice> => (set, get) =
   };
 };
 
+// Timeout is two days. However, in order to prevent identifying oneself by clock skew,
+// timeout time is rounded up to the nearest 10 minute interval.
+// Reference in core: https://github.com/penumbra-zone/penumbra/blob/1376d4b4f47f44bcc82e8bbdf18262942edf461e/crates/bin/pcli/src/command/tx.rs#L1066-L1067
+export const currentTimePlusTwoDaysRounded = (currentTimeMs: number): bigint => {
+  const tenMinsMs = 1000 * 60 * 10;
+  const twoDaysMs = 1000 * 60 * 60 * 24 * 2;
```
jessepinho, line 97: I typically pull never-changing constants out of the function body since they don't depend on function input, e.g., 

```suggestion
const tenMinsMs = 1000 * 60 * 10;
const twoDaysMs = 1000 * 60 * 60 * 24 * 2;

// Timeout is two days. However, in order to prevent identifying oneself by clock skew,
// timeout time is rounded up to the nearest 10 minute interval.
// Reference in core: https://github.com/penumbra-zone/penumbra/blob/1376d4b4f47f44bcc82e8bbdf18262942edf461e/crates/bin/pcli/src/command/tx.rs#L1066-L1067
export const currentTimePlusTwoDaysRounded = (currentTimeMs: number): bigint => {
```

jessepinho (approved):

turbocrime (approved):

PR #958, opened 2024-04-19
`@penumbra-zone/keys`

turbocrime:
pulled out keys diff from #955 since it's technically unrelated

~~@grod220 the 'keys' package should not be versioned with changeset, as it corresponds directly to the penumbra core repo version. but i'm not sure how publishing works when we're not using changeset to manage~~

keys now versioned normally and packaged with changeset


grod220:
Can you give a description of what the code change does? It's not really possible to discern from reading the shell script. Unfortunately, a shell script that goes beyond very basic commands will be essentially unmaintainable. Doesn't have to be this PR but eventually we should convert this to either typescript or rust.

> the 'keys' package should not be versioned with changeset, as it corresponds directly to the penumbra core repo version

Isn't the version we pull defined in `defaultKeysVersion` variable? I don't think this npm package needs to be pinned to the same version as the keys. That prevents us making changes to this package once mainnet hits given key version freezes.

turbocrime:
okay. at one point the 'default' version was pulled from package json but you're right it doesn't have to be pinned

grod220 (approved):

PR #968, opened 2024-04-22
add license key to every `package.json`

turbocrime:
i expected publishing to integrate the workspace root license key and license files due to documentation at https://pnpm.io/cli/publish however this seems to not be the case

### DIFF
--- a/packages/wasm/package.json
+++ b/packages/wasm/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@penumbra-zone/wasm",
   "version": "3.0.0",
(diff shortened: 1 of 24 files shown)


packages/wasm/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@penumbra-zone/wasm",
   "version": "3.0.0",
-  "license": "MIT",
```
turbocrime, line 4: missed this one somehow

grod220 (approved):

Issue #944, opened 2024-04-16
UI for scheduling Dutch auctions


TalDerei:
**Minifront M0**

Landing page amongst other UI-related auction features.

PR #970, opened 2024-04-22
Refactor <AssetOutSelector /> and <SelectTokenModal />

jessepinho:
While working on #944, I refactored these two components:
- `<AssetOutSelector />` and `<SelectTokenModal />` are now named `<AssetSelector />` and `<BalanceSelector />`, respectively.
- They now take `value`/`onChange` props instead of the previous use case-specific names for their props.
- They had some spacing issues that are fixed.
- They both have search boxes that actually work.

The two components still have a lot of similarities, but they're still separate components since they are used for different data types. `AssetSelector` allows you to choose from various `Metadata`s, while `BalanceSelector` allows you to choose from various `BalancesResponse`s — i.e., the UI presents not only the asset symbol and icon, but also your balance of it and which account # it's held in.

I figured I'd split this into its own PR to make the #944 PR smaller.


apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -0,0 +1,126 @@
+import {
```
jessepinho, line 1: Almost all of this is copied from `AssetOutSelector`, but with different prop names. I'll highlight the changes.

apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -0,0 +1,126 @@
...
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { localAssets } from '@penumbra-zone/constants/src/assets';
+import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
+import { useEffect, useMemo, useState } from 'react';
+import { IconInput } from '@penumbra-zone/ui/components/ui/icon-input';
+import { MagnifyingGlassIcon } from '@radix-ui/react-icons';
+
+interface AssetSelectorProps {
+  value?: Metadata;
+  onChange: (metadata: Metadata) => void;
```
jessepinho, line 21: `value` and `onChange` are renamed from `assetOut` and `setAssetOut`.

apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -0,0 +1,126 @@
...
+  if (!filter || !value) return;
+
+  if (!filter(value)) {
+    const firstAssetThatPassesTheFilter = assets.find(filter);
+    if (firstAssetThatPassesTheFilter) onChange(firstAssetThatPassesTheFilter);
+  }
+};
+
+const useFilteredAssets = ({ value, onChange, filter }: AssetSelectorProps) => {
+  const [search, setSearch] = useState('');
```
jessepinho, line 53: This search functionality is new — it used to only be in `<SelectTokenModal />` (which is now renamed to `<BalanceSelector />`).

apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <DialogContent>
+        <div className='flex max-h-screen flex-col'>
+          <DialogHeader>Select asset</DialogHeader>
+          <div className='flex flex-col gap-2 overflow-auto p-4'>
+            <IconInput
+              icon={<MagnifyingGlassIcon className='size-5 text-muted-foreground' />}
+              value={search}
+              onChange={setSearch}
+              placeholder='Search assets...'
+            />
```
jessepinho, line 105: this search UI is new

apps/minifront/src/components/swap/asset-out-selector.tsx
jessepinho, line 1: Renamed to `<AssetSelector />` above

apps/minifront/src/components/shared/input-token.tsx
```diff
@@ -17,6 +17,7 @@ interface InputTokenProps extends InputProps {
   setSelection: (selection: BalancesResponse) => void;
   validations?: Validation[];
   balances: BalancesResponse[];
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
```
jessepinho, line 20: In the interest of not letting consumers mess with this component too much (via `className`/etc.), this limits the props to only the ones we want to allow consumers to set.

packages/ui/components/ui/icon-input.tsx
```diff
@@ -0,0 +1,32 @@
+import { ReactNode } from 'react';
+import { Input } from './input';
+
+const className = 'flex w-full gap-2 rounded-lg border bg-background px-3 py-2';
+
+/**
+ * Use this to render an input with an icon to its left, such as a search field
+ * with a magnifying glass.
+ */
+export const IconInput = ({
```
jessepinho, line 10: image

apps/minifront/src/components/shared/balance-selector.tsx
```diff
@@ -12,50 +12,60 @@ import { cn } from '@penumbra-zone/ui/lib/utils';
...
-                <MagnifyingGlassIcon className='size-5 text-muted-foreground' />
-              </div>
-              <Input
-                className='pl-10'
-                value={search}
-                onChange={e => setSearch(e.target.value)}
-                placeholder='Search asset...'
-              />
-            </div>
-          </div>
```
jessepinho, line 51: Extracted this to `<IconInput />`

Valentine1898 (commented):
It seems to me that our swap design is a bit different from the standards of other DEX applications 

1. Users are probably used to being able to swap selectors with a single click (examples with uniswap, 1inch, osmosis)
image
image
image



2. Users probably expect to be able to select both tokens they have on their balance and tokens they don't have on their balance in both selectors. (e.g. to estimate the current exchange rate) (examples with uniswap, 1inch, osmosis).
image
image
image


Based on this, I think we should not have two selectors AssetSelector and BalanceSelector, but one universal selector.

jessepinho:
@Valentine1898 would you be opposed to this being merged as a step in that direction? (They're already two separate components anyway, so this doesn't make the problem worse.)

We currently have a baked-in assumption in a lot of our code that the only assets you can choose when sending/swapping/etc. are ones you have a balance in. I think it's a great idea to revisit that assumption, but that'd need to be part of a larger refactor.

Valentine1898 (approved):
Yes, I agree. I will add a new issue to improve the swap design and we can do that later.

TalDerei (approved):

PR #976, opened 2024-04-23
Fix issue with shifting account header

### LINKED ISSUE #966: Account switcher heading movement
Toggling between account 0 and 1 makes the title hop a bit left and right. Likely due to arrow not being there for padding.

https://github.com/penumbra-zone/web/assets/16624263/54cc3019-2186-47b1-b0f4-9fc281629607

jessepinho:
From #966:

> Toggling between account 0 and 1 makes the title hop a bit left and right. Likely due to arrow not being there for padding.
>
> https://github.com/penumbra-zone/web/assets/16624263/54cc3019-2186-47b1-b0f4-9fc281629607

Now, the next/previous buttons are always showing, but are disabled if we're on the last/first account, respectively:
image

Closes #966

### DIFF
--- a/packages/ui/components/ui/account-switcher.test.tsx
+++ b/packages/ui/components/ui/account-switcher.test.tsx
@@ -56,10 +56,10 @@ describe('<AccountSwitcher />', () => {
       expect(onChange).toHaveBeenCalledWith(122);
     });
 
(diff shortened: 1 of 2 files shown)


grod220 (approved):

TalDerei (approved):
LGTM

Issue #778, opened 2024-03-16
bech32 package does not validate properly


turbocrime:
during reorganization of bech32 utilities into their own package, i discovered they improperly handle length, and in some cases may be improperly truncating data.

edit:

bech32m format described here https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki

splitting out bech32 package, i strove to organize but maintain exact functionality. some of the functionality was acknowledged incorrect.

previously, we were using only a couple functions for calculating/checking/encoding, with `LIMIT` parameters that did not necessarily match the actual data we were handling. notably, a generically named function was used in many places. while incorrect, it was functioning, so i maintained the exact behavior when splitting out the package, while providing clearer names.

https://github.com/penumbra-zone/web/blob/1e4476eb0d4d1ba13c483bc0686179794b6ce68c/packages/types/src/address.ts#L17-L20

it looks like some work has since been done to expand some of that in https://github.com/penumbra-zone/web/pull/846. specifically, lengths for most things are now present.

https://github.com/penumbra-zone/web/blob/69e02d5810ede7ad39bcb218235669fd6e067d3b/packages/bech32/src/penumbra-bech32.ts#L1-L12

the bech32 package we depend on contains a hard-coded default `LIMIT` parameter of 90 characters total. this is appropriate for bitcoin segwit: a human-readable part of 'bc', separator, data containing 20 bytes, checksum suffix is well below 90.

https://github.com/bitcoinjs/bech32/blob/fd92fdfba8cf7a7de6cab5cec683d5576d5bb2f1/src/index.ts#L103-L105

many of our encoded lengths are longer. ignoring the 90-char maximum in spec, the appropriate string length for any given prefix and data works out to

```ts
const bech32Length = (prefix: string, data: Uint8Array) =>
  prefix.length +
  1 + // separator
  Math.ceil((8 / 5) * data.length) +
  6; // checksum
```

not all of the lengths defined in the exports conform to this, and the lengths are not consistently used to confirm appropriate input/output size.

also, as the package is intended for string manipulation and not messaging, we should not import buf packages. the types do not provide any type safety in this package - every type imported describes a trivial `{ inner: Uint8Array }`. ideally, this should be a utility package that anyone wanting validation or basic conversion could import with minimal dependency graph. buf packages are available in the workspace but not present in `package.json`, so as currently published it is broken.

objectives

1. possibly rename everything. we're using bech32m, not bech32. also the to/from utility names are very similar
[...]

PR #979, opened 2024-04-24
bech32m

turbocrime:
fixing #778 and providing typesafe, named functions

also updating most of the tests to use more realistic data, now that lengths and checksums are validated everywhere.

this package is configured to build with tsc for publish, despite discussion of not building - it's a relatively simple package, the main export is all constant primitives, and it may be used in simpler contexts than the larger and more complex packages that depend on message types.

the diff is huge, because these functions are used absolutely everywhere.


packages/types/src/string.ts
```diff
@@ -17,8 +15,3 @@ export const shorten = (str: string, endsLength = 4) => {
     return str.slice(0, endsLength) + '…' + str.slice(endsLength * -1);
   }
 };
-
-export const shortenAddress = (str: string) => {
-  const charsToDisplay = 24;
-  return str.slice(0, PENUMBRA_BECH32_ADDRESS_PREFIX.length + 1 + charsToDisplay) + '…';
-};
```
turbocrime, line 24: this `shortenAddress` fn wasn't actually used anywhere

packages/ui/components/ui/identity-key-component.tsx
```diff
@@ -15,15 +16,24 @@ const PREFIX_AND_SEPARATOR = PENUMBRA_BECH32_IDENTITY_PREFIX + '1';
...
 export const IdentityKeyComponent = ({ identityKey }: { identityKey: IdentityKey }) => {
+  const [ik, setIk] = useState<string>('');
+
+  useEffect(() => {
+    try {
+      setIk(bech32mIdentityKey(identityKey));
+    } catch {
+      /* noop */
+    }
+  }, [identityKey]);
```
turbocrime, line 27: it's possible the new functions throw on invalid input, so use may be a little more awkward in some contexts

apps/minifront/src/state/staking/index.test.ts
```diff
@@ -270,6 +275,7 @@ describe('Staking Slice', () => {
      * before validator 3 at the end: we have a 0 balance of both, but validator
      * 4 has more voting power.
      */
+    console.log('here we goe');
```
grod220, line 278: can delete

apps/minifront/src/state/ibc.ts
```diff
@@ -193,16 +193,18 @@ export const ibcValidationErrors = (state: AllSlices) => {
...
   };
 };
 
-const validateAddress = (chain: Chain | undefined, address: string): boolean => {
+const validateUnknownAddress = (chain: Chain | undefined, address: string): boolean => {
   if (!chain || address === '') return false;
-  return bech32IsValid(address, chain.addressPrefix);
+  const { prefix, words } =
+    bech32.decodeUnsafe(address, Infinity) ?? bech32m.decodeUnsafe(address, Infinity) ?? {};
+  return !words || prefix !== chain.addressPrefix;
```
grod220, line 207: This feels more appropriately placed in the bech32 package. Also, what's the deal with `Infinity`? Some code comments may help here.

apps/minifront/src/state/send.ts
```diff
@@ -19,7 +19,16 @@ import {
...
+import { addressFromBech32m } from '@penumbra-zone/bech32m/penumbra';
+
+const isPenumbraAddress = (addr: string): addr is `penumbra1${string}` => {
+  try {
+    addressFromBech32m(addr);
+    return true;
+  } catch {
+    return false;
+  }
+};
```
grod220, line 31: Given we have duplicates of this, think we should relocate to bech32 package

packages/bech32m/src/format/prefix.ts
```diff
@@ -0,0 +1,15 @@
+export const Prefixes = {
+  passet: 'passet',
+  pauctid: 'pauctid',
+  penumbra: 'penumbra',
+  //penumbracompat1: 'penumbracompat1',
```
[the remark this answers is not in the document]
grod220: Can you add this a code comment?

packages/bech32m/src/format/prefix.ts
```diff
@@ -0,0 +1,15 @@
...
+  penumbrafullviewingkey: 'penumbrafullviewingkey',
+  penumbragovern: 'penumbragovern',
+  penumbraspendkey: 'penumbraspendkey',
+  penumbravalid: 'penumbravalid',
+  penumbrawalletid: 'penumbrawalletid',
+  plpid: 'plpid',
+} as const;
+
+export type Prefix = keyof typeof Prefixes;
+export const isPenumbraPrefix = (prefix: string): prefix is Prefix => prefix in Prefixes;
```
grod220, line 15: Unused?

packages/bech32m/src/format/size.ts
grod220, line 1: Can we combine all of these objects to one? Something like:

```typescript
type PenumbraPrefix =
  | 'passet'
  | 'pauctid'
  | 'penumbra'
  | 'penumbrafullviewingkey'
  | 'penumbragovern'
  | 'penumbraspendkey'
  | 'penumbravalid'
  | 'penumbrawalletid'
  | 'plpid';

type Bech32Info = {
  [key in PenumbraPrefix]: {
    inner: 'inner' | 'gk' | 'ik';
    prefix: key;
    stringLength: number;
    byteLength: number;
  };
};
```

packages/bech32m/src/index.ts
grod220, line 1: Is it necessary to expose these? If so, think the combined object would be a bit easier to do so.

packages/bech32m/src/penumbra.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Inner } from './format/inner';
+import { Prefixes } from './format/prefix';
+
+const innerName = Inner.penumbra;
+const prefix = Prefixes.penumbra;
+
+export const bech32mAddress = ({ [innerName]: bytes }: { [innerName]: Uint8Array }) =>
+  toBech32m(bytes, prefix);
+
+export const addressFromBech32m = (penumbra: string): { [innerName]: Uint8Array } => ({
```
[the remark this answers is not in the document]
grod220: Realizing this feedback can apply to all of the different asset types in this package

packages/ui/components/ui/identity-key-component.tsx
grod220: Shouldn't we throw though if someone tries to pass an invalid key? If not, we should add another `useState` for an error state and show a different UI below.

grod220 (commented):
Nice tests!

packages/bech32m/src/penumbra.ts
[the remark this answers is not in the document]
turbocrime: the reason for not doing this is described in the readme. we must do it in many places because many of our parameters and props are typed incorrectly. i would prefer to correct those instead but that is large

packages/bech32m/src/penumbra.ts
```diff
@@ -0,0 +1,15 @@
...
+  [innerName]: fromBech32m(penumbra as `${typeof prefix}1${string}`, prefix),
```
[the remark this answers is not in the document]
turbocrime: then the caller must assert type. this way allows caller to use for check

apps/minifront/src/state/ibc.ts
turbocrime: our bech32m package may be consumed to validate penumbra addresses.

this is validating an address from an unknown chain in an unknown format.

other formats are the responsibility of the application.

packages/ui/components/ui/identity-key-component.tsx
turbocrime: > Shouldn't we throw though if someone tries to pass an invalid key? If not, we should add another `useState` for an error state and show a different UI below. 

the caller initially passes undefined and then populates immediately.

throwing from this component on invalid input breaks the app. there was no previous invalid state handling, it just rendered an empty string. in current use, adding a visual error indication would not create any practical or visible change: it would always activate and then always deactivate within a fraction of a second.

this choice maintains existing function with minimal refactor.

packages/bech32m/src/penumbra.ts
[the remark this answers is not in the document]
turbocrime: > Can we rename param to `(penumbraAddr: string)`?

all of these to/from function parameters are named after the prefix of the expected input

apps/minifront/src/state/ibc.ts
turbocrime: added comment explaining

packages/getters/src/metadata.test.ts
turbocrime, line 1: tests for removed function deleted

packages/getters/src/metadata.ts
turbocrime, line 1: use replaced with `getValidatorIdentityKey` plus `bech32IdentityKey`

packages/bech32m/src/penumbra.ts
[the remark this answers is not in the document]
turbocrime: now these are named like `addressFromBech32m(penumbra1: string)`

packages/bech32m/src/format/prefix.ts
turbocrime: removed

packages/bech32m/src/format/size.ts
turbocrime: these were separate for reasons and mostly used internally, but i'm now exporting a collated object

packages/bech32m/src/format/index.ts
```diff
@@ -0,0 +1,70 @@
...
+type PenumbraBech32mSpec = Required<{
+  readonly [p in Prefix]: {
+    readonly prefix: (typeof Prefixes)[p];
+    readonly stringLength: (typeof StringLength)[p];
+    readonly byteLength: (typeof ByteLength)[p];
+    readonly innerName: (typeof Inner)[p];
+  };
+}>;
+
+export default {
```
grod220, line 15: We should avoid using default exports in general

packages/bech32m/src/index.ts
```diff
@@ -0,0 +1,30 @@
+import SPEC from './format';
+
+export default SPEC;
```
grod220, line 3: Same here, let's name our exports

packages/bech32m/src/format/index.ts
grod220, line 1: This package seems to have multiple ways its accessing the same information

```
A)
import { Prefixes } from './format/prefix';
const prefix = Prefixes.passet;

B)
import SPEC from './format';
export const PENUMBRA_BECH32M_ASSETID_PREFIX = SPEC.passet.prefix;
```

Instead of defining all of these parameters in separate objects, it feels simpler to just put them all in this file. Aka, this:

```
export const bech32Spec = {
  passet: {
    prefix: 'passet',
    stringLength: 65,
    byteLength: 32,
    innerName: "inner",
  },
}

```

versus

```
import { Inner } from './inner';
import { StringLength } from './strings';
import { ByteLength } from './bytes';
import { Prefixes } from './prefix';

export const bech32Spec = {
  passet: {
    prefix: Prefixes.passet,
    stringLength: StringLength.passet,
    byteLength: ByteLength.passet,
    innerName: Inner.passet,
```

This will allow us also to delete `packages/bech32m/src/index.ts` entirely

packages/bech32m/src/passet.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const innerName = Inner.passet;
+const prefix = Prefixes.passet;
+
+export const bech32mAssetId = ({ [innerName]: bytes }: { [innerName]: Uint8Array }) =>
+  toBech32m(bytes, prefix);
+
+export const assetIdFromBech32m = (passet1: string): { [innerName]: Uint8Array } => ({
+  [innerName]: fromBech32m(passet1 as `${typeof prefix}1${string}`, prefix),
+});
```
grod220, line 13: The whole purpose of this package is to allow us easy ways to convert to/from bech32 to/from Penumbra protobuf types. Returning approximations of the inside of these types feels unnecessary, especially given all consumers will then be required to construct the type manually right after.  The more natural API is to simply return what they need. I think we should do this:

```diff
- export const assetIdFromBech32m = (passet1: string): { [innerName]: Uint8Array }
+ export const assetIdFromBech32m = (passet1: string): AssetId
```

turbocrime: the whole purpose of the package is to convert from bech32m to binary data in a verifiable way.

the types are a convenience. i would prefer to remove the structure than to add the types.

Issue #800, opened 2024-03-20
Updated IBC ui


grod220:
Built out UI described in: https://discord.com/channels/824484045370818580/1044457038886998137/1216919208714047600

<img width="963" alt="Screenshot_2024-03-11_at_9 22 15_PM" src="https://github.com/penumbra-zone/web/assets/16624263/d8c7e483-3e7a-4348-91e9-b7eca764c8e8">

Some thought required on handling smaller screen sizes (likely just stack).

grod220:
Redundant to: https://github.com/penumbra-zone/web/issues/79

PR #982, opened 2024-04-25
Interchain UI -> chain selection

grod220:
Part 1 of a multi-part task of getting support for IBC'ing into Penumbra. This has been copied over and modified from the [asset lists interchain ui](https://cosmology.zone/products/create-cosmos-app) example. The subsequent PRs will be incorporating more and more from this example.

First up is to get support for a chain out selector.

https://github.com/penumbra-zone/web/assets/16624263/215bbd4b-724d-4a7e-8546-ba33d4f3796f

### DIFF
--- a/.github/workflows/turbo-ci.yml
+++ b/.github/workflows/turbo-ci.yml
@@ -51,7 +51,7 @@ jobs:
 
   turbo-build:
     name: Build
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: buildjet-8vcpu-ubuntu-2204
     needs: turbo-compile
     steps:
       - uses: actions/checkout@v4
--- a/apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
+++ b/apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
@@ -44,7 +44,7 @@ export const GrpcEndpointForm = ({
       <div className='flex flex-col gap-2'>
         <form className='flex flex-col gap-4' onSubmit={handleSubmit}>
           {rpcsQuery.isLoading && <LoadingIndicator />}
-          {rpcsQuery.error && (
+          {Boolean(rpcsQuery.error) && (
             <div className='text-red-700'>
               Error loading chain registry: {String(rpcsQuery.error)}
             </div>
--- a/apps/minifront/src/components/ibc/layout.tsx
+++ b/apps/minifront/src/components/ibc/layout.tsx
@@ -1,6 +1,6 @@
 import { Card } from '@penumbra-zone/ui/components/ui/card';
-import { IbcOutForm } from './ibc-out-form';
-import { IbcInForm } from './ibc-in-form';
+import { IbcOutForm } from './ibc-out/ibc-out-form';
+import { IbcInForm } from './ibc-in/ibc-in-form';
 import { ArrowLeftIcon, ArrowRightIcon } from '@radix-ui/react-icons';
 
 export const IbcLayout = () => {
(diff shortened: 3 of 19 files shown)


apps/extension/src/shared/components/grpc-endpoint-form/index.tsx
```diff
@@ -44,7 +44,7 @@ export const GrpcEndpointForm = ({
       <div className='flex flex-col gap-2'>
         <form className='flex flex-col gap-4' onSubmit={handleSubmit}>
           {rpcsQuery.isLoading && <LoadingIndicator />}
-          {rpcsQuery.error && (
+          {Boolean(rpcsQuery.error) && (
```
grod220, line 47: Getting linting errors for this

apps/minifront/src/components/ibc/layout.tsx
```diff
@@ -1,18 +1,18 @@
...
     <>
       <div className="fixed inset-0 z-[-100] size-full bg-[url('penumbra-logo.svg')] bg-[length:160vmax] bg-fixed bg-[top_50%_left_25vw] bg-no-repeat" />
       <div className='flex flex-1 flex-col gap-4 md:flex-row md:place-content-around'>
-        <Card light className='relative overflow-visible md:self-start'>
+        <Card light className='relative overflow-visible md:self-end'>
           <ArrowRightIcon className='absolute inset-y-0 right-0 z-10 my-auto -mr-14 size-16 text-stone-300' />
           <IbcInForm />
         </Card>
-        <Card gradient className='relative overflow-visible md:self-end'>
+        <Card gradient className='relative overflow-visible md:self-start'>
```
grod220, line 15: Given the ibc chain side will grow to be much longer than the ibc out, put the right side to be on top so it doesn't get pushed down.

Screenshot 2024-04-25 at 3 52 30 PM

packages/ui/components/ui/tx/view/address-view.tsx
```diff
@@ -32,7 +32,7 @@ export const AddressViewComponent = ({ view, copyable = true }: AddressViewProps
       {accountIndex !== undefined ? (
         <>
           <AddressIcon address={view.addressView.value.address} size={14} />
-          <span className='font-bold break-keep'>
+          <span className='break-keep font-bold'>
```
grod220, line 35: A lint warning here

.github/workflows/turbo-ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
   turbo-build:
     name: Build
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: buildjet-8vcpu-ubuntu-2204
```
grod220, line 54: Out of memory issues in CI for this task. Gotta boost it up.

apps/minifront/src/components/ibc/layout.tsx
turbocrime: the downward placement was a significant part of the design concept, providing an intuitive sense of descending into a place of concealment. cc @hdevalence 

if you used my style code, on wide enough viewports it should not compete for vertical space with the "Shield Assets" card - it will be pushed to the bottom of the screen, no more. 

on narrower viewports that cannot accomodate the cards side-by-side, it responsively switches to a column layout. the order for that is configurable much like the other pages

hdevalence: the design thought of the original concept was that when a user clicks on "Shield Funds" it's probably because they don't already have funds on Penumbra and so the first thing that should appear is the deposit dialog.

the reason to offset the two cards was so that the arrows could extend "into" the other domain, without pointing directly at the other card, rather than a sense of descent. the idea was just that the deposit card would have an arrow that goes into the inside of the eclipse. the thing that actually matters was having a two-column layout where one column is inside the eclipse and the other is outside.

> Given the ibc chain side will grow to be much longer than the ibc out

i think i'm missing something, i'm not sure why the length matters here? we can just have two columns, no?

hdevalence: cc @bibnub

hdevalence: prev context https://github.com/penumbra-zone/web/issues/800

grod220: Helpful context. With the flexbox start/end layout, as the "start" element grew beyond the size of the window (not shown in the screenshot), the "end" element got pushed below view. Swapping them was a fix, but I see it doesn't fix the concept so well. The easiest solution is simply to add top margin to the right side.

Screenshot 2024-04-26 at 2 53 34 PM

We may need to evaluate these arrows again if the left side box more space than given.

turbocrime:
main started failing CI after this merge

Issue #924, opened 2024-04-11
improve `HandlerContext`


turbocrime:
everything in `ctx/prax.ts` is prax-specific and should be abstracted in the interest of third-party consumption

https://github.com/penumbra-zone/web/blob/ed38353a302995447a975120123ab917f70a1315/packages/router/src/ctx/prax.ts#L11-L13

- [x] https://github.com/penumbra-zone/web/issues/826
- [x] https://github.com/penumbra-zone/web/issues/730
- [x] https://github.com/penumbra-zone/web/issues/923
- [x] https://github.com/penumbra-zone/web/issues/731

PR #983, opened 2024-04-25
don't block services init on storage key presence

turbocrime:
servicesCtx and fvkCtx beocme promises, so services init no longer blocks on storage key presence - a function is provided via context to query storage at runtime.

this is part of improving handler context #924 and makes initialization asynchronous and much more robust, which has benefits used in #998 to allow services to run entirely without network connection.

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -13,74 +13,45 @@ import './listeners';
 
 // services
 import { Services } from '@penumbra-zone/services-context';
-import { localExtStorage } from '@penumbra-zone/storage/chrome/local';
-
-// adapter
-import { ConnectRouter, createContextValues, PromiseClient } from '@connectrpc/connect';
-import { CRSessionManager } from '@penumbra-zone/transport-chrome/session-manager';
-import { createDirectClient } from '@penumbra-zone/transport-dom/direct';
-import { connectChannelAdapter } from '@penumbra-zone/transport-dom/adapter';
-import { jsonOptions } from '@penumbra-zone/protobuf';
-
-// context
-import { CustodyService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/custody/v1/custody_connect';
-import { QueryService as StakingService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/core/component/stake/v1/stake_connect';
-import { approverCtx } from '@penumbra-zone/services/ctx/approver';
-import { custodyCtx } from '@penumbra-zone/services/ctx/custody';
-import { servicesCtx } from '@penumbra-zone/services/ctx/prax';
-import { stakingClientCtx } from '@penumbra-zone/services/ctx/staking-client';
-import { approveTransaction } from './approve-transaction';
+import { backOff } from 'exponential-backoff';
 
 // all rpc implementations, local and proxy
 import { getRpcImpls } from './get-rpc-impls';
-import { backOff } from 'exponential-backoff';
+
+// adapter
+import { ConnectRouter, createContextValues, PromiseClient } from '@connectrpc/connect';
+import { jsonOptions } from '@penumbra-zone/protobuf';
+import { CRSessionManager } from '@penumbra-zone/transport-chrome/session-manager';
+import { connectChannelAdapter } from '@penumbra-zone/transport-dom/adapter';
+
+// context
+import { approverCtx } from '@penumbra-zone/services/ctx/approver';
+import { fvkCtx } from '@penumbra-zone/services/ctx/full-viewing-key';
+import { servicesCtx } from '@penumbra-zone/services/ctx/prax';
+import { approveTransaction } from './approve-transaction';
+import { getFullViewingKey } from './ctx/full-viewing-key';
+
+// context clients
+import { QueryService as StakingService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/core/component/stake/v1/stake_connect';
+import { CustodyService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/custody/v1/custody_connect';
+import { custodyClientCtx } from '@penumbra-zone/services/ctx/custody-client';
+import { stakingClientCtx } from '@penumbra-zone/services/ctx/staking-client';
+import { createDirectClient } from '@penumbra-zone/transport-dom/direct';
+
+// storage
 import {
   FullViewingKey,
   WalletId,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-import { fvkCtx } from '@penumbra-zone/services/ctx/full-viewing-key';
-import { WalletJson } from '@penumbra-zone/types/wallet';
-
-/**
- This fixes an issue where some users do not have 'grpcEndpoint' set after they have finished onboarding
- */
-const fixEmptyGrpcEndpointAfterOnboarding = async () => {
-  //TODO change to mainnet default RPC
-  const DEFAULT_GRPC_URL = 'https://grpc.testnet.penumbra.zone';
-  const grpcEndpoint = await localExtStorage.get('grpcEndpoint');
-  const wallets = await localExtStorage.get('wallets');
-  if (!grpcEndpoint && wallets[0]) {
-    await localExtStorage.set('grpcEndpoint', DEFAULT_GRPC_URL);
-  }
-};
-
-/**
- * When a user first onboards with the extension, they won't have chosen a gRPC
- * endpoint yet. So we'll wait until they've chosen one to start trying to make
- * requests against it.
- */
-const waitUntilGrpcEndpointExists = async () => {
-  const grpcEndpointPromise = Promise.withResolvers<void>();
-  const grpcEndpoint = await localExtStorage.get('grpcEndpoint');
-
-  if (grpcEndpoint) {
-    grpcEndpointPromise.resolve();
-  } else {
-    const listener = (changes: Record<string, { newValue?: unknown }>) => {
-      if (changes['grpcEndpoint']?.newValue) {
-        grpcEndpointPromise.resolve();
-        localExtStorage.removeListener(listener);
-      }
-    };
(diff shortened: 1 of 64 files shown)


jessepinho (commented):
Could you please add some context about this change? Why it's needed, what it fixes, etc.

turbocrime:
> Could you please add some context about this change? Why it's needed, what it fixes, etc.

this kind of happened accidentally as i was setting up spendkey `HandlerContext`, so i made it into its own PR

packages/services/src/view-service/assets.test.ts
```diff
@@ -12,7 +12,7 @@ import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { servicesCtx } from '../ctx/prax';
 import { assets } from './assets';
 import { IndexedDbMock, MockServices } from '../test-utils';
-import { Services } from '@penumbra-zone/services-context/src/index';
+import { ServicesInterface } from '@penumbra-zone/types/src/services';
```
turbocrime, line 15: using the interface in tests now

turbocrime:
related to #998

apps/extension/src/service-worker.ts
```diff
@@ -13,99 +13,68 @@ import './listeners';
...
+        fullViewingKey: FullViewingKey.fromJsonString(wallet.fullViewingKey),
+      });
+      await services.initialize();
+      return services;
     },
-  });
+    {
+      retry: (e, attemptNumber) => {
+        if (process.env['NODE_ENV'] === 'development')
+          console.warn('services-context could not start', attemptNumber, e);
```
grod220, line 67: Please add brackets here

apps/extension/src/storage/onboard.ts
```diff
@@ -0,0 +1,51 @@
+import { ExtensionStorage } from '@penumbra-zone/storage/chrome/base';
+import { localExtStorage } from '@penumbra-zone/storage/chrome/local';
+import { SessionStorageState } from '@penumbra-zone/storage/chrome/session';
+import { LocalStorageState } from '@penumbra-zone/storage/chrome/types';
+
+export const waitForStorageKey = async <
```
grod220, line 6: Can you add a code comment above? and/or inline in the function?

apps/extension/src/storage/onboard.ts
```diff
@@ -0,0 +1,51 @@
...
+): Promise<I extends undefined ? S[K] : S[K][I]> => {
+  type Stored = I extends undefined ? S[K] : S[K][I];
+  const { promise, resolve, reject } = Promise.withResolvers<Stored>();
+
+  const storedObject: S[K] = await storage.get(storageKey);
+  const storedValue = (innerKey != null ? storedObject[innerKey] : storedObject) as
+    | Stored
+    | undefined;
+  if (storedValue != null) resolve(storedValue);
+  else {
```
grod220, line 23: Let's use brackets for if/else's

apps/extension/src/storage/onboard.ts
```diff
@@ -0,0 +1,51 @@
...
+import { LocalStorageState } from '@penumbra-zone/storage/chrome/types';
+
+export const waitForStorageKey = async <
+  S extends LocalStorageState | SessionStorageState,
+  K extends keyof S,
+  I extends keyof S[K],
+>(
+  storage: ExtensionStorage<S>,
+  storageKey: K & string,
+  innerKey?: I,
```
grod220, line 13: what's innerKey?

apps/extension/src/storage/onboard.ts
```diff
@@ -0,0 +1,51 @@
...
+): Promise<I extends undefined ? S[K] : S[K][I]> => {
+  type Stored = I extends undefined ? S[K] : S[K][I];
+  const { promise, resolve, reject } = Promise.withResolvers<Stored>();
+
+  const storedObject: S[K] = await storage.get(storageKey);
+  const storedValue = (innerKey != null ? storedObject[innerKey] : storedObject) as
```
grod220, line 19: Can we convert all equality checks to be strict equality pls? It makes it harder to discern what the type coercision is. In general, folks in the industry try not to use it.

Also, shouldn't these be undefined checks? aka `innerKey !== undefined`

apps/extension/src/storage/onboard.ts
grod220: Ah, I see now. I think a more flexible api would be something like

```diff
storage: ExtensionStorage<S>,
- storageKey: K & string,
- innerKey?: I,
+ checkFn: (changes: StorageChanges) => Promise<boolean> // or maybe pass storage into the fn?
```

apps/extension/src/storage/onboard.ts
```diff
@@ -0,0 +1,51 @@
...
+    | Stored
+    | undefined;
+  if (storedValue != null) resolve(storedValue);
+  else {
+    const storageListener = (changes: Record<string, { newValue?: unknown }>) => {
```
grod220, line 24: Instead of manually typing the Record type for changes, think we should have a type we can import here from `packages/storage/src/chrome/base.ts`

apps/extension/src/service-worker.ts
```diff
@@ -126,14 +95,13 @@ const getServiceHandler = async () => {
...
 
       return Promise.resolve({ ...req, contextValues });
     },
   });
 };
 
 await fixEmptyGrpcEndpointAfterOnboarding();
-await waitUntilGrpcEndpointExists();
-const handler = await getServiceHandler();
+const handler = getServiceHandler(await onboardGrpcEndpoint());
```
grod220, line 106: Maybe it's my IDE, but it seems like `await onboardGrpcEndpoint()` is returning a `never` type

packages/services/src/ctx/full-viewing-key.ts
```diff
@@ -1,4 +1,6 @@
-import { createContextKey } from '@connectrpc/connect';
+import { Code, ConnectError, createContextKey } from '@connectrpc/connect';
 import { FullViewingKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 
-export const fvkCtx = createContextKey<FullViewingKey | undefined>(undefined);
+export const fvkCtx = createContextKey<() => Promise<FullViewingKey>>(() =>
+  Promise.reject(new ConnectError('No full viewing key available', Code.FailedPrecondition)),
+);
```
grod220, line 6: What is this doing? This may need a code comment for context.

packages/services/src/custody-service/authorize/index.ts
```diff
@@ -4,18 +4,19 @@ import { approverCtx } from '../../ctx/approver';
...
 import { isControlledAddress } from '@penumbra-zone/wasm/address';
 import { AuthorizeRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/custody/v1/custody_pb';
 import { fvkCtx } from '../../ctx/full-viewing-key';
+import { FullViewingKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 
 export const authorize: Impl['authorize'] = async (req, ctx) => {
   if (!req.plan) throw new ConnectError('No plan included in request', Code.InvalidArgument);
 
-  assertValidRequest(req, ctx);
+  assertValidRequest(req, await ctx.values.get(fvkCtx)());
```
grod220, line 19: A bit hard to read, can you unnest the await?

```
const fvk = await ctx.values.get(fvkCtx);
assertValidRequest(req, fvk);
```

packages/services/src/view-service/transaction-info.ts
```diff
@@ -3,16 +3,12 @@ import { servicesCtx } from '../ctx/prax';
...
 export const transactionInfo: Impl['transactionInfo'] = async function* (req, ctx) {
-  const services = ctx.values.get(servicesCtx);
+  const services = await ctx.values.get(servicesCtx);
   const { indexedDb } = await services.getWalletServices();
 
-  const fullViewingKey = ctx.values.get(fvkCtx);
-  if (!fullViewingKey) {
-    throw new ConnectError('Cannot access full viewing key', Code.Unauthenticated);
-  }
+  const fvk = ctx.values.get(fvkCtx);
```
grod220, line 11: Is it possible to get the same await syntax as servicesCtx? Aka:

```
const fvk = await ctx.values.get(fvkCtx);
```

apps/extension/src/storage/onboard.ts
turbocrime: i spent some time on this, but it gets complex quickly, so i'm no longer implementing this reuseable method. considering it out of scope.

packages/services/src/ctx/full-viewing-key.ts
turbocrime: much like the other context keys in this package, this calls `createContextKey` with a type and default value.

`createContextKey` is explained in ConnectRPC documentation.

https://connectrpc.com/docs/node/interceptors#context-keys

packages/services/src/view-service/transaction-info.ts
turbocrime: this would complicate things on the configuration side.

i think making servicesCtx a bare promise was probably an incorrect choice, after experimenting with both techniques, so i'll align them both to be functions.

apps/extension/src/service-worker.ts
```diff
@@ -13,86 +13,64 @@ import './listeners';
...
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-import { fvkCtx } from '@penumbra-zone/services/ctx/full-viewing-key';
-import { WalletJson } from '@penumbra-zone/types/src/wallet';
-
-/**
- * When a user first onboards with the extension, they won't have chosen a gRPC
- * endpoint yet. So we'll wait until they've chosen one to start trying to make
- * requests against it.
- */
-const waitUntilGrpcEndpointExists = async () => {
```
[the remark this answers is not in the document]
turbocrime: no longer a reuseable fn, but still relocated to onboard.ts

grod220 (approved):

TalDerei:
add changeset

PR #984, opened 2024-04-25
HandlerContext: spendKey

turbocrime:
rebased now that #983 is merged

adds spendKey to HandlerContext, simplifies approver context by providing a default, simplifies tests


apps/extension/src/service-worker.ts
```diff
@@ -109,17 +111,19 @@ const getServiceHandler = async () => {
       custodyClient ??= createDirectClient(CustodyService, handler, transportOptions);
       stakingClient ??= createDirectClient(StakingService, handler, transportOptions);
 
+      if (new URL(req.url).pathname.startsWith('/penumbra.custody.v1.Custody'))
+        contextValues.set(skCtx, getSpendKey);
```
Valentine1898, line 115: It might be worth considering splitting view-service and custody-service into different packages, because they need very different contexts

turbocrime: yes! as context improves this becomes way more possible.

turbocrime:
chained into #983

packages/services/src/custody-service/authorize.test.ts
turbocrime, line 1: todo: some of this test functionality should be relocated to extension's implementation of skCtx

apps/extension/src/ctx/spend-key.ts
```diff
@@ -0,0 +1,23 @@
...
+export const getSpendKey = async () => {
+  const pk = await sessionExtStorage.get('passwordKey');
+  if (!pk) throw new ConnectError('User must login to extension', Code.Unauthenticated);
+
+  const wallet0 = (await localExtStorage.get('wallets'))[0];
+  if (!wallet0) throw new ConnectError('No wallet found', Code.FailedPrecondition);
+
+  const decryptedSeedPhrase = await (
+    await Key.fromJson(pk)
+  ).unseal(Box.fromJson(wallet0.custody.encryptedSeedPhrase));
```
grod220, line 17: Can you unnest the double await here for better readability?

grod220 (approved):

PR #985, opened 2024-04-25
faster dev init, fix wasm clean

turbocrime:
jesse complained about dev startup slowing down so changing dep from build to compile should accelerate it.

i believe the build issues (package exports) present in https://github.com/penumbra-zone/web/commit/4ef2ef7c3874b9bc85a632455cf6456af46f208d that prompted this change are by #995, so there should be no negative effects of this, but watch out for broken HMR or other issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ node_modules
 
 # testing
 coverage
-vite.config.ts.timestamp-*-*.mjs
+vite*.config.ts.timestamp-*-*.mjs
 
 # build output
 .next/
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "all-check": "pnpm clean && pnpm install && pnpm compile && pnpm lint && pnpm lint:rust && pnpm build && pnpm test && pnpm test:rust",
     "build": "turbo build",
     "clean": "turbo clean",
-    "clean:vitest-mjs": "find . -type f -name 'vite.config.ts.timestamp-*-*.mjs' -ls -delete",
+    "clean:vitest-mjs": "find . -type f -name 'vite*.config.ts.timestamp-*-*.mjs' -ls -delete",
     "compile": "turbo compile",
     "dev": "turbo dev",
     "format": "turbo format",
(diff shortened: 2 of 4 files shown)


turbocrime:
threw in a couple script tweaks

.gitignore
```diff
@@ -5,7 +5,7 @@ node_modules
 
 # testing
 coverage
-vite.config.ts.timestamp-*-*.mjs
+vite*.config.ts.timestamp-*-*.mjs
```
turbocrime, line 8: this was missing a `vitest.config.ts.timestamp`...

package.json
```diff
@@ -8,7 +8,7 @@
     "all-check": "pnpm clean && pnpm install && pnpm compile && pnpm lint && pnpm lint:rust && pnpm build && pnpm test && pnpm test:rust",
     "build": "turbo build",
     "clean": "turbo clean",
-    "clean:vitest-mjs": "find . -type f -name 'vite.config.ts.timestamp-*-*.mjs' -ls -delete",
+    "clean:vitest-mjs": "find . -type f -name 'vite*.config.ts.timestamp-*-*.mjs' -ls -delete",
```
turbocrime, line 11: same here

packages/wasm/package.json
```diff
@@ -4,11 +4,11 @@
   "license": "(MIT OR Apache-2.0)",
   "type": "module",
   "scripts": {
-    "clean": "rm -v wasm/index*",
+    "clean": "rm -fv wasm/index*",
```
turbocrime, line 7: prevents failing if this is already clean

turbo.json
```diff
@@ -34,16 +34,19 @@
     },
     "dev": {
       "cache": false,
-      "dependsOn": ["build"],
+      "dependsOn": ["^compile"],
       "persistent": true
     },
```
turbocrime, line 39: dev requires compile of dependencies, and vite will dynamically build all the typescript, so we don't need to depend on build. depending on build only adds a startup delay and demands a perfect build, instead of allowing warnings like dev. so this change allows faster development.

full build requirements are still present in all-check and are enforced by ci, so we lose nothing.

jessepinho (approved):
EXCELLENT. Thank you! 🙏🏻  😭

turbocrime:
oh hmm i think the build issues are not solved. i am working on it now

turbocrime:
this needs to wait for #995

Issue #762, opened 2024-03-14
Dropping the /src/ 


grod220:
In https://github.com/penumbra-zone/web/pull/672 we removed a number of barrel files. While this is more performant, we are now appending `/src/` to a bunch of imports:

```typescript
import { createChannelTransport } from '@penumbra-zone/transport-dom/src/create';
```

It would be nice to be able to drop this and have: 

```typescript
import { createChannelTransport } from '@penumbra-zone/transport-dom/create';
```

Also, this is largely not an issue with external code imports as they can do:

```typescript
    "exports": {
      ".": "./dist/prax.js",
      "./global": "./dist/global.js",
      "./get-port": "./dist/get-port.js"
    }
```

However, doing like so with internal code renders a type error
```typescript
    "exports": {
      ".": "./src/prax.ts",
      "./global": "./src/global.ts",
      "./get-port": "./src/get-port.ts"
    }
```

turbocrime:
this can be handled by publishConfig

turbocrime:
there's actually a couple of these left, have some local work on it

PR #995, opened 2024-04-26
fix all package exports, remove `/src/` import path segments

turbocrime:
#762 

resolves dev/build issues relevant to #985

this removes all `/src/` import path segments from everything by providing the correct exports config in package.json

many imports throughout the repo have to be modified as a result


packages/polyfills/package.json
```diff
@@ -9,15 +9,22 @@
...
-      "./src/*": "./dist/*.js",
-      "./src/ReadableStream[Symbol.asyncIterator]": {
+      "./ReadableStream[Symbol.asyncIterator]": {
         "import": "./dist/ReadableStream_Symbol.asyncIterator_.js",
-        "types": "./dist/ReadableStream[Symbol.asyncIterator].d.ts"
+        "types": "./src/ReadableStream_Symbol.asyncIterator_.d.ts"
+      },
+      "./*": {
+        "import": "./dist/*.js",
+        "types": "./dist/*.d.ts"
```
turbocrime, line 27: replacing brackets with underscores for path compatibility concerns

packages/polyfills/src/Array.fromAsync.test.ts
```diff
@@ -1,7 +1,7 @@
 import { expect, describe, it } from 'vitest';
 
 import Array from './Array.fromAsync';
-import './ReadableStream[Symbol.asyncIterator]';
+import './ReadableStream_Symbol.asyncIterator_';
```
turbocrime, line 4: internally, underscore paths are literal, everywhere else they remain

packages/bech32m/tsconfig.json
turbocrime, line 1: bech32m build was broken due to failing to exclude dist

packages/query/package.json
```diff
@@ -8,6 +8,10 @@
     "lint": "eslint \"**/*.ts*\"",
     "test": "vitest run"
   },
+  "exports": {
+    "./*": "./src/*.ts",
+    "./quieriers": "./src/queriers/*.ts"
```
jessepinho, line 13:
```suggestion
    "./queriers": "./src/queriers/*.ts"
```

Also, isn't this case handled by `./*` above it?

jessepinho (approved):
NOICE (one lil typo, then merge)

packages/query/package.json
turbocrime: it does seem to function without that, so removed

Issue #996, opened 2024-04-26
main branch fails CI after b8c8260


turbocrime:
main failing ci after b8c8260 cc @grod220

PR #997, opened 2024-04-26
larger buildjet for publish and release ci

turbocrime:
attempt to address #996

### DIFF
--- a/.github/workflows/extension-publish.yml
+++ b/.github/workflows/extension-publish.yml
@@ -21,7 +21,7 @@ jobs:
   publish:
     environment: ext-publish
     name: Publish extension
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: buildjet-16vcpu-ubuntu-2204
     needs: turbo-compile
 
     steps:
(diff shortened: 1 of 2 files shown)


jessepinho (approved):
assuming this isn't gonna cost us $ (or that it's approved if so), ship it!

conorsch:
> assuming this isn't gonna cost us $ (or that it's approved if so), ship it!

We pay a flat fee for access to 164 concurrent vcpus. We can dip into that pool at any time. If we overprovision CI jobs, then the result is that the jobs will fight for access to vcpus, and get stuck in a "Queued" state while other jobs finish and free up more cpus.

For release-related workflows, higher concurrency is generally better. There's a lot to be said for caching, too, but that's more useful on every-PR jobs than on release workflows. 

So, this won't cost us more monthly, but it _may_ cause queuing. I don't think it'll be a problem, but if you notice more queuing, holler and I'll either fix it or just bump up the limit again.

turbocrime:
> So, this won't cost us more monthly, but it _may_ cause queuing. I don't think it'll be a problem, but if you notice more queuing, holler and I'll either fix it or just bump up the limit again.

we have some cache logic but it could probably be improved. we do experience significant queuing already - often it's the majority of any given CI time. it's still manageable.

i think for publish/release flows it's fine to take a huge bite, they are less frequent.

Issue #971, opened 2024-04-23
Don't require a network connection


turbocrime:
initializeWalletServices will throw if it can't make a network request at startup

https://github.com/penumbra-zone/web/blob/54d32866d6f9655fdaa5d6c35b908627152a6acd/packages/services-context/src/index.ts#L95-L96

this should gracefully degrade

Valentine1898:
The thing is that indexed-db is strictly tied to chain-id, which means that we can't replace calling `this.querier.app.appParams()` with reading data from indexed-db, because we first need to somehow find out the chain-id and only then we can initialize indexed-db. 

I can see an option that we can store the last used chain-id in local storage.
Then if we can't get the chain-id from the rpc, we can use the chain-id from local storage

Valentine1898:
But I think the more reliable way is that we can't provide reliable services without network, even if we solve the problem of getting `appParams()`, we get the problem of calling registry. 
Perhaps the best solution is to endlessly retry `initializeWalletServices` until the network is up

turbocrime:
fixed in #998

PR #998, opened 2024-04-26
start up without chainId (or network)

turbocrime:
this allows prax to init without a known chain id, by querying chain id from remote RPC, or using chain id from local storage.

as a result prax is more robust and can function while remote RPC is unavailable.

ui components such as sync progress also now function properly when the RPC is unavailable.


turbocrime:
#971

packages/ui/components/ui/block-sync-status/condensed.tsx
turbocrime, line 1: block sync status components probably need to be using BigInt but that's out of scope for this pr

packages/ui/components/ui/block-sync-status/large.tsx
turbocrime, line 1: seems like this is unused

packages/storage/src/chrome/types.ts
```diff
@@ -15,9 +17,10 @@ export interface OriginRecord {
...
-  grpcEndpoint?: string;
+  grpcEndpoint: string | undefined;
   frontendUrl: string;
-  passwordKeyPrint?: KeyPrintJson;
-  fullSyncHeight?: number;
+  passwordKeyPrint: KeyPrintJson | undefined;
+  fullSyncHeight: number | undefined;
   knownSites: OriginRecord[];
+  params: Jsonified<AppParameters> | undefined;
 }
```
turbocrime, line 26: in order to successfully return undefined from the storage getter, undefined had to be explicitly added to the state type.

packages/query/src/queriers/shielded-pool.ts
```diff
@@ -14,8 +14,9 @@ export class ShieldedPoolQuerier implements ShieldedPoolQuerierInterface {
...
 
-  async assetMetadataById(assetId: AssetId): Promise<Metadata | undefined> {
-    const res = await this.client.assetMetadataById({ assetId });
-    return res.denomMetadata;
-  }
+  assetMetadataById = (assetId: AssetId): Promise<Metadata | undefined> =>
+    this.client.assetMetadataById({ assetId }).then(
+      ({ denomMetadata }) => denomMetadata,
+      () => undefined,
+    );
```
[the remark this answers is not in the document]
grod220: I'd say the same for `latestBlockHeight()` below too

packages/services-context/src/index.ts
```diff
@@ -88,12 +90,32 @@ export class Services implements ServicesInterface {
...
+    const queriedParams = await this.querier.app.appParams().catch((e: unknown) => void e);
+
+    // prefer fetched params
+    const chainId = queriedParams?.chainId ?? storedParams?.chainId;
+    const epochDuration =
+      queriedParams?.sctParams?.epochDuration ?? storedParams?.sctParams?.epochDuration;
+
+    // store chainId if it was missing
+    if (!storedParams?.chainId && queriedParams?.chainId)
+      await localExtStorage.set('params', queriedParams.toJson() as Jsonified<AppParameters>);
```
grod220, line 108: Brackets here please

packages/services-context/src/index.ts
```diff
@@ -88,12 +90,32 @@ export class Services implements ServicesInterface {
...
+    const storedParams = await localExtStorage
+      .get('params')
+      .then(j => j && AppParameters.fromJson(j));
+    // try to fetch params from network
+    const queriedParams = await this.querier.app.appParams().catch((e: unknown) => void e);
+
+    // prefer fetched params
+    const chainId = queriedParams?.chainId ?? storedParams?.chainId;
+    const epochDuration =
+      queriedParams?.sctParams?.epochDuration ?? storedParams?.sctParams?.epochDuration;
```
grod220, line 104: Can we move the asserts closer to where the variables are set? Something like:

```ts
    const chainId = queriedParams?.chainId ?? storedParams?.chainId;
    if (!chainId) throw new Error('Cannot initialize without chainId');
    // fail if chainId is mismatched
    if (storedParams?.chainId && storedParams.chainId !== chainId)
      throw new Error(
        'Stored chainId does not match the remote chainId. Your local state may\
        be invalid, and you cannot use this remote endpoint without clearing\
        your cached chain.',
      );

    const epochDuration =
      queriedParams?.sctParams?.epochDuration ?? storedParams?.sctParams?.epochDuration;
    if (!epochDuration) throw new Error('Cannot initialize without epoch duration');

    // store chainId if it was missing
    if (!storedParams?.chainId && queriedParams?.chainId) {
      await localExtStorage.set('params', queriedParams.toJson() as Jsonified<AppParameters>);
    }
```

packages/services/src/view-service/status-stream.ts
```diff
@@ -4,18 +4,14 @@ import { servicesCtx } from '../ctx/prax';
...
-  }
+  // This should stream forever unless cancelled.
+  let remoteBlockHeight: bigint | undefined;
+  for await (const { value: syncHeight } of indexedDb.subscribe('FULL_SYNC_HEIGHT'))
+    if ((remoteBlockHeight ??= await services.querier.tendermint.latestBlockHeight()))
+      yield {
+        latestKnownBlockHeight: syncHeight <= remoteBlockHeight ? remoteBlockHeight : syncHeight,
+        partialSyncHeight: syncHeight,
+        fullSyncHeight: syncHeight,
+      };
```
grod220, line 16: The lack of brackets make this hard to understand. Don't think we should ever do bracket-less loops.

packages/services/src/view-service/status-stream.ts
```diff
@@ -4,18 +4,14 @@ import { servicesCtx } from '../ctx/prax';
...
-    yield {
-      latestKnownBlockHeight: syncHeight >= latestBlockHeight ? syncHeight : latestBlockHeight,
-      partialSyncHeight: syncHeight,
-      fullSyncHeight: syncHeight,
-    };
-  }
+  // This should stream forever unless cancelled.
+  let remoteBlockHeight: bigint | undefined;
+  for await (const { value: syncHeight } of indexedDb.subscribe('FULL_SYNC_HEIGHT'))
+    if ((remoteBlockHeight ??= await services.querier.tendermint.latestBlockHeight()))
```
grod220, line 11: Think this is a bit more understandable if we can pull this out, aka:

```ts
  for await (const { value: syncHeight } of indexedDb.subscribe('FULL_SYNC_HEIGHT')) {
    remoteBlockHeight ??= await services.querier.tendermint.latestBlockHeight();
    if (remoteBlockHeight) {
      yield {
```

packages/services-context/src/index.ts
turbocrime: i'm going to make a PR fixing the eslint config to reenable react-hooks, and i'll add a linter rule for braces at that point

packages/services-context/src/index.ts
turbocrime: idb can start up without epochDuration, so they are deliberately separated. i'll pull out the params request into a method to simplify it conceptually

packages/services-context/src/index.ts
turbocrime, line 1: full config is now guaranteed at init, so it doesn't need the promise logic anymore

PR #999, opened 2024-04-27
document client

turbocrime:
brief docs demonstrating client init, just to have something we can link.

### DIFF
--- a/docs/extension-services.md
+++ b/docs/extension-services.md
@@ -38,20 +38,39 @@ import { ViewService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/
 const viewClient = createPraxClient(ViewService);
 ```
 
+An incredibly simple use might be something like this.
+
+```ts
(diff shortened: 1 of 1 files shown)


docs/extension-services.md
```diff
@@ -38,20 +38,39 @@ import { ViewService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/
...
+```
+
+Use is identical.
+
+```ts
+const { fullViewingKey } = await custodyClient.exportFullViewingKey({});
+void fetch('https://example.com/iamveryclever', {
+  method: 'POST',
+  body: fullViewingKey.toJsonString(),
+});
```
hdevalence, line 73: probably we don't want this to be the example since the prax custody service doesn't expose this api on purpose

turbocrime: lol ok updated

hdevalence (approved):

Issue #1004, opened 2024-04-29
Minifront cannot issue swap claims


grod220:
Error message thrown:
```
Error: [unknown] Tendermint error 1: failed to deliver transaction: check_stateless failed: 
a swap claim proof did not verify: proof did not verify
```

Tested on https://github.com/penumbra-zone/web/pull/1002 where proving keys are on the latest version

PR #1006, opened 2024-04-29
Support v0.73.0 swap claims

grod220:
Closes https://github.com/penumbra-zone/web/issues/1004


apps/extension/.env.testnet
```diff
@@ -1,4 +1,4 @@
 CHAIN_ID=penumbra-testnet-deimos-6
-IDB_VERSION=34
+IDB_VERSION=35
```
grod220, line 2: So we can get those old swaps stored with sct positions

packages/transport-dom/src/create.ts
```diff
@@ -184,7 +185,7 @@ export const createChannelTransport = ({
             Code.OutOfRange,
           );
       } else {
-        const stream = ReadableStream.from(input).pipeThrough(
+        const stream: ReadableStream<JsonValue> = ReadableStream.from(input).pipeThrough(
```
grod220, line 188: New lint error from protobuf pkg update

packages/wasm/crate/src/tree.rs
```diff
@@ -0,0 +1,18 @@
...
+#[wasm_bindgen]
+pub fn sct_position(block_height: u64, epoch_bytes: &[u8]) -> WasmResult<u64> {
+    let epoch = Epoch::decode(epoch_bytes)?;
+    let epoch_index = u16::try_from(epoch.index)?;
+    // The block index is determined by looking at how many blocks have elapsed since
+    // the start of the epoch.
+    let block_index = u16::try_from(block_height - epoch.start_height)?;
+
+    let position = Position::from((epoch_index, block_index, 0u16));
+    Ok(position.into())
```
Valentine1898, line 17: Seems like we could have done it without wasm?

Valentine1898 (approved):
LGTM!

packages/wasm/crate/src/tree.rs
grod220: The tree position is serialized using bitwise operators: https://github.com/penumbra-zone/penumbra/blob/8d4dc0330a8bf86a4334987a02f0fa04ac40baa0/crates/crypto/tct/src/index.rs#L228-L238.

I thought about implementing a version of this in typescript, but I figured 1) it's easier to use what they created and 2) we'll be attempting to migrate this to wasm soon anyway.

Valentine1898: Makes sense

PR #1010, opened 2024-04-29
don't throw if metadata unavailble, just return empty response

### LINKED ISSUE #1005: Minifront should not fail if no asset metadata is found
If you clear the cache in extension and go to the swap page, you will get the error 
```typescript
ConnectError: [unknown] No denom metadata found for asset: {"inner":"CKBQapu+DkQpsKyTfKESLTV19/NPWR5sNZtvQsd3Hw8="}
``` 
image

Most likely this error is related to displaying swaps that were not claimed, so you can only reproduce it if you have some swaps that were not claimed

It may also be due to the fact that when clearing the cache we delete all data from indexed-db but do not restart service-worker, in this case data from asset registry will not be written to indexed-db (not confirmed).

turbocrime:
fixes #1005 

not sure how to check this functionality works as intended, on the minifront side


Valentine1898 (approved):
LGTM! 
This fixes all the bugs for me, but it might be worth improving the balance display a bit more  
image

packages/services/src/view-service/asset-metadata-by-id.ts
```diff
@@ -1,19 +1,25 @@
...
+  if (localMetadata) return { denomMetadata: localMetadata };
+
+  const remoteMetadata = await querier.shieldedPool.assetMetadataById(assetId);
+  if (remoteMetadata) {
+    void indexedDb.saveAssetsMetadata(remoteMetadata);
+    return { denomMetadata: remoteMetadata };
+  }
 
-  return { denomMetadata };
+  return { denomMetadata: undefined };
```
Valentine1898, line 24: image
Based on the documentation rpc to buf is the correct behavior, we should not throw an error

PR #1012, opened 2024-04-29
Expose `batch-swap-output-data` getters to package consumers

ejmg:
# Summary

It looks like the getters defined for `BatchSwapOutputData` are exported but never exposed in `vite.config.ts` for downstream users to access. This PR is a one line change to `vite.config.ts` that changes that.


grod220 (approved):
LGTM

grod220:
Happy we are sharing these now :)

Issue #942, opened 2024-04-16
Tracking issue: Dutch auction support in the web 


TalDerei:
Support the ability to perform Gradual Dutch Actions (GDAs). References https://github.com/penumbra-zone/penumbra/issues/4196

- [x] #944
- [x] https://github.com/penumbra-zone/web/issues/978
- [x] #1018
- [x] #1019 
- [x] #1032
- [x] #943
- [x] #980
- [x] #1020
- [x] #1059
- [x] https://github.com/penumbra-zone/web/issues/981
- [x] #1081
- [x] #1088
- [x] #1155
- [x] #1146
- [x] #1157
- [x] https://github.com/penumbra-zone/web/issues/1181
- [x] #1156
- [x] #1198
- [x] #1284
- [x] #1277
- [x] #1285
- [x] #1090
- [x] https://github.com/penumbra-zone/web/issues/1212
- [x] #1166

TalDerei:
currently blocked until **pcli M0** in dependency graph is complete.

Issue #980, opened 2024-04-24
prax: implement `auctions` RPC method


TalDerei:
- [x] implement `queryLatestState`
- [x] implement `includeInactive`

PR #1015, opened 2024-04-30
Process Dutch auction NFT metadata in the block processor

jessepinho:
As part of #942, I've made a massive branch with a ton of changes to the Dutch auction UI. To make the review process more manageable, I'm breaking the changes into smaller atomic PRs.

This PR modifies the block processor to identify Dutch auctions, and save their NFTs' metadata to the database when it encounters them. Future PRs will build on top of this.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -14,7 +14,10 @@ import {
   CommitmentSource,
   Nullifier,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/sct/v1/sct_pb';
-import { Transaction } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import {
+  Action,
+  Transaction,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 import { TransactionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/txhash/v1/txhash_pb';
 import { StateCommitment } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/crypto/tct/v1/tct_pb';
 import {
@@ -37,6 +40,7 @@ import { PRICE_RELEVANCE_THRESHOLDS } from '@penumbra-zone/constants/assets';
 import { toDecimalExchangeRate } from '@penumbra-zone/types/amount';
 import { ValidatorInfoResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/stake/v1/stake_pb';
 import { uint8ArrayToHex } from '@penumbra-zone/types/hex';
+import { getAuctionId, getAuctionNftMetadata } from '@penumbra-zone/wasm/auction';
 
 declare global {
   // `var` required for global declaration (as let/const are block scoped)
@@ -52,9 +56,15 @@ interface QueryClientProps {
   stakingTokenMetadata: Metadata;
 }
 
-const blankTxSource = new CommitmentSource({
+const BLANK_TX_SOURCE = new CommitmentSource({
   source: { case: 'transaction', value: { id: new Uint8Array() } },
 });
+
+const POSITION_STATES: PositionState[] = [
+  new PositionState({ state: PositionState_PositionStateEnum.OPENED }),
+  new PositionState({ state: PositionState_PositionStateEnum.CLOSED }),
+  new PositionState({ state: PositionState_PositionStateEnum.WITHDRAWN, sequence: 0n }),
+];
 
 export class BlockProcessor implements BlockProcessorInterface {
   private readonly querier: RootQuerier;
@@ -142,7 +152,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         if (txCommitments.some(txCommitment => stateCommitment.equals(txCommitment))) {
           txId ??= new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
           relevantTx.set(txId, tx);
-          if (blankTxSource.equals(spendableNoteRecord.source)) {
+          if (BLANK_TX_SOURCE.equals(spendableNoteRecord.source)) {
             spendableNoteRecord.source = new CommitmentSource({
               source: { case: 'transaction', value: { id: txId.inner } },
             });
@@ -261,7 +271,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         // - detect LpNft position opens
         // - generate all possible position state metadata
         // - update idb
-        await this.identifyLpNftPositions(blockTx);
+        await this.processTransactions(blockTx);
 
         // at this point txinfo can be generated and saved. this will resolve
         // pending broadcasts, and populate the transaction list.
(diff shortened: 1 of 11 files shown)


packages/query/src/block-processor.ts
```diff
@@ -52,10 +56,16 @@ interface QueryClientProps {
   stakingTokenMetadata: Metadata;
 }
 
-const blankTxSource = new CommitmentSource({
+const BLANK_TX_SOURCE = new CommitmentSource({
```
jessepinho, line 59: Just a rename to make it clear that this is a never-changing constant

packages/query/src/block-processor.ts
```diff
@@ -52,10 +56,16 @@ interface QueryClientProps {
...
   source: { case: 'transaction', value: { id: new Uint8Array() } },
 });
 
+const POSITION_STATES: PositionState[] = [
```
jessepinho, line 63: Sort of the same here — extracting this constant from inside of a function, where it was being defined on every function call despite never changing.

packages/query/src/block-processor.ts
```diff
@@ -361,45 +371,66 @@ export class BlockProcessor implements BlockProcessorInterface {
     return spentNullifiers;
   }
 
-  private async identifyLpNftPositions(txs: Transaction[]) {
```
jessepinho, line 364: You'll probably want to review the rest of this file with "Hide whitespace" turned on. In short, I:

1. created a new `processTransactions` method that iterates over each transaction, then each action of each transaction, and calls `this.identifyAuctionNfts()` and `this.identifyLpNftPositions()` on each _action_.
2. created a new `identifyAuctioNfts()` method that saves auction NFT metadata when it appears.
3. modified `identifyLpNftPositions()` to process individual _actions_, rather than processing an array of transactions.

packages/wasm/crate/src/metadata.rs
```diff
@@ -73,21 +83,32 @@ mod test_helpers {
 
     use super::*;
 
-    pub fn get_metadata_for(display_denom: &str) -> Metadata {
+    pub fn get_metadata_for(display_denom: &str, base_denom_is_display_denom: bool) -> Metadata {
```
jessepinho, line 86: This is just a test helper. Auction NFTs only have one denom unit, so the base denom is the same as the display denom. So I added this arg to this function so that I can generate the right type of metadata for auction NFTs as well.

packages/wasm/crate/src/planner.rs
```diff
@@ -101,7 +101,7 @@ impl ActionList {
         for value in self.balance().provided() {
             self.change_outputs.insert(
                 value.asset_id,
-                OutputPlan::new(&mut OsRng, value, change_address),
+                OutputPlan::new(&mut OsRng, value, change_address.clone()),
```
jessepinho, line 104: For some reason, Rust was complaining about ownership issues here, despite there being no other changes to the planner.

packages/wasm/crate/src/tx.rs
```diff
@@ -298,21 +298,24 @@ pub async fn transaction_info_inner(
         match action_view {
             ActionView::Spend(SpendView::Visible { note, .. }) => {
                 let address = note.address();
```
jessepinho, line 300: More fixes to Rust ownership issues in this file

packages/wasm/crate/src/tx.rs
```diff
@@ -297,37 +297,48 @@ pub async fn transaction_info_inner(
         };
         match action_view {
             ActionView::Spend(SpendView::Visible { note, .. }) => {
-                let address = note.address();
```
jessepinho, line 300: More changes due to ownership issues.

Also, note that `address_views` used to be a `BTreeMap` of addresses to address views. Now, it's a map of addresses-encoded-to-byte-arrays to address views. That's because [a recent change](https://github.com/penumbra-zone/penumbra/pull/4285/files#diff-bb4c349bac5ab2231b6774ebbfdaabe48a985d24599f872525f90916a3e9fb76R37) to `Address` resulted in `Address` structs being mutable, since they hold a property of type `OnceLock<...>`. Clippy complains about the use of mutable keys, since their serialization can change over time when they're mutable, thus destroying the integrity of the key.

To work around this, @aubrika and I realized we could use the byte-encoded addresses as keys, rather than the `Address` objects themselves.

packages/query/src/block-processor.ts
```diff
@@ -361,45 +371,66 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      const metadata = getAuctionNftMetadata(
+        action.value.auctionId,
+        // Always a sequence number of 1 when ending a Dutch auction
+        1n,
+      );
+      await this.indexedDb.saveAssetsMetadata(metadata);
+    }
+    /**
+     * @todo Handle `actionDutchAuctionWithdraw`, and figure out how to
+     * determine the sequence number if there have been multiple withdrawals.
```
erwanor, line 401: I think one way to do that is to structure the `Action['action']` with a sequence number, rather than adding it ex-post in this method, unsure

erwanor (approved):

packages/query/src/block-processor.ts
TalDerei: this a great design choice as it's extensible to other metadata-related processing we'll need to do in the future!

packages/query/src/block-processor.ts
```diff
@@ -261,7 +271,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         // - detect LpNft position opens
         // - generate all possible position state metadata
         // - update idb
-        await this.identifyLpNftPositions(blockTx);
```
TalDerei, line 264: can we update the description above `processTransactions(blockTx)` to accurately describe this function? Can we also move these existing comments to `identifyLpNftPositions()` and add comments to `identifyAuctionNfts()`.

packages/wasm/crate/Cargo.toml
```diff
@@ -15,19 +15,22 @@ default = ["console_error_panic_hook"]
...
-penumbra-proof-params  = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-proof-params", default-features = false }
-penumbra-proto         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-proto", default-features = false }
-penumbra-sct           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-sct", default-features = false }
-penumbra-shielded-pool = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-shielded-pool", default-features = false }
-penumbra-stake         = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-stake", default-features = false }
-penumbra-tct           = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-tct" }
-penumbra-transaction   = { git = "https://github.com/penumbra-zone/penumbra.git", tag = "v0.73.0", package = "penumbra-transaction", default-features = false }
+# TODO: Use `tag` instead of `rev` once auctions land in a tagged release of
```
TalDerei, line 18: 👍

packages/wasm/crate/src/planner.rs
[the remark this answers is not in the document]
TalDerei: yeah was getting lot of these copy errors too without cloning

TalDerei (approved):

TalDerei:
> save their NFTs' metadata to the database

we're currently only saving the auction metadata to the `ASSETS` table in [indexed-db](https://github.com/penumbra-zone/web/blob/c73a8e2cac816b44540003eabe3a06ce876127ea/packages/storage/src/indexed-db/index.ts#L210), where the `assetId` is used as a key for the table. 

can you create a separate issue in the tracking issue for creating a auction state table which stores the auction id and note commitment?
- Add an index-db table (https://github.dev/penumbra-zone/web/blob/c73a8e2cac816b44540003eabe3a06ce876127ea/packages/types/src/indexed-db.ts#L51-L52). Reference the view server PR for the sql construction. Make the `AuctionId` the key in the table.
- Define methods for reading, writing, and updating data in the table (https://github.dev/penumbra-zone/web/blob/c73a8e2cac816b44540003eabe3a06ce876127ea/packages/storage/src/indexed-db/index.ts#L79-L80)

TalDerei:
The functionality seems to be all here, LGTM!

packages/wasm/crate/src/metadata.rs
jessepinho: Yeah — the display denom is the one you see in the UI (e.g., "penumbra") but the base denom is the smallest possible unit (e.g., "upenumbra"). Usually they're different, just like US dollars and cents are different, but auction NFTs apparently only have a single denom unit — which makes sense, since you'll never hold more than 1 NFT for a given auction.

packages/query/src/block-processor.ts
jessepinho: Cool, thanks, added a link to this comment in the ticket (#1020)

packages/query/src/block-processor.ts
jessepinho: Whoops, good catch — fixed.

jessepinho:
@TalDerei 
> can you create a separate issue in the tracking issue for that builds on this?

Yeah, that's covered by #980. I didn't specifically fill in the details since I know I'll be working on it myself anyway (plus I'll be referencing the pairing doc you created).

PR #1016, opened 2024-04-30
publish `@penumbra-zone/protobuf`

turbocrime:
this breaks the protobuf message `typeRegistry` out of the types package, in preparation for some packaging work discovered necessary during writing of the client demo app. follow-on PRs soon


packages/protobuf/package.json
```diff
@@ -0,0 +1,42 @@
...
+        "types": "./dist/registry.d.ts"
+      }
+    }
+  },
+  "bundleDependencies": [
+    "@buf/cosmos_ibc.bufbuild_es",
+    "@buf/cosmos_ibc.connectrpc_es",
+    "@buf/penumbra-zone_penumbra.bufbuild_es",
+    "@buf/penumbra-zone_penumbra.connectrpc_es"
+  ],
```
turbocrime, line 29: this field directs npm to pack a `node_modules` with these dependencies in the emitted tarball. this means users should not have to configure the buf registry in order to use this package.

packages/transport-dom/src/create.ts
```diff
@@ -49,8 +49,8 @@ export interface ChannelTransportOptions
 
 export const createChannelTransport = ({
   getPort,
-  defaultTimeoutMs,
   jsonOptions,
+  defaultTimeoutMs = 10_000,
```
turbocrime, line 53: applied a default here, instead of using a shared configuration object repo-wide. this seems more consumeable

packages/constants/package.json
turbocrime, line 1: even before this PR, constants was no longer using these packages.

packages/client/package.json
turbocrime, line 1: client no longer depends on `@penumbra-zone/types`

packages/protobuf/README.md
```diff
@@ -0,0 +1,35 @@
...
+### Simple example
+
+```ts
+import { jsonOptions } from '@penumbra-zone/protobuf';
+
+import { createChannelTransport } from '@penumbra-zone/transport-dom';
+import type { ServiceType } from '@bufbuild/protobuf';
+
+// unsafely get first available provider
+const getPort = () => Object.values(window[Symbol.for('penumbra')])[0].connect();
```
grod220, line 27: Shouldn't we be encouraging them to use `getAnyPenumbraPort` from the client package here instead?

grod220 (approved):

TalDerei (approved):
LGTM

packages/protobuf/README.md
turbocrime: this example intentionally demonstrates use in isolation.

developers using the client package will use the features in the client package that conceal this, and will be redirected by the callout section before this example. anyone consuming this package instead will only be confused or distracted by example references to the client package.

Issue #978, opened 2024-04-24
prax: add auction planner support


TalDerei:
As a precursor to the more involved planner refactor tracked by https://github.com/penumbra-zone/web/issues/876, this will modify the existing web planner logic in wasm to include **auction** actions to unblock the prax auction work.

PR #1017, opened 2024-04-30
Account for Dutch auction schedules/ends in the WASM planner

jessepinho:
Continuing to split out PRs related to #942 into smaller, reviewable chunks.

This PR accounts for `ActionDutchAuctionSchedule` and `ActionDutchAuctionEnd` actions in a transaction planner request. Thanks to @TalDerei for pairing with me on this.

### DIFF
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -3,8 +3,12 @@ use std::collections::BTreeMap;
 use anyhow::anyhow;
 use ark_ff::UniformRand;
 use decaf377::{Fq, Fr};
-use penumbra_asset::asset::Metadata;
+use penumbra_asset::asset::{Id, Metadata};
 use penumbra_asset::{asset, Balance, Value};
+use penumbra_auction::auction::dutch::{
+    ActionDutchAuctionEnd, ActionDutchAuctionSchedule, DutchAuctionDescription,
+};
+use penumbra_auction::auction::AuctionId;
 use penumbra_dex::swap_claim::SwapClaimPlan;
 use penumbra_dex::{
     swap::{SwapPlaintext, SwapPlan},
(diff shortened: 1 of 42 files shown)


packages/wasm/crate/src/planner.rs
```diff
@@ -384,6 +388,54 @@ pub async fn plan_transaction(
...
+            ActionDutchAuctionSchedule {
+                description: DutchAuctionDescription {
+                    start_height: description.start_height,
+                    end_height: description.end_height,
+                    step_count: description.step_count,
+                    input,
+                    output_id,
+                    min_output,
+                    max_output,
+                    nonce: [0; 32], // TODO: Use a real nonce
```
erwanor, line 421: I think this needs to be filled out right?

jessepinho: Ah, scratch that — I just read the very next line 🤦🏻‍♂️ will fix, one sec...

jessepinho: @erwanor [fixed](https://github.com/penumbra-zone/web/pull/1017/commits/d111fea61c0b59cbe2e7bb8c7be1d7a13ac92e0a)

erwanor: The reason we can't set the nonce to zero is that doing so would restrict the ability to submit an auction with those parameters to only the first person who tries (that would be funny actually)

grod220 (approved):

packages/wasm/crate/src/planner.rs
```diff
@@ -384,6 +388,56 @@ pub async fn plan_transaction(
...
+        let auction_id: AuctionId = auction_id
+            .ok_or_else(|| anyhow!("missing auction ID in Dutch auction end action"))?
+            .try_into()?;
+
+        actions.push(ActionPlan::ActionDutchAuctionEnd(ActionDutchAuctionEnd {
+            auction_id,
+        }));
+    }
+
+    // TODO: Handle Dutch auction withdraws
```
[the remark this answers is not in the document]
jessepinho: Yep, see [here](https://github.com/penumbra-zone/web/issues/1020)

TalDerei:
References https://github.com/penumbra-zone/web/issues/978

PR #1021, opened 2024-05-01
See interchain balances [part 3]

grod220:
Allows us to see user balances of chains we have ibc connections to. This has been copied over and modified from the [asset lists interchain ui](https://cosmology.zone/products/create-cosmos-app) example.

Next PR will then use the balances list to construct a form that allows for IBC'ing in. 

https://github.com/penumbra-zone/web/assets/16624263/329de322-59c5-4d42-bb8c-eb038443324d


apps/extension/package.json
```diff
@@ -35,7 +35,7 @@
     "@penumbra-zone/types": "workspace:*",
     "@penumbra-zone/ui": "workspace:*",
     "@penumbra-zone/wasm": "workspace:*",
-    "@tanstack/react-query": "^5.28.9",
+    "@tanstack/react-query": "4.36.1",
```
grod220, line 38: This is somewhat sad. `osmos-query` is currently pinned on an old version. This means we need to downgrade until the team has updated theirs. We've flagged this with them and they are prioritizing it.

apps/minifront/src/components/ibc/layout.tsx
```diff
@@ -13,7 +13,7 @@ export const IbcLayout = () => {
             direction='right'
             // Negative calculated margin giving lint issue
             /* eslint-disable-next-line tailwindcss/enforces-negative-arbitrary-values */
-            className='invisible absolute -top-44 right-0 z-0 -mr-[calc(30vw-3px)] size-[30vw] text-stone-300 md:visible'
+            className='invisible absolute -top-32 right-0 z-0 -mr-80 size-80 text-stone-300 md:visible'
```
grod220, line 16: The resizing of these arrows are causing strange behaviors. Going to make them fixed sizes.

Valentine1898 (approved):
LGTM

PR #1022, opened 2024-05-02
Update web code to handle latest auction-related Protobuf changes

jessepinho:
🚨 18 of the changed files are just `package.json`s and `Cargo.toml/lock`.

A `TransactionView`'s `bodyView.actionViews` property is an array of [`ActionView`s](https://github.com/penumbra-zone/penumbra/blob/main/proto/penumbra/penumbra/core/transaction/v1/transaction.proto#L194-L236) containing either A) an action object, or B) a `*View` object representing the action. In the case of B, the `*View` object usually contains additional data needed to render the action, such as asset metadata, etc.

A [recent core PR](https://github.com/penumbra-zone/penumbra/pull/4295/files#diff-3539cae533f3b760a0c57c6e348615c7db70291e6df9a1698c6e869714753fea) updated the types for Dutch auction schedule and Dutch auction end actions so that they are now represented by `*View` objects containing necessary metadata/etc.

This PR takes advantage of that change, filling in the `*View` fields for Dutch auction-related actions and then rendering the correct metadata in the transaction view.


apps/minifront/src/components/tx-details/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { Card } from '@penumbra-zone/ui/components/ui/card';
 import { FadeTransition } from '@penumbra-zone/ui/components/ui/fade-transition';
-import { TxViewer } from './hash-parser';
```
jessepinho, line 3: Not strictly part of this PR, but it was always bothering me how that file was named 😆

packages/wasm/crate/src/planner.rs
```diff
@@ -410,19 +410,27 @@ pub async fn plan_transaction(
...
+            nonce,
+        };
+
+        save_auction_nft_metadata_if_needed(
+            &description,
+            &storage,
+            // When scheduling a Dutch auction, the sequence number is always 0
+            0,
+        )
+        .await?;
```
jessepinho, line 430: Now that we're actually able to render a `ValueView` of the action NFT in the transaction approval dialog, we need to optimistically save the NFT metadata to the database so that the dialog has metadata to render.

grod220 (approved):
LGTM

turbocrime:
i think it was necessary to changeset all packages that depend on the types packages, to trigger a release. trying to consume our published packages is resulting in peer conflicts because some of them have not been released with these new versions.

packages/constants/package.json
turbocrime, line 1: is there a reason these were added

packages/types/package.json
turbocrime, line 1: it seems like extra deps were also added here

jessepinho: Huh... good catch. Looks like I didn't review these super closely when I was self-reviewing. That said, I believe this would only have happened due to `fix-mismatches` — I didn't install these extra deps manually myself

turbocrime: i think it was probably due to rebase/merge details, i removed these immediately before your branch pushed to main

PR #1023, opened 2024-05-02
fix polyfills package, don't publish polyfills package

turbocrime:
`Array.fromAsync` declarations fixed, no longer requiring duplication.

our current version of typescript provides `Map.groupBy` and `Promise.withResolvers` in the `ESnext` lib, so these are removed, and we target `ESnext`. consumers of our packages will be required to use `typescript@^5.4.3` or provide their own polyfill.

the remaining packages that depend directly on polyfills are 

- `apps/minifront` - not a consumable package, ok to depend on internal monorepo package
- `@penumbra-zone/services` - intended for third party consumption. no build, already broken publish
- `@penumbra-zone/storage` - unsuitable for third party consumption, no build, already broken publish
- `@penumbra-zone/transport-dom` - intended for third party consumption, vite build, already broken publish

transport-dom is necessary to fix for client package consumption. the vite build already inlines polyfills - polyfills are a dev dependency, not a runtime dependency, so it's fine to have a private package in the development monorepo for this purpose.

transport-dom fixes to follow in another PR.

services remains broken pending a large amount of other work.


turbocrime:
the existing `@penumbra-zone/polyfills` on npm may require an `npm unpublish` or it will linger

packages/eslint-config-custom/index.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
     '@typescript-eslint/no-confusing-void-expression': ['error', { ignoreArrowShorthand: true }],
     '@typescript-eslint/switch-exhaustiveness-check': 'error',
     '@typescript-eslint/no-unnecessary-condition': ['error', { allowConstantLoopConditions: true }],
+    '@typescript-eslint/no-invalid-void-type': 'off',
```
jessepinho, line 38: This is weird — according to [the docs](https://typescript-eslint.io/rules/no-invalid-void-type/#allowingenerictypearguments), it shouldn't give a linting error when void is used in a generic type argument. Did you look into this? I just tried re-enabling the rule with `allowInGenericTypeArguments: true`, and it's still giving an error.

Not necessary to fix in this PR, but if you know the reason, it'd be great to not have to disable this rule.

jessepinho (approved):
LGTM, with a single comment I was wondering about.

Just so I understand, does targeting `ESNext` mean these polyfills get compiled into the codebase?

Nice work!

packages/eslint-config-custom/index.js
turbocrime: notice the docs are regarding type definitions, not function calls - in `Promise.withResolvers<void>()`, it's not a generic, you're parameterizing the types of actual values on the returned object

turbocrime:
> Just so I understand, does targeting `ESNext` mean these polyfills get compiled into the codebase?

targeting esnext means that the polyfills are not polyfilled - they are referenced literally in the build output, as the target environment is assumed to support them.

jessepinho:
@turbocrime 
> the target environment is assumed to support them

Got it. Just a reminder to check something like [CanIUse.com](https://caniuse.com/?search=Array.fromAsync) for a given browser API to make sure that there's plenty of global adoption of the browser version that has that API. (Looks like we're safe for `Promise.withResolvers` and `Map.groupBy`, though.)

turbocrime:
we're able to demand a chrome version in the extension manifest, and we should do that

PR #1024, opened 2024-05-02
fix transport-dom package

turbocrime:
- correct `publishConfig` paths
- correct location of `files` field
- `tsconfig.json` excludes build output
- `prepack` script runs build
- `prebuild` script runs clean
- `clean` script removes pack outputs
- vite build reconfigured
  - `vite-plugin-externalize-deps` to actually use dependencies
  - except for `@penumbra-zone/polyfills` which is configured to inline

### DIFF
--- a/.changeset/five-baboons-itch.md
+++ b/.changeset/five-baboons-itch.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/transport-dom': major
+---
+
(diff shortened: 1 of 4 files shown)


turbocrime:
chained to #1023, retargeting to main on merge

jessepinho (approved):

PR #1025, opened 2024-05-02
`ViewService#auctions` RPC method and basic auctions UI

jessepinho:
image

This PR does two things: 1) implements `ViewService#auctions`, and 2) creates a basic UI for viewing your existing auctions (basically as a proof of concept that the RPC method works).

## In this PR
- Create a basic UI for rendering the user's auctions. This will grow over time as we add features to it, like ending/withdrawing from auctions.
- Implement part of the `ViewService#auctions` RPC method (except `queryLatestState` and `includeInactive` — see below).
- Create an `AUCTIONS` IndexedDB table to store `DutchAuctionDescription`s and their corresponding NFTs' `noteCommitment`s, keyed by `AuctionId`.
- Update the block processor to save auction data to the database when it encounters that data.
- Extract a `<DutchAuctionComponent />` that's shared by `<ActionDutchAuctionScheduleViewComponent />` and `<Auctions />` (the latter renders the auctions list on the new auction page).

## Not in this PR
- We can't yet get the latest auction state from the fullnode (via `queryLatestState`) because of Issues™ — that'll come in a later PR.
- I'll add `includeInactive` support (and the corresponding `auctionState` seq number in the database) in a follow-up PR.

Relates to #980 and #1018 (but closes neither, since there's more work to be done)

### DIFF
--- a/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
@@ -1,0 +1,14 @@
+import { useStore } from '../../../state';
+
+export const Auctions = () => {
+  const auctions = useStore(state => state.dutchAuction.auctions);
+  return (
+    <div className='flex flex-col gap-2'>
+      {!auctions.length && <>No auctions</>}
+      {!!auctions.length &&
+        auctions.map(auction => (
+          <div key={auction.description?.nonce}>{auction.description?.startHeight.toString()}</div>
+        ))}
+    </div>
+  );
+};
--- a/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
@@ -25,8 +25,6 @@ const auctionsSelector = (state: AllSlices) => ({
 export const Auctions = () => {
   const { auctions, metadataByAssetId } = useStoreShallow(auctionsSelector);
 
-  if (!auctions.length) return null;
-
   return (
     <>
       <p className='mb-2 bg-text-linear bg-clip-text font-headline text-xl font-semibold leading-[30px] text-transparent md:text-2xl md:font-bold md:leading-9'>
@@ -34,6 +32,8 @@ export const Auctions = () => {
       </p>
 
       <div className='flex flex-col gap-2'>
+        {!auctions.length && "You don't currently have any auctions running."}
+
         {auctions.map(auction => (
           <ViewBox
             key={auction.description?.nonce.toString()}
--- a/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
@@ -1,6 +1,12 @@
 import { ViewBox } from '@penumbra-zone/ui/components/ui/tx/view/viewbox';
 import { useStore } from '../../../state';
 import { DutchAuctionComponent } from '@penumbra-zone/ui/components/ui/dutch-auction-component';
+import {
+  AssetId,
+  Metadata,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+
+const getMetadata = (penumbraAssetId?: AssetId) => new Metadata({ penumbraAssetId });
 
 export const Auctions = () => {
   const auctions = useStore(state => state.dutchAuction.auctions);
@@ -13,7 +19,13 @@ export const Auctions = () => {
           <ViewBox
             key={auction.description?.nonce.toString()}
             label='Dutch Auction'
-            visibleContent={<DutchAuctionComponent dutchAuction={auction} />}
+            visibleContent={
+              <DutchAuctionComponent
+                dutchAuction={auction}
+                inputMetadata={getMetadata(auction.description?.input?.assetId)}
+                outputMetadata={getMetadata(auction.description?.outputId)}
+              />
+            }
           />
         ))}
     </div>
--- a/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
@@ -1,15 +1,20 @@
+import { ViewBox } from '@penumbra-zone/ui/components/ui/tx/view/viewbox';
 import { useStore } from '../../../state';
+import { DutchAuctionComponent } from '@penumbra-zone/ui/components/ui/dutch-auction-component';
 
 export const Auctions = () => {
   const auctions = useStore(state => state.dutchAuction.auctions);
   return (
     <div className='flex flex-col gap-2'>
       {!auctions.length && <>No auctions</>}
+
       {!!auctions.length &&
         auctions.map(auction => (
-          <div key={auction.description?.nonce.toString()}>
-            {auction.description?.startHeight.toString()}
-          </div>
+          <ViewBox
+            key={auction.description?.nonce.toString()}
+            label='Dutch Auction'
+            visibleContent={<DutchAuctionComponent dutchAuction={auction} />}
+          />
         ))}
     </div>
   );
--- a/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/auctions.tsx
@@ -1,15 +1,30 @@
 import { ViewBox } from '@penumbra-zone/ui/components/ui/tx/view/viewbox';
-import { useStore } from '../../../state';
+import { AllSlices } from '../../../state';
 import { DutchAuctionComponent } from '@penumbra-zone/ui/components/ui/dutch-auction-component';
 import {
   AssetId,
   Metadata,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { useStoreShallow } from '../../../utils/use-store-shallow';
+import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
 
-const getMetadata = (penumbraAssetId?: AssetId) => new Metadata({ penumbraAssetId });
+const getMetadata = (metadataByAssetId: Record<string, Metadata>, assetId?: AssetId) => {
+  let metadata: Metadata | undefined;
+  if (assetId && (metadata = metadataByAssetId[bech32mAssetId(assetId)])) {
+    return metadata;
+  }
+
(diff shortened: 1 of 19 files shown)


packages/bech32m/src/format/convert.ts
```diff
@@ -16,7 +16,7 @@ export const fromBech32m = <P extends Prefix>(
   prefix: Prefix = b32mStr.slice(0, b32mStr.lastIndexOf('1')) as P,
 ): Uint8Array => {
   const p = b32mStr.slice(0, b32mStr.lastIndexOf('1'));
-  if (p !== prefix) throw new TypeError('Unexpected prefix');
+  if (p !== prefix) throw new TypeError(`Unexpected prefix: expected ${prefix}, got ${p}`);
```
jessepinho, line 19: Ran into these errors when using these helpers with test data, so I added some detail to the errors to make them easier to parse.

packages/storage/src/chrome/test-utils/tests-setup.js
```diff
@@ -23,3 +23,8 @@ globalThis.chrome = {
 
 globalThis.DEFAULT_GRPC_URL = 'https://rpc.example.com/';
 globalThis.MINIFRONT_URL = 'https://app.example.com';
+globalThis.process = {
+  env: {
+    NODE_ENV: 'test',
```
jessepinho, line 28: The new `IndexedDb#iterateAuctions()` method calls `base64ToUint8Array()`, which calls `validateSchema()`, which calls `isDevEnv()`, which calls `getEnv()`, which expects `process.env.NODE_ENV` to exist :)

packages/protobuf/src/registry.ts
```diff
@@ -52,6 +54,11 @@ export const typeRegistry: IMessageTypeRegistry = createRegistry(
   SctService,
   ShieldedPoolService,
   StakeService,
+
+  // Types not explicitly referenced by any above services should be added here.
+  // Otherwise, it will not be possible to serialize/deserialize these types if,
+  // e.g., they're used in an `Any` protobuf.
+  DutchAuction,
```
jessepinho, line 61: It seems that this registry walks through the entire tree of Protobuf types, starting with services, then RPC methods, then data types served by those methods.

But `ViewService#auctions` doesn't actually reference `DutchAuction` — it just references `Any`, since multiple auction types can be served by that endpoint. So I have to manually add `DutchAuction` to the registry or else the registry won't know about it.

packages/query/src/block-processor.ts
```diff
@@ -274,6 +278,10 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.saveTransactions(compactBlock.height, relevantTx);
       }
 
+      for (const spendableNoteRecord of flush?.newNotes ?? []) {
+        await this.maybeUpsertAuctionWithNoteCommitment(spendableNoteRecord);
+      }
+
```
jessepinho, line 284: I... really don't like this.

You can see on [lines 238-9 above](https://github.com/penumbra-zone/web/pull/1025/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR238-R239) that we're already iterating over `flush.newNotes`. So why don't I put this call to `this.maybeUpsertAuctionWithNoteCommitment()` inside that earlier `for` loop?

The problem is, we need to [call `this.processTransactions()`](https://github.com/penumbra-zone/web/pull/1025/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR272) _before_ calling `this.maybeUpsertAuctionWithNoteCommitment()`, because `this.processTransactions()` is what saves the auction NFT metadata to the database. `this.maybeUpsertAuctionWithNoteCommitment()` [depends on that auction NFT metadata being saved already](https://github.com/penumbra-zone/web/pull/1025/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR465) to be able to detect whether a given note is for an auction NFT; only then will it save the note's commitment to the `AUCTIONS` table.

"So why not just move `this.processTransactions()` to before the block where we handle `flush.newNotes`?" Because `this.processTransactions()` should only run _after_ we've handled `flush.newNotes`, since [we depend on the result of the flush](https://github.com/penumbra-zone/web/pull/1025/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR253) to determine whether there are transactions to process in the first place. It's a catch-22.

This isn't a problem in core because core isn't going back and forth between Rust and TypeScript like we are. If and when we move the block processor into Rust, I believe this issue will be resolved. In the meantime, I can't think of a better solution — but I'm also tired, so reviewers, feel free to make suggestions.

packages/query/src/block-processor.ts
```diff
@@ -215,8 +218,9 @@ export class BlockProcessor implements BlockProcessorInterface {
       };
 
       const recordsByCommitment = new Map<StateCommitment, SpendableNoteRecord | SwapRecord>();
+      let flush: ScanBlockResult | undefined;
       if (Object.values(flushReasons).some(Boolean)) {
-        const flush = this.viewServer.flushUpdates();
+        flush = this.viewServer.flushUpdates();
```
jessepinho, line 223: See my comment below on why I'm making this change

packages/ui/components/ui/dutch-auction-component.tsx
```diff
@@ -0,0 +1,64 @@
...
+    valueView: {
+      case: 'knownAssetId',
+      value: {
+        amount,
+        metadata,
+      },
+    },
+  });
+
+export const DutchAuctionComponent = ({
```
jessepinho, line 21: image

apps/minifront/src/components/swap/dutch-auction/auctions.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return new Metadata({ penumbraAssetId: assetId });
+};
+
+const auctionsSelector = (state: AllSlices) => ({
+  auctions: state.dutchAuction.auctions,
+  metadataByAssetId: state.dutchAuction.metadataByAssetId,
+});
+
+export const Auctions = () => {
```
jessepinho, line 25: Screenshot 2024-05-03 at 11 08 17 PM

packages/storage/src/indexed-db/index.ts
```diff
@@ -669,4 +674,55 @@ export class IndexedDb implements IndexedDbInterface {
...
+        auction,
+        noteCommitment,
+      },
+    });
+  }
+
+  async *iterateAuctions() {
+    yield* new ReadableStream({
+      start: async controller => {
+        let cursor = await this.db.transaction('AUCTIONS', 'readonly').store.openCursor();
```
grod220, line 707: Can we use for await syntax instead?

```typescript
for await (const cursor of this.db.transaction('AUCTIONS').store) {
          controller.enqueue({
            id: new AuctionId({ inner: base64ToUint8Array(cursor.key) }),
            value: {
              auction: cursor.value.auction
                ? DutchAuctionDescription.fromJson(cursor.value.auction)
                : undefined,
              noteCommitment: cursor.value.noteCommitment
                ? StateCommitment.fromJson(cursor.value.noteCommitment)
                : undefined,
            },
          });
        }
```

apps/minifront/src/components/swap/dutch-auction/auctions.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
+
+const getMetadata = (metadataByAssetId: Record<string, Metadata>, assetId?: AssetId) => {
+  let metadata: Metadata | undefined;
+  if (assetId && (metadata = metadataByAssetId[bech32mAssetId(assetId)])) {
+    return metadata;
+  }
+
+  return new Metadata({ penumbraAssetId: assetId });
+};
```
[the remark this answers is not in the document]
jessepinho: Not exactly. I'm falling back to a default `Metadata` object with `penumbraAssetId` populated, if we can't find the existing metadata in `metadataByAssetId`. At the moment, I think we just render "Unknown asset" for that, but it at least allows us to render a `knownAssetId` view.

grod220 (approved):
Went through together offline. Think there are some more AI's here, but let's save the data loading stuff for a follow up later.

packages/protobuf/src/registry.ts
jessepinho: Good call, done

packages/query/src/block-processor.ts
jessepinho: Yep, for sure. Done.

packages/storage/src/indexed-db/index.ts
```diff
@@ -669,4 +674,55 @@ export class IndexedDb implements IndexedDbInterface {
       txs.add({ table: 'SWAPS', value: n.toJson() as Jsonified<SwapRecord> });
     }
   }
+
+  // As more auction types are created, add them to T as a union type.
+  async upsertAuction<T extends DutchAuctionDescription>(
+    auctionId: AuctionId,
+    value: {
+      auction?: T;
+      noteCommitment?: StateCommitment;
```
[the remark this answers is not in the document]
jessepinho: Yeah, because we save to the `AUCTIONS` table twice: once with the auction, and once with the note commitment. It's a little awkward, but making methods specific to which value is being saved (e.g., `upsertAuctionWithNoteCommitment` felt messier.

packages/services/src/view-service/auctions.ts
```diff
@@ -0,0 +1,86 @@
...
+  const { includeInactive, queryLatestState, accountFilter } = req;
+
+  if (queryLatestState) {
+    throw new ConnectError('`queryLatestState` not yet implemented', Code.Unimplemented);
+  }
+  if (includeInactive) {
+    throw new ConnectError('`includeInactive` not yet implemented', Code.Unimplemented);
+  }
+
+  const auctionsThisUserControls = await getAuctionsThisUserControls(ctx, accountFilter);
```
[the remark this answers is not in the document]
jessepinho: Ahh, good call. To address this, I:
- Removed `indexedDb.iterateAuctions()`
- Created `indexedDb.getAuction()` (which takes an `auctionId` arg)
- Used a single `for await` loop in `ViewService#auctions` that just iterates over balances the user holds of auction NFTs, and then `yield`s `AuctionsResponse`s with the same data as before.

Should be faster! 🏎️

packages/storage/src/indexed-db/index.ts
jessepinho: (Moot point now that I've deleted this method)

PR #1027, opened 2024-05-03
fix client package

turbocrime:
- correct build
- correct paths
- change name of injection interface to prevent confusion
- add createPraxTransport export

some more client package work in another PR but this is enough for the demo


turbocrime:
working demo with this client package here https://github.com/turbocrime/nextjs-penumbra-client-example

packages/client/src/index.ts
```diff
@@ -5,7 +5,7 @@ export enum PenumbraRequestFailure {
 
 export const PenumbraSymbol = Symbol.for('penumbra');
 
-export interface PenumbraProvider {
+export interface PenumbraInjection {
```
grod220, line 8: What's the thinking behind the rename? Is this standard with other wallets? I feel like the `Provider` naming is more common in things I've encountered.

packages/client/package.json
```diff
@@ -5,23 +5,29 @@
...
   "type": "module",
   "scripts": {
-    "build": "tsc && vite build",
-    "clean": "rm -rfv dist",
-    "lint": "eslint \"**/*.ts*\""
+    "build": "tsc",
+    "clean": "rm -rfv dist package penumbra-zone-client-*.tgz",
+    "lint": "eslint \"**/*.ts*\"",
+    "prebuild": "$npm_execpath run clean",
+    "prepack": "$npm_execpath run build"
```
grod220, line 12: What's `$npm_execpath` do here?

grod220 (approved):

packages/client/package.json
turbocrime: `$npm_execpath` will refer to npm/pnpm/yarn/whatever. it seems like this is the only cross-package-manager way for one package script to call another. these scripts probably don't matter because a consumer won't call them outside of our monorepo, but in general when writing a package for publication it's better to not name an explicit npm/pnpm/yarn in scripts

packages/client/src/index.ts
turbocrime: `Provider` naming is common for react components that provide something to child components,  which felt awkward while writing the demo and using other `Provider`-named things, so i renamed this.

turbocrime:
i'm going to add a changeset patch bump to bech32m to calm down the peer dependency, and then merge this

PR #1034, opened 2024-05-04
Insert missing asset IDs for ActionDutchAuctionSchedule

jessepinho:
## Before
image

(Note that the input metadata was present because the TXP had been filled with that metadata when processing the "Spend" action. However, we don't want to rely on that.)

## After
image

## Punting to future PRs
- Add the same handling for `*Withdraw`/`*End` actions. Punting for now since that flow isn't yet supported on web.

### DIFF
--- a/packages/wasm/crate/src/tx.rs
+++ b/packages/wasm/crate/src/tx.rs
@@ -2,6 +2,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::convert::TryInto;
 
 use anyhow::anyhow;
+use penumbra_auction::auction::dutch::actions::view::ActionDutchAuctionScheduleView;
 use penumbra_dex::BatchSwapOutputData;
 use penumbra_keys::keys::SpendKey;
 use penumbra_keys::FullViewingKey;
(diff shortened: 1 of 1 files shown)


packages/wasm/crate/src/tx.rs
```diff
@@ -341,6 +342,13 @@ pub async fn transaction_info_inner(
                 address_views.insert(address.encode_to_vec(), address_view);
                 asset_ids.insert(note.asset_id());
             }
+            ActionView::ActionDutchAuctionSchedule(ActionDutchAuctionScheduleView {
+                action,
+                ..
+            }) => {
+                asset_ids.insert(action.description.output_id);
+                asset_ids.insert(action.description.input.asset_id);
+            }
```
jessepinho, line 351: This is a part of a match statement inside `transaction_info_inner`, in which we iterate over each action in the transaction, adding relevant pieces of information to the transaction perspective that will be needed to generate the transaction view. In this case, we need to add the asset IDs used for the auction to `asset_ids`, so that later on in this function, we can fetch those assets' metadata.

TalDerei (approved):

PR #1035, opened 2024-05-04
WIP: Support `queryLatestState` in `ViewService#auctions`

jessepinho:


TalDerei (changes requested):
I don't think we should be making an additional RPC request here. The point of `query_latest_state` was to make it easier for clients to get the latest state _without_ making additional RPC calls. If `include_inactive` in `AuctionsRequest` is set to true, then `AuctionsResponse` will include the latest auction state. cc @erwanor

erwanor:
@TalDerei That's right but afaict that's what this PR does? no? why not? it setup a view service implementation that offer an option for clients to request the latest auction state, which is what we want. This makes it easy to improve later on because every query service call is mediated by the user's view server from the get go

erwanor (approved):

erwanor:
This LGTM, @TalDerei re: `include_inactive`: that's not necessarily the case that we will return concrete auction state if that flag is set.

The view service implementation contract is like this:
- `include_inactive` -> include note records for auctions that are closed or withdrawn
- `query_latest_state` -> _somehow_ return concrete auction state / DEX positions (in practice, via rpc calls to the auction/dex query services)

It's completely valid to ask for `include_inactive` with `query_latest_state` set to false

TalDerei:
> If `include_inactive` in `AuctionsRequest`

oops typo here, meant If `query_latest_state ` in `AuctionsRequest` is set to true.

TalDerei (approved):
@erwanor thanks for the clarification, this makes sense! I re-reviewed the PR and the logic seems to be all there

Issue #1033, opened 2024-05-03
Bundle complete asset registry into Prax


turbocrime:
The compiled asset registry should be bundled with Prax at build time, to avoid making network requests at any later time.

turbocrime:
Assets unknown at build time would still need to be queried.

PR #1044, opened 2024-05-06
Update to latest registry [bundled version]

grod220:
Pairs with: https://github.com/prax-wallet/registry/pull/29
Closes: https://github.com/penumbra-zone/web/issues/1033

The new version bundles the registry inside. No need for async fetching. Also uses the new `version()` method to check if it's necessary to save (discussed in today's sync).


apps/extension/package.json
```diff
@@ -20,7 +20,7 @@
     "@bufbuild/protobuf": "^1.9.0",
     "@connectrpc/connect": "^1.4.0",
     "@connectrpc/connect-web": "^1.4.0",
-    "@penumbra-labs/registry": "^5.1.0",
+    "@penumbra-labs/registry": "npm:@penumbra-zone/test-publish@0.7.0",
```
grod220, line 23: To be swapped with `7.0.0` when published

turbocrime: can we simply use `latest`

grod220: It's not a bad idea, but given we are on v7, we've had 7 breaking changes of the registry api to date. Think this is worth revisiting when that stabilizes a bit. Otherwise, we'd risk dev env breaking on a registry update.

packages/storage/src/indexed-db/index.ts
```diff
@@ -254,9 +256,17 @@ export class IndexedDb implements IndexedDbInterface {
   }
 
   // creates a local copy of the asset list from registry (https://github.com/prax-wallet/registry)
-  async saveRegistryAssets(assets: Metadata[]) {
+  async saveRegistryAssets(registryClient: ChainRegistryClient, chainId: string) {
+    const { commit } = registryClient.version();
+    const lastPosition = await this.db.get('REGISTRY_VERSION', 'commit');
+
+    // Registry version already saved
+    if (lastPosition === commit) return;
```
Valentine1898, line 264: An older registry commit can overwrite a newer one. But it seems this can never occur

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -53,16 +53,19 @@ import {
   DutchAuctionDescription,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1alpha1/auction_pb';
 import { StateCommitment } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/crypto/tct/v1/tct_pb';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
 
 describe('IndexedDb', () => {
   // uses different wallet ids so no collisions take place
+  const registryClient = new ChainRegistryClient();
+  const chainId = 'penumbra-testnet-deimos-6';
```
Valentine1898, line 61: Optional
You could use mock for `ChainRegistryClient` instead, then the indexed-db tests would not depend on registry. And the tests won't break if the registry no longer contains data for the test `chain-id`.

Valentine1898 (approved):
LGTM

packages/storage/src/indexed-db/index.ts
grod220: Yes this is true, but thinking this is not a likely scenario for us

packages/storage/src/indexed-db/indexed-db.test.ts
grod220: I started down this track, but realized that we'd need to implement an interface, but the registry wasn't currently exposing one. Given time constraints, just used a chain id we know is there. Can revisit this later though.

PR #1047, opened 2024-05-07
Handle `includeInactive` in `ViewService#auctions`

jessepinho:
`AuctionsRequest`s have an `includeInactive` property that was not yet supported by our `auctions` implementation. This PR adds support for that property by A) keeping track of auctions' sequence numbers to determine whether they're active, and B) filtering out inactive auctions unless `includeInactive` is `true`.

## In this PR
- Add a `seqNum` column to the `AUCTIONS` table (to mimic the `state` column in [core](https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/storage/schema.sql#L136-L140)'s implementation)
- Upsert the `seqNum` in the block processor when we encounter a Dutch auction-related action.
- Add support for `includeInactive` in the `ViewService#auctions` RPC method.
- Bump the IndexedDB version number.

Relates to #980 (but doesn't close it yet, since `queryLatestState` is not yet supported)

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -419,24 +419,29 @@ export class BlockProcessor implements BlockProcessorInterface {
   private async identifyAuctionNfts(action: Action['action']) {
     if (action.case === 'actionDutchAuctionSchedule' && action.value.description) {
       const auctionId = getAuctionId(action.value.description);
-      const metadata = getAuctionNftMetadata(
-        auctionId,
-        // Always a sequence number of 0 when starting a Dutch auction
-        0n,
-      );
+
+      // Always a sequence number of 0 when starting a Dutch auction
+      const seqNum = 0n;
+
+      const metadata = getAuctionNftMetadata(auctionId, seqNum);
+
       await Promise.all([
         this.indexedDb.saveAssetsMetadata(metadata),
         this.indexedDb.upsertAuction(auctionId, {
           auction: action.value.description,
+          seqNum,
         }),
       ]);
     } else if (action.case === 'actionDutchAuctionEnd' && action.value.auctionId) {
-      const metadata = getAuctionNftMetadata(
-        action.value.auctionId,
-        // Always a sequence number of 1 when ending a Dutch auction
-        1n,
-      );
-      await this.indexedDb.saveAssetsMetadata(metadata);
+      // Always a sequence number of 1 when ending a Dutch auction
+      const seqNum = 1n;
+
+      const metadata = getAuctionNftMetadata(action.value.auctionId, seqNum);
+
+      await Promise.all([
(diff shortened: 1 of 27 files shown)


packages/query/src/block-processor.ts
```diff
@@ -419,29 +419,42 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      const metadata = getAuctionNftMetadata(
-        action.value.auctionId,
-        // Always a sequence number of 1 when ending a Dutch auction
-        1n,
-      );
-      await this.indexedDb.saveAssetsMetadata(metadata);
+      // Always a sequence number of 1 when ending a Dutch auction
+      const seqNum = 1n;
+
+      const metadata = getAuctionNftMetadata(action.value.auctionId, seqNum);
```
TalDerei, line 439: won't this request metadata for closed auctions since it's passing in `seqNum` of 1? Shouldn't it pass `seqNum` of 0 for active auctions, and only after save the metadata and update the auction sequence number? Or does it not matter here?

packages/services/src/view-service/auctions.ts
```diff
@@ -33,6 +33,8 @@ const getBech32mAuctionId = (
   return captureGroups.auctionId;
 };
 
+const isInactive = (seqNum?: bigint) => (seqNum === undefined ? false : seqNum > 0n);
```
TalDerei, line 36: why would the sequence number ever be undefined?

jessepinho: I made it optional since `value` (used on line 61 below) may have an undefined `seqNum` property if we haven't saved it yet for some reason.

TalDerei (approved):

packages/query/src/block-processor.ts
jessepinho: @TalDerei 

> won't this request metadata for closed auctions since it's passing in seqNum of 1?

Correct — although it's not exactly _requesting_ the metadata, but rather _generating_ it. (`getAuctioNftMetadata()` calls a WASM function under the hood that generates metadata for an asset that doesn't exist yet.)

> Shouldn't it pass seqNum of 0 for active auctions, and only after save the metadata and update the auction sequence number?

Hm — we could request the NFT metadata for the open auction, and the modify its sequence number. But that feels a bit messy to me — if the algorithm for generating NFT metadata changes in the future such that closed-auction NFTs have different metadata than open-auction NFTs, this would break. Easier to just generate it from scratch for each auction state.

TalDerei: yeah was thinking about that, I think generating the metadata is the right design choice here.

TalDerei: > requesting

yeah wrong word choice here 😅

Issue #1043, opened 2024-05-06
minifront: cannot claim unbonding tokens


erwanor:
A user cannot claim their undelegation tokens using minifront.

STR:

1. Checkout main
2. Delegate to a validator
3. Undelegate from validator
4. Wait for unbonding delay to elapse
5. Try to click the "Claim" button 

Approximate cause:

This is because `getUndelegateClaimPlannerRequest` cannot recover the `IdentityKey` from the token's value view. The reason for that is that the `ValueView` is missing `extended_metadata` which only get populated for bonding tokens.

PR #1058, opened 2024-05-07
Skip unknownAssetId when checking isUnbondingTokenBalance()

Valentine1898:
While working on https://github.com/penumbra-zone/web/issues/1043 I discovered a issue 

If a user has unknown tokens in their balance, they will receive a log error and invalid Unbonding amount and Claimable amount values

in addition to https://github.com/penumbra-zone/web/pull/1010

image

Before:
image
After: 
image

### DIFF
--- a/packages/services/src/view-service/unbonding-tokens-by-address-index/helpers.ts
+++ b/packages/services/src/view-service/unbonding-tokens-by-address-index/helpers.ts
@@ -11,11 +11,8 @@ import { status } from '../status';
 import { appParameters } from '../app-parameters';
 
 export const isUnbondingTokenBalance = (balancesResponse: PartialMessage<BalancesResponse>) => {
-  if (balancesResponse.balanceView?.valueView?.case === 'unknownAssetId') return false;
-
-  return assetPatterns.unbondingToken.matches(
-    getDisplayFromBalancesResponse(new BalancesResponse(balancesResponse)),
-  );
+  const display = getDisplayFromBalancesResponse(new BalancesResponse(balancesResponse));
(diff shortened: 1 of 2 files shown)


packages/services/src/view-service/unbonding-tokens-by-address-index/helpers.ts
```diff
@@ -10,10 +10,13 @@ import { getDisplayFromBalancesResponse } from '@penumbra-zone/getters/balances-
 import { status } from '../status';
 import { appParameters } from '../app-parameters';
 
-export const isUnbondingTokenBalance = (balancesResponse: PartialMessage<BalancesResponse>) =>
-  assetPatterns.unbondingToken.matches(
+export const isUnbondingTokenBalance = (balancesResponse: PartialMessage<BalancesResponse>) => {
+  if (balancesResponse.balanceView?.valueView?.case === 'unknownAssetId') return false;
```
grod220, line 14: Perhaps we can make `getDisplayFromBalancesResponse` have an `optional()` on the getter instead?

Valentine1898: `getDisplayFromBalancesResponse` is not used anywhere else, so yes, it makes sense. Done

TalDerei (approved):
LGTM

packages/services/src/view-service/unbonding-tokens-by-address-index/helpers.ts
grod220: LGTM 👍

PR #1060, opened 2024-05-07
UI for ending auctions

### LINKED ISSUE #1019: Create UI for ending an existing auction

jessepinho:
Now that auctions can be scheduled and viewed, it's time to allow users to end them!

https://github.com/penumbra-zone/web/assets/1121544/de0dcec0-0b31-461f-ad29-994e3d7ae9f2

## In this PR
- Modify `<DutchAuctionComponent />` to render an "End auction" button.
- Create `<ActionDutchAuctionEndComponent />` to render the action of ending an auction.
- Change metadata symbols for auction NFTs to include the sequence numbers (before: `auction(abcd1234...)`; after: `auction@1(abcd1234...)`
  - While I was at it, I changed all the customized symbols to use ellipses, since those display more nicely.
    Before:
    image

    After:
    image
- Change the shape of the `dutchAuction` state slice; now, there's an `auctionInfos` array instead of an `auctions` array. Each `AuctionInfo` contains the auction ID and the `DutchAuction` (because the `auctions` array was just of type `DutchAuction[]`, which doesn't contain the auction ID, which we need for ending an auction).
- Rename `assembleRequest()` to `assembleScheduleRequest()`, to differentiate from the other types of requests that get assembled in the Dutch auction slice.
- Create new transaction classifications for Dutch auction withdraws and ends.

Closes #1019

### DIFF
--- a/apps/extension/.env.testnet
+++ b/apps/extension/.env.testnet
@@ -1,4 +1,4 @@
 CHAIN_ID=penumbra-testnet-deimos-7
-IDB_VERSION=38
+IDB_VERSION=39
 MINIFRONT_URL=https://app.testnet.penumbra.zone
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -8,7 +8,7 @@ import {
   Metadata,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { planBuildBroadcast } from '../helpers';
-import { assembleRequest } from './assemble-request';
+import { assembleScheduleRequest } from './assemble-schedule-request';
 import { DurationOption } from './constants';
 import {
   AuctionId,
@@ -101,7 +101,7 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
     });
 
     try {
-      const req = await assembleRequest(get().dutchAuction);
+      const req = await assembleScheduleRequest(get().dutchAuction);
       await planBuildBroadcast('dutchAuctionSchedule', req);
 
       get().dutchAuction.setAmount('');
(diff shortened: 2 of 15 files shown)


apps/extension/.env.testnet
```diff
@@ -1,4 +1,4 @@
 CHAIN_ID=penumbra-testnet-deimos-7
-IDB_VERSION=38
+IDB_VERSION=39
```
jessepinho, line 2: Required because I made changes to auction NFT symbols

apps/minifront/src/state/dutch-auction/index.ts
```diff
@@ -93,6 +105,7 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
       await planBuildBroadcast('dutchAuctionSchedule', req);
 
       get().dutchAuction.setAmount('');
+      void get().dutchAuction.loadAuctionInfos();
```
jessepinho, line 108: I realized that we should be reloading auctions after successfully scheduling an auction, so that the latest data shows up in the auctions list.

apps/minifront/src/state/dutch-auction/index.ts
```diff
@@ -135,4 +148,20 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
...
 
   metadataByAssetId: {},
+
+  endAuction: async auctionId => {
+    set(state => {
+      state.dutchAuction.txInProgress = true;
+    });
+
+    try {
+      const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }] });
```
jessepinho, line 158: No `assemble*Request()` helper needed here, because creating the request is so simple. 

However, I still renamed the other helper to `assembleScheduleRequest()` in this PR to make it clear which request type is being assembled.

packages/ui/components/ui/auction-id-component.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Renders an auction's `AuctionId` as a bech32-encoded string, along with a
+ * copy button.
+ *
+ * @example
+ * ```tsx
+ * <AuctionIdComponent auctionId={auctionId} />
+ * ```
+ */
+export const AuctionIdComponent = ({ auctionId }: { auctionId?: AuctionId }) => {
```
jessepinho, line 17: image

apps/minifront/src/state/dutch-auction/index.ts
grod220: Could this potentially give them a success, but still show the same auctions screen? Aka, should we await this instead?

grod220 (approved):
LGTM

PR #1061, opened 2024-05-07
externally connectable

turbocrime:
wip status: 

set env `PRAX_EXTERNAL` to something json-truthy to activate externally connectable in minifront and prax

todo: penumbra global is stubbed out. 'request' happens automatically on connection.

it might make sense to serve mainworld global injection script as a web-accessible resource.

potential route for content-script-less announcement is a redirect that applies headers.


grod220:
Ah, I didn't anticipate externally connectable would have so many changes. Thought we were going for a small side usage in attempt to pass the chrome review. If we are doing a larger change, can you provide details on:
- what this is accomplishing and how it's beneficial to migrate to
- how it's different than our current messaging
- how users or web apps will experience a difference (if at all)

PR #1066, opened 2024-05-07
1043 minifront cannot claim unbonding tokens

### LINKED ISSUE #1043: minifront: cannot claim unbonding tokens

Valentine1898:
> This is because getUndelegateClaimPlannerRequest cannot recover the IdentityKey from the token's value view. The reason for that is that the ValueView is missing extended_metadata which only get populated for bonding tokens.

Now extended_metadata fills in for unbonding tokens as well

close #1043

### DIFF
--- a/packages/services/src/staking-service/get-validator-info.ts
+++ b/packages/services/src/staking-service/get-validator-info.ts
@@ -1,0 +1,18 @@
+import { Impl } from './index';
+import { servicesCtx } from '../ctx/prax';
+import { Code, ConnectError } from '@connectrpc/connect';
+
+export const getValidatorInfo: Impl['getValidatorInfo'] = async (req, ctx) => {
+  if (!req.identityKey) {
+    throw new ConnectError('Missing identityKey in request', Code.InvalidArgument);
+  }
+  const services = await ctx.values.get(servicesCtx)();
+  const { indexedDb } = await services.getWalletServices();
+
+  let validatorInfo = await indexedDb.getValidatorInfo(req.identityKey);
+
+  if (!validatorInfo) {
+    throw new ConnectError('No found validator info', Code.NotFound);
+  }
+  return { validatorInfo };
+};
--- a/packages/services/src/staking-service/get-validator-info.ts
+++ b/packages/services/src/staking-service/get-validator-info.ts
@@ -1,4 +1,4 @@
-import { Impl } from './index';
+import { Impl } from '.';
 import { servicesCtx } from '../ctx/prax';
 import { Code, ConnectError } from '@connectrpc/connect';
 
@@ -9,7 +9,7 @@ export const getValidatorInfo: Impl['getValidatorInfo'] = async (req, ctx) => {
   const services = await ctx.values.get(servicesCtx)();
   const { indexedDb } = await services.getWalletServices();
 
-  let validatorInfo = await indexedDb.getValidatorInfo(req.identityKey);
+  const validatorInfo = await indexedDb.getValidatorInfo(req.identityKey);
 
   if (!validatorInfo) {
     throw new ConnectError('No found validator info', Code.NotFound);
--- a/packages/services/src/staking-service/index.ts
+++ b/packages/services/src/staking-service/index.ts
@@ -7,7 +7,10 @@ import { getValidatorInfo } from './get-validator-info';
 
 export type Impl = ServiceImpl<typeof StakingService>;
 
-export const stakingImpl: Pick<Impl, 'getValidatorInfo' | 'validatorInfo' | 'validatorPenalty'> = {
+export const stakingImpl: Omit<
+  Impl,
+  'currentValidatorRate' | 'validatorStatus' | 'validatorUptime'
+> = {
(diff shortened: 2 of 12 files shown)


packages/services/src/staking-service/get-validator-info.ts
```diff
@@ -0,0 +1,18 @@
+import { Impl } from '.';
+import { servicesCtx } from '../ctx/prax';
+import { Code, ConnectError } from '@connectrpc/connect';
+
+export const getValidatorInfo: Impl['getValidatorInfo'] = async (req, ctx) => {
```
Valentine1898, line 5: I added the `getValidatorInfo` implementation because it would not be optimal to iterate through the list of all validators for each unbonding balance

packages/services/src/view-service/unbonding-tokens-by-address-index/index.test.ts
```diff
@@ -71,13 +83,25 @@ const mockBalancesResponse3 = new BalancesResponse({
...
+const mockBalancesResponse4 = new BalancesResponse({
+  balanceView: {
+    valueView: {
+      case: 'unknownAssetId',
+      value: {
+        amount: { hi: 0n, lo: 3n },
+      },
+    },
+  },
+});
```
Valentine1898, line 103: added to make sure that having balances with unknownAssetId does not break anything

packages/services/src/view-service/unbonding-tokens-by-address-index/index.ts
```diff
@@ -31,9 +44,29 @@ export const unbondingTokensByAddressIndex: Impl['unbondingTokensByAddressIndex'
...
+      const extendedMetadata = new Any({
+        typeUrl: ValidatorInfo.typeName,
+        value: validatorInfo.toBinary(),
+      });
+
+      const withValidatorInfo = getBalanceView(new BalancesResponse(balancesResponse)).clone();
+      if (withValidatorInfo.valueView.case !== 'knownAssetId')
+        throw new Error(`Unexpected ValueView case: ${withValidatorInfo.valueView.case}`);
+
+      withValidatorInfo.valueView.value.extendedMetadata = extendedMetadata;
```
Valentine1898, line 65: Same way as for delegation tokens 
https://github.com/penumbra-zone/web/blob/b97e2102767369a835e4fc6b1df49ce0ab30f9f3/packages/services/src/view-service/delegations-by-address-index.ts#L77-L84

packages/services/src/staking-service/get-validator-info.ts
TalDerei: This is great 👍

packages/services/src/staking-service/index.ts
```diff
@@ -3,10 +3,12 @@ import { ServiceImpl } from '@connectrpc/connect';
 
 import { validatorInfo } from './validator-info';
 import { validatorPenalty } from './validator-penalty';
+import { getValidatorInfo } from './get-validator-info';
 
 export type Impl = ServiceImpl<typeof StakingService>;
 
-export const stakingImpl: Pick<Impl, 'validatorInfo' | 'validatorPenalty'> = {
+export const stakingImpl: Pick<Impl, 'getValidatorInfo' | 'validatorInfo' | 'validatorPenalty'> = {
+  getValidatorInfo,
```
grod220, line 11: Would kinda be nice if this was an `Omit<Impl, ...>`, so that it was kinda in our face what we have not implemented yet

packages/services/src/view-service/unbonding-tokens-by-address-index/index.ts
```diff
@@ -31,9 +44,29 @@ export const unbondingTokensByAddressIndex: Impl['unbondingTokensByAddressIndex'
...
+      const validatorInfoResponse = await stakingClient.getValidatorInfo({
+        identityKey: identityKeyFromBech32m(regexResult.idKey),
+      });
+      const validatorInfo = getValidatorInfo(validatorInfoResponse);
+      const extendedMetadata = new Any({
+        typeUrl: ValidatorInfo.typeName,
+        value: validatorInfo.toBinary(),
+      });
+
+      const withValidatorInfo = getBalanceView(new BalancesResponse(balancesResponse)).clone();
```
grod220, line 61: Why do we need to clone()?

grod220 (approved):
LGTM

packages/services/src/view-service/unbonding-tokens-by-address-index/index.ts
Valentine1898: Indeed, it's unnecessary

Issue #1040, opened 2024-05-06
view: create a markdown file of the action views


TalDerei:
Produce a markdown specification detailing the essential information to be understood during the signing of a penumbra transaction. This requires enumerating the protobuf fields, categorizing them into visible and opaque groups, and providing a brief explanation on why the omitted fields are safely and desirably excluded and why the visible fields should be public. This is important because the UI rendering and approval policies are linked.

TalDerei:
noting that these only include actions currently **_used_** by prax. The other actions can be audited separately

PR #1068, opened 2024-05-08
Let users withdraw funds from auctions

### LINKED ISSUE #1020: Create UI/planner support for withdrawing from an existing auction
Related: [this comment](https://github.com/penumbra-zone/web/pull/1015#discussion_r1587690202)

jessepinho:
https://github.com/penumbra-zone/web/assets/1121544/d0c38812-51b1-44ea-8abe-48f8eccc4a41

## In this PR
- Refactor `<DutchAuctionComponent />` to take `buttonType` and `onClickButton` props, rather than `endButton` and `onClickEndButton` props (since now the button may be a withdraw or end button).
- Create a new `outstandingReserves` property on the `AUCTIONS` table for storing outstanding reserves for an auction that has ended but not been withdrawn from.
- Handle withdraw actions in the planner.
- Extract a `byte_array_to_base64` helper and use it throughout `storage.rs`.
- Extract a `processActionDutchAuctionEnd()` helper from the block processor.
  - Processing the action that ends Dutch auctions was getting complex and was entirely untested (like everything else in the block processor 😬 😬 ), so I extracted it to a helper and wrote tests.

Closes #1020

### DIFF
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -13,7 +13,7 @@ import { DurationOption } from './constants';
 import {
   AuctionId,
   DutchAuction,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1alpha1/auction_pb';
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1/auction_pb';
 import { viewClient } from '../../clients';
 import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
 
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -44,6 +44,7 @@ export interface DutchAuctionSlice {
   loadMetadata: (assetId?: AssetId) => Promise<void>;
   metadataByAssetId: Record<string, Metadata>;
   endAuction: (auctionId: AuctionId) => Promise<void>;
+  withdraw: (auctionId: AuctionId, currentSeqNum: bigint) => Promise<void>;
 }
 
 export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (set, get) => ({
@@ -157,6 +158,24 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
     try {
       const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }] });
       await planBuildBroadcast('dutchAuctionEnd', req);
+      void get().dutchAuction.loadAuctionInfos();
+    } finally {
+      set(state => {
+        state.dutchAuction.txInProgress = false;
+      });
+    }
+  },
+
+  withdraw: async (auctionId, currentSeqNum) => {
+    set(state => {
+      state.dutchAuction.txInProgress = true;
+    });
+
+    try {
+      const req = new TransactionPlannerRequest({
+        dutchAuctionWithdrawActions: [{ auctionId, seq: currentSeqNum + 1n }],
+      });
+      await planBuildBroadcast('dutchAuctionWithdraw', req);
       void get().dutchAuction.loadAuctionInfos();
     } finally {
       set(state => {
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -121,7 +121,7 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
       state.dutchAuction.auctionInfos = [];
     });
 
-    for await (const response of viewClient.auctions({})) {
+    for await (const response of viewClient.auctions({ includeInactive: true })) {
       if (!response.auction || !response.id) continue;
       const auction = DutchAuction.fromBinary(response.auction.value);
       const auctions = [...get().dutchAuction.auctionInfos, { id: response.id, auction }];
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -122,6 +122,7 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
       state.dutchAuction.auctionInfos = [];
     });
 
+    /** @todo: Sort by... something? */
     for await (const response of viewClient.auctions({ includeInactive: true })) {
       if (!response.auction || !response.id) continue;
       const auction = DutchAuction.fromBinary(response.auction.value);
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -152,36 +152,16 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
   metadataByAssetId: {},
 
   endAuction: async auctionId => {
-    set(state => {
-      state.dutchAuction.txInProgress = true;
-    });
-
-    try {
-      const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }] });
-      await planBuildBroadcast('dutchAuctionEnd', req);
-      void get().dutchAuction.loadAuctionInfos();
-    } finally {
-      set(state => {
-        state.dutchAuction.txInProgress = false;
-      });
-    }
+    const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }] });
+    await planBuildBroadcast('dutchAuctionEnd', req);
+    void get().dutchAuction.loadAuctionInfos();
   },
 
   withdraw: async (auctionId, currentSeqNum) => {
-    set(state => {
-      state.dutchAuction.txInProgress = true;
+    const req = new TransactionPlannerRequest({
+      dutchAuctionWithdrawActions: [{ auctionId, seq: currentSeqNum + 1n }],
     });
-
-    try {
-      const req = new TransactionPlannerRequest({
-        dutchAuctionWithdrawActions: [{ auctionId, seq: currentSeqNum + 1n }],
-      });
-      await planBuildBroadcast('dutchAuctionWithdraw', req);
-      void get().dutchAuction.loadAuctionInfos();
-    } finally {
-      set(state => {
-        state.dutchAuction.txInProgress = false;
-      });
-    }
+    await planBuildBroadcast('dutchAuctionWithdraw', req);
+    void get().dutchAuction.loadAuctionInfos();
   },
 });
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -44,6 +44,7 @@ import { getAuctionId, getAuctionNftMetadata } from '@penumbra-zone/wasm/auction
 import { AuctionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1alpha1/auction_pb';
 import { auctionIdFromBech32 } from '@penumbra-zone/bech32m/pauctid';
 import { ScanBlockResult } from '@penumbra-zone/types/state-commitment-tree';
+import { processActionDutchAuctionEnd } from './helpers/process-action-dutch-auction-end';
 
 declare global {
   // `var` required for global declaration (as let/const are block scoped)
@@ -433,15 +434,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         }),
       ]);
     } else if (action.case === 'actionDutchAuctionEnd' && action.value.auctionId) {
-      // Always a sequence number of 1 when ending a Dutch auction
-      const seqNum = 1n;
-
-      const metadata = getAuctionNftMetadata(action.value.auctionId, seqNum);
-
-      await Promise.all([
-        this.indexedDb.saveAssetsMetadata(metadata),
-        this.indexedDb.upsertAuction(action.value.auctionId, { seqNum }),
-      ]);
+      await processActionDutchAuctionEnd(action.value, this.querier.auction, this.indexedDb);
     } else if (action.case === 'actionDutchAuctionWithdraw') {
       const auctionId = action.value.auctionId;
       if (!auctionId) return;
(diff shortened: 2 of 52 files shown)


apps/minifront/src/state/dutch-auction/index.ts
```diff
@@ -150,18 +152,16 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
...
 
-    try {
-      const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }] });
-      await planBuildBroadcast('dutchAuctionEnd', req);
-      void get().dutchAuction.loadAuctionInfos();
-    } finally {
-      set(state => {
-        state.dutchAuction.txInProgress = false;
-      });
-    }
```
jessepinho, line 165: Realized we don't need to set `txInProgress`, since that's really just for controlling the submit button of the auction form. No need to disable it while an end/withdraw transaction is in progress.

packages/services/src/view-service/auctions.ts
```diff
@@ -70,8 +70,8 @@ export const auctions: Impl['auctions'] = async function* (req, ctx) {
       auction = new Any({
         typeUrl: DutchAuction.typeName,
         value: new DutchAuction({
+          state: state ?? { seq: value.seqNum },
```
jessepinho, line 73: If we aren't querying latest state, we still want to return a state partial with the sequence number so that the UI can determine whether to render end/withdraw buttons.

packages/ui/components/ui/tx/view/action-dutch-auction-withdraw-view.tsx
```diff
@@ -0,0 +1,33 @@
+import { ActionDutchAuctionWithdrawView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1alpha1/auction_pb';
+import { ViewBox } from './viewbox';
+import { ActionDetails } from './action-details';
+import { AuctionIdComponent } from '../../auction-id-component';
+import { ValueViewComponent } from './value';
+import { getDisplayDenomFromView } from '@penumbra-zone/getters/value-view';
+
+export const ActionDutchAuctionWithdrawViewComponent = ({
```
jessepinho, line 8: image

packages/wasm/crate/src/storage.rs
```diff
@@ -168,10 +171,7 @@ impl IndexedDBStorage {
         let store = tx.object_store(&self.constants.tables.assets)?;
 
         Ok(store
-            .get_owned(base64::Engine::encode(
-                &base64::engine::general_purpose::STANDARD,
-                id.to_proto().inner,
-            ))?
+            .get_owned(byte_array_to_base64(&id.to_proto().inner))?
```
jessepinho, line 174: A lot of this file's line changes are just refactoring to use this `byte_array_to_base64` helper. The meat of the changes is down below.

packages/wasm/crate/src/storage.rs
```diff
@@ -379,4 +358,39 @@ impl IndexedDBStorage {
...
+    ) -> WasmResult<OutstandingReserves> {
+        self.db
+            .transaction_on_one(&self.constants.tables.auctions)?
+            .object_store(&self.constants.tables.auctions)?
+            .get::<JsValue>(&byte_array_to_base64(&auction_id.to_proto().inner).into())?
+            .await?
+            .map(|auction| {
+                serde_wasm_bindgen::from_value::<AuctionRecord>(auction)?
+                    .outstanding_reserves
+                    .ok_or(WasmError::Anyhow(Error::msg("could not find reserves")))
```
jessepinho, line 374: Is there a clean way to not have to repeat this error line below on line 376?

packages/wasm/crate/src/tx.rs
```diff
@@ -349,6 +351,12 @@ pub async fn transaction_info_inner(
                 asset_ids.insert(action.description.output_id);
                 asset_ids.insert(action.description.input.asset_id);
             }
+            ActionView::ActionDutchAuctionWithdraw(ActionDutchAuctionWithdrawView {
+                reserves,
+                ..
+            }) => reserves.iter().for_each(|reserve| {
+                asset_ids.insert(reserve.asset_id());
+            }),
```
jessepinho, line 359: This doesn't do anything yet, since core currently creates a stubbed withdraw view with [an empty reserves array](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/is_action.rs#L500).

This is a [known issue](https://github.com/penumbra-zone/penumbra/issues/4219#issuecomment-2088801880) and will eventually be fixed. In the meantime, though, I'm writing the code here that will automatically populate the necessary metadata once core is updated.

packages/query/src/block-processor.ts
```diff
@@ -452,6 +445,7 @@ export class BlockProcessor implements BlockProcessorInterface {
         this.indexedDb.saveAssetsMetadata(metadata),
         this.indexedDb.upsertAuction(auctionId, {
           seqNum: action.value.seq,
+          outstandingReserves: undefined,
```
turbocrime, line 448: this is in a couple places. i understand what it's doing here in idb,

https://github.com/penumbra-zone/web/blob/670034adff328654f49867f68ea28e6450907408/packages/storage/src/indexed-db/index.ts#L714-L721

but i think it would be more appropriate to use some other mechanism, or use a value like `null` rather than exploiting the subtle distinction between an undefined value and a value that is undefined.

jessepinho: Hmmm... I tend to agree. But as I started to implement your suggestion of using `null`, I realized I'd either have to:

1. Make `outstandingReserves` required and type it as `outstandingReserves?: { input: ..., output: ... } | null` — meaning it can be undefined OR null, depending on whether `outstandingReserves` has previously been set and then cleared, OR
2. Pass `outstandingReserves: null` everywhere that `upsertAuction` is called, which sort of defeats the purpose of an upsert. (I'd designed the method so that you can only pass the properties you want to update.)

Alternatively, instead of using `null` or `undefined`, I could create a separate `OUTSTANDING_AUCTION_RESERVES` table that we add to / delete from at the appropriate times. I don't _love_ it, since outstanding reserves are essentially a child of auction records and thus should probably be in the auctions table. But... 🤷🏻‍♂️ 

Thoughts? Suggestions?

packages/wasm/crate/src/storage.rs
turbocrime: these are technically different errors - i would put 'could not find auction' here, and 'could not find reserves' is appropriate at the other.

consider also rewriting it in the pattern of other fns in this file, which seem a bit clearer.

packages/query/src/block-processor.ts
turbocrime: >Make outstandingReserves required and type it as outstandingReserves?: { input: ..., output: ... } | null — meaning it can be undefined OR null, depending on whether outstandingReserves has previously been set and then cleared

In storage, it's okay to use a simple optional on the stored type, as you'll never need to distinguish 'was never defined' from 'was previously defined but then cleared'

but your upsert parameter definitely needs three possible values, as you're taking three possible actions:

- preserve existing data
- write new data
- delete data

and these deserve clarity.

null and undefined are distinct values - loosely equal but not strictly equal, so something like
`outstandingReserves === null` can be a useful conditional.

notably some of this is converted to json - strictly, json has no `undefined` type, but does have `null`. double-check how bufbuild serializes each

part of the issue also is you're essentially doing a deep object merge.

some options:

- break out outstandingReserves into multiple values to avoid deep comparison. the null/undefined problem remains, but is simpler
- use null value to clear outstandingReserves, and undefined to preserve it
- use null value to clear fields inside outstandingReserves, and undefined to preserve them
- use an optional object for outstandingReserves, optional fields inside outstandingReserves, and a present but empty object or nullish values to clear outstandingReserves

turbocrime: i think keeping a table of active auctions is reasonable, as the table of all auctions may eventually be large.

grod220: I think we should be careful to not assign meaning to `null` vs `undefined` from an API perspective. This has been a long debated thing in javascript land and there isn't a plain-meaning consensus on it. I'm personally in the camp that including `null` in the language was a mistake (yet another win for Rust).

Think explicit indicators for update versus deletion would be nice. Perhaps something like:

```typescript
type UpdateField<T> = { type: 'UPDATE'; value: T } | { type: 'DELETE' };

outstandingReserves?: UpdateField<{ input: Value; output: Value }>;
```

or perhaps a new unset field:

```typescript
  upsertAuction<T extends DutchAuctionDescription>(
    auctionId: AuctionId,
    value: UpdateFields<T>,
    unset?: (keyof UpdateFields<T>)[],
  ): Promise<void>;
```

or even an explicit method:

```typescript
  deleteAuctionFields<T extends DutchAuctionDescription>(
    auctionId: AuctionId,
    fields: (keyof UpdateFields<T>)[],
  ): Promise<void>;
```

turbocrime: the amount of discussion generated here makes me think we should not provide any merging upsert, but just expect the caller to construct the object they intend to store, and then clobber whatever's there

packages/wasm/crate/src/tx.rs
TalDerei: flagging that I'll also be inspecting these views closely per https://github.com/penumbra-zone/web/issues/1040, possibly changing what we display in our TxP / TxV

erwanor: Yeah it's hardcoded for now, my sense is that we can get to it a bit later, not next week but the one after, let me know if you think this should be prioritized and i'll get to it sooner

apps/minifront/src/state/dutch-auction/index.ts
```diff
@@ -121,7 +122,8 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
       state.dutchAuction.auctionInfos = [];
     });
 
-    for await (const response of viewClient.auctions({})) {
+    /** @todo: Sort by... something? */
```
TalDerei, line 125: what's the motivation here?

packages/query/src/block-processor.ts
```diff
@@ -433,15 +434,7 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      const seqNum = 1n;
-
-      const metadata = getAuctionNftMetadata(action.value.auctionId, seqNum);
-
-      await Promise.all([
-        this.indexedDb.saveAssetsMetadata(metadata),
-        this.indexedDb.upsertAuction(action.value.auctionId, { seqNum }),
-      ]);
+      await processActionDutchAuctionEnd(action.value, this.querier.auction, this.indexedDb);
     } else if (action.case === 'actionDutchAuctionWithdraw') {
```
TalDerei, line 438: I think helper functions for decluttering the block processor is great. For continuity, can we make `processActionDutchAuctionSchedule` and `processActionDutchAuctionWithdraw` helpers as well?

packages/services/src/view-service/auctions.ts
TalDerei: will this throw if state is `undefined`?

packages/wasm/crate/src/planner.rs
```diff
@@ -449,6 +450,27 @@ pub async fn plan_transaction(
...
     }
 
+    for tpr::ActionDutchAuctionWithdraw { auction_id, seq } in
+        request.dutch_auction_withdraw_actions
+    {
+        let auction_id: AuctionId = auction_id
+            .ok_or_else(|| anyhow!("missing auction ID in Dutch auction withdraw action"))?
+            .try_into()?;
+
+        save_auction_nft_metadata_if_needed(auction_id, &storage, seq).await?;
```
TalDerei, line 460: can you remind me again we make this call in the planner to save the metadata, if we're already doing this in the block processor?

packages/services/src/view-service/auctions.ts
jessepinho: Nope — the `??` is the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing), which defaults to the right side of the operator if `state` is null or undefined.

packages/wasm/crate/src/planner.rs
jessepinho: Huh, I actually didn't think of this. But now that you raise it, I think we'll still need to do it this way for these two use cases:

### 1. The user creates the auction via `pcli`, then views it via a Prax-connected dapp like minifront.

In this case, we need to save the metadata in the block processor, so it won't yet exist in IndexedDB.

### 2. The user creates the auction via a Prax-connected dapp like minifront.

In this case, we need to save the metadata during the planning request, or else the auction NFT will show up as "Unknown Asset" in the "Output" action in the txn approval dialog.

---

Fortunately, these are keyed by ID, so there won't be duplicate assets in the database.

packages/query/src/block-processor.ts
jessepinho: Good call — done!

apps/minifront/src/state/dutch-auction/index.ts
jessepinho: Eventually, I'll probably want to sort by e.g., most recent auction descending, or something.

packages/wasm/crate/src/planner.rs
hdevalence: The planner should not have _any_ write access to view server state, precisely for this reason. The data flow should be that the view server state is only modified during sync, precisely so that it doesn’t matter where the action was performed, because the data in the view server should just be a replica of the chain state. Otherwise we have tricky synchronization problems depending on where a transaction was performed — eg what if the user uses prax on two computers?

packages/wasm/crate/src/planner.rs
```diff
@@ -446,15 +446,15 @@ pub async fn plan_transaction(
             .try_into()?;
 
         save_auction_nft_metadata_if_needed(auction_id, &storage, seq).await?;
-        let oustanding_reserves: OutstandingReserves =
+        let outstanding_reserves: OutstandingReserves =
```
jessepinho, line 449: Typo :)

packages/query/src/block-processor.ts
jessepinho: OK see [this commit](https://github.com/penumbra-zone/web/pull/1068/commits/95da77016f260e15d02d4c41fd47dd4c4a0470d5).

I like keeping `upsertAuction` as a merge function since there are three fields that need to be updated separately. So I broke out outstanding reserves into a separate table, and simply added/deleted records as the user ends/withdraws from an auction, respectively.

(I added a commit afterward to upgrade to the latest buf build/etc. packages, but you can safely ignore that.)

jessepinho: (cc @grod220 / @Valentine1898 / @turbocrime / @TalDerei )

TalDerei (approved):

Issue #819, opened 2024-03-22
Extension "trusted testers" publish workflow


grod220:
Currently, after we have a release in the repo, [extension-publish.yml](https://github.com/penumbra-zone/web/blob/main/.github/workflows/extension-publish.yml) runs. It adds the extension to the review queue, which upon approval, immediately publishes the new version publicly.

In the attempt to build a more robust beta-testing setup, we want to change this up a bit. It looks like this:

- We are ready to release a new extension version
- We create a repo release which runs `extension-publish.yml`
- It adds to review queue and sets the [publish-target](https://github.com/mnao305/chrome-extension-upload/blob/cbba77bc133c3cdc6e0e4248abb5f093ec7e3bd1/action.yml#L31) to `trustedTesters`
- We inform the tester group to review the new update
- When they've deemed it all good, we click publish in the developer dashboard, which immediately makes it public

We should sync with Park about securing the gmail's of this particular group of volunteers from discord + likely a dedicated channel for feedback. Their emails need to be manually added within the developer dashboard.

TalDerei:
https://github.com/penumbra-zone/web/issues/138 for visibility

PR #1070, opened 2024-05-08
Beta extension deployment

grod220:
Closes https://github.com/penumbra-zone/web/issues/819

We now have a google group for prax beta testers: https://groups.google.com/a/penumbralabs.xyz/g/prax-beta-testers. This will allow us to push a private beta version of the extension, garner feedback from external contributors (i.e. passionate discord members). This serves as a kind of staging environment. Details of workflow in docs.

Note new colors for Beta extension.

Screenshot 2024-05-08 at 3 09 16 PM


Valentine1898:
We probably need to have not only the beta version of extension, but also deploy the beta version of minifront. 
In fact, extension has very limited functionality, and there is almost nothing to test manually.  
At the same time, if you test beta version of extension with prod version of minifront, it may create non-existing bugs, and new functionality cannot be tested.

grod220:
> but also deploy the beta version of minifront

Yep! Will create an issue for this. Should be quite easy to do.

Valentine1898 (approved):
LGTM

I wonder if it won't cause any conflicts when users have two wallet extensions installed (beta and prod) 
At least it is not obvious which extension will be used

TalDerei:
Screenshot 2024-05-08 at 10 21 13 AM

oh you have to pay a registration fee, good revenue source there

TalDerei (approved):
what time window are we targeting between how long beta testers will have to review the beta extension features and when the production version will publish? Won't this create time-delays not only for new features, but also for fixes for recently discovered bugs? For the latter case, is there a way to manually publish the production version and bypass the review?

grod220:
@TalDerei 

- Access to the chrome dev dashboard at the moment will be restricted to only a few internal team members fyi
- It's entirely up to us to determine how long we wish for our extension to be tested in the new "staging" step via this private beta extension. For hot fixes, we can always set the prod-version of the extension to go live immediately after approval (like we are doing now). But with the new beta version, we have more flexibility to test new features and/or chain upgrades before they go live for everyone.

PR #1071, opened 2024-05-08
filter out delegationTokens and unbondingTokens from swap asset selector

Valentine1898:
Before: 
image

After: 
image


jessepinho (approved):

TalDerei (approved):
thanks for the decluttering!

PR #1074, opened 2024-05-08
Update WASM crate to point to SHA of core that includes correct middleware

jessepinho:
Core was previously missing a necessary middleware for its auction endpoint. As a result, calls to that endpoint would fail due to a missing trailer.

This was [recently fixed](https://github.com/penumbra-zone/penumbra/pull/4351), so this PR points the WASM crate to the SHA that fixed that issue, so that calls to get the latest auction state from the fullnode will succeed.

This PR also addresses a couple typings changes from core that resulted from the update.

### DIFF
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -78,20 +78,18 @@ impl ActionList {
         // to the fee is ideally small, hopefully it doesn't matter.
         let mut gas = Gas::zero();
         for action in &self.actions {
-            // TODO missing AddAssign
-            gas = gas + action.gas_cost();
+            gas += action.gas_cost();
         }
         for action in self.change_outputs.values() {
-            // TODO missing AddAssign
             // TODO missing GasCost impl on OutputPlan
-            gas = gas + ActionPlan::from(action.clone()).gas_cost();
+            gas += ActionPlan::from(action.clone()).gas_cost();
         }
 
         gas
     }
 
     fn fee_estimate(&self, gas_prices: &GasPrices, fee_tier: &FeeTier) -> Fee {
-        let base_fee = Fee::from_staking_token_amount(gas_prices.fee(&self.gas_estimate()));
+        let base_fee = gas_prices.fee(&self.gas_estimate());
         base_fee.apply_tier(*fee_tier)
     }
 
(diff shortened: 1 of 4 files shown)


packages/wasm/crate/src/planner.rs
```diff
@@ -91,7 +91,7 @@ impl ActionList {
     }
 
     fn fee_estimate(&self, gas_prices: &GasPrices, fee_tier: &FeeTier) -> Fee {
-        let base_fee = Fee::from_staking_token_amount(gas_prices.fee(&self.gas_estimate()));
+        let base_fee = gas_prices.fee(&self.gas_estimate());
```
jessepinho, line 94: `GasPrices::fee()` now returns a `Fee` instead of an `Amount`. I figured `Fee::from_staking_token_amount()` isn't needed, since the fee from a `GasPrices` should always be in the staking token.

packages/wasm/crate/src/planner.rs
```diff
@@ -538,14 +538,14 @@ pub async fn plan_transaction(
     };
 
     if let Some(pb_memo_plaintext) = request.memo {
-        plan.memo = Some(MemoPlan::new(&mut OsRng, pb_memo_plaintext.try_into()?)?);
+        plan.memo = Some(MemoPlan::new(&mut OsRng, pb_memo_plaintext.try_into()?));
```
jessepinho, line 541: Getting rid of the final `?` since it's no longer needed

packages/wasm/crate/src/planner.rs
```diff
@@ -78,20 +78,18 @@ impl ActionList {
         // to the fee is ideally small, hopefully it doesn't matter.
         let mut gas = Gas::zero();
         for action in &self.actions {
-            // TODO missing AddAssign
-            gas = gas + action.gas_cost();
+            gas += action.gas_cost();
         }
         for action in self.change_outputs.values() {
-            // TODO missing AddAssign
```
jessepinho, line 85: Implemented here: https://github.com/penumbra-zone/penumbra/pull/4312/files#diff-97a56abd45353fa2b49c0d9e3ffe5c618326ae35755d2ba69e6d0486d1862d2fR46-R51

grod220 (approved):

erwanor:
We're doing a tagged release today and will bump the auction protos to `v1` so this is great preliminary work towards that

packages/wasm/crate/src/planner.rs
TalDerei: planner logic will need to change again, so that's alright

TalDerei (approved):

PR #1077, opened 2024-05-09
Fix unknown assets for ibc-out

Valentine1898:
The ibc page is broken if you have unknown assets on your balance
telegram-cloud-photo-size-2-5325986573956797599-y

### DIFF
--- a/apps/minifront/src/state/ibc-out.ts
+++ b/apps/minifront/src/state/ibc-out.ts
@@ -8,6 +8,7 @@ import { ClientState } from '@buf/cosmos_ibc.bufbuild_es/ibc/lightclients/tender
 import { Height } from '@buf/cosmos_ibc.bufbuild_es/ibc/core/client/v1/client_pb';
 import { ibcChannelClient, ibcClient, ibcConnectionClient, viewClient } from '../clients';
 import {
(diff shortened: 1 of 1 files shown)


grod220 (approved):
Nice fix 👍

apps/minifront/src/state/ibc-out.ts
```diff
@@ -242,7 +243,6 @@ export const filterBalancesPerChain = (
   const assetIdsToCheck = [penumbraAssetId, ...assetsWithMatchingChannel];
 
   return allBalances.filter(({ balanceView }) => {
-    const metadata = getMetadata(balanceView);
-    return assetIdsToCheck.some(assetId => assetId.equals(metadata.penumbraAssetId));
+    return assetIdsToCheck.some(assetId => assetId.equals(getAssetIdFromValueView(balanceView)));
```
Valentine1898, line 246: Unknown assets also have an `assetId`, but no `metadata`. 

Theoretically, this code allows for an unknown asset to be sent via IBC. But in fact, we can never match an unknown asset with an asset from some ibc channel, because in this case the asset is available in the registry and cannot be an unknown asset

PR #1078, opened 2024-05-09
eliminate @penumbra-zone/constants

Valentine1898:
registry has almost completely eliminated `@penumbra-zone/constants `

In this PR:
- `assets.ts` and `assets.test.ts` moved to `@penumbra-zone/types` ( other packages are not suitable for moving, and there is no resonance in keeping them in a separate package)
- `@penumbra-zone/constants` completely removed 
- minor refactoring of `getUnbondingStartHeight` to avoid cyclic dependency between `@penumbra-zone/getters` and `@penumbra-zone/types`.


turbocrime:
i wonder if `assetPatterns` should conceptually be part of bech32m package

packages/types/src/assets.ts
turbocrime, line 1: what is the source of truth for `PRICE_RELEVANCE_THRESHOLDS`? they are constants here, but where does the number come from?

turbocrime (approved):

packages/types/src/assets.ts
turbocrime: is there a reference link or doc?

Valentine1898: https://github.com/penumbra-zone/web/issues/556#issue-2137977913

Valentine1898:
> i wonder if `assetPatterns` should conceptually be part of bech32m package
> 
> edit: regex exports was mentioned in #778 but eventually omitted from the PR #979. asset patterns are slightly different but related - it could be a submodule of bech32m
> 
> should also add a specified length to the data segment, instead of simply using a greedy selector

perceived `bech32m` as a package that is limited to `bech32m` encoding/decoding only, but it seems that `assetPatterns` may well be part of it. 
Ok, I'll try to moving it if it doesn't create cyclic dependencies

turbocrime:
ok with me to make relocating a separate issue. merge when ready

Issue #1053, opened 2024-05-07
nobletestnet IBC-in error


grod220:
Steps to reproduce:
1 - Get money from [usdc faucet](https://faucet.circle.com/). Send to your noble address (can use keplr/leap).
2 - Go to ibc section in minfront (https://localhost:5173/#/ibc) and issue an ibc-in from noble with your new funds.
3 - See error after approving tx

```
Error: Broadcasting transaction failed with code 1 (codespace: undefined). 
Log: error decoding address (penumbra1g2fpeqzsv702alg33aamp5l8rq56de0h5x87lyjn0mn98jekhfy0dsuecy8364h5qqar8ngkzcsxyc4pt4gg3whtll6dx4gherwz9td7vtl5vnthevxchxknaxnsmglvefvjre): 
decoding bech32 failed: invalid checksum (expected v4u7xm got efvjre)
```

hdevalence:
For Noble specifically we need to use a Bech32 encoding rather than Bech32m, because Noble currently has a middleware that decodes as Bech32. This sucks but when we picked Bech32m we didn't really think about it being different than other Cosmos address encodings. Noble plans to change this at some point in the future but until then we need to use a special encoding just for Noble specifically:

1. Decode the Bech32m address to bytes (or start with bytes)
2. Encode the inner bytes as Bech32 with prefix "penumbracompat"
3. All places where Rust code parses addresses will recognize the prefix and parse as Bech32 instead of Bech32m, so it should Just Work

@avahowell has tested this flow with the CLI so we expect it to work.

PR #1079, opened 2024-05-09
Support compat addresses for ibc-in

grod220:
Closes https://github.com/penumbra-zone/web/issues/1053

For noble, we need to encode our addresses to bech32 (not bech32m). In core, we have a special prefix for this specific case: https://github.com/penumbra-zone/penumbra/blob/main/crates/proto/src/serializers/bech32str.rs#L146.

Amended our bech32m package to support this new encoding method.


Valentine1898:
How long will the ibc transfer from noble take? 
I sent a transaction, but my balance in noble has changed only by the value of the fee  🤔

grod220:
Currently the client state has expired. We'll need to work with the protocol team in getting these back up before we get a successful transfer. However, after this PR, we are no longer getting the bech32 error like we had before. Thanks for testing though!

apps/minifront/src/state/ibc-in.ts
```diff
@@ -93,6 +98,17 @@ export const createIbcInSlice = (): SliceCreator<IbcInSlice> => (set, get) => {
   };
 };
 
+/**
+ * For Noble specifically we need to use a Bech32 encoding rather than Bech32m,
+ * because Noble currently has a middleware that decodes as Bech32.
+ * Noble plans to change this at some point in the future but until then we need
+ * to use a special encoding just for Noble specifically.
+ */
```
turbocrime, line 106: i would add this information to the readme as well

packages/bech32m/src/penumbracompat1.ts
turbocrime, line 1: rename to just `penumbracompat.ts` to match other files in this dir

turbocrime (changes requested):

turbocrime:
suggestions in https://github.com/penumbra-zone/web/pull/1083

packages/bech32m/src/penumbracompat1.ts
grod220: But aren't these files named according to the prefix? For some reason, this prefix uniquely has the 1 in it: https://github.com/penumbra-zone/penumbra/blob/d13d731fadd665d7f2ec38c3bde57f53b2e86891/crates/proto/src/serializers/bech32str.rs#L146

grod220: I also wonder if it has this number in it because we may need to have `penumbracompat2` for some other chain in the future.

turbocrime: interesting. 1 is a valid prefix character, but it is also the separator character, and this may be a bug in core or has some unexplained meaning like version as you suggest.

i can't find any examples or documentation on this.

cc @avahowell you committed the `penumbracompat1` in core?

turbocrime (approved):

Issue #965, opened 2024-04-19
Re-thinking react state patterns


grod220:
The [GrpcEndpointForm](https://github.com/penumbra-zone/web/blob/3ecc7a3ac0e4e1cbc2b1c7955187ff7f34285e92/apps/extension/src/shared/components/grpc-endpoint-form/index.tsx#L14) is a perfect testbed for re-thinking some of our data+state management patterns. It dips into many areas:
- async data fetching with react-query
- async event handler via grpc client (with local state for errors)
- custom hook usage
- local state via Zustand
- derived state via selectors
- 6x useState hooks
- 2x useEffects for reactions

The problem is that the cognitive load of understanding all of these side effects is high. It makes the likeliness of a bug slipping in to be higher, and debugging a slog. Sadly, the default tools of react are insufficient of managing this problem sanely. 

As this grows in requirements, every additional react hook grows the mental complexity by even more. For example, [here is run-of-the-mill hook](https://github.com/osmosis-labs/osmosis-frontend/blob/354481a6ca57339567b878dae1e2eb5cb8dd168c/packages/web/hooks/input/use-amount-input.ts) in the osmosis codebase. The effort required to understand the exact relationships between all of its effects are quite high.

### Ideas

- What if we considered using finite state machines as a model for laying out these relationships? ([shameless plug for my talk](https://www.youtube.com/watch?v=NG-fP_nR7Lk)). We could map out all of the states beforehand in an enum and define the events that go between them. Perhaps a class per state? 🤔 . There are also libraries like xstate that could help with something like this.
- Perhaps we should create our own query store within zustand so we don't have to merge zustand+react-hook state in awkward ways.
- Perhaps another state manager (like mobx) could make it easier to map out side effects and derived state (we'd still have to write a custom query store though).

jessepinho:
@grod220 and I discussed on a separate call the possibility of essentially porting some of the key features of [React Query](https://tanstack.com/query/latest) to Zustand, so that, instead of React Query holding its own state separately from Zustand like it does now, _all_ of our state could live inside Zustand.

I'm proposing doing a spike on a solution to this. I was starting to research what a solution would look like, but then realized that's what a spike would be for.

Here's the requirements of a `ZQuery` (Zustand Query) implementation:
1. Easy way to add it to the store:
    ```
    interface MySlice {
      metadata: ZQuery<Metadata>;
    }

    const mySlice = createMySliceSlice((set, get) => ({
      metadata: zQuery(),
    }))
[...]

PR #1092, opened 2024-05-10
Redesign Duration display for Dutch auctions with live-updating price visualization.

### LINKED ISSUE #1088: Redesign "duration" field in auction view to show a live-updating slider
We know an auction's start/end heights and step count from `DutchAuctionDescription`. We know the current block height from the `status` endpoint. From these pieces of data, we can show a live-updating slider on the view of an auction to depict how far along the auction is toward completion:

$`(currentHeight - startHeight) / (endHeight - startHeight)`$

We can also deduce how many blocks are between each step, in case we want to e.g. show a dot on the graph for each step:

$`(endHeight - startHeight) / stepCount`$

This ticket is to render a UI using this data.

Algorithm for price at each step:

https://github.com/penumbra-zone/penumbra/blob/e0291fb/crates/core/component/auction/src/component/dutch_auction.rs#L548-L572

jessepinho:
Scrub through this video to get a timelapse of part of a 10-minute auction.

https://github.com/penumbra-zone/web/assets/1121544/ad04a0cf-3020-4560-939d-0bd69a15020f

This is a bit of a frilly PR that I prob shouldn't have spent as much time on as I did... Basically it visualizes to the user what the current price is of the input asset in terms of the output asset, as well as how far along we are in the auction.

Again, this is by no means a finalized design, but it's at least an improvement on what we had before.

## In this PR
- Design a live-updating display for Dutch auctions.
- Create a new Zustand `status` slice for storing the sync status (`fullSyncHeight` and `latestKnownBlockHeight`) so it can be used both in the header and in places like Dutch auctions.
- Simplify `<Progress />` to only accept a few specific props, rather than the full default set (which we neither need nor want).
- Add size variants to both `<AssetIcon />` and `<ValueViewComponent />`.

Closes #1088

### DIFF
--- a/apps/minifront/src/components/Status.tsx
+++ b/apps/minifront/src/components/Status.tsx
@@ -1,0 +1,14 @@
+import { useEffect } from 'react';
+import { useStore } from '../state';
+
+/**
+ * A component that simply keeps the Zustand state up to date with the latest
+ * output of `viewClient.status()`.
+ */
+export const Status = () => {
+  const start = useStore(state => state.status.start);
+
+  useEffect(() => void start(), [start]);
+
+  return null;
+};
--- a/packages/ui/components/ui/dutch-auction-component/get-price.ts
+++ b/packages/ui/components/ui/dutch-auction-component/get-price.ts
@@ -45,13 +45,9 @@ export const getPrice = (
   // The input, scaled up by `step_count` to match.
   const inputScaled = (stepCount - 1n) * input;
 
-  // The trading function interpolates between (input, 0) and (0, target_output)
-  const p = targetOutputScaled;
-  const q = inputScaled;
-
   const inputDisplayDenomExponent = getDisplayDenomExponent.optional()(inputMetadata);
   const multiplier = 10 ** (inputDisplayDenomExponent ?? 0);
-  const price = Math.round((Number(p) / Number(q)) * multiplier);
+  const price = Math.round((Number(targetOutputScaled) / Number(inputScaled)) * multiplier);
 
   return new Amount(splitLoHi(BigInt(price)));
 };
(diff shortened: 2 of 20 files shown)


packages/ui/lib/toast/transaction-toast.test.tsx
```diff
@@ -67,13 +67,9 @@ describe('TransactionToast', () => {
...
         expect.objectContaining({
           description: (
-            <Progress
-              status='in-progress'
-              background='stone'
-              value={50}
-              size='sm'
-              className='mt-2'
-            />
+            <div className='mt-2'>
```
jessepinho, line 70: I removed `className` from `<Progress />`'s props, so we need to wrap it in a div with a margin-top.

apps/minifront/src/components/Status.tsx
grod220, line 1: Ah! Another case for https://github.com/penumbra-zone/web/issues/965.

packages/ui/components/ui/dutch-auction-component/get-price.ts
```diff
@@ -0,0 +1,57 @@
...
+  // The target output, scaled up by `step_count` to avoid divisions.
+  // Linearly interpolate between `max_output` at `step_index = 0`
+  //                          and `min_output` at `step_index = step_count - 1`.
+  const targetOutputScaled = (stepCount - stepIndex - 1n) * maxOutput + stepIndex * minOutput;
+  // The input, scaled up by `step_count` to match.
+  const inputScaled = (stepCount - 1n) * input;
+
+  // The trading function interpolates between (input, 0) and (0, target_output)
+  const p = targetOutputScaled;
+  const q = inputScaled;
```
grod220, line 50: Do we need these reassigned? Can we just use `targetOutputScaled` and `inputScaled` inline below?

packages/ui/components/ui/dutch-auction-component/get-progress.ts
```diff
@@ -0,0 +1,16 @@
+const clampToDecimal = (value: number) => Math.min(Math.max(value, 0), 1);
```
grod220, line 1: What does clamp mean here? Maybe a code comment here? (same in other usage too)

grod220 (approved):
Nice! Looks good.

packages/ui/components/ui/dutch-auction-component/get-price.ts
jessepinho: Ah, good catch — this was a leftover from copying core's implementation.

packages/ui/components/ui/dutch-auction-component/get-progress.ts
jessepinho: [Clamping](https://en.wikipedia.org/wiki/Clamping_(graphics)) refers to limiting a value between a given min and max. Will add a comment.

Issue #1049, opened 2024-05-07
Counterparty chain gas prices


grod220:
In order to issue an IBC request into Penumbra, we need to be able to calculate gas prices on the counter party chain. At the moment, we are hardcoding in two places:

https://github.com/penumbra-zone/web/blob/bf28b67ed2b974fa4c513c7f0182d343a763f721/apps/minifront/src/components/ibc/ibc-in/chain-provider.tsx#L19-L32

and

https://github.com/penumbra-zone/web/blob/bf28b67ed2b974fa4c513c7f0182d343a763f721/apps/minifront/src/state/ibc-in.ts#L126-L130

This is not mainnet ready and assumes the transferred token is also the gas token.

The cosmology team has pointed to here as an example: https://github.com/cosmology-tech/create-cosmos-app/blob/e7ed6c946ded6f719e70a4d050f3f356bd553629/examples/nft/pages/_app.tsx#L39.

Whatever method, ideally we can use it injunction with the `estimateFee()` hook from `useChain()` to get some automation for the calculations.

PR #1106, opened 2024-05-14
Properly estimate ibc gas fees

grod220:
Closes https://github.com/penumbra-zone/web/issues/1049

Before we were hardcoding a standard fee. This doesn't work generally across chains. Now we are doing tx simulations to get these estimates.

### DIFF
--- a/apps/minifront/src/state/ibc-in.ts
+++ b/apps/minifront/src/state/ibc-in.ts
@@ -11,11 +11,13 @@ import { cosmos, ibc } from 'osmo-query';
 import { Toast } from '@penumbra-zone/ui/lib/toast/toast';
 import { tendermintClient } from '../clients';
 import { currentTimePlusTwoDaysRounded } from './ibc-out';
-import { StdFee } from '@cosmjs/stargate';
+import { calculateFee, GasPrice, SigningStargateClient } from '@cosmjs/stargate';
 import { getChainId } from '../fetchers/chain-id';
 import { BLOCKS_PER_HOUR } from './dutch-auction/constants';
 import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { bech32CompatAddress } from '@penumbra-zone/bech32m/penumbracompat1';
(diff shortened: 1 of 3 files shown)


apps/minifront/src/state/ibc-in.ts
```diff
@@ -109,6 +111,28 @@ const getCompatibleBech32 = (chainName: string, address: Address): string => {
...
+  client,
+  signerAddress,
+  message,
+}: {
+  chainId: string;
+  client: SigningStargateClient;
+  signerAddress: string;
+  message: EncodeObject;
+}) => {
+  const feeToken = chains.find(({ chain_id }) => chain_id === chainId)?.fees?.fee_tokens[0];
```
Valentine1898, line 125: can there be several `feeTokens`? can there be a case when among 2 `feeTokens` we choose the one which is not on the user's balance?

apps/minifront/src/state/ibc-in.ts
```diff
@@ -109,6 +111,28 @@ const getCompatibleBech32 = (chainName: string, address: Address): string => {
...
+  const avgGasPrice = feeToken?.average_gas_price;
```
Valentine1898, line 126: Is this the accepted approach? If gas prices suddenly rise, all transactions we initiate will fail?

Valentine1898 (approved):
Looks good to me, just a few clarifying questions

apps/minifront/src/state/ibc-in.ts
grod220: In theory yes. However, in practice, the cosmos chain registry for the chains we are connected to all have one. We may need to revisit this later if we connect to a chain and need to build in logic for choosing a gas token.

apps/minifront/src/state/ibc-in.ts
grod220: This is what is employed by cosmos kit (using the average price published in the chain registry). Nearly every example of using cosmjs hardcodes this value. However, I share your sentiment. It feels non-flexible. Going to follow up with cosmology team and get their feedback on this. Will merge this for now though.

PR #1114, opened 2024-05-14
onboarding hotfix

turbocrime:
onboarding storage subscriptions were busted, fixed now.

needs tests, probably. better typed wrapper in another pr

PR #1115, opened 2024-05-14
improve storage wrapper

turbocrime:
wip

### DIFF
--- a/apps/extension/src/service-worker.ts
+++ b/apps/extension/src/service-worker.ts
@@ -53,6 +53,7 @@ import {
 } from './storage/onboard';
 
 const startServices = async (wallet: WalletJson) => {
+  console.log('startServices');
   const grpcEndpoint = await onboardGrpcEndpoint();
   const services = new Services({
     idbVersion: IDB_VERSION,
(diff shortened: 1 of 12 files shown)


turbocrime:
storage wrapper is of debatable utility imo

turbocrime:
better solution than #1114  but more elaborate

packages/storage/src/chrome/base.ts
```diff
@@ -23,26 +18,36 @@ type Version = string;
...
+      if (key in this.defaults) return this.defaults[key as K & D];
+      else return (result as Partial<T>)[key as Exclude<K, D>];
+    }
+
+    // old version in storage
+    if (result[key].version !== this.version) return await this.migrate(key, result[key]);
+
+    // normal case
+    return result[key].value;
   }
```
turbocrime, line 47: ideally, any PR doing this would type the getter such that keys in `this.defaults` are guaranteed and don't need a nullish check by the caller

apps/extension/src/service-worker.ts
```diff
@@ -53,6 +53,7 @@ import {
 } from './storage/onboard';
 
 const startServices = async (wallet: WalletJson) => {
+  console.log('startServices');
```
TalDerei, line 56: remove extraneous log statements

Issue #1052, opened 2024-05-07
IBC-in broadcast toast state


grod220:
The current IBC-in success toast looks like this:

<img width="425" alt="Screenshot 2024-05-07 at 4 57 36 PM" src="https://github.com/penumbra-zone/web/assets/16624263/33f8b20a-45a5-42e2-a862-a925ac8ce6ec">

Corresponding code:

https://github.com/penumbra-zone/web/blob/bf28b67ed2b974fa4c513c7f0182d343a763f721/apps/minifront/src/state/ibc-in.ts#L80-L88

We should make this more useful by adding an action button that goes to mintscan to view the transaction.

Also, we should explore seeing if we can view the state of the transaction, not just if it was successfully broadcast.

PR #1117, opened 2024-05-15
Ibc tx link to mintscan

grod220:
Closes https://github.com/penumbra-zone/web/issues/1052

Note, there may be a delay on tx being picked up by mintscan

https://github.com/penumbra-zone/web/assets/16624263/b73d7927-403d-417f-b425-3eea74f481ab

### DIFF
--- a/apps/minifront/src/state/ibc-in/constants.ts
+++ b/apps/minifront/src/state/ibc-in/constants.ts
@@ -1,7 +1,0 @@
-// Chains using specifically bech32 encoding (not bech32m)
-export const bech32Chains = ['noble', 'nobletestnet'];
-
-export const blockExplorerTxBase: Record<string, string> = {
-  'osmo-test-5': 'https://www.mintscan.io/osmosis-testnet/tx',
-  'grand-1': 'https://www.mintscan.io/noble-testnet/tx',
-};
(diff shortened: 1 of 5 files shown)


apps/minifront/src/state/ibc-in/index.tsx
```diff
@@ -128,7 +144,7 @@ const estimateFee = async ({
   if (!avgGasPrice) throw new Error(`Average gas price not found for ${chainId}`);
 
   const estimatedGas = await client.simulate(signerAddress, [message], '');
-  const gasLimit = estimatedGas * 1.5; // Give some padding to the limit due to fluctuations
+  const gasLimit = Math.round(estimatedGas * 1.5); // Give some padding to the limit due to fluctuations
```
grod220, line 147: Bug fix 🐛 

The gas limit needs to be an integer

apps/minifront/src/state/ibc-in/constants.ts
```diff
@@ -0,0 +1,7 @@
+// Chains using specifically bech32 encoding (not bech32m)
+export const bech32Chains = ['noble', 'nobletestnet'];
+
+export const blockExplorerTxBase: Record<string, string> = {
+  'osmo-test-5': 'https://www.mintscan.io/osmosis-testnet/tx',
+  'grand-1': 'https://www.mintscan.io/noble-testnet/tx',
```
Valentine1898, line 6: we can get it from the cosmos registry
https://github.com/cosmos/chain-registry/blob/2e7e7f67d58c71cad1bc9d82d9342b6e48a72026/noble/chain.json#L234-L257

grod220: Ah! That is indeed better. Going to revise this PR to use this.

PR #1121, opened 2024-05-16
idb plainmessage gasprices

turbocrime:
cc @TalDerei

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -63,7 +63,7 @@ import {
   DutchAuctionDescription,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1alpha1/auction_pb';
 import { ChainRegistryClient } from '@penumbra-labs/registry';
-import { PartialMessage, toPlainMessage } from '@bufbuild/protobuf';
+import { PartialMessage, PlainMessage, toPlainMessage } from '@bufbuild/protobuf';
 
 interface IndexedDbProps {
   dbVersion: number; // Incremented during schema changes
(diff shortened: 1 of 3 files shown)


TalDerei:
@turbocrime bumped idb to enable gas fees on web

TalDerei (approved):

packages/storage/src/indexed-db/index.ts
```diff
@@ -388,12 +392,16 @@ export class IndexedDb implements IndexedDbInterface {
     return SwapRecord.fromJson(json);
   }
 
-  async getGasPrices(): Promise<GasPrices | undefined> {
-    return this.db.get('GAS_PRICES', 'gas_prices');
+  async getGasPrices(): Promise<PlainMessage<GasPrices> | undefined> {
+    return await this.db.get('GAS_PRICES', 'gas_prices');
```
grod220, line 396: Can we serialize/deserialize through json here?

grod220: Also, think `GasPrices` type is better to passing around the application

packages/types/src/indexed-db.ts
```diff
@@ -194,7 +195,7 @@ export interface PenumbraDb extends DBSchema {
   };
   GAS_PRICES: {
     key: 'gas_prices';
-    value: GasPrices;
+    value: PlainMessage<GasPrices>;
```
grod220, line 198: Let's make this `Jsonified<GasPrices>`

turbocrime: `PlainMessage` should provide better type safety than the de-typed output of `Jsonified` which is simply `JsonObject`

packages/storage/src/indexed-db/index.ts
TalDerei: this seems to just work finely

TalDerei:
For context and visibility, the flow is:

1. Bump IDB version, clearing the database on upgrade,
2. Manually set default gas prices on indexDB initialization,
3. `initGasPrices` will only be called once on initialization, and condition will fail on subsequent extension startups,
4. Once non-zero gas fees are enabled in core, a single compact block's pb message will have a `gasPrices` field that won't be undefined,
5. In the block processor, it will call `saveGasPrices` and override the default value in indexDB,
6. remove the `initGasPrices` call

grod220 (approved):

PR #1123, opened 2024-05-16
exclude auctionNft assets from dashboard, send and swap

Valentine1898:


apps/minifront/src/components/dashboard/assets-table/helpers.ts
```diff
@@ -0,0 +1,15 @@
+import { assetPatterns } from '@penumbra-zone/types/assets';
+import { getDisplay } from '@penumbra-zone/getters/metadata';
+import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+import { getMetadata } from '@penumbra-zone/getters/value-view';
+
+const hiddenAssetPatterns = [assetPatterns.auctionNft];
```
grod220, line 6: Can you add a code comment about this being assets we are filtering out?

apps/minifront/src/components/dashboard/assets-table/helpers.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const isUnknown = (balancesResponse: BalancesResponse) =>
+  balancesResponse.balanceView?.valueView.case === 'unknownAssetId';
+
+export const isVisible = (balance: BalancesResponse) =>
```
grod220, line 11: instead of `isVisible` can we do something like `shouldDisplay`? (just so there isn't an confusion this is a private data thing)

apps/minifront/src/components/send/helpers.ts
```diff
@@ -8,3 +14,18 @@ export const penumbraAddrValidation = (): Validation => {
...
   };
 };
+
+const nonTransferableAssetPatterns = [assetPatterns.proposalNft, assetPatterns.auctionNft];
+
+export const isTransferable = (metadata: Metadata) =>
+  nonTransferableAssetPatterns.every(pattern => !pattern.matches(getDisplay(metadata)));
+
+const isUnknown = (balancesResponse: BalancesResponse) =>
+  balancesResponse.balanceView?.valueView.case === 'unknownAssetId';
```
grod220, line 24: Can we import the other one instead of re-defining?

grod220 (approved):

apps/minifront/src/components/dashboard/assets-table/index.tsx
```diff
@@ -64,16 +65,18 @@ export default function AssetsTable() {
...
-                <TableCell>
-                  <ValueViewComponent view={assetBalance.balanceView} />
-                </TableCell>
-                <TableCell>
-                  <EquivalentValues valueView={assetBalance.balanceView} />
-                </TableCell>
-              </TableRow>
-            ))}
+            {account.balances
+              .filter(balance => isVisible(balance))
```
jessepinho, line 69:
```suggestion
              .filter(isVisible)
```

PR #1125, opened 2024-05-16
Create UI to allow users to query latest auction state

### LINKED ISSUE #1059: Create UI for querying latest state of auctions
We don't automatically use `queryLatestState` when calling `ViewService#auctions`, because that results in a call to a fullnode. We should have a button (or something) that the user can click when viewing their auctions, that will fetch the latest auction state from the fullnode.

jessepinho:
screencap

## In this PR
- Add a button to the auctions list to query the latest auction state.
- Show the current reserves of the auction if we have them.
- Extract a `<GradientHeader />` component, to avoid having to copy a bunch of Tailwind classes.
- Show an auction's outstanding reserves if there is an entry in the `AUCTION_OUTSTANDING_RESERVES` IDB table (i.e., when an auction has ended but its reserves haven't yet been claimed).
  - I realized we could have been doing this all along without even querying the latest state, since we know an auction's reserves when it ends.

Closes #1059

### DIFF
--- a/apps/minifront/src/components/swap/dutch-auction/query-latest-state-button.tsx
+++ b/apps/minifront/src/components/swap/dutch-auction/query-latest-state-button.tsx
@@ -1,0 +1,33 @@
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@penumbra-zone/ui/components/ui/tooltip';
+import { ReloadIcon } from '@radix-ui/react-icons';
+import { useStore } from '../../../state';
+
+export const QueryLatestStateButton = () => {
+  const loadAuctionInfos = useStore(state => state.dutchAuction.loadAuctionInfos);
+  const handleClick = () => void loadAuctionInfos(true);
+
+  return (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger
+          onClick={handleClick}
+          aria-label='Get the current auction reserves (makes a request to a fullnode)'
+        >
+          <div className='p-2'>
+            <ReloadIcon />
+          </div>
+        </TooltipTrigger>
+        <TooltipContent>
+          Get the current auction reserves
+          <br />
+          (makes a request to a fullnode)
+        </TooltipContent>
+      </Tooltip>
+    </TooltipProvider>
+  );
+};
--- a/apps/minifront/src/state/dutch-auction/index.ts
+++ b/apps/minifront/src/state/dutch-auction/index.ts
@@ -40,7 +40,7 @@ export interface DutchAuctionSlice {
   onSubmit: () => Promise<void>;
   txInProgress: boolean;
   auctionInfos: AuctionInfo[];
-  loadAuctionInfos: () => Promise<void>;
+  loadAuctionInfos: (queryLatestState?: boolean) => Promise<void>;
   loadMetadata: (assetId?: AssetId) => Promise<void>;
   metadataByAssetId: Record<string, Metadata>;
   endAuction: (auctionId: AuctionId) => Promise<void>;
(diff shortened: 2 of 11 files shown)


packages/ui/components/ui/gradient-header.tsx
```diff
@@ -0,0 +1,8 @@
+/**
+ * A header with text whose color is a gradient of brand colors.
+ */
+export const GradientHeader = ({ children }: { children: string }) => (
```
jessepinho, line 4: image

apps/minifront/src/components/swap/dutch-auction/query-latest-state-button.tsx
```diff
@@ -0,0 +1,33 @@
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '@penumbra-zone/ui/components/ui/tooltip';
+import { ReloadIcon } from '@radix-ui/react-icons';
+import { useStore } from '../../../state';
+
+export const QueryLatestStateButton = () => {
```
jessepinho, line 10: image

packages/ui/components/ui/dutch-auction-component/reserves.tsx
```diff
@@ -0,0 +1,43 @@
...
+    valueView: {
+      case: 'knownAssetId',
+      value: {
+        amount,
+        metadata,
+      },
+    },
+  });
+
+export const Reserves = ({
```
jessepinho, line 21: image

apps/minifront/src/state/dutch-auction/index.ts
```diff
@@ -117,13 +117,13 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
   txInProgress: false,
 
   auctionInfos: [],
-  loadAuctionInfos: async () => {
+  loadAuctionInfos: async (queryLatestState = false) => {
```
grod220, line 120: In what case would we want to not query latest state?

packages/services/src/view-service/auctions.ts
```diff
@@ -65,12 +65,18 @@ export const auctions: Impl['auctions'] = async function* (req, ctx) {
     let state: DutchAuctionState | undefined;
     if (queryLatestState) state = (await querier.auction.auctionStateById(id))?.state;
```
grod220, line 66: can you add a bracket here? I find this is easier to understand:

```typescript
    if (queryLatestState) {
      const auction = await querier.auction.auctionStateById(id);
      state = auction?.state;
    }
```

packages/services/src/view-service/auctions.ts
```diff
@@ -65,12 +65,18 @@ export const auctions: Impl['auctions'] = async function* (req, ctx) {
...
 
+    const outstandingReserves = await indexedDb.getAuctionOutstandingReserves(id);
```
grod220, line 68: Should this be moved into the if clause given it's only needed if `!!value.auction || state` is true?

packages/ui/components/ui/gradient-header.tsx
grod220: Nice abstracting this out

grod220 (approved):

apps/minifront/src/state/dutch-auction/index.ts
jessepinho: We generally don't want to query latest state unless the user explicitly chooses to. Otherwise, we'd be making a request to the fullnode every time the user visits the auction page, which leaks some data about their possible intentions.

packages/services/src/view-service/auctions.ts
jessepinho: Good call. Fixed.

packages/services/src/view-service/auctions.ts
jessepinho: Good catch. Fixed.

PR #1130, opened 2024-05-17
Upgrade to eslint 9.0

grod220:
Primary goal is to bump our eslint dep to the latest. However, this version has a new breaking API change that introduces [flat configs](https://eslint.org/blog/2022/08/new-config-system-part-2/). So did a number of things:
- .eslintrc.js -> eslint.config.mjs format change requirement
- Removed all top level eslint packages in `package.json` and moved them into `packages/eslint-config/package.json`
- Updated eslint config to new api requirements


apps/minifront/src/components/ibc/layout.tsx
```diff
@@ -11,17 +11,13 @@ export const IbcLayout = () => {
         <Card light className='relative z-10 overflow-visible md:self-start'>
           <LongArrowIcon
             direction='right'
-            // Negative calculated margin giving lint issue
-            /* eslint-disable-next-line tailwindcss/enforces-negative-arbitrary-values */
```
grod220, line 15: Any deleted comments in this PR were because they were now flagged as unused

packages/eslint-config/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@penumbra-zone/eslint-config",
+  "version": "1.0.0",
```
grod220, line 3: Renamed this and also removed private. This way, external consumers (like our registry) can make use of this.

apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -24,9 +30,19 @@ export const popupIndexLoader = async (): Promise<Response | PopupLoaderData> =>
...
+  (wallet?: Wallet) =>
+  (index: number, ephemeral: boolean): Address => {
+    if (!wallet) throw new Error('No active wallet');
+
+    const fullViewingKey = FullViewingKey.fromJsonString(wallet.fullViewingKey);
+    return ephemeral
+      ? getEphemeralByIndex(fullViewingKey, index)
+      : getAddressByIndex(fullViewingKey, index);
+  };
+
```
grod220, line 43: Somewhat related to linting, but wanted to remove the exhaustive deps lint exclusion from `packages/ui/components/ui/select-account.tsx`. Moved this from zustand to here. By changing this function to take a wallet (instead of all slices) we don't have the constant re-rendering issue that messed with ephemeral addresses like before.

turbocrime:
eslint 9 supports monorepo/workspaces, so we could get rid of the package entirely and have one top level config

grod220:
One benefit of the package is that we can expose it externally, which is kinda nice especially when we migrate the extension stuff out.

I've also yet to find a way to get files/ignore patterns to work with our configuration outside of the CLI command in package.json (would be required to move it up to the parent). 

It's not a bad idea though, but think we can make a follow up PR to do it sometime later.

turbocrime (approved):

PR #1131, opened 2024-05-17
Fix issue with how error splash page was rendered

### LINKED ISSUE #960: Error splash screen render bug
We seemed to have had a regression from the work done here: https://github.com/penumbra-zone/web/pull/752

Our error splash page seems to be interleaved with our assets table now:

https://github.com/penumbra-zone/web/assets/16624263/7b28b6cd-ffa0-4c4e-8fca-20a1f746b356

jessepinho:
## Before (from #960)
https://github.com/penumbra-zone/web/assets/16624263/7b28b6cd-ffa0-4c4e-8fca-20a1f746b356

## After
image

Closes #960

### DIFF
--- a/apps/minifront/src/components/root-router.tsx
+++ b/apps/minifront/src/components/root-router.tsx
@@ -34,12 +34,10 @@ export const rootRouter = createHashRouter([
             index: true,
             loader: AssetsLoader,
             element: <AssetsTable />,
-            errorElement: <ErrorBoundary />,
           },
           {
             path: PagePath.TRANSACTIONS,
             element: <TransactionTable />,
-            errorElement: <ErrorBoundary />,
           },
         ],
       },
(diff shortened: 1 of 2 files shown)


apps/minifront/src/components/root-router.tsx
```diff
@@ -34,12 +34,10 @@ export const rootRouter = createHashRouter([
             index: true,
             loader: AssetsLoader,
             element: <AssetsTable />,
-            errorElement: <ErrorBoundary />,
```
jessepinho, line 37: Adding error boundaries lower down in the route tree means that the error boundary will render _within_ that route's `<Outlet />`. I've removed these so that all errors bubble up to the top-level boundary.

grod220 (approved):

PR #1132, opened 2024-05-17
Add "Inspect address" UI to extension so users can see if an address belongs to them

### LINKED ISSUE #716: Add address viewing capability to extension popup - so a user can identify control of an ephemeral address
The extension popup currently just has an account switcher and a progress bar.

It also needs to have ability to view/inspect addresses. This should live in the same popup as the current address selector.

There should be a label, I'd propose "Inspect Address" with a magnifying glass. Below that should be a text box the user can paste into.  On paste, the space below the text box should be filled with red X's or green checkmarks, depending:

- if the address is not visible to the wallet, there should be a red X with "address does not belong to this wallet"
- if the address is visible to the wallet, there should be a green checkmark with "address belongs to this wallet". below that:
  - it should show the account number
  - if the address is an IBC deposit address, that should appear below the account number. (If the address is not an IBC deposit address, nothing should be shown).
  
This feature is extremely important to enable IBC flows, because it allows a user to verify that the random address they're sending funds to from their cosmos wallet is actually theirs.

jessepinho:
This PR creates a UI in the extension to detect whether a given address belongs to the current user (and if so, whether it's an IBC/ephemeral address).

https://github.com/penumbra-zone/web/assets/1121544/8192072a-6889-4745-8704-8b9e6deae5f0

## In this PR
- Create `<InspectAddress />` component w/ all necessary functionality.
- Extract a new `<Box />` component and use it in place of styled `<div />`s throughout the codebase.

Closes #716


apps/extension/src/routes/popup/home/inspect-address/result.tsx
```diff
@@ -0,0 +1,34 @@
+import { BadgeAlert, BadgeCheck } from 'lucide-react';
+import type { Result as TResult } from './types';
+
+export const Result = ({ result }: { result?: TResult }) => {
```
jessepinho, line 4: image

apps/extension/src/routes/popup/home/inspect-address/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { Box } from '@penumbra-zone/ui/components/ui/box';
+import { IconInput } from '@penumbra-zone/ui/components/ui/icon-input';
+import { useEffect, useState } from 'react';
+import { MagnifyingGlassIcon } from '@radix-ui/react-icons';
+import { Result as TResult } from './types';
+import { getResultFromBech32mAddress } from './get-result-from-bech32m-address';
+import { Result } from './result';
+
+export const InspectAddress = () => {
```
jessepinho, line 9: image

apps/minifront/src/components/shared/input-block.tsx
```diff
@@ -14,20 +15,22 @@ export const InputBlock = ({ label, className, validations, value, children }: I
...
-        vResult?.type === 'error' && 'border-red-400',
-        vResult?.type === 'warn' && 'border-yellow-300',
-        className,
-      )}
-    >
-      <div className='flex items-center gap-2'>
-        <p className='text-base font-bold'>{label}</p>
-        {vResult ? <div className={cn('italic', 'text-red-400')}>{vResult.issue}</div> : null}
-      </div>
+    <Box>
```
jessepinho, line 18: No change here other than removing the `Box`-specific classes from the div, and wrapping everything in a `<Box />`

apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -105,13 +106,17 @@ export const AssetSelector = ({ assets, onChange, value, filter }: AssetSelector
...
+
           <div className='flex flex-col gap-2 overflow-auto p-4'>
-            <IconInput
-              icon={<MagnifyingGlassIcon className='size-5 text-muted-foreground' />}
-              value={search}
-              onChange={setSearch}
-              placeholder='Search assets...'
-            />
+            <Box spacing='compact'>
+              <IconInput
```
jessepinho, line 112: I removed the black background and gray border from `<IconInput />`, so I had to wrap all instances of `<IconInput />` in the codebase with `<Box />`

packages/ui/components/ui/icon-input.tsx
```diff
@@ -19,7 +17,7 @@ export const IconInput = ({
   placeholder?: string;
 }) => {
   return (
-    <div className={className}>
+    <div className='flex w-full items-center gap-2'>
```
jessepinho, line 20: Got rid of the `<Box />`-like classes, so I just pulled the rest of them inline

packages/ui/components/ui/select-account.tsx
```diff
@@ -38,17 +39,21 @@ export const SelectAccount = ({ getAddrByIndex }: SelectAccountProps) => {
...
           <AccountSwitcher account={index} onChange={setIndex} />
 
-          <div className='mt-4 flex items-center justify-between gap-1 break-all rounded-lg border bg-background px-3 py-4'>
-            <div className='flex items-center gap-[6px] overflow-hidden'>
-              <div className='shrink-0'>
-                <AddressIcon address={address} size={24} />
-              </div>
+          <div className='mt-4'>
+            <Box>
+              <div className='flex items-center justify-between gap-1 break-all'>
```
jessepinho, line 44: Review this with whitespace hidden. I added an `mt-4` wrapper above the box, then removed the `<Box />`-specific classes from this div.

apps/extension/src/routes/popup/home/inspect-address/index.tsx
grod220: Could we add CSS that:
1 - Only causes the inspect box to expand downward and not cause the box to input to raise up on verification?
2 - Not affect the position of the account selector on verification expansion?

apps/extension/src/routes/popup/home/inspect-address/types.ts
```diff
@@ -0,0 +1,3 @@
+export type Result =
```
grod220, line 1: Can we use a more specific name than Result here?

apps/extension/src/routes/popup/home/inspect-address/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [address, setAddress] = useState('');
+  const [result, setResult] = useState<TResult | undefined>();
+
+  useEffect(() => {
+    void getResultFromBech32mAddress(address).then(setResult);
+  }, [address]);
+
+  return (
+    <Box label='Inspect address'>
```
grod220, line 18: Can we have a border color that changes that would indicate an input is not valid? It would be nice to get some type of response from the UI if I put an invalid address.

grod220, line 18: Is "inspect" the best word here? I wonder if `Validate Address` would be better.

apps/extension/src/routes/popup/home/inspect-address/get-result-from-bech32m-address.ts
```diff
@@ -0,0 +1,23 @@
...
+    const { addressIndex } = await viewClient.indexByAddress({
+      address: new Address({ altBech32m: bech32mAddress }),
+    });
+
+    if (!addressIndex) return { belongsToWallet: false };
+
+    return {
+      addressIndexAccount: addressIndex.account,
+      belongsToWallet: true,
+      ibc: !!addressIndex.randomizer.length,
```
grod220, line 18: I wonder if a name like `isEphemeralAddr` would be better than `ibc`

grod220 (approved):

apps/extension/src/routes/popup/home/inspect-address/types.ts
jessepinho: Yep, renamed to `AddressOwnershipInfo`

apps/extension/src/routes/popup/home/inspect-address/index.tsx
jessepinho: 👍🏻 done

apps/extension/src/routes/popup/home/inspect-address/get-result-from-bech32m-address.ts
jessepinho: Done — I renamed it to `isEphemeral`

apps/extension/src/routes/popup/home/inspect-address/index.tsx
jessepinho: I like it! Here's what I came up with:

screencap

Issue #990, opened 2024-04-26
Extension alert toasts


grod220:
At the moment, we do not have a way to communicate errors _within_ the extension. Quite frequently, we've asked users to do `right click -> inspect -> console` to see if there is anything there. It's made troubleshooting quite a challenge. Though, we do have one element that does communicate error states and gives users an indication something is wrong, the block sync bar:

<img width="397" alt="Screenshot 2024-04-26 at 10 58 24 AM" src="https://github.com/penumbra-zone/web/assets/16624263/e31d2ee7-bc03-4254-a841-afc531914c8d">

However, https://github.com/penumbra-zone/web/issues/963 has illuminated the need to be able to trigger some type of toast or popup or indicator if there is an issue the user should be flagged about. Possible design direction:

<img width="486" alt="Screenshot 2024-04-26 at 11 02 44 AM" src="https://github.com/penumbra-zone/web/assets/16624263/c073fa87-f8e5-4143-8191-283f59516542">

Some ideas:
- Create a high level <ErrorBoundary> component that catches all errors that bubble up and put it in a toast
- Expose an API that allows for manual toast triggering (say in a try/catch)
- Consider having a high level service worker error catcher that can propagate errors back to the extension for toast triggering

PR #1134, opened 2024-05-19
WIP: 962 unavailable page links to app not node status page

### LINKED ISSUE #962: Unavailable page links to app, not node status page
If our extension is unavailable, [we encourage](https://github.com/penumbra-zone/web/blob/bfb101310e5695274d151b36ea1b31fb43dd8df3/apps/minifront/src/components/extension-unavailable.tsx#L18-L25) users to visit the node status page.

However, [this link](https://github.com/penumbra-zone/web/blob/bfb101310e5695274d151b36ea1b31fb43dd8df3/apps/minifront/src/components/extension-unavailable.tsx#L4-L5) points to the app page (https://app.testnet.penumbra.zone) and not node page (https://grpc.testnet.penumbra.zone). 

We should change this to link to the node status page (link retrieved from their set grpc endpoint in the network slice).

Valentine1898:
WIP 
This essentially fixes #962, but needs to also solve #990 to be able to show the user in the extension toast with a link to the node status page

### DIFF
--- a/apps/minifront/src/components/extension-unavailable.tsx
+++ b/apps/minifront/src/components/extension-unavailable.tsx
@@ -16,12 +16,7 @@ export const ExtensionUnavailable = () => {
           and see if that fixes the issue.
         </p>
         <p>
-          If it doesn&apos;t, the RPC node that you&apos;re connected to could be down. Check{' '}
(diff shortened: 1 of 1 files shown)


grod220 (approved):
Yes, think we need a better toast system for the extension. But for the minifront side of things, think this is fine.

PR #1136, opened 2024-05-20
Redesign the swap form with swap routes

### LINKED ISSUE #560: Swap execution sankey diagram
For any trading pair, we are able to query for the swap execution trace. It serves to display the pool routes that are expected for the specific swap.

Protos: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1#penumbra.core.component.dex.v1.QueryService.SwapExecution

Add a "Show Execution" button to swap dialog that reveals a [sankey diagram](https://gojs.net/latest/samples/sankey.html). Can also reference other defi apps like Sushi and Uniswap to see how they handle displaying expected routes.

Also, reference the pcli implementation for help: `pcli q dex simulate-trade`.

jessepinho:
This is not a finalized design, as @bibnub will have new designs finalized soon. But this at least shows the user their swap routes in a very lo-fi way until we have better designs.

## In this PR
- Extract a `<TokenSwapInput />` component for entering an amount, an asset in, and asset out. Use it in both the Dutch auction form and the swap form.
- Break out `<AssetOutBox />` into separate component files underneath `<SimulateSwapResult />`.
- Redesign how the filled/unfilled amounts and price impact are rendered.
- Render the traces returned by the swap simulation service underneath the swap form.

image

Closes #560


packages/ui/components/ui/balance-value-view.tsx
```diff
@@ -0,0 +1,15 @@
+import { ValueView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { WalletIcon } from 'lucide-react';
+import { ValueViewComponent } from './tx/view/value';
+
+/**
+ * Renders a `ValueView` as a balance with a wallet icon.
+ */
+export const BalanceValueView = ({ valueView }: { valueView: ValueView }) => {
```
jessepinho, line 8: image

apps/minifront/src/components/swap/dutch-auction/price.tsx
```diff
@@ -50,15 +43,6 @@ export const Price = () => {
           />
         </div>
       </div>
-
-      <div className='w-min'>
```
jessepinho, line 54: `<Price />` no longer renders an asset selector, since that's now part of `<TokenSwapInput />`

apps/minifront/src/components/swap/token-swap-input/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+const isValidAmount = (amount: string, assetIn?: BalancesResponse) =>
+  Number(amount) >= 0 && (!assetIn || !amountMoreThanBalance(assetIn, amount));
+
+/**
+ * Exposes a UI with three interactive elements: an asset selector for the user
+ * to choose which asset to swap _from_, an asset selector for the user to
+ * choose which asset to swap _to_, and a text field for the user to enter an
+ * amount.
+ */
+export const TokenSwapInput = ({
```
jessepinho, line 47: image

apps/minifront/src/components/swap/swap/swap-form/simulate-swap-result/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { useStore } from '../../../../../state';
+import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
+import { PriceImpact } from './price-impact';
+import { Trace } from './trace';
+import { Box } from '@penumbra-zone/ui/components/ui/box';
+
+export const SimulateSwapResult = () => {
```
jessepinho, line 7: image

apps/minifront/src/state/swap.ts
```diff
@@ -32,11 +33,32 @@ import { getAmountFromValue, getAssetIdFromValue } from '@penumbra-zone/getters/
...
 import { amountMoreThanBalance } from './send';
+import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
+
+const getMetadataByAssetId = async (
+  traces: SwapExecution_Trace[] = [],
+): Promise<Record<string, Metadata>> => {
+  const map: Record<string, Metadata> = {};
+
+  for (const trace of traces) {
+    for (const value of trace.value) {
```
grod220, line 44: Probably not a huge performance buff, but you could do:

```typescript
 const promises = traces.flatMap(trace =>
    trace.value.map(async value => {
      if (!value.assetId || map[bech32mAssetId(value.assetId)]) return;

      const { denomMetadata } = await viewClient.assetMetadataById({ assetId: value.assetId });

      if (denomMetadata) {
        map[bech32mAssetId(value.assetId)] = denomMetadata;
      }
    }),
  );

  await Promise.all(promises);
```

apps/minifront/src/state/swap.ts
```diff
@@ -32,11 +33,32 @@ import { getAmountFromValue, getAssetIdFromValue } from '@penumbra-zone/getters/
...
+      if (!value.assetId || map[bech32mAssetId(value.assetId)]) continue;
+
+      const { denomMetadata } = await viewClient.assetMetadataById({ assetId: value.assetId });
+
+      if (denomMetadata) map[bech32mAssetId(value.assetId)] = denomMetadata;
```
grod220, line 49: I find variable assignment in a non-bracketed conditional a bit hard to read. Can you add brackets to this?

grod220 (approved):
Can you double check this doesn't impact our IBC out form?

apps/minifront/src/state/swap.ts
jessepinho: Ah, good call — fixed.

jessepinho:
@grod220:

> - Shouldn't our identicon render `gn` and `gm` with different colors?

Hmm, yeah. I'll make a ticket for that — I wasn't touching colors for this PR, but that's definitely something to look into.

> - Can we add the price somewhere? (aka 10test_usd/1penumbra). Maybe at the every end of each row or on hover of the last pill? This may help displaying because the routes, I believe, are ordered from best price chunk to worse. 

Done:

image

I'll also make a separate PR to add some explainer text about how it's going from most to least beneficial route, but that'll need review separately.

> - I feel like a proper design here needs more width, but with our current layout I see the scroll need. Can we make the scroll bar thinner though?

Yeah, definitely will be better with a wider layout. Re: the scrollbar: this is a very weird issue, and I don't know why Chrome makes the scrollbars so wide?! But I fixed it using a CSS property I didn't know about: `scrollbar-width: thin` 🤯 

> - Could we make the input denom flush left (like it is) and the final output denom flush right? That means the middle would be flexible (sort of like the dotted arrows from the other design).

Ah yeah, I was having trouble getting all the rows to be flush inside a scroll container and punted for now. But given your comment, I put some more effort into it and got it fixed:

screencap

PR #1143, opened 2024-05-22
expose wasm asset id generation

turbocrime:
this allows a caller to convert from an altBaseDenom string to the binary asset id, by using the core class's acceptance of altBaseDenom as hash input


TalDerei (approved):
LGTM

grod220:
Can you give context on where this is needed?

PR #1144, opened 2024-05-22
basic externally connectable

turbocrime:
adds permission to manifest for `<all_urls>` and provides a riddle for users at the endpoint

### DIFF
--- a/apps/extension/src/listeners/message-external.ts
+++ b/apps/extension/src/listeners/message-external.ts
@@ -1,0 +1,44 @@
+/* eslint-disable */
+
+import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
+
+// this is temporary code to use the externally_connectable permission,
+// also providing a secret puzzle for curious users
+chrome.runtime.onMessageExternal.addListener((m, _, um) => {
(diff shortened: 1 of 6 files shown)


apps/extension/src/listeners/message-external.ts
```diff
@@ -0,0 +1,33 @@
+import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
+
+chrome.runtime.onMessageExternal.addListener((msg, _, respond) => {
```
grod220, line 3: So not as to confuse later devs that stumble upon this file, can you add a code comment above about externally connectable and that this is temporary code?

apps/extension/eslint.config.mjs
```diff
@@ -1,13 +1,18 @@
...
+      parserOptions: {
+        project: true,
+        tsconfigRootDir: import.meta.dirname,
+      },
     },
   },
-});
+  {
+    ignores: ['**/message-external.ts'],
+  },
```
grod220, line 17: Alternatively, think you could add `/* eslint-disable */` to the top of the file.

grod220 (approved):

PR #1145, opened 2024-05-22
externally_connectable, iframe+cookie discovery

turbocrime:
how it would work.

some initial detection detail still needs work


TalDerei:
closing as not planned

PR #1150, opened 2024-05-23
fix sync status section

### LINKED ISSUE #1084: minifront may improperly show 'genesis state syncing'
Screenshot 2024-05-09 at 11 01 21

i believe this is related to

1. a querier that asserts a possible undefined https://github.com/penumbra-zone/web/blob/4208c5074ce1c87a759af0150e9adac1ef29f54c/packages/query/src/queriers/tendermint.ts#L15-L18
3. the sync status component treating undefined the same as 0n

this mainly occurs when the tendermint endpoint is unavailable.

i already have some work on this that happened related to #998 that could resolve this

Valentine1898:
close #1084

### DIFF
--- a/apps/minifront/src/state/status.ts
+++ b/apps/minifront/src/state/status.ts
@@ -27,7 +27,6 @@ export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
       }
     } catch (error) {
       set(state => {
-        console.error(error);
         state.status.error = error;
       });
     }
--- a/apps/minifront/src/state/status.ts
+++ b/apps/minifront/src/state/status.ts
@@ -11,7 +11,9 @@ export interface StatusSlice {
 export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
   start: async () => {
     try {
-      //set the current status using unary request, since statusStream does not guarantee that we will get the status
+      // statusStream sends new data to stream only when a new block is detected.
+      // This can take up to 5 seconds (time of new block generated).
+      // Therefore, we need to do a unary request to start us off.
       const status = await viewClient.status({});
       set(state => {
         state.status.fullSyncHeight = status.fullSyncHeight;
--- a/apps/minifront/src/state/status.ts
+++ b/apps/minifront/src/state/status.ts
@@ -10,9 +10,15 @@ export interface StatusSlice {
 
 export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
   start: async () => {
-    const stream = viewClient.statusStream({});
-
     try {
+      //set the current status using unary request, since statusStream does not guarantee that we will get the status
+      const status = await viewClient.status({});
+      set(state => {
+        state.status.fullSyncHeight = status.fullSyncHeight;
+        if (!status.catchingUp) state.status.latestKnownBlockHeight = status.fullSyncHeight;
+      });
+
+      const stream = viewClient.statusStream({});
       for await (const result of stream) {
         set(state => {
           state.status.fullSyncHeight = result.fullSyncHeight;
@@ -21,6 +27,7 @@ export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
       }
     } catch (error) {
       set(state => {
+        console.error(error);
         state.status.error = error;
       });
     }
--- a/packages/services/src/view-service/status.ts
+++ b/packages/services/src/view-service/status.ts
@@ -6,10 +6,9 @@ export const status: Impl['status'] = async (_, ctx) => {
   const { indexedDb } = await services.getWalletServices();
   const latestBlockHeight = await services.querier.tendermint.latestBlockHeight();
   const fullSyncHeight = await indexedDb.getFullSyncHeight();
-  if (!fullSyncHeight) throw new Error('Last block synced not in db');
 
   return {
-    catchingUp: fullSyncHeight < latestBlockHeight,
+    catchingUp: fullSyncHeight ? fullSyncHeight < latestBlockHeight : true,
     partialSyncHeight: fullSyncHeight,
     fullSyncHeight: fullSyncHeight,
   };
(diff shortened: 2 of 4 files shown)


apps/minifront/src/state/status.ts
```diff
@@ -10,9 +10,15 @@ export interface StatusSlice {
...
-
     try {
+      //set the current status using unary request, since statusStream does not guarantee that we will get the status
+      const status = await viewClient.status({});
+      set(state => {
+        state.status.fullSyncHeight = status.fullSyncHeight;
+        if (!status.catchingUp) state.status.latestKnownBlockHeight = status.fullSyncHeight;
+      });
+
+      const stream = viewClient.statusStream({});
```
Valentine1898, line 21: The point is that `statusStream` sends new data to stream only when a new block is detected. 
This means that under normal sync waiting for data from `statusStream` can take up to 5 seconds (time of new block generation). 
And if the network is halted, for example, you will never get data from `statusStream`

packages/services/src/view-service/status.ts
```diff
@@ -6,10 +6,9 @@ export const status: Impl['status'] = async (_, ctx) => {
   const { indexedDb } = await services.getWalletServices();
   const latestBlockHeight = await services.querier.tendermint.latestBlockHeight();
   const fullSyncHeight = await indexedDb.getFullSyncHeight();
-  if (!fullSyncHeight) throw new Error('Last block synced not in db');
```
Valentine1898, line 9: It is almost universally accepted that fullSyncHeight may be undefined during genesis sync . So we should not throw an error in this case

packages/ui/components/ui/block-sync-status/condensed.tsx
```diff
@@ -12,8 +12,9 @@ export const CondensedBlockSyncStatus = ({
   error,
 }: Partial<BlockSyncProps>) => {
   if (error) return <BlockSyncErrorState error={error} />;
+  console.log(latestKnownBlockHeight, fullSyncHeight);
   if (!latestKnownBlockHeight || !fullSyncHeight)
-    return <AwaitingSyncState genesisSyncing={!fullSyncHeight} />;
+    return <AwaitingSyncState genesisSyncing={!!latestKnownBlockHeight} />;
```
Valentine1898, line 16: Minifront incorrectly showed "Genesis state syncing..." instead of "Loading sync state..." for the case when both `fullSyncHeight` and `latestKnownBlockHeight` are undefined

apps/minifront/src/state/status.ts
```diff
@@ -10,9 +10,15 @@ export interface StatusSlice {
...
 export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
   start: async () => {
-    const stream = viewClient.statusStream({});
-
     try {
+      //set the current status using unary request, since statusStream does not guarantee that we will get the status
+      const status = await viewClient.status({});
+      set(state => {
+        state.status.fullSyncHeight = status.fullSyncHeight;
+        if (!status.catchingUp) state.status.latestKnownBlockHeight = status.fullSyncHeight;
```
grod220, line 18: Brackets on this plz

apps/minifront/src/state/status.ts
```diff
@@ -10,9 +10,15 @@ export interface StatusSlice {
 
 export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
   start: async () => {
-    const stream = viewClient.statusStream({});
-
     try {
+      //set the current status using unary request, since statusStream does not guarantee that we will get the status
+      const status = await viewClient.status({});
```
grod220, line 15: Suggestion, but maybe this is an overoptimization, up to you:

```typescript

type Setter = (arg: (state: WritableDraft<AllSlices>) => void) => void;

const setStatusInState = (set: Setter, status: StatusResponse | StatusStreamResponse) => {
  set(state => {
    state.status.fullSyncHeight = status.fullSyncHeight;
    state.status.latestKnownBlockHeight = status.fullSyncHeight;
  });
};

export const createStatusSlice = (): SliceCreator<StatusSlice> => set => ({
  start: async () => {
    try {
      //set the current status using unary request, since statusStream does not guarantee that we will get the status
      const status = await viewClient.status({});
      setStatusInState(set, status);

      const stream = viewClient.statusStream({});
      for await (const status of stream) {
        setStatusInState(set, status);
      }
    } catch (error) {
      set(state => {
        state.status.error = error;
      });
    }
  },
});

```

apps/minifront/src/state/status.ts
grod220: Ah, that's a good catch! Think this github comment is clearer than the code comment though. 

```diff
- //set the current status using unary request, since statusStream does not guarantee that we will get the status
+ // statusStream sends new data to stream only when a new block is detected. This can take up to 5 seconds (time of new block generated). Therefore, we need to do a unary request to start us off.
```

grod220 (approved):

apps/minifront/src/state/status.ts
grod220: Not sure we need `if (!status.catchingUp)`, do we?

apps/minifront/src/state/status.ts
Valentine1898: Unlike `StatusStreamResponse` there is no `latestKnownBlockHeight` field in `StatusResponse` (which is strange really) 
Instead of 
```
state.status.latestKnownBlockHeight = status.fullSyncHeight;
```
we need 
```
state.status.latestKnownBlockHeight = status.latestKnownBlockHeight;
```
And that requires checking that status `instanceof StatusStreamResponse`, so I'll just leave it as it is

apps/minifront/src/state/status.ts
Valentine1898: > Not sure we need `if (!status.catchingUp)`, do we?

This is not necessary, but it allows us to display that we are fully synchronized before we get the first message from the `statusStream` 
Also interesting that the `StatusStream` documentation says `Stream sync status updates until the view service has caught up with the chain`

https://github.com/penumbra-zone/penumbra/blob/2c9c3f3694440e651325306f2345f74796285157/proto/penumbra/penumbra/view/v1/view.proto#L29-L31

In our implementation, streaming continues even after we've caught up with chain (and I think it's the right thing to do) 
But theoretically, there could be other view service implementations where streaming just doesn't happen when we catch up with the chain

TalDerei:
thanks for fixing this!

Issue #1069, opened 2024-05-08
Support swapping `max` in swap form


grod220:
_Idea from @erwanor_ 

For Dexes, it's quite common to want to swap an entire balance of a token. Either there is a dedicated max button or the wallet balance is clickable. See Sushi:

https://github.com/penumbra-zone/web/assets/16624263/38632456-50eb-43c3-892b-aa15b1d56583

This feature is particularly helpful in cases there there are many digits (e.g. `324.979884960959`). 

In our case, is seems easy enough to make this clickable to insert the max balance:

<img width="194" alt="Screenshot 2024-05-08 at 10 58 57 AM" src="https://github.com/penumbra-zone/web/assets/16624263/aeb911c2-9b93-4fcc-b7a1-7fca80ae29da">

Valentine1898:
It might make sense to do all the DEX ui improvements together https://github.com/penumbra-zone/web/issues/972

PR #1151, opened 2024-05-23
Select max input option

grod220:
Closes https://github.com/penumbra-zone/web/issues/1069

Allows the ibc, send, and swap forms to click the balance to insert max amount (quite helpful for amounts with many digits).

https://github.com/penumbra-zone/web/assets/16624263/421f13a2-d572-42a2-b815-ba27b618a2f0

### DIFF
--- a/packages/ui/components/ui/balance-value-view.test.tsx
+++ b/packages/ui/components/ui/balance-value-view.test.tsx
@@ -1,0 +1,87 @@
+import { describe, expect, test, vi } from 'vitest';
+import { fireEvent, render, screen } from '@testing-library/react';
+import {
+  Metadata,
+  ValueView,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { BalanceValueView } from './balance-value-view';
+import { base64ToUint8Array } from '@penumbra-zone/types/base64';
+
+// Mocking the WalletIcon component
+vi.mock('./icons/wallet', () => ({
+  WalletIcon: ({ className }: { className: string }) => (
+    <div data-testid='wallet-icon' className={className}></div>
+  ),
+}));
+
+describe('<BalanceValueView />', () => {
+  const penumbraMetadata = new Metadata({
+    base: 'upenumbra',
+    display: 'penumbra',
+    symbol: 'UM',
+    penumbraAssetId: {
+      inner: base64ToUint8Array('KeqcLzNx9qSH5+lcJHBB9KNW+YPrBk5dKzvPMiypahA='),
+    },
+    images: [
+      {
+        png: 'https://raw.githubusercontent.com/penumbra-zone/web/main/apps/minifront/public/favicon.png',
+      },
+    ],
+    denomUnits: [
+      {
+        denom: 'penumbra',
+        exponent: 6,
+      },
+      {
+        denom: 'mpenumbra',
+        exponent: 3,
+      },
+      {
+        denom: 'upenumbra',
+        exponent: 0,
+      },
+    ],
+  });
+
+  const valueView = new ValueView({
+    valueView: {
+      case: 'knownAssetId',
+      value: {
+        amount: {
+          hi: 0n,
+          lo: 123_456_789n,
+        },
+        metadata: penumbraMetadata,
+      },
+    },
+  });
+
+  test('renders value and wallet icon without cursor-pointer', () => {
+    const { container } = render(<BalanceValueView valueView={valueView} />);
+
+    expect(container).toHaveTextContent(`123.456789`);
+    expect(screen.getByTestId('wallet-icon')).toBeInTheDocument();
+    const clickableElement = screen.getByTestId('wallet-icon').parentElement;
+    expect(clickableElement).not.toHaveClass('cursor-pointer');
+  });
+
+  test('renders with cursor-pointer class', () => {
+    const handleClick = vi.fn();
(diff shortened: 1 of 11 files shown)


packages/types/src/amount.ts
```diff
@@ -77,9 +77,17 @@ export const formatNumber = (number: number, options: FormatOptions): string =>
...
+  withCommas = false,
+}: {
+  amount: Amount;
+  exponent?: number;
+  withCommas?: boolean;
+}) =>
   fromBaseUnitAmount(amount, exponent)
-    .toFormat(6)
+    .toFormat(6, { decimalSeparator: '.', groupSeparator: withCommas ? ',' : '', groupSize: 3 })
     .replace(/(\.\d*?[1-9])0+$|\.0*$/, '$1');
```
grod220, line 91: Before, this was defaulting to returning a string with commas. Seeing that this amount is manipulated in some places later, I think it's safer for consumers to request commas if desired.

packages/ui/components/ui/tx/view/value/index.tsx
```diff
@@ -5,6 +5,19 @@ import { Amount } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/nu
...
+  if (v.valueView.case === 'knownAssetId' && v.valueView.value.metadata) {
+    const { amount = new Amount(), metadata } = v.valueView.value;
+    const exponent = getDisplayDenomExponent.optional()(metadata);
+    return formatAmount({ amount, exponent, withCommas });
+  } else if (v.valueView.case === 'unknownAssetId') {
+    const { amount = new Amount() } = v.valueView.value;
+    return formatAmount({ amount, withCommas });
+  }
+  throw new Error(`Cannot derive formatted amount from value view: ${JSON.stringify(v.toJson())}`);
+};
```
grod220, line 19: Extracted from below so it can be used outside

grod220:
@jessepinho think this touches a lot of the code you were making edits to this week. Can you review? 🙏

packages/ui/components/ui/balance-value-view.tsx
```diff
@@ -1,13 +1,24 @@
...
+  onClick,
+}: {
+  valueView: ValueView;
+  onClick?: (valueView: ValueView) => void;
+}) => {
   return (
-    <div className='flex items-start gap-1 truncate'>
+    <div
+      className={cn('flex items-start gap-1 truncate select-none', onClick && 'cursor-pointer')}
+      onClick={onClick ? () => onClick(valueView) : undefined}
```
jessepinho, line 20:
```suggestion
      onClick={onClick ? () => onClick(valueView) : undefined}
      role={onClick ? 'button' : undefined}
```

packages/ui/components/ui/balance-value-view.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  test('renders value and wallet icon without cursor-pointer', () => {
+    const { container } = render(<BalanceValueView valueView={valueView} />);
+
+    expect(container).toHaveTextContent(`123.456789UM`);
+    expect(screen.getByTestId('wallet-icon')).toBeInTheDocument();
+    const clickableElement = screen.getByTestId('wallet-icon').parentElement;
+    expect(clickableElement).not.toHaveClass('cursor-pointer');
+  });
+
+  test('renders with cursor-pointer class and calls onClick when clicked', () => {
```
jessepinho, line 68: I generally think it's a good idea to split out tests into the single "units" they're testing — i.e., one test that asserts it has the `cursor-pointer` class when `onClick` is passed, and another that asserts that `onClick` is called on click. That gives you better visibility into what's breaking in your unit tests, if something fails. (For example, let's say _both_ the `cursor-pointer` class is missing _and_ the `onClick` handler is not working. This test will fail, but you'll only see the failed expectation for the `cursor-pointer` class, and you won't immediately see that `onClick` isn't working either.)

packages/ui/components/ui/tx/view/value/index.tsx
jessepinho: Hm, can we put this somewhere else, like in `packages/types`? IMO, other components shouldn't be importing universal helpers from a single component's file.

jessepinho (approved):
LGTM! Just a few small comments. Feel free to merge when ready.

hdevalence:
Does this actually do "send max" behavior? It looks like it's setting the input to the exact balance, but this won't work in the presence of fees. However, the recent planner work _does_ provide a way to do "send max", though the plumbing may not be there in the `TransactionPlannerRequest`:

- set the change address to the destination address
- request spending all notes in the account

This gives a true "send max" even in the presence of fees, since the gas will be deducted from the _output_ rather than the input.

grod220:
@hdevalence does the flow you describe pertain to the user that wants to zero out their entire balance of every asset and transfer/swap it?

As I understood it, the userflow that @erwanor originally proposed was specifically for these forms on minifront where a specific asset (from a specific account) has been chosen and the user wants to enter in the max of that specific balance---but typing that many digits feels onerous:

Screenshot 2024-05-24 at 5 45 47 PM

So clicking it just puts it into the input.

Given our fees are in penumbra this doesn't seem like an issue except in the case that you are trying to send the max of specifically penumbra right? And also, it must be the last penumbra in _all_ their accounts (right?) because if a user has penumbra in say account 12 and 99, when attempting to send the max penumbra of account 12, it succeeds because it takes the fees from the other account (?).

If this is true, it seems like we can take detect the case of send-max-of-remaining-penumbra and update the change address to the destination addr (after a proto modification to `TransactionPlannerRequest`)

PR #1158, opened 2024-05-24
publish built typescript in `@penumbra-zone/wasm`

turbocrime:
turns out consuming typescript packages is really annoying, so we'll just do everyone a favor

confirmed to work in the dex app.


packages/wasm/package.json
```diff
@@ -4,19 +4,34 @@
   "license": "(MIT OR Apache-2.0)",
   "type": "module",
   "scripts": {
-    "clean": "rm -fv wasm/index*",
+    "build": "tsc --build",
+    "clean": "rm -rfv dist wasm package penumbra-zone-wasm-*.tgz",
     "clean:rust": "cargo clean --manifest-path ./crate/Cargo.toml",
-    "compile": "cd crate && wasm-pack build --no-pack --target bundler --out-name index --out-dir ../wasm",
+    "compile": "cd crate && wasm-pack build --no-pack --target bundler --out-name index --out-dir ../wasm && touch ../wasm/.npmignore",
```
turbocrime, line 10: this `.npmignore` existed previously, and is used to disable `npm pack`'s aggressive ignoring of directory contents due to `wasm-pack`'s autogenerated `.gitignore` (which can't be disabled). now that `clean` just blows away the directory, we have to restore it after compile

packages/wasm/.npmignore
turbocrime, line 1: this `.npmignore` is new

packages/wasm/package.json
grod220: How does `packages/wasm/wasm/.npmignore` relate to `packages/wasm/.npmignore`? Also, kinda feel like the double wasm dir is kinda confusing.

What purpose does the empty .npmignore serve? Does it make npm not publish the output?

grod220 (approved):

TalDerei (approved):

packages/wasm/package.json
turbocrime: the empty npmignore makes npm actually pack the directory, instead of using the wildcard gitignore to ignore the files. this already existed, i just changed the clean behavior at the former line 7 and thus now it needs recreation in some cases.

the higher level ignore is for not packing the tests.

i think the double wasm directory is also a bad choice, but i didn't make that choice in this PR.

PR #1159, opened 2024-05-24
Deploy minifront on merge to main

grod220:
Now that our APIs are stable, moving back to continuous deployment


Valentine1898 (approved):
That's awesome
But that means we won't change our APIs anymore? And when adding new API methods, we should always handle the case that they may not be implemented in the installed version of the extension?

grod220:
> But that means we won't change our APIs anymore? 

We should do our best, but I'm sure we will

> And when adding new API methods, we should always handle the case that they may not be implemented in the installed version of the extension?

Generally yes. We should be mindful to add support for the extension first before the feature is enabled in minifront. That means we'll have to feature-gate it in minfront or hold the PR until the extension is released.

PR #1160, opened 2024-05-24
Update to latest registry [Starling addition]

grod220:
Update registry to latest:
- Fixes Penumbra labs icons+asset ids
- Adds metadata for Starling

@plaidfinch, registries are loaded into the extension. So when we deploy that early next week, this will be visible 🤩

Screenshot 2024-05-24 at 1 43 16 PM


Valentine1898 (approved):

grod220:
@plaidfinch new extension version just went out. As soon as folks get their updates, they should see the branding now in the staking page.

PR #1173, opened 2024-05-27
optimize animation

### LINKED ISSUE #1067: high cpu from svg animation
Screenshot 2024-05-07 at 16 18 58

my m3 pro is dropping most frames, so i think it's safe to consider this too heavy.

currently, the animation calculates a new path, and mutates the path attribute every frame, causing a rerender.

the animation is a loop so paths could be cached. some components of the animation could easily be a css animation, such as the constant rotation.

options

1. substantially reduce the framerate
3. calculate all paths once, render the frames, and animate that
4. calculate all paths once, and use them as keyframes for an animated svg path
5. use the same path logic to render paths on a canvas, instead of in svg
6. use canvas and gpu-rendered geometry

(3) will be the simplest but ime real svg animation still results in high cpu and may not be worth it

turbocrime:
fixes #1067

i tried a few techniques but this is the simplest and most effective. frames take 4ms on my laptop.

it's actually too fast, now. will have to slow it down.

https://github.com/penumbra-zone/web/assets/134443988/fa0f6ae3-2bef-412f-8aef-a553bbfe3190

### DIFF
--- a/packages/ui/components/ui/logo/animated-penumbra.tsx
+++ b/packages/ui/components/ui/logo/animated-penumbra.tsx
@@ -41,249 +41,32 @@ export const AnimatedPenumbra = ({ className }: { className: string }) => {
           <stop offset='67%' style={{ stopColor: '#f77e14', stopOpacity: 1 }} />
           <stop offset='87%' style={{ stopColor: '#8be4d9', stopOpacity: 1 }} />
         </radialGradient>
-        <mask id='plot1Mask'>
-          <path
-            id='plot1'
-            fill='#fff'
-            stroke='#fff'
-            strokeWidth='0.25'
-            fillRule='evenodd'
-            filter='url(#shiny-filter)'
-          />
-        </mask>
-        <mask id='plot2Mask'>
-          <path
-            id='plot2'
-            fill='#fff'
-            stroke='#fff'
-            strokeWidth='0.25'
-            fillRule='evenodd'
-            filter='url(#shiny-filter)'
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
frame rates much better on my machine!

turbocrime:
should probably use a package like `fast-simplex-noise` instead of this weird diy perlin but whatever

TalDerei:
> instead of this weird diy perlin

diy entropy ftw

packages/ui/components/ui/logo/animation-logic.ts
```diff
@@ -18,92 +20,26 @@ const numPlots = 22; //this number corresponds to the number of plots in the svg
...
-  p: number[],
   octaves: number,
 ) => {
   const baseRad = ccp * baseRadFactor; //the radius of the smallest ring.
   const r =
     initialValue +
     baseRad +
     noiseIntensity *
-      PerlinNoise(p, wobblyConstant * Math.cos(theta), wobblyConstant * Math.sin(theta), octaves);
+      scale(
```
grod220, line 40: Was this the source of the performance improvement? Scaling the noise?

grod220:
Cool to see this improved! It seems like my version runs slower than the one in your video clip? 

https://github.com/penumbra-zone/web/assets/16624263/5fdedaa9-1d82-417b-8826-3ea9cee3fc50

Not a priority thing, but I wonder if the timing is now performance determined (?)

PR #1175, opened 2024-05-27
keys v0.76.0

turbocrime:
https://www.youtube.com/watch?v=8ViiEsVp0xE


jessepinho (approved):
LGTM — how'd you generate the `.shasum` file? (Could you add that to the README when you get a chance?) 

Thanks!

grod220:
@jessepinho after you download the new keys, you can run:

```
shasum -a 256 *.bin
```

And add this to a new .shasum file. We should probably document this process.

jessepinho:
See https://github.com/penumbra-zone/web/pull/1196

PR #1184, opened 2024-05-28
change NOTES in db to ADVICE_NOTES

### LINKED ISSUE #1152: Change `NOTES` in db to `ADVICE_NOTES`
Got confused the other day thinking this table was spendable notes. It's also a bit confusing `penumbra.view.v1.ViewService.Notes` returns spendable notes.

We should rename this to be more specific. Places needing changed:

https://github.com/penumbra-zone/web/blob/d6b8a238bca9eb7e8d177c45c4d5432c37f34efd/packages/types/src/indexed-db.ts#L285

https://github.com/penumbra-zone/web/blob/d6b8a238bca9eb7e8d177c45c4d5432c37f34efd/packages/storage/src/indexed-db/index.ts#L115

https://github.com/penumbra-zone/web/blob/d6b8a238bca9eb7e8d177c45c4d5432c37f34efd/packages/types/src/indexed-db.ts#L196-L203

And bumping idb version:
https://github.com/penumbra-zone/web/blob/d6b8a238bca9eb7e8d177c45c4d5432c37f34efd/apps/extension/.env.testnet#L2

Valentine1898:
close #1152 

This also emphasizes the difference between rust view service data storage, where a relational database is used, and extension view service where a non-relational indexed-db is used. 

The point is that in the relational model, each `SPENDABLE_NOTES` must have a corresponding record in the `NOTES` table (`note_commitment` is used as a foreign key). Therefore, in rust view service the `NOTES` table stores both notes that do not yet have an associated record in the `SPENDABLE_NOTES` table (we call them advices) and notes that already have an associated record in `SPENDABLE_NOTES`

In indexed-db, we store advices separately in the `ADVICE_NOTES` table, and store spendable notes along with nested notes in the `SPENDABLE_NOTES` table

### DIFF
--- a/apps/extension/.env.testnet
+++ b/apps/extension/.env.testnet
@@ -1,4 +1,4 @@
 CHAIN_ID=penumbra-testnet-deimos-8
-IDB_VERSION=40
+IDB_VERSION=41
 MINIFRONT_URL=https://app.testnet.penumbra.zone
 PRAX=lkpmkhpnhknhmibgnmmhdhgdilepfghe
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -112,7 +112,7 @@ export class IndexedDb implements IndexedDbInterface {
         db.createObjectStore('FMD_PARAMETERS');
         db.createObjectStore('APP_PARAMETERS');
 
-        db.createObjectStore('NOTES');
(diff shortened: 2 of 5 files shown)


grod220 (approved):
That is very helpful context in your PR description. Can we capture this somewhere in the codebase? Maybe `packages/types/src/indexed-db.ts` above where ADVICE_NOTES is defined?

PR #1197, opened 2024-05-28
Fix small bugs with auction UIs

### LINKED ISSUE #1181: state issues in auction row rendering
Screenshot 2024-05-27 at 7 45 50 PM

Some issues/papercuts I wanted to note to make sure they were tracked (feel free to deduplicate):

- [x] The first active auction is shown as having "25s of 5m remaining", but it already closed at price 1.000, the UI was never informed about this
- [x] The last auction is already canceled (hence "withdraw"), but this is not reflected in the UI. Probably we want to have a red ⊗ symbol (maybe literally `\U+2297`, idk?) that indicates that the auction was cancelled. (We can distinguish the two "completed" states by whether or not there is any of the input asset in the reserves, if so, it's the failure case, if there is none, it's the success case)
- [x] The expanded details should put "block" next to the block number so that it's more clear what the number means
- [x] The expanded details should have a line showing the auction ID with a copy paste button (for xref'ing on explorers)

jessepinho:
## In this PR
- Show "block" next to block number in auction details.
- Fix issue where auction ID was screwing up layouts by taking up too much horizontal space.
- Show auction ID in auction expanded details.
- Fix auction progress bar logic. Now, a progress bar shows an auction as 100% complete when a user ends an auction early (or even before it's started), whereas before it was showing an auction with "starts in X days" even if the user had already ended it. It also shows a red X if the auction has ended but there are input reserves remaining.
- Fix bug where multiple messages would show up simultaneously saying you don't have any auctions.
- Reorganize files a bit so that helpers/etc. used by only one component are in that component's directory.

Closes #1181


packages/ui/components/ui/box.tsx
```diff
@@ -1,7 +1,7 @@
 import { VariantProps, cva } from 'class-variance-authority';
 import { PropsWithChildren } from 'react';
 
-const variants = cva('rounded-lg border bg-background', {
+const variants = cva('overflow-hidden rounded-lg border bg-background', {
```
jessepinho, line 4: Ensures that children with `truncate`/etc. classes actually get truncated.

packages/ui/components/ui/dutch-auction-component/expanded-details.tsx
```diff
@@ -37,14 +43,14 @@ export const ExpandedDetails = ({
...
-    <div className='flex w-full flex-col'>
+    <div className='flex w-full flex-col overflow-hidden'>
       {maxPrice && (
         <Row label='Maximum'>
           {formatAmount(maxPrice, outputExponent)}
           {outputMetadata && (
             <span className='font-mono text-xs'>
               {' '}
-              {outputMetadata.symbol} / {inputMetadata?.symbol} @{' '}
+              {outputMetadata.symbol} / {inputMetadata?.symbol} @ block{' '}
```
jessepinho, line 53: image

packages/ui/components/ui/dutch-auction-component/expanded-details.tsx
```diff
@@ -88,6 +95,12 @@ export const ExpandedDetails = ({
           {outputMetadata && <span className='font-mono text-xs'> {outputMetadata.symbol}</span>}
         </Row>
       )}
+
+      {auctionId && (
```
jessepinho, line 99: image

packages/ui/components/ui/dutch-auction-component/progress-bar/indicator/get-progress.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, expect, it } from 'vitest';
+import { getProgress } from './get-progress';
+
+describe('getProgress()', () => {
```
jessepinho, line 4: Moved from one folder up. The only new tests here are the ones involving `seqNum`, but git thinks it's a whole new file.

packages/ui/components/ui/dutch-auction-component/progress-bar/indicator/index.tsx
```diff
@@ -0,0 +1,43 @@
+import { DutchAuction } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1/auction_pb';
+import { CircleArrowRight, CircleCheck, CircleX } from 'lucide-react';
+import { getProgress } from './get-progress';
+import { getDescription } from '@penumbra-zone/getters/dutch-auction';
+import { isZero } from '@penumbra-zone/types/amount';
+
+export const Indicator = ({
```
jessepinho, line 7: Moved from one folder up, but there were enough changes that git thinks it's a whole new file. Basically I just pulled more logic into `<Indicator />`.

grod220 (approved):

PR #1199, opened 2024-05-29
Create `ZQuery` package for managing data in Zustand

jessepinho:
`ZQuery` is a new package for managing remote data in Zustand. 

## Navigating this PR

In this PR, I've created the ZQuery package. The bulk of its code is [here](https://github.com/penumbra-zone/web/pull/1199/files#diff-e959855b23b27c5606383a84627732476ac61f4daf74bb56b01f10639321be94), and its types are [here](https://github.com/penumbra-zone/web/pull/1199/files#diff-17ffb5084a8c3404ac15f568398c19768442ab6c269d9dd02eb891d5e6b8860f).

This PR is a bit large because I've also implemented its use in a few places throughout the app to test it:
- The sync status:
  - Updated the [`<SyncStatusSection />` component](https://github.com/penumbra-zone/web/pull/1199/files#diff-dfbe890e82cc7c1d3f48a5c019f5496108bafc754e2e7d224078204be2236844)
  - Deleted the [`<Status />` component](https://github.com/penumbra-zone/web/pull/1199/files#diff-cc88999f1fdbc43337ce26d610ce0d9e7ecb43fff724fd6ac36e18c095622117) that was being used to sync status data to Zustand, since this is now handled automatically by hooks
  - Created a [status fetcher](https://github.com/penumbra-zone/web/pull/1199/files#diff-a916bd214d4af578812a71a542618c565af076785e6576200da0483615bb043b)
  - Simplified the status [Zustand slice](https://github.com/penumbra-zone/web/pull/1199/files#diff-153d57544bebed1b179bedfc904e2ab291e2ccc9af6bcdd9bee74a504d782cc3) to use ZQuery.
  - [Updated `<AuctionList />`](https://github.com/penumbra-zone/web/pull/1199/files#diff-36f96b273b871d8dc2c9e0cdc6eb75c2d65b63376652fc7e89e1839f63b3a9b7R43) to use `useStatus()` rather than getting the status from the store
  - [Updated `<TokenSwapInput />`](https://github.com/penumbra-zone/web/pull/1199/files#diff-eba73967353d1d0a2110102b4d3dfdd93a8c16e95c1c5221b79c18f3626facb3) to use `useStatus()` rather than getting the status from the store
- Unclaimed swaps:
  - [Removed unclaimed swaps from the swap loader](https://github.com/penumbra-zone/web/pull/1199/files#diff-e8569cb3cca9a2671e2fdac07e46cc214c5f52d39845ccb5dd532ee0407f900f), since they'll now be loaded using ZQuery
  - [Added a ZQuery object to the `unclaimedSwaps` Zustand slice](https://github.com/penumbra-zone/web/pull/1199/files#diff-3720f9ccd465ba75b008e79248d05bcd6b62c83601deafaa6da5a82b09568e70)
  - Consolidated logic related to the [unclaimed swaps fetcher](https://github.com/penumbra-zone/web/pull/1199/files#diff-1648548db6e93435861014b8d889e0278c9a73020f26e5e5a5dc1484e4cf7701)
  - [Updated and simplified the `<UnclaimedSwaps />` component](https://github.com/penumbra-zone/web/pull/1199/files#diff-89f339099b7673059b189a64d860d14ca13c9d2cebdfc1b3d789eb503ca43760)
- Auctions:
  - [Simplified the `dutchAuction` Zustand slice](https://github.com/penumbra-zone/web/pull/1199/files#diff-59f377e1755cd460f2acc198e470113649cbc814a887c38a6b930896ca30e7b8) to use ZQuery for the `auctionInfos` array.
  - [Extract the fetching logic](https://github.com/penumbra-zone/web/pull/1199/files#diff-4ba907eee9cabe5b64c838d07f1baec6eee94fc6adc210fb8e5c7a8df89c431f) to a new fetcher
  - [Update the `<AuctionList />` component](https://github.com/penumbra-zone/web/pull/1199/files#diff-36f96b273b871d8dc2c9e0cdc6eb75c2d65b63376652fc7e89e1839f63b3a9b7) to use the new `useAuctionInfos()` hook

## Main missing features for future improvements
- The biggest one: we don't currently cache based on parameters. Ideally, we should be able to do something like this: `const auctionInfos = useAuctionInfos({ cacheKey: 'latestState', fetchArgs: [{queryLatestState: true }] });`.  Then, the results of calling the `auctionInfos` fetcher with `{queryLatestState:true}` would be cached under the `latestState` cache key. If you just call `const auctionInfos = useAuctionInfos()` — i.e., without any arguments — the results would be stored under e.g., a `_zQueryDefaultCacheKey` cache key, and would be returned any time a consumer calls `useAuctionInfos()` without any arguments.

    This is actually a bit complex to implement, though, and this PR is a bit large already, so I figured it's best to ship this as-is and revisit this later if needed.
- We're not using a finite state machine yet. That is, it's possible for `error` and `data` and `loading` to all be truthy at the same time. In the interest of limiting this PR to a reasonable scope, I'm pushing that off till later.
- We don't truly abort calls when a new request is started. We [call an abort controller's `abort()` method](https://github.com/penumbra-zone/web/pull/1199/files#diff-e959855b23b27c5606383a84627732476ac61f4daf74bb56b01f10639321be94R185), but all that does is [stop iterating over the results](https://github.com/penumbra-zone/web/pull/1199/files#diff-e959855b23b27c5606383a84627732476ac61f4daf74bb56b01f10639321be94R247). I'm not familiar enough with generator functions to know if that actually stops the server from streaming data, but I'm pretty sure it doesn't? Could be wrong, though. In the future, we'll likely want to pass abort controllers (or at least their `signal` property) to the `fetch` callback so that we can abort remote calls properly.


apps/minifront/src/state/status.ts
```diff
@@ -1,36 +1,37 @@
...
 });
+
+export interface StatusSlice {
+  status: ZQueryState<{
+    fullSyncHeight?: bigint;
+    latestKnownBlockHeight?: bigint;
+  }>;
+}
+
+export const createStatusSlice = (): SliceCreator<StatusSlice> => () => ({
```
VanishMax, line 35: [thought](https://conventionalcomments.org/): As a user, I'd expect to use the state inside Zustand with selectors as we always do, so I'd like for ZQuery to just return a slice that I can simply plug into the store and play with it. I don't think the return from ZQuery `{ status, useStatus }` is really relevant. Can we make ZQuery just a kinda create-and-forget type, so all the interactions with it would be through Zustand?

apps/minifront/src/state/status.ts
```diff
@@ -1,36 +1,37 @@
...
-      });
-    }
+export const { status, useStatus } = createZQuery({
+  name: 'status',
+  fetch: getStatusStream,
+  stream: (_prevState: Status | undefined, item: Status): Status => ({
+    fullSyncHeight: item.fullSyncHeight,
+    latestKnownBlockHeight: item.latestKnownBlockHeight,
+  }),
+  getUseStore: () => useStore,
```
VanishMax, line 18: [question](https://conventionalcomments.org/): why do we actually need the whole store for manipulating with just one slice? Aren't getters and setters enough?

apps/minifront/src/state/status.ts
```diff
@@ -1,36 +1,37 @@
...
+  get: state => state.status.status,
+  set: setter => {
+    const newState = setter(useStore.getState().status.status);
+    useStore.setState(state => {
+      state.status.status = newState;
+    });
```
VanishMax, line 24: [question](https://conventionalcomments.org/): from the DX perspective, can `setter` accept a `newState` argument with the `getter()` return, so that we won't have to query it every time?

VanishMax (commented):
Impressive work! 

Overall, ZQuery feels like a complicated multi-layer abstraction but the one I could personally get used to. I left some comments in form on the [conventional comments](https://conventionalcomments.org/). None of them actually prompt you to change anything but we could have a discussions in the comment threads to clarify some points

apps/minifront/src/state/status.ts
jessepinho: That was actually my initial design for it, but one of the goals of this project was to integrate into the React lifecycle hooks so that data would get fetched on component mount, and any in-flight requests would be aborted on component unmount. Using just selectors means that we would miss out on that benefit.

I do understand the desire for something that "feels" more like Zustand, though. If we decide to merge this PR as-is, we could still revisit that question in the future.

apps/minifront/src/state/status.ts
jessepinho: Ugh, unfortunately we need it for actually calling the hook (e.g., [here](https://github.com/penumbra-zone/web/blob/jessepinho/z-query-web-965/packages/zquery/src/index.ts#L185-L195)). I wanted to avoid this but, since we're actually hooking into the React component lifecycle, we need to actually use the `useStore` hook (rather than just the getter/setter).

apps/minifront/src/state/status.ts
jessepinho: Good call. I'd originally used the design you described, then had to switch to this design for some TypeScript reason. I think that reason has since become moot, so I should be able to switch it back. Will look into that and follow up. Thanks for catching that! (This DX is indeed quite confusing.)

jessepinho: OK, wasn't able to resolve this. For other reviewers' benefit, I've messaged Max separately about pairing on resolving the TS issues related to this; until then, though, I'd still love a PR review on this!

packages/ui/components/ui/dutch-auction-component/expanded-details/index.tsx
```diff
@@ -39,8 +39,8 @@ export const ExpandedDetails = ({
     fullSyncHeight >= description.startHeight &&
     fullSyncHeight <= description.endHeight;
 
-  const inputExponent = getDisplayDenomExponent(inputMetadata);
-  const outputExponent = getDisplayDenomExponent(outputMetadata);
+  const inputExponent = getDisplayDenomExponent.optional()(inputMetadata);
+  const outputExponent = getDisplayDenomExponent.optional()(outputMetadata);
```
grod220, line 43: Bug fix?

grod220 (approved):
Holy smokes, this is a full blown professional library addon. Think we should pitch Zustand-core for an integration 😅 

Think it's definitely in a place where we can feel confident about merging and brainstorming (in a more incremental way) updates, new features, api changes, etc.

Cool stuff!

packages/ui/components/ui/dutch-auction-component/expanded-details/index.tsx
jessepinho: Yep — weirdly, some asset metadata was missing in testnet that I'd had in my wallet.

PR #1200, opened 2024-05-29
add disconnect method to page api

turbocrime:
adds disconnect method to page api

page injection is refactored as a class to more explicitly represent and control injection state

adds listeners in prax to handle disconnect

for some reason, connection stalls on the ibc page, but works everywhere else.

minifront ui in subsequent pr

### DIFF
--- a/apps/extension/src/content-scripts/injected-penumbra-global.ts
+++ b/apps/extension/src/content-scripts/injected-penumbra-global.ts
@@ -34,7 +34,7 @@ class PraxInjection {
   private static singleton?: PraxInjection = new PraxInjection();
 
   public static get penumbra() {
-    return PraxInjection.singleton!.injection;
+    return new PraxInjection().injection;
   }
 
   private manifestUrl = `${PRAX_ORIGIN}/manifest.json`;
@@ -80,7 +80,14 @@ class PraxInjection {
   private state(): boolean | undefined {
     if (this.disconnectState) return false;
     if (this.requestState === 'rejected') return false;
-    return this.connectState && this.connectState === 'fulfilled';
+    switch (this.connectState) {
+      case 'rejected':
+        return false;
+      case 'fulfilled':
+        return true;
+      case undefined:
+        return undefined;
+    }
   }
 
   // this listener will resolve the connection promise AND request promise when
@@ -122,13 +129,14 @@ class PraxInjection {
   // 3. request
   private get connectionFailure() {
     // Promise.race checks in order of the list index. so if more than one
-    // promise is settled, Promise.race responds with the earlier index
+    // promise in the list is already settled, it responds with the result of
+    // the earlier index
     return Promise.race([
-      // rejects with disconnect success, rejects with disconnect failure
+      // rejects with disconnect failure, or 'Disconnected' if disconnect was successful
       this._disconnect.promise.then(() => Promise.reject(Error('Disconnected'))),
-      // ignores connection success, rejects with connection failure
+      // rejects with connect failure, never resolves
       this._connect.promise.then(() => new Promise<never>(() => null)),
-      // rejects with previous success, rejects with previous failure
+      // rejects with previous failure, or 'Disconnected' if request was successful
       this._request.promise.then(() => Promise.reject(Error('Disconnected'))),
       // this should be unreachable
       Promise.reject(Error('Unknown failure')),
@@ -189,13 +197,15 @@ class PraxInjection {
   }
 }
 
-// if the global isn't present, create it.
-if (!window[PenumbraSymbol])
-  Object.defineProperty(window, PenumbraSymbol, { value: {}, writable: false });
-
-// reveal
-Object.defineProperty(window[PenumbraSymbol], PRAX_ORIGIN, {
-  value: PraxInjection.penumbra,
-  writable: false,
-  enumerable: true,
-});
+// inject prax
+Object.defineProperty(
+  window[PenumbraSymbol] ??
+    // create the global if not present
+    Object.defineProperty(window, PenumbraSymbol, { value: {}, writable: false })[PenumbraSymbol],
+  PRAX_ORIGIN,
+  {
+    value: PraxInjection.penumbra,
+    writable: false,
+    enumerable: true,
+  },
+);
--- a/apps/extension/src/content-scripts/injected-penumbra-global.ts
+++ b/apps/extension/src/content-scripts/injected-penumbra-global.ts
@@ -5,54 +5,101 @@
  *
  * The global is identified by `Symbol.for('penumbra')` and consists of a record
  * with string keys referring to `PenumbraInjection` objects that contain a
- * simple api. The identifiers on this record should be unique and correspond to
- * an id in a manifest, and providers should provide a link to the manifest in
- * their record entry.
+ * simple API. The identifiers on this record should be unique, and correspond
+ * to a browser extension id. Providers should provide a link to their extension
+ * manifest in their record entry.
  *
- * The global is frozen to prevent mutation, but you should consider that the
+ * The global is frozen to discourage mutation, but you should consider that the
  * global and everything on it is only as trustable as the scripts running on
- * the page. Imports, includes, and packages your webapp depends on could all
- * mutate or preempt the global. User-agent injections like userscripts or other
- * content scripts could interfere or intercept connections.
+ * the page. Imports, requires, includes, script tags, packages your webapp
+ * depends on, userscripts, or other extensions' content scripts could all
+ * mutate or preempt this, and all have the power to interfere or intercept
+ * connections.
  */
 
 import { PenumbraInjection, PenumbraRequestFailure, PenumbraSymbol } from '@penumbra-zone/client';
-import { isPraxFailureMessageEvent, isPraxPortMessageEvent, PraxMessage } from './message-event';
+
+import {
+  isPraxFailureMessageEvent,
+  isPraxPortMessageEvent,
+  PraxMessage,
+  unwrapPraxMessage,
+} from './message-event';
 
 import { PraxConnection } from '../message/prax';
 
-const request = Promise.withResolvers<void>();
+type PromiseSettledResultStatus = PromiseSettledResult<unknown>['status'];
 
-// this is just withResolvers, plus a sync-queryable state attribute
-const connection = Object.assign(Promise.withResolvers<MessagePort>(), { state: false });
-connection.promise.then(
-  () => {
-    connection.state = true;
-    request.resolve();
-  },
-  () => {
-    connection.state = false;
-    request.reject();
-  },
-);
+class PraxInjection {
+  private static singleton?: PraxInjection = new PraxInjection();
 
-// this resolves the connection promise when the isolated port script indicates
-const connectionListener = (msg: MessageEvent<unknown>) => {
-  if (isPraxPortMessageEvent(msg) && msg.origin === window.origin) {
-    // @ts-expect-error - ts can't understand the injected string
-    const praxPort: unknown = msg.data[PRAX];
-    if (praxPort instanceof MessagePort) connection.resolve(praxPort);
+  public static get penumbra() {
+    return PraxInjection.singleton!.injection;
   }
-};
-window.addEventListener('message', connectionListener);
-void connection.promise.finally(() => window.removeEventListener('message', connectionListener));
 
-// declared outside of postRequest to prevent attaching multiple identical listeners
-const requestResponseListener = (msg: MessageEvent<unknown>) => {
-  if (msg.origin === window.origin) {
-    if (isPraxFailureMessageEvent(msg)) {
-      // @ts-expect-error - ts can't understand the injected string
-      const status = msg.data[PRAX] as PraxConnection;
+  private manifestUrl = `${PRAX_ORIGIN}/manifest.json`;
+  private _request = Promise.withResolvers<void>();
+  private _connect = Promise.withResolvers<MessagePort>();
+  private _disconnect = Promise.withResolvers<void>();
+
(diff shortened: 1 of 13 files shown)


apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -5,54 +5,101 @@
...
+    void this._request.promise.then(
+      () => (this.requestState = 'fulfilled'),
+      () => (this.requestState = 'rejected'),
+    );
+  }
+
+  private state(): boolean | undefined {
+    if (this.disconnectState) return false;
+    if (this.requestState === 'rejected') return false;
+    return this.connectState && this.connectState === 'fulfilled';
```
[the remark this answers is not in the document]
grod220: It would be nice if the PraxInjection exposed state like:

```ts
enum ConnectionState {
  Idle = 'Idle',
  Connecting = 'Connecting',
  Connected = 'Connected',
  Disconnected = 'Disconnected',
  Disconnecting = 'Disconnecting',
  Failed = 'Failed',
}
```

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -5,54 +5,101 @@
...
+
+  private constructor() {
+    if (PraxInjection.singleton) return PraxInjection.singleton;
+
+    window.addEventListener('message', this.connectionListener);
+
+    void this._connect.promise
+      .then(
+        () => (this.connectState ??= 'fulfilled'),
+        () => (this.connectState ??= 'rejected'),
```
grod220, line 65: Why do we use `??=` here? Shouldn't we respect the latest value?

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
 
-// Called to request a connection to the extension.
-const postRequest = () => {
-  if (!connection.state) {
-    window.addEventListener('message', requestResponseListener);
+  private postRequest() {
+    const state = this.state();
+    if (state === true)
+      // connection is already active
+      this._request.resolve();
```
grod220, line 142: Brackets plz

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
+    // attempt actual disconnect
+    void this._connect.promise
+      .then(
+        port => {
+          port.postMessage(false);
+          port.close();
+        },
+        (e: unknown) => console.warn('Could not attempt disconnect', e),
+      )
+      .catch((e: unknown) => console.error('Disconnect failed', e));
```
grod220, line 176: If disconnect fails, shouldn't we simply just throw here?

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -5,54 +5,101 @@
...
+  }
+
+  private manifestUrl = `${PRAX_ORIGIN}/manifest.json`;
+  private _request = Promise.withResolvers<void>();
+  private _connect = Promise.withResolvers<MessagePort>();
+  private _disconnect = Promise.withResolvers<void>();
+
+  private connectState?: PromiseSettledResultStatus;
+  private requestState?: PromiseSettledResultStatus;
+  private disconnectState?: PromiseSettledResultStatus;
```
grod220, line 47: In general, it's quite difficult trying to crawl the lifecycle of these promises in the code. Is there a different paradigm or abstraction we could use to make this more maintainable? Maybe even consider a finite state machine pattern. Or maybe create helper classes of some kind.

apps/extension/src/listeners/message-prax-request.ts
```diff
@@ -8,29 +8,37 @@ import { UserChoice } from '@penumbra-zone/types/user-choice';
...
-        }
-        if (e instanceof ConnectError && e.code === Code.Unauthenticated) {
-          respond(PraxConnection.NeedsLogin);
-        } else {
-          respond(PraxConnection.Denied);
-        }
-      },
-    );
-    return true; // instruct chrome to wait for the response
+    console.log('prax request', req, sender);
```
grod220, line 11: Can remove

apps/extension/src/listeners/message-prax-request.ts
```diff
@@ -8,29 +8,37 @@ import { UserChoice } from '@penumbra-zone/types/user-choice';
...
+    switch (req) {
+      case PraxConnection.End:
+        void removeOrigin(sender);
+        respond(PraxConnection.End);
+        return true; // instruct chrome to wait for the response
+      case PraxConnection.Request:
```
grod220, line 17: If `PraxConnection.Request` tries to get an approval, think we should rename this to `PraxConnection.RequestApproval`

apps/extension/src/origin.ts
```diff
@@ -11,6 +11,29 @@ export const originAlreadyApproved = async (url: string): Promise<boolean> => {
...
+  const { origin: senderOrigin } = sender;
+
+  if (!senderOrigin) {
+    console.warn('No record for origin', sender);
+    return;
+  }
+
+  const urlOrigin = new URL(senderOrigin).origin;
+  const knownSites = await localExtStorage.get('knownSites');
+  const sansDeletant = knownSites.filter(site => site.origin !== urlOrigin);
```
grod220, line 25: What's a deletant? Can we use another word?

apps/extension/src/origin.ts
```diff
@@ -11,6 +11,29 @@ export const originAlreadyApproved = async (url: string): Promise<boolean> => {
...
+
+  if (sansDeletant.length === knownSites.length) {
+    console.warn('No site record found:', urlOrigin);
+    return;
+  }
+
+  if (!sansDeletant.length) console.warn('No site records remaining:', sansDeletant);
```
grod220, line 32: This state is fine. Don't think we need a log for it.

packages/transport-chrome/src/session-client.ts
```diff
@@ -23,8 +23,35 @@ import {
 import { ChannelLabel, nameConnection } from './channel-names';
 import { isTransportInitChannel, TransportInitChannel } from './message';
 import { PortStreamSink, PortStreamSource } from './stream';
-import { Code, ConnectError } from '@connectrpc/connect';
-import { errorToJson } from '@connectrpc/connect/protocol-connect';
+
+const localErrorJson = (e: unknown, m?: unknown) =>
```
grod220, line 27: Can you change name to `(e: unknown, message?: unknown)`? Or is `m` suppose to be something else?

packages/transport-chrome/src/session-client.ts
```diff
@@ -59,33 +86,37 @@ export class CRSessionClient {
...
+  };
+
   private clientListener = (ev: MessageEvent<unknown>) => {
     try {
-      if (isTransportMessage(ev.data)) this.servicePort.postMessage(ev.data);
+      if (ev.data === false) this.disconnectService();
+      else if (isTransportMessage(ev.data)) this.servicePort.postMessage(ev.data);
       else if (isTransportStream(ev.data))
         this.servicePort.postMessage(this.requestChannelStream(ev.data));
       else console.warn('Unknown item from client', ev.data);
```
grod220, line 105: Please add brackets to all of these and the if-clause below

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -5,54 +5,101 @@
...
+
+  private injection: Readonly<PenumbraInjection> = Object.freeze({
+    disconnect: () => this.endConnection(),
+    connect: () => (this.state() !== false ? this._connect.promise : this.connectionFailure),
+    isConnected: () => this.state(),
+    request: () => this.postRequest(),
```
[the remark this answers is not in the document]
turbocrime: this is the same client api we had before.

request is a method that initiates a request for origin approval

packages/client/src/index.ts
```diff
@@ -5,6 +5,7 @@ export const PenumbraSymbol = Symbol.for('penumbra');
 export interface PenumbraInjection {
   readonly connect: () => Promise<MessagePort>;
   readonly request: () => Promise<void>;
+  readonly disconnect: () => Promise<void>;
   readonly isConnected: () => boolean | undefined;
```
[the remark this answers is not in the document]
turbocrime: this is the same client api we had before.

undefined means that a connection state has not been resolved - it is unattempted or pending. after attempting to connect, or after closing a connection, it may be 'false'.

a caller can understand that a truthy response indicates the connection is active, and a falsy value indicates the connection is not active.

apps/extension/src/content-scripts/injected-penumbra-global.ts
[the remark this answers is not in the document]
turbocrime: this is a private internal method that helps identify the appropriate boolean to return for the isConnected() interface.

i think adding an additional state variable that must be managed according to three other state variables is more complex than simply resolving the value upon demand.

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
-      request.reject(failure);
+      this._request.reject(failure);
     }
+  };
+
+  // always reject with the most important reason at time of access
+  // 1. disconnect
+  // 2. connection failure
+  // 3. request
+  private get connectionFailure() {
```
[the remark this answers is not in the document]
turbocrime: this provides rejection reasons for the promise returned by the connect() interface. it's used to provide a rejection for calls to the connect() interface if the connection is already closed.

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
+    // Promise.race checks in order of the list index. so if more than one
```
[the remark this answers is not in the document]
turbocrime: expanded comment

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
+    // promise is settled, Promise.race responds with the earlier index
+    return Promise.race([
+      // rejects with disconnect success, rejects with disconnect failure
+      this._disconnect.promise.then(() => Promise.reject(Error('Disconnected'))),
+      // ignores connection success, rejects with connection failure
+      this._connect.promise.then(() => new Promise<never>(() => null)),
+      // rejects with previous success, rejects with previous failure
```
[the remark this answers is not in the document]
turbocrime: elaborated comments

apps/extension/src/content-scripts/injected-penumbra-global.ts
```diff
@@ -65,44 +112,90 @@ const requestResponseListener = (msg: MessageEvent<unknown>) => {
...
+  // 2. connection failure
+  // 3. request
+  private get connectionFailure() {
+    // Promise.race checks in order of the list index. so if more than one
+    // promise is settled, Promise.race responds with the earlier index
+    return Promise.race([
+      // rejects with disconnect success, rejects with disconnect failure
+      this._disconnect.promise.then(() => Promise.reject(Error('Disconnected'))),
+      // ignores connection success, rejects with connection failure
+      this._connect.promise.then(() => new Promise<never>(() => null)),
```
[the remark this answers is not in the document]
turbocrime: it's a no-op function. this promise never resolves.

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: these state variables are for tracking promise state synchronously. a promise can only fulfill or reject once, so the variables may only be set once.

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: the call to this promise is voided, so throwing does not affect our own behavior, but unhandled rejections may cause problems elsewhere if they are ignored.

this is a purely internal failure - if this method has been called, we won't provide the connection to any caller again anyway, so the success or failure of these actions is effectively concealed. if we conveyed this rejection to the caller, the rejection would be misleading and meaningless.

so this is used to suppress the failure which won't affect our future behavior, and also log the difference between a failure that can't attempt to disconnect, and a failure that attempted to disconnect but couldn't.

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: this is the helper class. previously all of this behavior was in a flat file.

we can return to that if you prefer.

apps/extension/src/origin.ts
turbocrime: renamed to `knownSitesWithoutDeletedSite`

apps/extension/src/content-scripts/injected-penumbra-global.ts
turbocrime: the `state` function was three lines long and essentially contained all of the synchronous state resolution logic from these three state inputs. it's expanded now in response to comments, but functions the same.

the `connectionFailure` getter contains logic to provide an asynchronous failure state from the promises whose state is tracked by these three state inputs.

was waiting to write documentation until after review, as writing documentation is significant effort and most of my suggested changes to this interface are not accepted.

PR #1202, opened 2024-05-29
fix prax not installed error

### LINKED ISSUE #1191: Install Prax bug
In an incognito window visit https://app.testnet.penumbra.zone/. The web app is not able to detect install status of the extension and requests to connect instead:

Screenshot 2024-05-28 at 5 36 33 PM

The UI should instead display a link to go to the chrome store to download the app.

Valentine1898:
close #1191

Before :
image
<hr>
After:
image

### DIFF
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -11,13 +11,11 @@ import {
   TableRow,
 } from '@penumbra-zone/ui/components/ui/table';
 import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
-import { throwIfPraxNotConnectedTimeout } from '@penumbra-zone/client/prax';
 import { EquivalentValues } from './equivalent-values';
 import { Fragment } from 'react';
 import { shouldDisplay } from './helpers';
 
 export const AssetsLoader: LoaderFunction = async (): Promise<BalancesByAccount[]> => {
-  await throwIfPraxNotConnectedTimeout();
   return await getBalancesByAccount();
 };
 
--- a/packages/client/src/prax.ts
+++ b/packages/client/src/prax.ts
@@ -22,7 +22,7 @@ export class PraxManifestError extends Error {}
 
 export const getPraxPort = async () => {
   const provider = window[PenumbraSymbol]?.[prax_origin];
-  if (!provider) throw new Error('Prax not installed');
+  if (!provider) throw new PraxNotInstalledError('Prax not installed');
   return provider.connect();
 };
 
--- a/packages/client/src/prax.ts
+++ b/packages/client/src/prax.ts
@@ -23,6 +23,9 @@ export class PraxManifestError extends Error {}
 export const getPraxPort = async () => {
   const provider = window[PenumbraSymbol]?.[prax_origin];
   if (!provider) throw new PraxNotInstalledError('Prax not installed');
+
+  await throwIfPraxNotConnectedTimeout();
+
   return provider.connect();
 };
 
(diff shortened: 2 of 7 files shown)


apps/minifront/src/components/dashboard/assets-table/index.tsx
```diff
@@ -11,13 +11,11 @@ import {
...
 } from '@penumbra-zone/ui/components/ui/table';
 import { ValueViewComponent } from '@penumbra-zone/ui/components/ui/tx/view/value';
-import { throwIfPraxNotConnectedTimeout } from '@penumbra-zone/client/prax';
 import { EquivalentValues } from './equivalent-values';
 import { Fragment } from 'react';
 import { shouldDisplay } from './helpers';
 
 export const AssetsLoader: LoaderFunction = async (): Promise<BalancesByAccount[]> => {
-  await throwIfPraxNotConnectedTimeout();
   return await getBalancesByAccount();
```
Valentine1898, line 19: I don't know the exact reason why `throwIfPraxNotConnectedTimeout()` was added to `AssetsLoader`. 

It was this call that caused `PraxNotConnectedError` to be thrown before `PraxNotInstalledError`.

We have no guarantee that the balance load will be earlier than any other prax request, so I see no reason why we shouldn't just remove the `throwIfPraxNotConnectedTimeout()` call

At least I didn't see it break anything

turbocrime: there are several of these calls among other loaders. i tried removing all of them and did see failures. i was unsure why some loaders seem to be executing out of order or in unexpected contexts

will fully review this pr soon

Valentine1898: Ah, missed that there are other loaders that call `throwIfPraxNotConnectedTimeout()`
I'm going to re-watch this PR

packages/client/src/prax.ts
```diff
@@ -22,7 +22,10 @@ export class PraxManifestError extends Error {}
 
 export const getPraxPort = async () => {
   const provider = window[PenumbraSymbol]?.[prax_origin];
-  if (!provider) throw new Error('Prax not installed');
+  if (!provider) throw new PraxNotInstalledError('Prax not installed');
+
+  await throwIfPraxNotConnectedTimeout();
+
   return provider.connect();
```
Valentine1898, line 29: Okay, if we remove `throwIfPraxNotConnectedTimeout()` from all loders and call it inside `getPraxPort()`, that solves all issues 

The only thing I don't really like is that we basically check that prax is connected first, and only then do we call `provider.connect()`
How can we be connected if we never call `connect()`?

turbocrime:
i would have expected the detection to happen in layout loader, before other loaders execute

https://github.com/penumbra-zone/web/blob/1191-install-prax-bug/apps/minifront/src/components/layout.tsx#L25-L30

Valentine1898:
> i would have expected the detection to happen in layout loader, before other loaders execute

But actually calling in other loaders has happened before layout loader

grod220:
I think you stumbled upon a performance feature of react-router. Loaders fire in parallel and thus avoid the waterfall problem of `render->fetch->render->fetch...` that React is known for.

turbocrime:
i have some work on this i will push to a new branch soon

packages/client/src/prax.ts
turbocrime: connection state is available from `isConnected` on the global. connection may resolve and activate before the actual `connect` function is called, if the page is already approved.

apps/minifront/src/components/dashboard/assets-table/index.tsx
turbocrime: loaders without `throwIfPraxNotConnectedTimeout` will block page state until their requests succeed or time out. the throws in `LayoutLoader` were prevented from surfacing the correct error page due to the blocking loaders.

turbocrime:
there are some subtleties to this due to the specified behavior of the injection, and the inability of react loaders to execute in an ordered or dependent way, or propagate errors correctly. i've opened #1208

turbocrime:
closed in favor of #1208

PR #1206, opened 2024-05-29
fix the display of some known assets

### LINKED ISSUE #1172: auction ui: unknown assets errors
Unknown assets should be properly handled and filtered out, otherwise it can throw an error in the extension.

https://github.com/penumbra-zone/web/assets/70081547/3b8cc09a-7695-4705-b615-0eb64f8a420b

Screenshot 2024-05-25 at 10 43 16 AM

@jessepinho

Valentine1898:
close #1172

 This PR also contains other asset display fixes

### DIFF
--- a/apps/minifront/src/components/shared/asset-selector.tsx
+++ b/apps/minifront/src/components/shared/asset-selector.tsx
@@ -127,7 +127,7 @@ export const AssetSelector = ({ assets, onChange, value, filter }: AssetSelector
                     onClick={() => onChange(metadata)}
                   >
                     <AssetIcon metadata={metadata} />
-                    <p className='truncate'>{metadata.symbol}</p>
+                    <p className='truncate'>{metadata.symbol || metadata.display}</p>
                   </div>
                 </DialogClose>
               </div>
--- a/apps/minifront/src/components/swap/helpers.ts
+++ b/apps/minifront/src/components/swap/helpers.ts
@@ -24,6 +24,7 @@ const nonSwappableAssetPatterns = [
   assetPatterns.proposalNft,
   assetPatterns.votingReceipt,
   assetPatterns.auctionNft,
+  assetPatterns.lpNft,
 
   // In theory, these asset types are swappable, but we have removed them for now to get a better UX
   assetPatterns.delegationToken,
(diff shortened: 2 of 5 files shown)


apps/minifront/src/components/swap/helpers.ts
```diff
@@ -24,6 +24,7 @@ const nonSwappableAssetPatterns = [
   assetPatterns.proposalNft,
   assetPatterns.votingReceipt,
   assetPatterns.auctionNft,
+  assetPatterns.lpNft,
```
Valentine1898, line 27: filter out lpnft them out of the asset selectors

packages/ui/components/ui/tx/view/value/index.tsx
```diff
@@ -28,7 +28,7 @@ export const ValueViewComponent = ({
     const { amount = new Amount(), metadata } = view.valueView.value;
     const exponent = getDisplayDenomExponent.optional()(metadata);
     const formattedAmount = formatAmount(amount, exponent);
-    const symbol = metadata.symbol || 'Unknown Asset';
+    const symbol = metadata.symbol || metadata.display;
```
Valentine1898, line 31: We mistakenly displayed assets with undefined symbol as “Unknown assets”
In fact, these assets are known, and this was causing confusion. 

All  IBC assets should have a `symbol` field defined, it is required in cosmos registry 
https://github.com/cosmos/chain-registry/blob/62b3fa019a0db5a277925859999ab430420a3fa6/assetlist.schema.json#L36

But some penumbra internal assets (such as lpnft) don't have this field, so we have to use `display`

TalDerei:
filtered UI looks much cleaner!

TalDerei (approved):

apps/minifront/src/components/shared/asset-selector.tsx
```diff
@@ -127,7 +127,7 @@ export const AssetSelector = ({ assets, onChange, value, filter }: AssetSelector
                     onClick={() => onChange(metadata)}
                   >
                     <AssetIcon metadata={metadata} />
-                    <p className='truncate'>{metadata.symbol}</p>
+                    <p className='truncate'>{metadata.symbol || metadata.display}</p>
```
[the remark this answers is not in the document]
grod220: Yeah, we don't want to show the display here, looks like you address this in https://github.com/penumbra-zone/web/pull/1213

PR #1207, opened 2024-05-29
minifront(gda): fix mistakes in recipe parameters

### LINKED ISSUE #1198: Poisson distribution function is calculating start heights VERY incorrectly
Our [Poisson distribution function](https://github.com/penumbra-zone/web/blob/v8.1.0/apps/minifront/src/state/swap/dutch-auction/get-poisson-distribution.ts) is used to schedule auction start heights semi-evenly across a given span of blocks. But for some reason, the start heights don't match what's expected.

For example, if I start a clean new devnet chain and schedule auctions with 48 sub-auctions (expected to distribute over 48 hours), they have start times from 17h from now to 33d from now.

erwanor:
This should fix #1198, I re-run data exploration, and noticed that I had incorrectly transcribed the parameters for certain auctions (ostensibly the `48h` recipe). This also removes incorrect scaling of the lambda parameter that would cause all the auctions to be very tightly clustered around a single region. We also sample from the distribution a bit differently to make the starting heights strictly increasing. This one is a bit more discretionary, the motivation is that it looks better.

See before/after:

before:
Screenshot 2024-05-29 at 4 52 29 PM

after:
Screenshot 2024-05-29 at 4 52 51 PM
Screenshot 2024-05-29 at 4 56 03 PM

(for the second chart, x = time, y = auction price)


erwanor:
Hmm I think a bug slipped when we ported the pcli logic in #1119 because the logic there differs significantly from what pcli does or what it should do? Maybe it's easier if we pair on this?


```typescript
const getSubAuctionStartHeights = (overallStartHeight: bigint, recipe: GdaRecipe): bigint[] => {
  const startHeights: bigint[] = [];
  const lambda = recipe.poissonIntensityPerBlock;
  let currentHeight = overallStartHeight;

  for (let i = 0n; i < recipe.numberOfSubAuctions; i++) {
    const fastForwardClock = getPoissonDistribution(lambda, Number(recipe.numberOfSubAuctions));
    console.log(`Fast forwarding clock by ${fastForwardClock} blocks`);
    currentHeight += BigInt(Math.ceil(fastForwardClock));
    startHeights.push(currentHeight);
  }

  return startHeights;
};

// ...

export const getPoissonDistribution = (lambda: number, length: number): number => {
  const sampleNextEventTime = () => {
    return Math.abs(Math.log(Math.random()) / lambda);
  };

  let time = 0;
  let nevents = 0;
  while (time < length) {
    const nextEventTime = sampleNextEventTime();
    time += nextEventTime;
    nevents += 1;
    console.log(`Event ${nevents} at time ${time} (ff=${nextEventTime})`);
  }
  return nevents - 1; // last event is after time_period expired
};
```

should be modeled after: https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pcli/src/command/tx/auction/dutch/gda.rs#L139 i.e. `getPoissonDistribution` needs to return a "distance" from the current height sampled from `sampleNextEventTime`

jessepinho (approved):
LGTM. I'll follow up with a small tweak in my next PR, but let's get this merged. Thanks for the work @erwanor !

By the way, I scheduled a 48-hour auction. The final auction was only 30h away (and runs for 2 hours). Is that a normal possibility with a Poisson distribution?

Issue #825, opened 2024-03-25
Create Interface for extension-specific methods


grod220:
We have an objective that eventually other wallets (leap, keplr, etc) will be able to serve as a Penumbra wallet. We already have a few issues with this in mind (https://github.com/penumbra-zone/web/issues/340,  https://github.com/penumbra-zone/web/issues/341).

At the moment, web apps have a pretty simple set of things to utilize a penumbra wallet:

```typescript
// rpc methods
import { createPraxClient } from '@penumbra-zone/client';
export const viewClient = createPraxClient(ViewService);

// extension-specific methods
import { requestPraxConnection, throwIfPraxNotAvailable, throwIfPraxNotInstalled } from '@penumbra-zone/client';
throwIfPraxNotAvailable();
await throwIfPraxNotInstalled();
await requestPraxConnection();
```

These rely upon the extension (ours or others) to implement a few things available globally:
- `isPraxInstalled` requires a specific manifest id exposed in 
- `throwIfPraxNotAvailable` requires the extension to inject a penumbra field on the `window` object
- `requestPraxConnection` expects a specific method on that window penumbra field
- Any rpc method requires specific listeners/connection handlers in the content scripts

All of this we handle for ourselves in [extension/src/content-scripts](https://github.com/penumbra-zone/web/tree/main/apps/extension/src/content-scripts). _Side note:_ I'm not quite sure what would happen if two Penumbra-compatible wallets were installed. 

### Proposal 

Create an interface to define the expectations for the extension serving as a Penumbra wallet. Here's an example interface:

```typescript
interface PenumbraWallet {
  requestPraxConnection(): Promise<void>;
  isPraxConnected(): boolean;
  isPraxInstalled(): boolean;
  throwIfPraxNotInstalled(): void;
  throwIfPraxNotConnected(): void;
  createPraxClient<T extends ServiceType>(serviceType: T): PromiseClient<T>;
}
```

And then perhaps the web app is required to pass it in themselves

```typescript
import { LeapPenumbraWallet } from 'leap';
import { Prax } from 'prax';

import { createClient } from '@penumbra-zone/client';

const leap = createClient(LeapPenumbraWallet)
export const leapViewClient = leap(ViewService);

const prax = createClient(Prax)
export const praxViewClient = prax(ViewService);
```

Another benefit of the interface, is that each method's doc string could inform the extension developer where it needs to go (content scripts, service worker, etc).

turbocrime:
[...]

PR #1208, opened 2024-05-30
fix install bug

### LINKED ISSUE #1191: Install Prax bug

turbocrime:
fixes #1191 

loaders block the entire page until complete, unless they throw locally. loaders were manually calling a connection failure test in order to mimic the layoutloader connection test, but this was incomplete, resulting inaccurate state detection in some situations.

removing those tests resulted in loader use of a client blocking the page, and blocking any thrown error from generating an error page, until the transport timeout. this obscured any early-thrown prax connection failure, and presented the last loader error, which was also misdetected.

so this pr reimplements a more complete temporary solution to fail loaders. ultimately, we should prefer a transport provider and connect-query.

also cleans up the client package a bit, makes the error page ui consistent

some work was also done on the protobuf package, as #1209, it now re-exports service definitions as well as the registry.


turbocrime:
relevant #1202

packages/client/src/prax.ts
```diff
@@ -15,82 +17,99 @@ const prax_id = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe';
...
-  getPraxManifest().then(
-    () => true,
-    () => false,
+export const getPraxManifest = async () => {
+  const json = await fetch(prax_manifest).then(
+    ({ ok, json }) => (ok ? json : Promise.reject(new PraxNotInstalledError())),
+    () => Promise.reject(new PraxNotInstalledError()),
   );
+  return (await json()) as JsonValue;
+};
```
turbocrime, line 52: manifest fetch is used to confirm install. extension asset fetch has a different behavior in incognito mode that was not resolving properly.

packages/client/src/prax.ts
```diff
@@ -15,82 +17,99 @@ const prax_id = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe';
...
 
-export const throwIfPraxNotInstalled = async () => {
-  const isInstalled = await isPraxInstalled();
-  if (!isInstalled) throw new PraxNotInstalledError('Prax not installed');
+export const getPraxPort = async () => {
+  await throwIfPraxNotInstalled();
+  throwIfPraxNotAvailable();
+  return window[PenumbraSymbol]![prax_origin]!.connect();
 };
```
turbocrime, line 58: ideally the page developer should be expected to confirm connection before creating a transport with this passed as `getPort`.

there is some wiggle-room as transport doesn't fail until it passes the configured timeout (default 10s) during which connection may init and succeed this promise.

apps/minifront/src/abort-loader.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Resolves fast if Prax is connected, or rejects if Prax is not connected after
+ * timeout. This is a temporary solution until loaders properly await Prax
+ * connection.
+ *
+ * @param timeout maximum wait in milliseconds (default 500)
+ */
+export const abortLoader = async (timeout = 500): Promise<void> => {
+  await throwIfPraxNotInstalled();
+  (await retry(() => isPraxConnected(), timeout)) || throwIfPraxNotConnected();
+};
```
grod220, line 34: Adds clarity if expanded:

```ts
  const isConnected = await retry(isPraxConnected, timeout);
  if (!isConnected) {
    throwIfPraxNotConnected();
  }
```

apps/minifront/src/abort-loader.ts
```diff
@@ -0,0 +1,34 @@
+import {
+  isPraxConnected,
+  throwIfPraxNotConnected,
+  throwIfPraxNotInstalled,
+} from '@penumbra-zone/client/prax';
+
+const retry = (fn: () => boolean, ms: number) =>
+  new Promise<boolean>(resolve => {
+    if (fn()) resolve(true);
```
grod220, line 9: We should exit function if success right? aka

```ts
    if (fn()) {
      resolve(true);
      return;
    }
```

apps/minifront/src/abort-loader.ts
```diff
@@ -0,0 +1,34 @@
...
+    const interval = setInterval(
+      () => {
+        if (!fn()) return;
+        clearInterval(interval);
+        resolve(true);
+      },
+      Math.max(ms / 10, 50),
```
grod220, line 16: Can you add some documentation about this:

```diff
const retry = (fn: () => boolean, ms: number) => {
+  const retryDelay = Math.max(ms / 10, 50); // do not retry faster than 50ms
...
- Math.max(ms / 10, 50),
+ retryDelay
```

apps/minifront/src/components/extension-not-connected.tsx
```diff
@@ -62,15 +56,15 @@ export const ExtensionNotConnected = () => {
     <>
       <HeadTag />
       <Toaster />
-      <SplashPage title='Welcome to Penumbra'>
+      <SplashPage title='Connect to Penumbra'>
```
grod220, line 59: Probably `Connect to Penumbra Wallet`?

apps/minifront/src/components/extension-not-connected.tsx
```diff
@@ -62,15 +56,15 @@ export const ExtensionNotConnected = () => {
...
         <div className='flex items-center justify-between gap-[1em] text-lg'>
           <div>To get started, connect the Penumbra Chrome extension.</div>
           {!result ? (
-            <Button variant='gradient' onClick={() => void request()} className='px-4'>
```
grod220, line 69: Were there design changes? Can you share screenshots on changes?

apps/minifront/src/components/header/top-row.tsx
```diff
@@ -13,20 +13,18 @@ import { PagePath } from '../metadata/paths';
...
 // Henry (@hdevalence) and thus tied to his Discord account, so reach out to him
 // if there are any problems with this link.
 const WEB_EXT_FEEDBACK_DISCORD_CHANNEL = 'https://discord.gg/XDNcrhKVwV';
 
 export const TopRow = () => {
-  const { isInstalled, isConnected } = useLoaderData() as LayoutLoaderResult;
   const [chainId, setChainId] = useState<string | undefined>();
 
   useEffect(() => {
-    if (isInstalled && isConnected) void getChainId().then(setChainId);
```
grod220, line 28: Is there a reason why we don't need this guard anymore? Is it now impossible to get this component unless already connected?

packages/client/src/prax.ts
grod220, line 52: let's not mix async await and then() syntax here. Also, don't need to do Promise.reject() with async functions. Maybe something like:


```ts
export const getPraxManifest = async (): Promise<JsonValue> => {
  try {
    const response = await fetch(prax_manifest);
    if (!response.ok) {
      throw new PraxNotInstalledError();
    }
    return await response.json() as JsonValue;
  } catch (error) {
    throw new PraxNotInstalledError();
  }
};

```

packages/client/src/prax.ts
grod220, line 1: Related to: https://github.com/penumbra-zone/web/issues/825, but I think this could use some encapsulation:

```ts
export class PraxWallet implements PenumbraWallet {
  transport: Transport | undefined = undefined;

  constructor() {}

  isAvailable(): boolean {}

  isInstalled(): boolean {}

  createWalletClient<T extends ServiceType>(serviceType: T): PromiseClient<T> {
    if (this.transport) return this.transport;
    // else instantiate...
  }

}
```

packages/client/src/prax.ts
```diff
@@ -15,82 +17,99 @@ const prax_id = 'lkpmkhpnhknhmibgnmmhdhgdilepfghe';
...
+
+export class PraxManifestError extends Error {
+  constructor(
+    message = 'Incorrect Prax manifest href',
+    public opts?: ErrorOptions,
+  ) {
+    super(message, opts);
+    this.name = 'PraxManifestError';
+  }
+}
```
grod220, line 115: Can we move these to an error file?

grod220 (commented):
I feel a bit uneasy about the `abortLoader()` pattern. It feels a bit too hacky of a workaround.

Isn't the main thing we want is simply to redirect if the user is not connected to Prax? Loaders can return [redirect responses](https://reactrouter.com/en/main/fetch/redirect). Aka, change all of these:

```diff
- await throwIfPraxNotConnectedTimeout();
+ await redirectIfPraxNotConnectedTimeout(); // 302 response
``` 

which sends it to our connect page. Then the order would not be important and the first to resolve is the first to issue the redirect.

turbocrime:
> I feel a bit uneasy about the `abortLoader()` pattern. It feels a bit too hacky of a workaround.

It's hacky, and it's a workaround, but it is equivalent to the current state of things - calling `throwIfPraxNotConnected` - but it is standardized and complete.
 
> Isn't the main thing we want is simply to redirect if the user is not connected to Prax? Loaders can return [redirect responses](https://reactrouter.com/en/main/fetch/redirect). Aka, change all of these:
> 
> ```diff
> - await throwIfPraxNotConnectedTimeout();
> + await redirectIfPraxNotConnectedTimeout(); // 302 response
> ```
> 
> which sends it to our connect page. Then the order would not be important and the first to resolve is the first to issue the redirect.

ah, okay. does the redirect certainly not block until all loaders complete? this would also require new routes in minifront as a sort of "onboarding flow". which is essentially a new feature so i'm inclined to call this out of scope, this pr is a minimal solution

apps/minifront/src/components/extension-not-connected.tsx
turbocrime: the error pages all had different font sizes and such. i just made styling between all equivalent.

packages/client/src/prax.ts
turbocrime: encapsulation is out of scope here and work is in #1200

apps/minifront/src/components/header/top-row.tsx
turbocrime: this header doesn't appear on the connection error pages, which should be the only thing that appears if it's not possible to connect to the extension.

grod220 (approved):
Once we move away from loaders, the `abortLoader()` can get scrapped. However, this does illuminate an issue with our API. Other consumers will need to have a way to cleanly await and definitively know whether the extension is installed or not.

turbocrime:
>Other consumers will need to have a way to cleanly await and definitively know whether the extension is installed or not.

this is surfaced by the page global, but the loaders as written expect to create and use a client without respecting that status.

PR #1209, opened 2024-05-30
modularize protobuf package

turbocrime:
work related to #1208

mostly internal organization

significant change is this adds a re-export of all service definitions, which simplifies imports anywhere clients are constructed.


packages/protobuf/src/index.ts
turbocrime, line 1: this is a barrel import, but everything re-exported here is also used in `typeRegistry`, the intended main export of the package. the independent components are also re-exported for convenience of consumers who likely want to access the service definitions &c

packages/protobuf/src/penumbra-core.ts
```diff
@@ -0,0 +1,12 @@
+export { QueryService as AppService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/core/app/v1/app_connect';
+
+export { QueryService as AuctionService } from '@buf/penumbra-zone_penumbra.connectrpc_es/penumbra/core/component/auction/v1/auction_connect';
```
turbocrime, line 3: `DutchAuction` was removed as a bare type as it's included in `AuctionService`

packages/protobuf/README.md
```diff
@@ -1,41 +1,44 @@
...
 import { jsonOptions } from '@penumbra-zone/protobuf';
-
 import { createChannelTransport } from '@penumbra-zone/transport-dom';
-import type { ServiceType } from '@bufbuild/protobuf';
 
-// unsafely get first available provider
-const getPort = () => Object.values(window[Symbol.for('penumbra')])[0].connect();
+// naively get first available provider
+const provider = Object.values(window[Symbol.for('penumbra')])[0];
+void provider.request();
```
grod220, line 37: Shouldn't they use client package for this?

packages/query/src/root-querier.ts
```diff
@@ -16,7 +16,7 @@ export class RootQuerier implements RootQuerierInterface {
   readonly tendermint: TendermintQuerier;
   readonly shieldedPool: ShieldedPoolQuerier;
   readonly ibcClient: IbcClientQuerier;
-  readonly staking: StakingQuerier;
+  readonly staking: StakeQuerier;
```
grod220, line 19: Should be renamed too?

```diff
- readonly staking: StakeQuerier;
+ readonly stake: StakeQuerier;
```

packages/services/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@penumbra-zone/services",
   "version": "3.2.1",
+  "private": true,
```
grod220, line 4: No need to publish anymore?

grod220 (approved):

.changeset/little-masks-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@penumbra-zone/protobuf': patch
+---
```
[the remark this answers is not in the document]
turbocrime: the protobuf package change adds new exports, but doesn't change existing exports. so those are a patchbump

after making this changeset file, i updated users of the package to use the new exports. that does need a new changeset with new version bumps

packages/protobuf/README.md
turbocrime: most developers will. this example is demonstrating use of the type registry this package, not the client package.

a section above this in the readme instructs, 

>If you are looking for a Penumbra extension client
>
>You should install `@penumbra-zone/client`. This package is provided for
developers interested in lower-level work or more detailed configuration.

.changeset/little-masks-watch.md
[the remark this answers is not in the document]
grod220: We don't need another PR to change this, but new features in a backward compatible way would be a minor: https://semver.org/

PR #1213, opened 2024-05-30
customize ICS20 assets symbol

Valentine1898:
Discussion regarding this PR: https://discord.com/channels/824484045370818580/1044457038886998137/1245709532701917336

image

### DIFF
--- a/packages/services/src/view-service/asset-metadata-by-id.ts
+++ b/packages/services/src/view-service/asset-metadata-by-id.ts
@@ -1,6 +1,6 @@
 import type { Impl } from '.';
 import { servicesCtx } from '../ctx/prax';
-import {assetPatterns} from "@penumbra-zone/types/assets";
+import { assetPatterns } from '@penumbra-zone/types/assets';
 
 export const assetMetadataById: Impl['assetMetadataById'] = async ({ assetId }, ctx) => {
   if (!assetId) throw new Error('No asset id passed in request');
--- a/packages/services/src/view-service/asset-metadata-by-id.ts
+++ b/packages/services/src/view-service/asset-metadata-by-id.ts
@@ -1,5 +1,6 @@
 import type { Impl } from '.';
 import { servicesCtx } from '../ctx/prax';
+import {assetPatterns} from "@penumbra-zone/types/assets";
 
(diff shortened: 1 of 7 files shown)


Valentine1898:
@grod220 
We have two types of unknown assets 
1. those for which we do not have `metadata`. We display them like this:
image
2. those for which we have a `metadata`, but in which the `symbol` value is not defined. These are, for example, ibc assets whose metadata is generated on the node, LPNFTs. We want to display them like this:
image

If we do not store ibc metadata that are not in the registry at the block processor layer as you suggest, then we will display them as the first type of unknown assets

If we want to display them as the second type of assets, we can either do it in components as before:
```typescript
 <p className='truncate'>{metadata.symbol || 'Unknown asset'}</p>
``` 
Or do metadata customization, but instead of `Unknown IBC asset (channel-2)` just `Unknown asset`

But it is also worth considering whether we need different types of unknown assets

grod220:
> those for which we do not have metadata. We display them like this

Ah, thanks for finding this. It feels like a miss that we handle these two cases differently. They both mean the same thing, "we don't know what this is". Also, I find the `passet` display denom confusing. Think we should harmonize both of these and just display "Unknown Asset" in both cases.

Valentine1898:
Ready to review, but will hold off until RPC is available

grod220 (approved):

packages/services/src/view-service/asset-metadata-by-id.ts
```diff
@@ -16,7 +17,10 @@ export const assetMetadataById: Impl['assetMetadataById'] = async ({ assetId },
   if (localMetadata) return { denomMetadata: localMetadata };
 
   const remoteMetadata = await querier.shieldedPool.assetMetadataById(assetId);
-  if (remoteMetadata) {
+
+  const isIbcAsset = remoteMetadata && assetPatterns.ibc.matches(remoteMetadata.display);
+
+  if (remoteMetadata && !isIbcAsset) {
```
Valentine1898, line 23: It turns out that in addition to the block processor, we also request metadata from the node in the `assetMetadataById` rpc. 
This doesn't seem right to me, but I will create a separate issue for this, since this PR is blocking the release

Issue #1183, opened 2024-05-28
Candlestick chart on estimate


grod220:
With https://github.com/penumbra-zone/penumbra/pull/4398, we now have a [Candlestick data](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1#penumbra.core.component.dex.v1.CandlestickData) that can be requested from the frontend.

This will be helpful in augmenting our swap & auction estimate component. After an estimate request is made, we should make a `CandlestickDataRequest`, transform the data as needed for our charting library, and add to the estimate view:

<img width="912" alt="Screenshot 2024-05-28 at 12 47 32 PM" src="https://github.com/penumbra-zone/web/assets/16624263/ff854a1d-d90c-4067-a6db-652c214ad160">

Libraries to consider

- visx has boxplot support: https://airbnb.io/visx/docs/stats
- chartjs has support: https://www.chartjs.org/chartjs-chart-financial/
- others? d3js?

grod220:
@turbocrime can you update the status on this?

turbocrime:
candlestick chart did merge in #1220, but is issued as broken by https://github.com/penumbra-zone/web/issues/1347. i started work on that but was instructed to focus on page api and react wrapper packages which have been through several refactor/wait/review iterations since then, with all available other attention taken by other high priority things

and then all the testnet infra was taken down, and the `penumbra-1` chain afaict has no swap activity to examine charts for. this is essentially impossible to work on without a chain containing swap data.

i rebased and spent some time with a local dev net today, i'll open a PR with my rebased work.

PR #1220, opened 2024-05-31
Candlesticks

turbocrime:
minimal shot at #1183

uses visx graph library. loads graph on selection of trading pair, renders in selection component.


hdevalence:
I don't think bundle size is a concern

turbocrime:
these are basically done, wrapping it up this evening i think

apps/extension/src/rpc/rethrow-impl-errors.ts
turbocrime, line 1: these debug logs are conditional and compile out in prod

package.json
```diff
@@ -9,6 +9,7 @@
     "buf-update": "pnpm update --latest \"@buf/*\" \"@bufbuild/*\" \"@connectrpc/*\" && pnpm syncpack fix-mismatches && pnpm install",
     "build": "turbo build",
     "clean": "turbo clean",
+    "clean-modules": "rm -rf node_modules apps/*/node_modules packages/*/node_modules pnpm-lock.yaml",
```
turbocrime, line 12: bonus script

packages/eslint-config/eslint.config.mjs
turbocrime, line 1: reactHooksPlugin crashes when linting the new candlesticks.tsx

packages/transport-dom/src/proxy.ts
turbocrime, line 1: This was a feature that got dropped at some point in the implementation of proxy.ts, it's used now to make TendermintProxyService safe

apps/extension/src/rpc/rethrow-impl-errors.ts
```diff
@@ -12,10 +12,12 @@ const wrapUnaryImpl =
       const result = methodImplementation(req, ctx);
       if (result instanceof Promise)
         return result.catch((e: unknown) => {
+          if (process.env['NODE_ENV'] === 'development') console.debug(ctx.method.name, req, e);
```
grod220, line 15: At first I thought the throw was within the if clause, can you add brackets to these?

apps/extension/src/rpc/tendermint-proxy.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * We should explore PRing bufbuild or monkey-patching the serialization.
+ */
+export const makeTendermintProxyZeroNanos = (
+  c: PromiseClient<typeof TendermintProxyService>,
+): Pick<ServiceImpl<typeof TendermintProxyService>, 'getBlockByHeight'> => ({
+  getBlockByHeight: async req => {
+    const r = await c.getBlockByHeight(req);
+    if (r.block?.header?.time?.nanos) r.block.header.time.nanos = 0;
+    if (r.block?.lastCommit?.signatures.length) r.block.lastCommit.signatures = [];
```
grod220, line 27: Can you add brackets? We should reserve no-bracket conditionals only for things that are one line and exit the function (return/throw).

apps/minifront/src/state/swap/index.ts
```diff
@@ -95,4 +110,50 @@ export const createSwapSlice = (): SliceCreator<SwapSlice> => (set, get, store)
...
+      assetOut,
+      candlestick: { loading, abort: abortOldLoad },
+    } = get().swap;
+
+    if (loading) abortOldLoad();
+
+    if (!height)
+      set(({ swap }) => {
+        swap.candlestick.data = [];
+      });
```
grod220, line 133: Please add brackets to these

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -42,7 +51,18 @@ export const TokenSwapInput = () => {
...
     setAssetOut,
     balancesResponses,
+    candlestickData,
+    loadCandlestick,
+    latestKnownBlockHeight,
   } = useStoreShallow(tokenSwapInputSelector);
+
+  useEffect(() => {
+    if (!assetIn || !assetOut) return;
+    const ac = new AbortController();
```
grod220, line 61: Are candlestick requests very expensive? Aka, is there any meaningful downside to excluding the abort controller stuff for the sake of simpler code?

packages/eslint-config/eslint.config.mjs
grod220, line 1: Can you provide context for the lint changes?

grod220 (approved):
Mostly brackets feedback. Otherwise LGTM.

packages/eslint-config/eslint.config.mjs
turbocrime: eslint is configured wrong. react-hooks and react plugins don't support v9, and need more work to run properly. try re-enabling it and then linting the project- eslint will crash.

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
turbocrime: they can be many thousands of records. right now it's fixed, but eventually the size will be configurable. the abortcontroller was previously in the slice, but moved here as i needed some more detailed control - i could probably move it back into the slice

packages/types/src/environment.ts
```diff
@@ -13,7 +13,7 @@ enum Environment {
 declare global {
   interface Window {
     // This file has many environments running it. Makes it hard to please every runtime.
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
```
turbocrime, line 17: not sure why the linter started complaining about this file that i didn't edit

PR #1222, opened 2024-05-31
proving keys v0.77.0

turbocrime:


TalDerei (approved):

TalDerei:
let's ship it, this should be the last time proving keys change

Issue #889, opened 2024-04-05
Improve swap view rendering


hdevalence:
Improvements to rendering for swap / swap claim views:

- [x] Instead of "Fee", both Swap and SwapClaim views should refer to it as "Prepaid Claim Fee" to distinguish it from the transaction fee.

~~The "public view" of Swap and SwapClaim is extremely misleading, to the point of lying to users (though this isn't our intent).~~ (EDIT: this was fixed between when I last pulled code and when I filed the issue, amazing!)

- [x] The swap inputs and outputs need to be shown in the public view for a `Swap`. The inputs are public because they are in the clear. The outputs are also known because they can be computed by anyone with the BSOD.
- [x] The prepaid claim fee should be shown only in the user's view and not the public view, because it is not publicly revealed in the Swap.
- [x] The swap inputs and outputs should _not_ be shown in the public view for a `SwapClaim`. The output notes are minted privately and the claim doesn't reveal which swap commitment it claimed.
- [x] The prepaid claim fee should be shown in the public view for a `SwapClaim`. This is because it is revealed publicly (and this is the reason it is _not_ revealed in the Swap).

TalDerei:
Tracked by https://github.com/penumbra-zone/web/issues/875

PR #1228, opened 2024-05-31
views: rendering opaque swap view

TalDerei:
References #889 and partially #1251. 

Fixes the unimplemented opaque views for swaps / swap claims to render the proper information. 

Code deduplication in the transaction view page is achieved by hooking into the existing [translator](https://github.com/penumbra-zone/web/tree/main/packages/perspective/translators) logic, enabling the `visible` view to be transformed into an `opaque` view using the _same_ code for both the approval dialogue and transaction view. 

Receiver view in the transaction view page will be handled in https://github.com/penumbra-zone/web/pull/1260.


packages/getters/src/swap-claim-view.ts
```diff
@@ -14,5 +14,17 @@ export const getOutput2 = createGetter((swapClaimView?: SwapClaimView) =>
...
+);
+
 export const getOutput1ValueOptional = getOutput1.optional().pipe(getValue);
 export const getOutput2ValueOptional = getOutput2.optional().pipe(getValue);
+
+export const getSwapClaimFeeOpaque = createGetter((swapClaimView?: SwapClaimView) =>
+  swapClaimView?.swapClaimView.case === 'opaque'
+    ? swapClaimView.swapClaimView.value.swapClaim?.body?.fee
+    : undefined,
+);
```
grod220, line 30: Can you combine these?

```ts
export const getSwapClaimFee = createGetter((swapClaimView?: SwapClaimView) =>
  swapClaimView?.swapClaimView.case === 'visible' || swapClaimView?.swapClaimView.case === 'opaque'
    ? swapClaimView.swapClaimView.value.swapClaim?.body?.fee
    : undefined,
);

```

packages/getters/src/swap-plaintext.ts
```diff
@@ -1,6 +1,11 @@
...
+  SwapBody,
+  SwapPlaintext,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/dex/v1/dex_pb';
 import { createGetter } from './utils/create-getter';
 
 export const getDelta1I = createGetter((swapPlaintext?: SwapPlaintext) => swapPlaintext?.delta1I);
 export const getDelta2I = createGetter((swapPlaintext?: SwapPlaintext) => swapPlaintext?.delta2I);
 export const getClaimFee = createGetter((swapPlaintext?: SwapPlaintext) => swapPlaintext?.claimFee);
+export const getDelta1IOpaque = createGetter((swapBody?: SwapBody) => swapBody?.delta1I);
+export const getDelta2IOpaque = createGetter((swapBody?: SwapBody) => swapBody?.delta2I);
```
grod220, line 11: Getter inputs and the file names should match. This means these should move to a file like:

```
packages/getters/src/swap-body.ts
```

grod220: Also, we should remove `Opaque` from the name

packages/getters/src/swap-view.ts
```diff
@@ -36,6 +42,31 @@ export const getClaimTx = createGetter((swapView?: SwapView) =>
...
+export const getDelta2IFromSwapViewOpaque = getSwapPlaintextOpaque.pipe(getDelta2IOpaque);
+
+export const getOutput1Opaque = createGetter((swapView?: SwapView) =>
+  swapView?.swapView.case === 'opaque' ? swapView.swapView.value.output1Value! : undefined,
+);
+export const getOutput2Opaque = createGetter((swapView?: SwapView) =>
+  swapView?.swapView.case === 'opaque' ? swapView.swapView.value.output2Value : undefined,
+);
+
+export const getOutput1ValueOpaqueView = getOutput1Opaque.optional().pipe(getValueOpaque);
```
grod220, line 58: Unused

packages/getters/src/swap-view.ts
grod220, line 1: Hmm. I am struggling with the opaque-specific getters here 🤔 

Could we change all of these to work generically and in the component logic, we do the guarding there to ensure the thing is opaque?

packages/perspective/transaction/perspective.ts
grod220, line 1: Nice job here 🧹

packages/types/src/swap.ts
```diff
@@ -100,3 +113,59 @@ export const getOneWaySwapValues = (
...
+ * amount. This function is useful for displaying a swap in a UI, and it assumes
+ * we're dealing with a one-way swap. If passed a two-way swap, it will throw.
+ */
+export const getOneWaySwapValuesOpaque = (
+  swapView: SwapView,
+): {
+  input: ValueView;
+  output: ValueView;
+} => {
+  if (!isOneWaySwapOpaque(swapView)) {
```
grod220, line 128: Ah here's the guard. I think this gives you license to use generic getters below that work for both opaque and visible cases.

packages/ui/components/ui/tx/view/swap/index.tsx
```diff
@@ -58,7 +62,24 @@ export const SwapViewComponent = ({ value }: { value: SwapView }) => {
   }
 
   if (value.swapView.case === 'opaque') {
-    return <UnimplementedView label='Swap' />;
+    const oneWaySwap = isOneWaySwapOpaque(value) ? getOneWaySwapValuesOpaque(value) : undefined;
+    if (!isOneWaySwapOpaque(value)) {
```
grod220, line 66: We can do this right?

```ts
    const oneWaySwap = isOneWaySwapOpaque(value) ? getOneWaySwapValuesOpaque(value) : undefined;
    if (!oneWaySwap) {
```

packages/ui/components/ui/tx/view/swap/index.tsx
```diff
@@ -58,7 +62,24 @@ export const SwapViewComponent = ({ value }: { value: SwapView }) => {
...
+      throw new Error(
+        'Attempted to get one-way swap values from a two-way swap. `getOneWaySwapValues()` should only be called with a `SwapView` containing a one-way swap -- that is, a swap with at least one `swapPlaintext.delta*` that has an amount equal to zero.',
+      );
```
grod220, line 69: Should we throw an error? This will crash the UI. How about we show an `Unimplemented` line of text. Looks like you are doing that here? 
https://github.com/penumbra-zone/web/pull/1228/files#diff-281812fff2a966e48bcea90f67cb77e3ca74d0d6ba16044ed78f49db1abbc999R78

If so, it doesn't seem we need to throw.

packages/getters/src/swap-claim-view.ts
TalDerei: good call, combined getters

packages/getters/src/swap-view.ts
TalDerei: removed

packages/ui/components/ui/tx/view/swap/index.tsx
TalDerei: that seems to work as well!

packages/getters/src/swap-view.ts
TalDerei: I think we may be able to use generic getters here, but it would require using explicit type casting in places, which seems fine. I will think through a design and modify the component logic to be generic.

packages/ui/components/ui/tx/view/swap/index.tsx
TalDerei: good call, removed as it will be correctly caught as you pointed out.

packages/getters/src/swap-plaintext.ts
TalDerei: moved to dedicated file

packages/getters/src/swap-view.ts
TalDerei: see latest commit where generic getters were implemented

TalDerei:
swaps / swap claim views are now using _generic_ getters, where a generic getter is defined as being agnostic to the `visible` or `opaque` nature of the view. The generics have strict type guards to check that the returned values are correctly typed.

packages/types/src/swap.ts
```diff
@@ -83,20 +81,16 @@ export const getOneWaySwapValuesGeneric = (
     },
   });
 
-  let output = (isZero(delta2I) ? output2 : output1) as ValueView;
```
TalDerei: @jessepinho can you sanity check that removing this logic inside the `getOneWaySwapValuesGeneric` function in fact correct? I was trying to satisfy the linter, and swap views still seem to function properly.

grod220:
Let's grab some time to pair 🍐 on this. Will DM!

packages/getters/src/swap-view.ts
```diff
@@ -25,56 +29,48 @@ export const getOutput2Value = createGetter((swapView?: SwapView) => {
...
-//         return swapView.swapView.value.swap?.body?.[property as keyof SwapBody] as T[K];
-//       default:
-//         return undefined;
-//     }
-//   });
-// };
+// Generic getter function that returns the value of a specified property from either 'swapPlaintext' or 'swapBody'.
+// This pattern utilizes parameterized types, 'T', to handle property access, 'K' within different nested objects based on the case
+// of 'SwapView'. These parameterized types can represent an intersection of multiple types.
+const createSwapGetter = <T, K extends keyof T>(property: K) => {
```
TalDerei, line 35: I modified to getter to ensure an `Amount` is returned to the caller. I quite like this pattern, and maybe it's extensible to other getter functions in the future.

packages/types/src/swap.ts
```diff
@@ -87,8 +87,6 @@ export const getOneWaySwapValues = (
 
   let output = isZero(delta2I) ? output2 : output1;
 
-  // TODO: Why is this needed?
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
TalDerei: fixing the getters naturally resolved this

packages/getters/src/swap-view.ts
grod220: Unfortunately, this re-introduces the bug 🐛 with `claimFee` in the opaque case.

Using `as T[K]` is basically saying that we guarantee every field of `SwapPlaintext` and `SwapBody` are the same. But it then allows for human error. This helper then only becomes useful if the consumer will only use it in the correct way.

In short, the abstraction is not worth it. Think we should revert to previous version.

TalDerei: would it? wouldn't the opaque case just return undefined because there isn't a `claimFee` field?

grod220 (approved):
Added one last comment, your call though 👍

packages/getters/src/swap-view.ts
TalDerei: that makes sense! yeah, the parameterization `T` represents an _intersection_ rather than a union. I think keeping this pattern is a good reference for eventually overhauling other getters to be generic over the `visible` and `opaque` views down the line 👍

TalDerei: I'll open an issue to add more detailed comments to our getters package

PR #1234, opened 2024-05-31
`formatAmount()` updated args [bug fix]

grod220:
For higher-precision decimal numbers, we were getting a `NaN` showing in the UI:

Screenshot 2024-05-31 at 6 35 48 PM

Turns out because we were using `formatNumber()` sometimes returns a comma separated string which cannot be put inside the `new BigNumber()` constructor. This PR changes the API for that function to be explicit with commas usage, which fixes it:
Screenshot 2024-05-31 at 6 39 14 PM

===

Extracted from https://github.com/penumbra-zone/web/pull/1151. Merging this will make that PR smaller.

### DIFF
--- a/apps/minifront/src/components/swap/swap-form/simulate-swap-result/trace/price.tsx
+++ b/apps/minifront/src/components/swap/swap-form/simulate-swap-result/trace/price.tsx
@@ -2,7 +2,7 @@ import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/
 import { SwapExecution_Trace } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/dex/v1/dex_pb';
 import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
 import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
-import { formatAmount } from '@penumbra-zone/types/amount';
+import { formatAmount, removeTrailingZeros } from '@penumbra-zone/types/amount';
 import { BigNumber } from 'bignumber.js';
 
 export const Price = ({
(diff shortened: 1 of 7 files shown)


apps/minifront/src/components/swap/swap-form/simulate-swap-result/trace/price.tsx
```diff
@@ -23,14 +23,22 @@ export const Price = ({
...
+        amount: outputValue.amount,
+        exponent: outputDisplayDenomExponent,
+      });
 
       const outputToInputRatio = new BigNumber(formattedOutputAmount)
         .dividedBy(formattedInputAmount)
         .toFormat(outputDisplayDenomExponent);
 
-      price = `1 ${firstValueMetadata.symbol} = ${outputToInputRatio} ${lastValueMetadata.symbol}`;
+      const outputToInputFormatted = removeTrailingZeros(outputToInputRatio);
```
grod220, line 39: Removing trailing zeros from the format so we don't have outputs like:

Screenshot 2024-05-31 at 6 29 23 PM

packages/types/src/amount.ts
```diff
@@ -77,10 +77,28 @@ export const formatNumber = (number: number, options: FormatOptions): string =>
...
-    .toFormat(6)
-    .replace(/(\.\d*?[1-9])0+$|\.0*$/, '$1');
+export const removeTrailingZeros = (strNum: string): string => {
+  return strNum.replace(/(\.\d*?[1-9])0+$|\.0*$/, '$1');
+};
+
+export const formatAmount = ({
+  amount,
+  exponent = 0,
+  commas = false,
```
grod220, line 87: I change the comma default to be `false`. Up for discussion if reviewer feels it should be opposite. Potentially safer this way (?).

TalDerei (approved):
LGTM

PR #1248, opened 2024-06-04
fix: #1241: Fix destination validation in the ibc-out form

### LINKED ISSUE #1241: IBC-out addresses incorrectly validating
It appears the boolean is opposite? 

Should work:
Screenshot 2024-06-03 at 12 45 48 PM

Should not work:
Screenshot 2024-06-03 at 12 45 54 PM

VanishMax:
Closes #1241 

I noticed the tests were skipped in the `ibc-out.test.ts` file. I un-skipped the passing tests and added a comment for the future to fix the broken tests

### DIFF
--- a/apps/minifront/src/state/ibc-out.test.ts
+++ b/apps/minifront/src/state/ibc-out.test.ts
@@ -110,6 +110,17 @@ describe('IBC Slice', () => {
 
       expect(validationErrors.recipientErr).toBeFalsy();
     });
+
+    test('destination address validation per selected chain fails with incorrect address', () => {
+      const osmoAddress = 'osmo1xxxxxx';
+
+      useStore.getState().ibcOut.setChain(chain);
+      useStore.getState().ibcOut.setDestinationChainAddress(osmoAddress);
+
+      const validationErrors = ibcValidationErrors(useStore.getState());
+
+      expect(validationErrors.recipientErr).toBeTruthy();
+    });
   });
 
   describe('setSelection', () => {
--- a/apps/minifront/src/state/ibc-out.test.ts
+++ b/apps/minifront/src/state/ibc-out.test.ts
@@ -12,9 +12,9 @@ import { produce } from 'immer';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
 import { addressFromBech32m } from '@penumbra-zone/bech32m/penumbra';
 import { Chain } from '@penumbra-labs/registry';
-import { currentTimePlusTwoDaysRounded } from './ibc-out';
+import { currentTimePlusTwoDaysRounded, ibcValidationErrors } from './ibc-out';
 
-describe.skip('IBC Slice', () => {
+describe('IBC Slice', () => {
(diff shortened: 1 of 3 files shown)


apps/minifront/src/state/ibc-out.ts
```diff
@@ -199,15 +199,16 @@ export const ibcValidationErrors = (state: AllSlices) => {
...
+      : validateUnknownAddress(state.ibcOut.chain, state.ibcOut.destinationChainAddress),
     amountErr: !state.ibcOut.selection
       ? false
       : amountMoreThanBalance(state.ibcOut.selection, state.ibcOut.amount),
   };
 };
 
 /**
- * we don't know what format foreign addresses are in. so this only checks:
+ * Matches the given address to the chain's address prefix. Returns `true` if the address is not valid.
```
grod220, line 210: Hmm, think there could be a bug in `validateUnknownAddress`. It should return `true` if the address _is_ valid. That is consistent with the first line:

```
if (!chain || address === '') return false; // not valid = false
```

We should rename this function to: `unknownAddrIsValid()` to add clarity.

VanishMax: Right, updated!

apps/minifront/src/state/ibc-out.test.ts
```diff
@@ -85,19 +86,30 @@ describe.skip('IBC Slice', () => {
...
+    test('destination address validation per selected chain', () => {
+      const osmoAddress = 'osmo1dyrr4r42ql4em7d46srcmnn5ymxk9asvcv95sg';
+
+      useStore.getState().ibcOut.setChain(chain);
+      useStore.getState().ibcOut.setDestinationChainAddress(osmoAddress);
+
+      const validationErrors = ibcValidationErrors(useStore.getState());
+
+      expect(validationErrors.recipientErr).toBeFalsy();
+    });
```
grod220, line 112: Can you add a test that triggers the validation error? Aka

```ts
   test('invalid address', () => {
      const osmoAddress = 'osmo1xxxxx';

      useStore.getState().ibcOut.setChain(chain);
      useStore.getState().ibcOut.setDestinationChainAddress(osmoAddress);

      const validationErrors = ibcValidationErrors(useStore.getState());

      expect(validationErrors.recipientErr).toBeTruthy();
    });
```

apps/minifront/src/state/ibc-out.ts
```diff
@@ -199,23 +199,24 @@ export const ibcValidationErrors = (state: AllSlices) => {
...
     amountErr: !state.ibcOut.selection
       ? false
       : amountMoreThanBalance(state.ibcOut.selection, state.ibcOut.amount),
   };
 };
 
 /**
- * we don't know what format foreign addresses are in. so this only checks:
+ * Matches the given address to the chain's address prefix.
+ * We don't know what format foreign addresses are in. so this only checks:
```
grod220, line 211:
```diff
- * We don't know what format foreign addresses are in. so this only checks:
+ * We don't know what format foreign addresses are in, so this only checks:
```

grod220 (approved):
Good job! A few more comments and then good to merge.

Issue #1225, opened 2024-05-31
Clicking the arrow in the trade widget should reverse assets


hdevalence:
**Is your feature request related to a problem? Please describe.**

It's annoying to have to select two assets from a list just to reverse a direction.

**Describe the solution you'd like**

<img width="827" alt="Screenshot 2024-05-30 at 11 56 30 PM" src="https://github.com/penumbra-zone/web/assets/44879/7cfce589-759b-44d5-9871-630f9dffaa54">

Clicking this button should swap the asset selections. Maybe to hint to the user that this is possible, we replace it with a reversed arrow on hover.

grod220:
Our account model makes this trickier than other protocols. In short, our input selector requires A) tokens the user already has and B) selecting which account. Valentine had a writeup here: https://github.com/penumbra-zone/web/issues/972.

Before we could add reverse, we'd have to do figure out this issue first: https://github.com/penumbra-zone/web/issues/1211, aka, how should our selector work without the account/balance restriction?

Issue #1211, opened 2024-05-30
Remove restriction input asset


grod220:
If you take a swapper UI like say, Sushi, there are no restrictions on A) the asset you can select and B) the amount you type in. Note the wallet balance in the component:

<img width="526" alt="Screenshot 2024-05-30 at 2 17 30 PM" src="https://github.com/penumbra-zone/web/assets/16624263/c8b40704-d247-4433-9165-39c4c8bd626a">

This is helpful in cases where a user wants to simulate trades for various pairs (price shopping) without actually having the asset (or enough of it) in their wallet.

Our current UI is:
- Only displaying _swap in_ assets that is in the user's wallet
- Prevents key presses for amounts greater than amount in their wallet

We should remove these restrictions:
- Pull assets (likely not delegation tokens?) from the registry and display them in swap-in
- Remove html key press validation
- Add validations in our slices and display errors like we do on the send form

This will require some rethinking on the asset select form:

<img width="494" alt="Screenshot 2024-05-30 at 2 22 05 PM" src="https://github.com/penumbra-zone/web/assets/16624263/f892ed76-c14a-47c9-a763-4f6e2ce33ee2">

Note how it displays account number. We'd need a way to handle cases where the user has none of the token and where they have the token in specific accounts. This is tricky, likely needs Sam input/support.

PR #1249, opened 2024-06-04
Properly format if no metadata

grod220:
Reported by @VanishMax

Follow up from https://github.com/penumbra-zone/web/pull/1151

When an asset is known, but has no metadata, we should also be able to format correctly (at the moment, an error bubbles up blocking page view). Added tests as well to cover this case.

### DIFF
--- a/.changeset/shaggy-knives-protect.md
+++ b/.changeset/shaggy-knives-protect.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/types': patch
+---
+
+Bug fix in getFormattedAmtFromValueView: support known assets without metadata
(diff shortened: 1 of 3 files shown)


Valentine1898 (approved):
LGTM

VanishMax:
@grod220 With no assets in the account, the trade block now renders like this:

image

grod220:
This relates to https://github.com/penumbra-zone/web/issues/1211 and https://github.com/penumbra-zone/web/issues/1225. At the moment, our UI does not support in/out assets from the perspective of an account with no assets. Definitely something we should prioritize fixing in an upcoming sprint.

PR #1250, opened 2024-06-04
fix: #1195: Hide the redundant scroll caused by long addresses

### LINKED ISSUE #1195: Delegation approval width bug 🐜
When making a delegation, the width of the approval window gets so large it requires horizontal scroll. Perhaps we need to make the validator identity key something that is truncated or has a wordbreak?

https://github.com/penumbra-zone/web/assets/16624263/44d49c15-7d06-4952-babf-c3dc6e860773

VanishMax:
Closes #1195 

Screenshot of the problem:

image

### DIFF
--- a/.changeset/old-candles-compete.md
+++ b/.changeset/old-candles-compete.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/ui': patch
+---
+
+fix delegation prompting window being too wide
--- a/packages/ui/components/ui/box.tsx
+++ b/packages/ui/components/ui/box.tsx
@@ -34,19 +34,21 @@ export const Box = ({
   layout,
   layoutId,
   headerContent,
+  className,
 }: PropsWithChildren<
   VariantProps<typeof variants> & {
     label?: ReactNode;
     layout?: boolean;
     layoutId?: string;
     headerContent?: ReactNode;
+    className?: string;
   }
 >) => {
   return (
     <motion.div
       layout={layout ?? !!layoutId}
       layoutId={layoutId}
-      className={cn('flex flex-col gap-4', variants({ spacing, state }))}
+      className={cn('flex flex-col gap-4', className, variants({ spacing, state }))}
       /**
        * Set the border radius via the style prop so it doesn't get distorted by framer-motion.
        *
--- a/packages/ui/components/ui/box.tsx
+++ b/packages/ui/components/ui/box.tsx
@@ -16,6 +16,9 @@ const variants = cva('rounded-lg border bg-background', {
       default: '',
       error: 'border-red',
     },
+    overflow: {
+      xHidden: 'overflow-x-hidden',
+    },
   },
   defaultVariants: {
     spacing: 'default',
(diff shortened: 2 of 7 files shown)


Valentine1898:
Can you add before and after screenshots to the PR? 
This would be very helpful for reviewers when it comes to UI changes

packages/ui/components/ui/tx/view/viewbox.tsx
```diff
@@ -17,7 +17,7 @@ export const ViewBox = ({ label, visibleContent, isOpaque }: ViewBoxProps) => {
   // if isOpaque is undefined, set it to !visibleContent
   isOpaque = isOpaque ?? !visibleContent;
   return (
-    <Box>
+    <Box className='overflow-hidden'>
```
grod220, line 20: This is tricky, but we need a way to display the full validator address here (without the window extending beyond the view of the screen). It will be important that users can validate this on this screen. Perhaps it's something that can be clicked to expand?

Screenshot 2024-06-04 at 1 19 37 PM

VanishMax: @grod220 @Valentine1898 Thanks for the comments. I added a picture of how it was before to the description of this PR. This time, I removed the `truncate` prop from the ActionDetails.Row component and created the ActionDetails.TruncatedText component that renders long text accessibly with keyboard navigation support.


https://github.com/penumbra-zone/web/assets/29180358/79cbaa59-67d1-46ce-8237-35c1ecc68851

TalDerei:
@VanishMax just flagging if there are any PRs you want the entire web team to review, you can select `web-reviewers`

.changeset/old-candles-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'@penumbra-zone/ui': patch
```
grod220, line 2: If we add new truncate functionality, think this would be a minor version bump

packages/ui/components/ui/box.tsx
```diff
@@ -34,19 +34,21 @@ export const Box = ({
...
   layoutId,
   headerContent,
+  className,
 }: PropsWithChildren<
   VariantProps<typeof variants> & {
     label?: ReactNode;
     layout?: boolean;
     layoutId?: string;
     headerContent?: ReactNode;
+    className?: string;
```
grod220, line 44: This has been one of the things that @jessepinho particularly has been hunting down to get rid of. By default, shadcn/ui takes a className-first kind of approach. This makes the consumer have more flexibility on passing any kind of style into the child. However, this also creates an API surface that is a bit more risky. Different parts our app can possibly customize too much. The solution we are converging toward is relying more upon variants and explicitly enumerated props passing.

In this case, we should find a way to accomplish the effect you are going for, but without requiring className as a prop.

packages/ui/components/ui/tx/view/undelegate.tsx
```diff
@@ -27,8 +27,10 @@ export const UndelegateComponent = ({ value }: { value: Undelegate }) => {
 
           {/** @todo: Render validator name/etc. after fetching? */}
           {!!value.validatorIdentity && (
-            <ActionDetails.Row label='Validator identity' truncate>
-              {bech32mIdentityKey(value.validatorIdentity)}
+            <ActionDetails.Row label='Validator identity'>
+              <ActionDetails.TruncatedText>
```
grod220, line 31: Nice, that's cool

packages/ui/components/ui/box.tsx
VanishMax: Right, got it. Changed to allow the `overflow` variant with type `xHidden` to the Box component, while removing className. Completely agree on the approach

grod220 (approved):

PR #1255, opened 2024-06-05
IBC `Any` ser/de

### LINKED ISSUE #1254: Use binary encoding to transfer TransactionPerspective and TransactionView
https://github.com/penumbra-zone/web/blob/20bb7acc5c9d50eb53157c10aaca7f4a5167bfaf/packages/wasm/crate/src/tx.rs#L176-L197

when containing an ibc action, they possess a google.protobuf.Any, which we haven't transferred across the wasm boundary before. serde_wasm_bindgen is not configured to handle this properly and creates fields as if we were transferring a standard protobuf message, rather than protojson as we expect on the other end.

this is one of the last areas we are using json to cross the boundary. so switching to binary encoding may provide a solution.

turbocrime:
fixes #1254 

relevant https://github.com/penumbra-zone/penumbra/issues/4563

### DIFF
--- a/apps/minifront/src/components/tx-details/tx-viewer.tsx
+++ b/apps/minifront/src/components/tx-details/tx-viewer.tsx
@@ -6,6 +6,7 @@ import type { Jsonified } from '@penumbra-zone/types/jsonified';
 import { viewFromEmptyPerspective } from '@penumbra-zone/perspective/transaction/perspective';
 import { useState } from 'react';
 import { SegmentedPicker } from '@penumbra-zone/ui/components/ui/segmented-picker';
+import { typeRegistry } from '@penumbra-zone/protobuf';
 
 export enum TxDetailsTab {
   PUBLIC = 'public',
@@ -33,7 +34,7 @@ export const TxViewer = ({ txInfo, hash }: TxDetailsLoaderResult) => {
           <TransactionViewComponent txv={txInfo.view!} />
           <div className='mt-8'>
             <div className='text-xl font-bold'>Raw JSON</div>
-            <JsonViewer jsonObj={txInfo.toJson() as Jsonified<TransactionInfo>} />
+            <JsonViewer jsonObj={txInfo.toJson({ typeRegistry }) as Jsonified<TransactionInfo>} />
           </div>
         </>
       )}
--- a/packages/wasm/crate/src/tx.rs
+++ b/packages/wasm/crate/src/tx.rs
@@ -161,11 +161,16 @@ pub fn build_parallel(
     Ok(tx.encode_to_vec())
 }
 
+/// Get transaction perspective, transaction view
+/// Arguments:
+///     full_viewing_key: `FullViewingKey` inner bytes
+///     tx: Binary-encoded `Transaction` message
+///     idb_constants: IndexedDbConstants
+/// Returns: `{ txp: Uint8Array, txv: Uint8Array }` representing binary-encoded `TransactionPerspective` and `TransactionView`
 #[wasm_bindgen(typescript_custom_section)]
 const TRANSACTION_VIEW_PERSPECTIVE_TS: &'static str = r#"
 export function transaction_perspective_and_view(full_viewing_key: Uint8Array, tx: Uint8Array, idb_constants: any): Promise<{ txp: Uint8Array, txv: Uint8Array }>;
 "#;
-
 #[wasm_bindgen(skip_typescript)]
 pub async fn transaction_perspective_and_view(
     full_viewing_key: &[u8],
--- a/packages/wasm/crate/src/tx.rs
+++ b/packages/wasm/crate/src/tx.rs
@@ -178,7 +178,7 @@ pub fn build_parallel(
 ///     full_viewing_key: `byte representation inner FullViewingKey`
 ///     tx: `pbt::Transaction`
 ///     idb_constants: IndexedDbConstants
-/// Returns: `TxInfoResponse`
+/// Returns: Javascript array of two numeric arrays
 #[wasm_bindgen]
 pub async fn transaction_info(
     full_viewing_key: &[u8],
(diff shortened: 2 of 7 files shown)


packages/wasm/crate/src/tx.rs
turbocrime, line 1: this is the worst part. i've fixed the serialization failure, but it's transmitting across the boundary as an array of two numeric arrays (not a binary vector)

would be nice to have some attention to typing this properly

apps/minifront/src/components/tx-details/tx-viewer.tsx
```diff
@@ -33,7 +34,7 @@ export const TxViewer = ({ txInfo, hash }: TxDetailsLoaderResult) => {
           <TransactionViewComponent txv={txInfo.view!} />
           <div className='mt-8'>
             <div className='text-xl font-bold'>Raw JSON</div>
-            <JsonViewer jsonObj={txInfo.toJson() as Jsonified<TransactionInfo>} />
+            <JsonViewer jsonObj={txInfo.toJson({ typeRegistry }) as Jsonified<TransactionInfo>} />
```
turbocrime, line 37: viewer should use the type registry to serialize to json

packages/wasm/src/transaction.ts
```diff
@@ -5,26 +5,28 @@ import {
...
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
 import type { IdbConstants } from '@penumbra-zone/types/indexed-db';
-import { JsonValue } from '@bufbuild/protobuf';
 import { FullViewingKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
-
-interface TxInfoWasmResult {
-  txp: JsonValue;
-  txv: JsonValue;
-}
+//import { typeRegistry } from '@penumbra-zone/protobuf';
```
grod220, line 9: Can delete

packages/wasm/src/transaction.ts
```diff
@@ -5,26 +5,28 @@ import {
...
-  const { txp, txv } = (await transaction_info(
+  const txInfo: unknown = await transaction_info(
     fullViewingKey.toBinary(),
     tx.toBinary(),
     idbConstants,
-  )) as TxInfoWasmResult;
+  );
+
+  if (!Array.isArray(txInfo) || !Array.isArray(txInfo[0]) || !Array.isArray(txInfo[1]))
+    throw new Error('Invalid transaction info');
```
grod220, line 23: Can you convert this to a type guard? Something like:

```ts
type TransactionInfo = [number[], number[]];

const isTransactionInfo = (data: unknown): data is TransactionInfo =>
  Array.isArray(data) &&
  data.length >= 2 &&
  Array.isArray(data[0]) &&
  data[0].every(item => typeof item === 'number') &&
  Array.isArray(data[1]) &&
  data[1].every(item => typeof item === 'number');
```

and then you can do:
```ts
  if (!isTransactionInfo(txInfo)) throw new Error('Invalid transaction info');

  const txpBin = new Uint8Array(txInfo[0]);
  const txvBin = new Uint8Array(txInfo[1]);
```

grod220 (approved):
Oh, strange behavior. Good catch though.

packages/wasm/src/transaction.ts
turbocrime: deserialization will throw if they're malformed. also, the arrays may be a few thousand items long, so mapping is excessive.

turbocrime: the correct thing would be to ensure wasm is returning TypedArrays, i'll spend some time on that in the morning

packages/wasm/crate/src/tx.rs
```diff
@@ -173,34 +161,48 @@ pub fn build_parallel(
...
+export function transaction_perspective_and_view(full_viewing_key: Uint8Array, tx: Uint8Array, idb_constants: any): Promise<{ txp: Uint8Array, txv: Uint8Array }>;
+"#;
+
+#[wasm_bindgen(skip_typescript)]
+pub async fn transaction_perspective_and_view(
     full_viewing_key: &[u8],
     tx: &[u8],
     idb_constants: JsValue,
-) -> WasmResult<JsValue> {
+) -> WasmResult<js_sys::Object> {
```
turbocrime, line 174: wasm-bindgen support for async is rather bad, so i had to resort to a manual declaration. this is better than the alternatives

packages/wasm/crate/src/tx.rs
```diff
@@ -173,34 +161,53 @@ pub fn build_parallel(
...
+const TRANSACTION_VIEW_PERSPECTIVE_TS: &'static str = r#"
+export function transaction_perspective_and_view(full_viewing_key: Uint8Array, tx: Uint8Array, idb_constants: any): Promise<{ txp: Uint8Array, txv: Uint8Array }>;
+"#;
+#[wasm_bindgen(skip_typescript)]
+pub async fn transaction_perspective_and_view(
     full_viewing_key: &[u8],
     tx: &[u8],
     idb_constants: JsValue,
-) -> WasmResult<JsValue> {
+) -> WasmResult<js_sys::Object> {
```
grod220, line 179: Think there is another more native-wasm-bindgeny way to do this. Will open a PR.

packages/wasm/crate/src/tx.rs
```diff
@@ -173,34 +161,53 @@ pub fn build_parallel(
...
-    let result = serde_wasm_bindgen::to_value(&response)?;
-    Ok(result)
+    let response = js_sys::Object::new();
+
+    js_sys::Reflect::set(
+        &response,
+        &"txp".into(),
+        &js_sys::Uint8Array::from(prost::Message::encode_to_vec(&txp).as_slice()),
+    )
+    .expect("Must set txp");
```
grod220, line 194: We want to avoid .expect() as it can panic and causes an exception we can't handle

packages/wasm/crate/src/tx.rs
grod220: https://github.com/penumbra-zone/web/pull/1261

PR #1259, opened 2024-06-06
Simplify Poisson distribution function

jessepinho:
@erwanor does this look OK to you? And does the comment describe it accurately?


erwanor (approved):
Yup, looks cleaner, thanks!

Issue #1251, opened 2024-06-04
Tx details & approval code de-dupe


grod220:
We have two areas in the codebase rendering UI from transaction views:

## 1. Tx approval dialog

<img width="395" alt="Screenshot 2024-06-04 at 2 12 11 PM" src="https://github.com/penumbra-zone/web/assets/16624263/e553aff3-d9b1-4667-a01f-1185fa20114f">

https://github.com/penumbra-zone/web/blob/c8e8d15ce1896f625452412b2d567790b0993646/apps/extension/src/state/tx-approval.ts#L84-L91

## 2. Tx details page

<img width="931" alt="Screenshot 2024-06-04 at 2 14 19 PM" src="https://github.com/penumbra-zone/web/assets/16624263/9c8ee37a-8471-40cf-9e2b-4ecd6a5dd45d">

https://github.com/penumbra-zone/web/blob/c8e8d15ce1896f625452412b2d567790b0993646/apps/minifront/src/components/tx-details/index.tsx#L32-L40

=====

They both use `TransactionViewComponent`, however, there are areas of code redundancy that can be de-duped:
- Tab usage. Both have different logic on how they handle tab selection.
- Differing views. Only the tx approval dialog has "Receiver View". This should also be available in the tx-details page.
- Utilities. `asPublicTransactionView` vs `viewFromEmptyPerspective` usage covering the same use cases.

Having one component that covers both use-cases and handle its own internal state will make this more maintainable.

hdevalence:
I don't think we want to deduplicate these views. In fact I think it's the opposite, the approval dialog should have its own separate view component distinct from the regular transaction view, because it's security critical and may want to have different rendering.

What I suggested deduplicating was the methods for attenuating the view, but this is already tracked by an existing issue.

TalDerei:
> but this is already tracked by an existing issue

can you link the relevant issue?

hdevalence:
Yup, sorry, I didn't have it handy when I wrote that comment, but here it is: https://github.com/penumbra-zone/web/issues/483

TalDerei:
> * Differing views. Only the tx approval dialog has "Receiver View". This should also be available in the tx-details page.

this is turning out to be specifically tricky because of the synchronicity of the `TxViewer` component, and our translators are asynchronous. I'm experimenting with a custom hook design.

Issue #483, opened 2024-02-08
Extract Receiver View code into common library and use it transaction views


hdevalence:
The approval dialog has code for attenuating the sender's view to the receiver's view, so it can show a three-state view: "Your View" / "Receiver's View" / "Public View".

This should also work in the transaction view page, using common code.

PR #1260, opened 2024-06-06
views: extract receiver view into transaction view

TalDerei:
References #483 and https://github.com/penumbra-zone/web/issues/1251#issuecomment-2150629725

@jessepinho the `asReceiverTransactionView` translator from the transaction approval dialog is reused in the transaction view page. Currently, i'm displaying _every_ transaction with an associated receiver view which obviously is **not** correct. 

There seems to be some misdirection in the `TransactionViewSwitcher` component that I couldn't quite parse, mainly where the actual filtering is happening that determines if the receiver view should be present or not. Is there a simple way to propagate this filtering to the transaction view page in this PR?

### DIFF
--- a/apps/minifront/src/components/tx-details/hooks.ts
+++ b/apps/minifront/src/components/tx-details/hooks.ts
@@ -1,0 +1,42 @@
+import { useEffect, useState } from 'react';
+import { asReceiverTransactionView } from '@penumbra-zone/perspective/translators/transaction-view';
+import { viewClient } from '../../clients';
+import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/transaction/v1/transaction_pb';
+import { TransactionInfo } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+
+interface UseReceiverViewResult {
+  receiverView: TransactionView | null;
+}
+
+const useReceiverView = (
+  txInfo: TransactionInfo | undefined,
+  option: string,
+): UseReceiverViewResult => {
+  const [receiverView, setReceiverView] = useState<TransactionView | null>(null);
+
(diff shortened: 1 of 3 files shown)


apps/minifront/src/components/tx-details/hooks.ts
grod220, line 1: Whenever we are doing async in react-land, there is a high likeliness that react-query can help. And this is a good candidate. Something like this I'd imagine:

```ts
const fetchReceiverView = async (txInfo: TransactionInfo): Promise<TransactionView> => {
  return await asReceiverTransactionView(txInfo.view, {
    isControlledAddress: async address =>
      viewClient.indexByAddress({ address }).then(({ addressIndex }) => Boolean(addressIndex)),
  });
};

export const TxViewer = ({ txInfo, hash }: TxDetailsLoaderResult) => {
  const [option, setOption] = useState(TxDetailsTab.PRIVATE);

  const { data: receiverView } = useQuery(
    ['receiverView', txInfo, option],
    () => fetchReceiverView(txInfo),
    {
      enabled: option === TxDetailsTab.RECIEVER && !!txInfo,
    },
  );
```

TalDerei: switched to using this very helpful pattern

TalDerei:
Rather than trying to generalize [ViewTabs](https://github.com/penumbra-zone/web/blob/main/apps/extension/src/routes/popup/approval/transaction/view-tabs.tsx#L7) and passing in `transactionClassification` as a prop, there's a simpler and more direct approach. Invoking the [classifyTransaction](https://github.com/penumbra-zone/web/blob/main/packages/perspective/transaction/classify.ts#L4) utility function found in the perspective package achieves the aforementioned filtering in the `TxViewer` component.

grod220 (approved):
👍

PR #1261, opened 2024-06-06
Return a TxInfo struct [#1255 follow up]

grod220:
Defining a wasm-bindgen struct allows for a simpler return value while keeping the same type signature.

Follow up from: https://github.com/penumbra-zone/web/pull/1255#pullrequestreview-2101210513

### DIFF
--- a/.changeset/mean-seahorses-run.md
+++ b/.changeset/mean-seahorses-run.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/wasm': patch
+---
+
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
the type signature is much more idiomatic!

turbocrime:
thanks

turbocrime (approved):

turbocrime:
however we probably shouldn't call it "txinfo" as it's too similar to the actual transactioninfo type that we have

PR #1262, opened 2024-06-06
Update onboarding re: new registry version

grod220:
Pairs with: https://github.com/prax-wallet/registry/pull/46

New registry has rpcs+frontends no longer indexed by chain id. This also removes the async nature of the rpc selection in the onboarding form.


VanishMax (approved):
Great improvement to not be dependent on chain ID!

grod220:
@VanishMax it's a bit of a trade off. Decoupling rpcs from chain ids make them easier to access on the consumer side, but that means it's possible that rpcs with different chain ids could be co-mingled. Something we can figure out later if the registry ends up running into this case.

Issue #1265, opened 2024-06-08
Invalid fee rendering


TalDerei:
Investigate the rendering of gas fees, which appear to be much higher than the actual fees being deducted in transactions.  

<img width="756" alt="Screenshot 2024-06-07 at 10 18 15 PM" src="https://github.com/penumbra-zone/web/assets/70081547/c081efeb-9f3d-4813-af81-1c1516d3e5df">
<img width="750" alt="Screenshot 2024-06-07 at 10 18 23 PM" src="https://github.com/penumbra-zone/web/assets/70081547/64dbd87b-97c4-48aa-85c8-21cdd3f6e19f">

Issue #1165, opened 2024-05-24
 multi-asset fees support in minifront


TalDerei:
We have multi-asset fee support at the protocol level, and the web planner implementation has achieved parity with core. There are supplemental UX changes that need to be implemented to fully realize this. 

Logic:
- if the user has UM, use that for fees
- otherwise, use the first alt fee token that they have balance of
  - tx approval warns about non-native fees as a privacy risk and recommends using native token
- otherwise, error that they have no available fee tokens

grod220:
Updated the description a bit. The Fee denom gets set to staking token in the default() initialization: https://github.com/penumbra-zone/web/blob/5b80e7cf988475fcbace89c62643d42f3a58135a/packages/wasm/crate/src/planner.rs#L205

It looks like, before this, we need to check for balances of UM. If none, initialize ActionList with the alt denom and I believe the methods below should work as is (?). Definitely needs testing out. Also, `transaction_parameters` needs likewise updating.

TalDerei:
folding in https://github.com/penumbra-zone/web/issues/1265 as relevant scope

PR #1268, opened 2024-06-09
fees: ux for multi-asset fees

TalDerei:
References #1165 and https://github.com/penumbra-zone/web/issues/1265. 

Multi-asset fees were implemented, allowing transactions to use an alternative token for fees if the user does not have UM tokens.

IndexedDB was modified to add an index on the asset ID of the note's value in the `SPENDABLE_NOTES` table. Indexes can improve query performance at the cost of additional disk space for storage.

Following the resolution of https://github.com/penumbra-zone/penumbra/issues/4306, the fee rendering in the transaction approval dialogue and transaction view page displayed values that appeared significantly higher than the actual fees being deducted in transactions. The values were divided by **10^6** to display the correct decimal fee amount. 

--------------------------------------------------------------------------------------------------------------------------

**before:**

Screenshot 2024-06-09 at 10 29 40 PM

**after:**

Screenshot 2024-06-09 at 11 30 38 PM

--------------------------------------------------------------------------------------------------------------------------

**alternative asset fees if the user doesn't have UM:**

Screenshot 2024-06-09 at 11 44 37 PM

--------------------------------------------------------------------------------------------------------------------------
**privacy warning for non-native alt fees**

https://github.com/penumbra-zone/web/assets/70081547/33ad2c63-3a07-4edc-9816-9004c7a98bd2


packages/services/src/view-service/fees.ts
```diff
@@ -0,0 +1,40 @@
...
+  ];
+
+  const nonEmptyField = fields.find(field => field.value.length > 0);
+
+  if (!nonEmptyField) {
+    console.warn('No non-empty field found in the request');
+    return undefined;
+  }
+
+  type PossibleTypes = TransactionPlannerRequest_Output | TransactionPlannerRequest_Swap;
```
TalDerei, line 31: `PossibleTypes` needs to be expanded to support other actions, for instance auctions where a user can perform price discovery on a specific asset _without_ needing UM tokens.

packages/services/src/view-service/fees.ts
```diff
@@ -0,0 +1,40 @@
+import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import {
+  TransactionPlannerRequest,
+  TransactionPlannerRequest_Output,
+  TransactionPlannerRequest_Swap,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+
+export const extractAltFee = (request: TransactionPlannerRequest): AssetId | undefined => {
```
TalDerei, line 8: need to still add warning message in tx approval for non-native assets.

packages/wasm/src/planner.ts
```diff
@@ -4,16 +4,19 @@ import { JsonValue } from '@bufbuild/protobuf';
 import { plan_transaction } from '../wasm';
 import type { IdbConstants } from '@penumbra-zone/types/indexed-db';
 import { FullViewingKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 
 export const planTransaction = async (
   idbConstants: IdbConstants,
   request: TransactionPlannerRequest,
   fullViewingKey: FullViewingKey,
+  assetId: AssetId,
```
Valentine1898, line 13: I think it makes sense to give a more meaningful name for this parameter, perhaps `feeAssetId`?

packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -8,7 +8,9 @@ export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
   if (!txv.bodyView) throw new Error('transaction view missing body view');
   if (!txv.bodyView.transactionParameters?.fee?.amount) throw new Error('Missing fee amount');
 
-  const fee = joinLoHiAmount(txv.bodyView.transactionParameters.fee.amount).toString();
+  const fee = (
+    Number(joinLoHiAmount(txv.bodyView.transactionParameters.fee.amount)) / 1000000
+  ).toString();
```
Valentine1898, line 13: We should probably use the `exponent` value from the registry metadata instead of the constant `1000000` here

packages/types/src/indexed-db.ts
```diff
@@ -141,6 +141,8 @@ export interface IndexedDbInterface {
   getAuctionOutstandingReserves(
     auctionId: AuctionId,
   ): Promise<{ input: Value; output: Value } | undefined>;
+
+  hasNativeAssetBalance(): Promise<boolean>;
```
Valentine1898, line 145: Are `NativeAsset` and `StakingAsset` the same thing? I'm not sure, but it might make sense to use the same naming throughout so it doesn't cause confusion. (although for me personally, `NativeAsset` is the clear naming)

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +784,12 @@ export class IndexedDb implements IndexedDbInterface {
       output: Value.fromJson(result.output),
     };
   }
+
+  async hasNativeAssetBalance(): Promise<boolean> {
+    const key = 'KeqcLzNx9qSH5+lcJHBB9KNW+YPrBk5dKzvPMiypahA=';
```
Valentine1898, line 789: same
you can get the `stakingAssetId` from the prax registry instead of a string constant

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +784,12 @@ export class IndexedDb implements IndexedDbInterface {
...
+    const jsonArray = await this.db.getAllFromIndex('SPENDABLE_NOTES', 'assetId', key);
+    if (jsonArray.length === 0) return false;
+
+    return true;
```
Valentine1898, line 793: you may have `notes` but they have all been spent, you should probably only get the unspent `notes` here

packages/services/src/view-service/fees.ts
Valentine1898: If `PossibleTypes` are defined not only for wasm planner but for the whole chain, it might be more convenient to import it from the core repo.

packages/services/src/view-service/transaction-planner/index.test.ts
```diff
@@ -77,47 +78,9 @@ describe('TransactionPlanner request handler', () => {
...
-    mockIndexedDb.getAppParams?.mockResolvedValueOnce(
-      new AppParameters({
-        chainId: 'penumbra-testnet-mock',
-        sctParams: new SctParameters({
-          epochDuration: 719n,
-        }),
-      }),
-    );
-    await expect(transactionPlanner(req, mockCtx)).rejects.toThrow('Gas prices is not available');
-  });
```
Valentine1898, line 122: any reason for that?

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -5,11 +5,29 @@ import { Code, ConnectError } from '@connectrpc/connect';
...
   const { indexedDb } = await services.getWalletServices();
 
+  // Query IndexedDB directly to check for the existence of spendable 'UM'
+  // notes (native asset balance)
+  const nativeToken = await indexedDb.hasNativeAssetBalance();
+
+  // Initialize the gas fee token using an native staking token's asset ID
+  let gasFeeToken = new AssetId({
+    inner: new Uint8Array(base64ToUint8Array('KeqcLzNx9qSH5+lcJHBB9KNW+YPrBk5dKzvPMiypahA=')),
+  });
```
Valentine1898, line 23: you can get the `stakingAssetId` from the prax registry instead of a string constant

Valentine1898 (commented):
Most of the comments have to do with the fact that we don't use registry in many places

Valentine1898:
Another thing that confuses me, you may have a small UM balance which may not be enough to pay the transaction fee. I assume that it is not enough to check just the balance, we need to make sure that this balance is more than necessary for the transaction

Valentine1898:
image
It also looks like a bug in your screenshot. Probably means `UM` instead of `upenumbra`

packages/services/src/view-service/fees.ts
TalDerei: `PossibleTypes` is a custom type I created to capture a subset of the _relevant_ action fields in `TransactionPlannerRequest`. The `assetId` will be extracted from that. I just need to extend this to extract the assetId from other types, for instance `TransactionPlannerRequest_ActionDutchAuctionSchedule`. For our purposes, I think this approach is sufficient.

packages/wasm/src/planner.ts
TalDerei: changed

packages/types/src/indexed-db.ts
TalDerei: yeah, good point. they're the same, but will stick with `StakingAsset` for consistency

packages/storage/src/indexed-db/index.ts
TalDerei: @grod220 I modified the function to query the registry, but this is happening for every call. Do we instead want to create a global variable `STAKING_TOKEN_ASSET_ID` since the staking token asset id won't change?

packages/storage/src/indexed-db/index.ts
TalDerei: I switched to using `viewService.balances(new BalancesRequest({ accountFilter: addressIndex, assetIdFilter: stakingAssetId }), ctx)` to retrieve the assetId balance.

packages/services/src/view-service/transaction-planner/index.ts
TalDerei: modified to 

```
let gasFeeToken = new AssetId({
    inner: stakingTokenId.inner,
});
```
  
  where `stakingTokenId` is retrieved from prax registry.

TalDerei:
> Another thing that confuses me, you may have a small UM balance which may not be enough to pay the transaction fee. I assume that it is not enough to check just the balance, we need to make sure that this balance is more than necessary for the transaction

If the user has an insufficient UM balance, the transaction will rightly fail. Rather than doing something more sophisticated, it's fine to let the transaction fail and require the user to fetch more UM. Multi-asset fees are intended primarily for IBC-in transfers where new users have no UM, and they can perform a swap or price discovery via the auction mechanism with their native bridged assets.

TalDerei:
> It also looks like a bug in your screenshot. Probably means `UM` instead of `upenumbra`

yes, following the exponent calculations, this should not be displaying `UM` instead of `upenumbra`. I will change the base denom to reflect this.

packages/storage/src/indexed-db/index.ts
TalDerei: @Valentine1898 further testing revealed that this using `viewService.balances` was in fact much slower because it's iterating over all the spendable notes in indexdb, rather than just the subset of UM notes we care about.

packages/storage/src/indexed-db/index.ts
TalDerei: and if the registry becomes unaccessible for some reason, won't this block transaction creation?

grod220: The registry is bundled within the NPM package. That means, it no longer makes a live http call. I just checks the existing bundle by chain Id. That means it should only fail if the chainId doesn't exist in the registry (which is not likely if they made it to this step). Also, it doesn't need to be async either, you can do:

```ts
  fetchStakingTokenId(): AssetId {
    const registryClient = new ChainRegistryClient();
    const registry = registryClient.get(this.chainId);
    const stakingTokenMetadata = registry.getMetadata(registry.stakingAssetId);
    return getAssetId(stakingTokenMetadata);
  }
```

packages/storage/src/indexed-db/index.ts
TalDerei: I've noticed this call takes ~`500ms`

packages/services/src/view-service/transaction-planner/index.test.ts
TalDerei: these aren't really doing anything in the context of the tests, and this will be caught in our transaction planner service impl in practice.

packages/storage/src/indexed-db/index.ts
TalDerei: on closer inspection, making this call incurs ~500ms - 1s in additional overhead that noticeably delayed the transaction approval popup. The `balances` query has extraneous logic and iterates over all the spendable notes rather than just inspecting the relevant staking asset notes. I instead opted for a _custom_ note processing scheme which takes a negligible amount of time to execute.

packages/services/src/view-service/transaction-planner/index.test.ts
```diff
@@ -77,47 +78,9 @@ describe('TransactionPlanner request handler', () => {
         compactBlockSpacePrice: 120n,
       }),
     );
+    mockIndexedDb.hasNativeAssetBalance?.mockResolvedValueOnce(true);
```
jessepinho, line 81: The mock has a `hasNativeAssetBalance` method, but the actual implementation has a `hasStakingAssetBalance` method. I assume this is a typo/old method name, but it makes me wonder if the mock is actually doing anything?

packages/services/src/test-utils.ts
```diff
@@ -32,6 +32,7 @@ export interface IndexedDbMock {
   getPricesForAsset?: Mock;
   getAuction?: Mock;
   getAuctionOutstandingReserves?: Mock;
+  hasNativeAssetBalance?: Mock;
```
jessepinho, line 35:
```suggestion
  hasStakingAssetBalance?: Mock;
```

packages/services/src/view-service/fees.ts
```diff
@@ -0,0 +1,40 @@
...
+  const nonEmptyField = fields.find(field => field.value.length > 0);
+
+  if (!nonEmptyField) {
+    console.warn('No non-empty field found in the request');
+    return undefined;
+  }
+
+  type PossibleTypes = TransactionPlannerRequest_Output | TransactionPlannerRequest_Swap;
+
+  const action = nonEmptyField.value[0] as PossibleTypes;
```
jessepinho, line 33: Unless I'm missing something, this looks type-unsafe. We're iterating over all the possible fields of a request, which means `nonEmptyField` could contain _any_ type of actions, not just outputs and swaps. And the other action types might not have an `assetId` property. So, although the code we have here technically works for now, it could mislead future developers who will think that `action` is definitely either an output or a swap. I'd prefer instead to use the `.find()` above to find those specific action types so that the types are valid.

pnpm-lock.yaml
jessepinho, line 1: `pnpm-lock.yaml` shouldn't have changed if none of your `package.json`s changed. Can you remove this file from the PR?

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +795,44 @@ export class IndexedDb implements IndexedDbInterface {
...
+  }
+
+  async hasStakingAssetBalance(assetId: AssetId): Promise<boolean> {
+    const spendableUMNotes = await this.db.getAllFromIndex(
+      'SPENDABLE_NOTES',
+      'assetId',
+      uint8ArrayToBase64(assetId.inner),
+    );
+
+    // Iterate over the spendable UM notes, and acrue balance for unspent notesz
```
jessepinho, line 815:
```suggestion
    // Iterate over the spendable UM notes, and accrue balance for unspent notes
```

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +795,44 @@ export class IndexedDb implements IndexedDbInterface {
...
+        const newAmount = addLoHi(
+          { lo: stakingTokenBalance.lo, hi: stakingTokenBalance.hi },
+          {
+            lo: BigInt(umNote.note?.value?.amount?.lo ?? 0n),
+            hi: BigInt(umNote.note?.value?.amount?.hi ?? 0n),
+          },
+        );
+        stakingTokenBalance.lo = newAmount.lo;
+        stakingTokenBalance.hi = newAmount.hi;
+      }
```
jessepinho, line 829: Couldn't this all be simplified a bit? e.g., 

```suggestion
        stakingTokenBalance = addLoHi(
          stakingTokenBalance,
          {
            lo: BigInt(umNote.note?.value?.amount?.lo ?? 0n),
            hi: BigInt(umNote.note?.value?.amount?.hi ?? 0n),
          },
        );
      }
```

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +795,44 @@ export class IndexedDb implements IndexedDbInterface {
...
+    const spendableUMNotes = await this.db.getAllFromIndex(
+      'SPENDABLE_NOTES',
+      'assetId',
+      uint8ArrayToBase64(assetId.inner),
+    );
+
+    // Iterate over the spendable UM notes, and acrue balance for unspent notesz
+    let stakingTokenBalance = new Amount();
+    for (const note of spendableUMNotes) {
+      let umNote = note as unknown as SpendableNoteRecord;
```
jessepinho, line 818:
```suggestion
      const umNote = SpendableNoteRecord.fromJson(note);
```

(right?)

packages/storage/src/indexed-db/index.ts
```diff
@@ -802,4 +795,44 @@ export class IndexedDb implements IndexedDbInterface {
...
+      if (umNote.heightSpent == undefined) {
```
jessepinho, line 819:
```suggestion
      if (umNote.heightSpent === undefined) {
```

However, if it's a class (as per my suggestion above of using `.fromJson()`), it will never be `undefined`, since Protobuf classes default to `0n` for undefined bigint values. So, hm, we'd need to think of another way to check this.

packages/ui/components/ui/tx/view/transaction.tsx
jessepinho: Definitely — especially if the fee can be in other currencies than UM

jessepinho (commented):
I haven't finished reviewing, but I want to pause first and make sure I'm understanding the approach here:

1. Are we replacing the staking token with any of the other tokens being used in the transaction, based on the assumption that the user has a balance in that token if it's being used in the transaction? If so, what if the user is already using up their entire balance of that token in the transaction, and thus has none left over for the fee? Shouldn't we instead be looking at the user's balances of other tokens to determine which token to use?
2. This may be my ignorance of how multi-asset fees work, but don't we need some guarantee of the value of another asset to be able to use it as a fee? i.e., don't we need to e.g., do a swap first into UM and then use it as the fee?
3. If we are truly using another asset type as the fee, we shouldn't hard-code `"UM"` in various places in the UI where fees are displayed, right?

hdevalence:
> I haven't finished reviewing, but I want to pause first and make sure I'm understanding the approach here:
> 
>     1. Are we replacing the staking token with any of the other tokens being used in the transaction, based on the assumption that the user has a balance in that token if it's being used in the transaction? If so, what if the user is already using up their entire balance of that token in the transaction, and thus has none left over for the fee? Shouldn't we instead be looking at the user's balances of other tokens to determine which token to use?

I think the simplest approach to get the desired behavior (users can deposit over IBC and immediately start using Penumbra) is to do the following:

1. if the user has an UM balance, use that;
2. otherwise, iterate through the list of allowed alt fee tokens and use the first one for which the user has a balance.

We already don't have a way to handle the "send max" use case in the UI, so there's no regression here. To fix that problem properly, we'd want to have a "MAX" button, and when it's set, the way the planner is invoked should change: rather than specifying _outputs_ and filling in spends to balance, it should specify _spends_ (spend every note available) and set the change address for the outputs to the destination address. This way fees are deducted from outputs rather than inputs. The broken behavior you're describing that would prevent that is already the case for the UM token, and most of the plumbing for a fix is in place. (This UX is not even possible on a chain like Ethereum, where there's no way to statically determine the exact fee ahead of time).

>     2. This may be my ignorance of how multi-asset fees work, but don't we need some guarantee of the value of another asset to be able to use it as a fee? i.e., don't we need to e.g., do a swap first into UM and then use it as the fee?

No, the initial multi-asset fee mechanism has independent prices for each token. The idea is that the prices in alternative tokens should be set higher (say 10x higher) so that the resource pricing is protected from price fluctuations. This also means that as training wheels, there's no dependence on the DEX prices being accurate, and no possibility that someone could manipulate prices on the DEX and turn that into a DoS attack on the chain. The intended design is that the fee component swaps the alternative fee token to the native token but this happens on the backend, internal to the fee component, and isn't implemented yet.

>     3. If we are truly using another asset type as the fee, we shouldn't hard-code `"UM"` in various places in the UI where fees are displayed, right?

Yes, that seems right. We might need to adjust the proto definitions for the `TransactionView` to this effect (to have a presupplied `Metadata`), or we could have the renderer do some janky searching through the rest of the transaction for a `Metadata` with matching asset ID.

packages/ui/components/ui/tx/view/transaction.tsx
TalDerei: modified to use `getStakingTokenMetaData` to retrieve the exponent relevant to the relevant fee asset_id in the TxV

pnpm-lock.yaml
TalDerei: local changeset adds registry dep to `ui` package's `package.json`, so we'll need to commit the lock file

packages/storage/src/indexed-db/index.ts
TalDerei: sloppy typos, nice catch

packages/storage/src/indexed-db/index.ts
TalDerei: this doesn't seem to work because of a type signature mismatch, ie. you're trying to assign `Required<LoHi>` to an `Amount`. I"ll make some changes to the variable names here.

packages/storage/src/indexed-db/index.ts
TalDerei: yeah exactly, made the same change locally!

packages/storage/src/indexed-db/index.ts
TalDerei: you're right, then can't you just do `umNote.heightSpent === 0n` to filter for unspent notes?

jessepinho: This is the thing that always confuses me a little bit about Protobuf's default values. `0n` is generally an actually meaningful value. It _happens_

jessepinho: Whoops, bumped "Comment." 

... It _happens_ not to be meaningful in this specific use case, because it's impossible to have spent any notes at block height `0n` since that's the genesis block. So you could safely use that check, yes. But that's... weird and janky? I kind of wish Protobufs just allowed u64 values to be undefined, rather than defaulting to `0n`.

But tl;dr: yes, that check should work.

packages/storage/src/indexed-db/index.ts
jessepinho: Ahh... I missed that `hi` was optional on a `LoHi`, and I'd assumed that `PartialMessage<Amount>` and `LoHi` were interchangeable. My mistake.

pnpm-lock.yaml
jessepinho: Before merging, please just confirm that this is still the case when you've upgraded your pnpm. I don't believe `pnpm-lock` should ever change is no `package.json`s have.

packages/services/src/view-service/fees.ts
TalDerei: good points, I [restricted](https://github.com/penumbra-zone/web/pull/1268/commits/89a24c183b4ab424250c0c08a35abb64086f1773) the scope to only supporting send / swap / auction actions. This means user's can only use alternative fees for that specified functionality, and returning `undefined` will cause the transaction to fail / prompt insufficient note balance because user doesn't have enough UM. Does that sufficiently address the concern here?

TalDerei: > Unless I'm missing something, this looks type-unsafe.

good point, I [restricted](https://github.com/penumbra-zone/web/pull/1268/commits/89a24c183b4ab424250c0c08a35abb64086f1773) the scope to only supporting send / swap / auction actions. This means user's can only use alternative fees for that specified functionality, and returning `undefined` will cause the transaction to fail / prompt insufficient note balance because user doesn't have enough UM. Does that sufficiently address the concern here?

TalDerei:
**update:** the remaining actions items here are **1** and **3**, which require proto-related modifications. Maybe we should tackle these in a separate PR to limit scope?

TalDerei:
> I think the simplest approach...

_update:_ the remaining action items are (1) supporting the "send max" functionality and (2) not hardcoding UM, which require proto-related modification. To limit scope, I suggest we fix this separately?

packages/storage/src/indexed-db/index.ts
```diff
@@ -824,4 +817,43 @@ export class IndexedDb implements IndexedDbInterface {
...
+        stakingTokenBalance.hi = noteAmount.hi;
+      }
+    }
+
+    if (stakingTokenBalance.lo > 0) {
+      return true;
+    }
+
+    return false;
+  }
```
Valentine1898, line 858: Can we simplify this?

```suggestion
async hasStakingAssetBalance(assetId: AssetId): Promise<boolean> {
  const spendableUMNotes = await this.db.getAllFromIndex(
    'SPENDABLE_NOTES',
    'assetId',
    uint8ArrayToBase64(assetId.inner),
  );

  return spendableUMNotes.some(note => {
    const umNote = SpendableNoteRecord.fromJson(note);
    return umNote.heightSpent === 0n && !isZero(getAmountFromRecord(umNote));
  });
}
```

packages/ui/components/ui/tx/view/registry.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const registryClient = new ChainRegistryClient();
+  const registry = registryClient.get(chainId);
+
+  if (assetId === undefined) {
+    assetId = registry.stakingAssetId;
+  }
+  const StakingTokenMetadata = registry.getMetadata(assetId);
+
+  return StakingTokenMetadata;
+}
```
Valentine1898, line 20: It's hard to figure out 
What does this function do? Does it return metadata for the `assetId` you passed as a parameter, or metadata for the staking token if you passed undefined?
I think it would be better to do a little refactoring and split this logic into two functions

packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -19,10 +29,7 @@ export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
         ))}
       </ViewSection>
       <ViewSection heading='Parameters'>
-        <ViewBox
-          label='Transaction Fee'
-          visibleContent={<div className='font-mono'>{fee} upenumbra</div>}
-        />
+        <ViewBox label='Fee' visibleContent={<div className='font-mono'>{fee} UM</div>} />
```
Valentine1898, line 32: Consider whether we can use `ValueViewComponent` here given that we have an `Metadata` and `Amount`

packages/ui/components/ui/tx/view/transaction.tsx
```diff
@@ -3,12 +3,22 @@ import { MemoViewComponent } from './memo-view';
...
+import { getDisplayDenomExponent } from '@penumbra-zone/getters/metadata';
 
 export const TransactionViewComponent = ({ txv }: { txv: TransactionView }) => {
   if (!txv.bodyView) throw new Error('transaction view missing body view');
   if (!txv.bodyView.transactionParameters?.fee?.amount) throw new Error('Missing fee amount');
 
-  const fee = joinLoHiAmount(txv.bodyView.transactionParameters.fee.amount).toString();
+  // Request the asset metadata
+  const chain_id = txv.bodyView.transactionParameters.chainId;
+  const asset_id = txv.bodyView.transactionParameters.fee.assetId;
```
Valentine1898, line 15: camelCase please

Valentine1898 (commented):
Some ui tests are broken, need to fix them

apps/minifront/src/components/send/helpers.ts
```diff
@@ -31,3 +32,21 @@ export const getTransferableBalancesResponses = async (): Promise<BalancesRespon
     balance => isUnknown(balance) || isTransferable(getMetadata(balance.balanceView)),
   );
 };
+
+export const hasStakingToken = (
+  assetBalances: BalancesResponse[],
+  feeAssetMetadata: Metadata,
```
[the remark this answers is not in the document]
TalDerei: changed to `stakingAssetMetadata`

apps/minifront/src/components/send/helpers.ts
```diff
@@ -31,3 +32,21 @@ export const getTransferableBalancesResponses = async (): Promise<BalancesRespon
...
+      const assetId = asset.balanceView.valueView.value.metadata?.penumbraAssetId?.inner;
+      const feeAssetId = feeAssetMetadata.penumbraAssetId?.inner;
+      if (assetId && feeAssetId && uint8ArrayToBase64(assetId) === uint8ArrayToBase64(feeAssetId)) {
+        stakingToken = true;
+      }
+    }
+  }
+
+  return stakingToken;
+};
```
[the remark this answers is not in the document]
TalDerei: modified to suggested approach, which is much cleaner and more idiomatic.

apps/minifront/src/components/send/send-form/index.tsx
```diff
@@ -100,6 +111,15 @@ export const SendForm = () => {
...
       />
+      {showWarning && (
+        <div className='rounded border border-yellow-500 bg-gray-800 p-4 text-yellow-500'>
+          <strong>Privacy Warning:</strong>
+          <span className='block'>
+            Using non-native tokens for transaction fees may pose a privacy risk. It is recommended
+            to use the native token (UM) for better privacy and security.
+          </span>
+        </div>
+      )}
```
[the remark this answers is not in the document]
TalDerei: Yeah, have a local change-set addressing this for other actions. I'll make the suggested change!

apps/minifront/src/components/send/send-form/index.tsx
```diff
@@ -51,6 +55,11 @@ export const SendForm = () => {
     sendTx,
     txInProgress,
   } = useStore(sendSelector);
+  // State to manage privacy warning display
+  const [showWarning, setShowWarning] = useState(false);
```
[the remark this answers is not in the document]
TalDerei: changed name

packages/services/src/view-service/fees.ts
```diff
@@ -0,0 +1,49 @@
...
+      return new AssetId({
+        inner: (action as TransactionPlannerRequest_ActionDutchAuctionEnd).auctionId?.inner,
+      });
+    case 'dutchAuctionWithdrawActions':
+      return new AssetId({
+        inner: (action as TransactionPlannerRequest_ActionDutchAuctionWithdraw).auctionId?.inner,
+      });
+    default:
+      console.warn('Unsupported action type.');
+      return undefined;
```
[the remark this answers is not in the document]
TalDerei: agreed, returned a `ConnectError` with `FailedPrecondition` argument

packages/services/src/view-service/fees.ts
```diff
@@ -0,0 +1,49 @@
...
+    { name: 'dutchAuctionScheduleActions', value: request.dutchAuctionScheduleActions },
+    { name: 'dutchAuctionEndActions', value: request.dutchAuctionEndActions },
+    { name: 'dutchAuctionWithdrawActions', value: request.dutchAuctionWithdrawActions },
+  ];
+
+  const nonEmptyField = fields.find(field => field.value.length > 0);
+
+  if (!nonEmptyField) {
+    console.warn('No non-empty field found in the request.');
+    return undefined;
```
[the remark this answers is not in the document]
TalDerei: made the same change as above

packages/storage/src/indexed-db/index.ts
```diff
@@ -292,21 +293,13 @@ export class IndexedDb implements IndexedDbInterface {
...
-        compactBlockSpacePrice: 1n,
-      });
-    }
+  async *iterateSpendableStakingNotes() {
+    yield* new ReadableStream(
+      new IdbCursorSource(
+        this.db.transaction('SPENDABLE_NOTES').store.openCursor(),
+        SpendableNoteRecord,
+      ),
+    );
```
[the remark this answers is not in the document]
TalDerei: good catch, this was residue that was supposed to be deleted

packages/storage/src/indexed-db/index.ts
```diff
@@ -292,21 +293,13 @@ export class IndexedDb implements IndexedDbInterface {
...
-    // Note: this is a temporary measure to enable gas prices in the web, but once
-    // https://github.com/penumbra-zone/penumbra/issues/4306 is merged, we can remove this.
-    if (!savedGasPrices) {
-      await this.saveGasPrices({
-        verificationPrice: 1n,
-        executionPrice: 1n,
-        blockSpacePrice: 1n,
-        compactBlockSpacePrice: 1n,
-      });
-    }
```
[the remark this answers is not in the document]
TalDerei: can you extrapolate? `GasPrices` has an `AssetId` field.

packages/storage/src/indexed-db/index.ts
TalDerei: this is better, returning when first note balance is detected. changed

packages/ui/components/ui/tx/view/registry.tsx
TalDerei: I'll add some clarifying comments. After the TxV is generated, the `assetId` for the associated fee may either be `undefined` (meaning it uses the native staking token) or defined as the alt staking token. In the undefined case, we set the assetId to the staking token. We want to check for the undefined case because in transactions.tsx we're using the relative exponent, `const exponent = getDisplayDenomExponent(getStakingTokenMetaData(chain_id, asset_id))` associated with the asset metadata.

packages/ui/components/ui/tx/view/transaction.tsx
TalDerei: changed to camelCase

packages/ui/components/ui/tx/view/transaction.tsx
TalDerei: thinking about something like this might work

```
  let feeValueView = new ValueView({
    valueView: {
      case: 'knownAssetId',
      value: { amount: txv.bodyView.transactionParameters.fee.amount, metadata: feeAssetMetadata },
    },
  });
```

and `<ViewBox label='Fee' visibleContent={<ValueViewComponent view={feeValueView} />} />`

TalDerei: Screenshot 2024-06-16 at 9 29 45 PM

apps/minifront/src/components/send/send-form/index.tsx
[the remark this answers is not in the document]
TalDerei: Screenshot 2024-06-16 at 9 31 20 PM

packages/storage/src/indexed-db/index.ts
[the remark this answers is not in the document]
Valentine1898: You are right, `GasPrices` does contain an `asset_id`, I missed that because `asset_id` is never defined. 
But the problem is that our `GasPrices` storage is deprecated and not suitable for multi-asset fees

Now we store `GasPrices` assuming that we can only have one record 
```
GAS_PRICES: {
    key: 'gas_prices';
    value: Jsonified<GasPrices>;
  };
``` 
Instead, we should use the `assetId` as the key
```
GAS_PRICES: {
    key: Jsonified<Required<GasPrices>['assetId']['inner']>;
    value: Jsonified<GasPrices>;
  };
``` 
But in that case, I expect we'll run into an error, because pd purposely omits the assetId for native `GasPrices`, but that's bad for us, because undefined can't be a key in indexed-db

We will have to manually add the `assetId` for native `GasPrices` or come up with another trick

Also, note that the getGasPrices function will also have to be modified 
```typescript
async getGasPrices(): Promise<GasPrices | undefined> {
    const jsonGasPrices = await this.db.get('GAS_PRICES', 'gas_prices');
    if (!jsonGasPrices) return undefined;
    return GasPrices.fromJson(jsonGasPrices);
  }
```  

Also need to change the **_gasPrices rpc_** for the view service so that it retrieves `alt_gas_prices`

And the block processor logic that should save `alt_gas_prices` to indexed-db when needed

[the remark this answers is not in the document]
Valentine1898: It might make sense to consider this as a separate PR afterwards (but keep in mind that alternative fees won't work correctly)

apps/minifront/src/components/send/send-form/index.tsx
```diff
@@ -100,11 +108,20 @@ export const SendForm = () => {
...
       />
+      {showNonNativeFeeWarning && (
+        <div className='rounded border border-yellow-500 bg-gray-800 p-4 text-yellow-500'>
+          <strong>Privacy Warning:</strong>
+          <span className='block'>
+            Using non-native tokens for transaction fees may pose a privacy risk. It is recommended
+            to use the native token (UM) for better privacy and security.
+          </span>
+        </div>
+      )}
```
Valentine1898, line 119: it makes sense to make this a separate component to avoid duplication

packages/services/src/view-service/fees.ts
Valentine1898, line 1: move this to `packages/services/src/view-service/transaction-planner/fees.ts` if you only use it for the planner

packages/storage/src/indexed-db/index.ts
```diff
@@ -824,4 +812,25 @@ export class IndexedDb implements IndexedDbInterface {
...
     };
   }
+
+  fetchStakingTokenId(): AssetId {
+    const registryClient = new ChainRegistryClient();
+    const registry = registryClient.get(this.chainId);
+    const stakingToken = registry.stakingAssetId;
+
+    return stakingToken;
+  }
```
Valentine1898, line 822: note that you can just use `this.stakingTokenAssetId`

packages/ui/components/ui/tx/view/registry.tsx
Valentine1898: Given that we can just change the planner, a good solution seems to be to have the planner always add the assetId to the fee.
But if we have reasons not to, I suggest renaming this function to `getFeeAssetMetadataOrDefault()` because the name `getStakingTokenMetaData()` may cause confusion with other functions that are named the same but have a different implementation

Valentine1898 (approved):
@TalDerei Great work!

packages/storage/src/indexed-db/index.ts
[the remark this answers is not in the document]
TalDerei: tracking in https://github.com/penumbra-zone/web/issues/1310

packages/storage/src/indexed-db/index.ts
TalDerei: simplified to using `this.stakingTokenAssetId` defined in the constructor

PR #1270, opened 2024-06-10
feat(extension): onboarding frontends selection

### LINKED ISSUE #1230: Onboarding frontend selection
In the extension, we have a link that goes to minifront. For most users, this will be their default gateway into Penumbra.

Screenshot 2024-05-31 at 3 32 20 PM

At the moment, we are hardcoding our own testnet url that serves as the default:
https://github.com/penumbra-zone/web/blob/4f8c150f0f4471567dc9418758d495d8bc9313bf/apps/extension/.env.testnet#L3

This can be updated in `Settings --> Connected Sites`

Screenshot 2024-05-31 at 3 35 35 PM

In order to change this, the user must connect to another penumbra frontend and then they can star ⭐️ it.

### Proposal

During user onboarding, similar to how we handle RPC selection, we also require users to select their "Default frontend".  We should remove the `connected sites` frontend setting and make it its own settings page that re-uses the onboarding components. 

Will require adding default choices to the [Prax Registry](https://github.com/prax-wallet/registry/blob/main/tools/compiler/src/parser.rs#L12-L19) which we will source from validators.

Also! Because every node will bundle a frontend (e.g. https://grpc.testnet.penumbra.zone/app/), at the bottom should be an option to "Use RPC frontend". Because this does not receive frequent updates, we shouldn't put this at the top though.

VanishMax:
Closes #1230 

The idea is the following:
1. Remove the hard-coded MINIFRONT_URL constant, use Prax registry to import the frontend options instead
2. Show a new onboarding screen for default frontend selection. The options are taken from the `frontends` field in the Prax registry plus an RPC field from the previous onboarding step. Also, a custom URL can be set and stored per the session
3. Reuse this onboarding screen for the same-named settings page
4. A redirect from the "manage portfolio" button to the settings page in case users don't have a frontend selected

https://github.com/penumbra-zone/web/assets/29180358/a7d0fc2c-4859-4252-9388-0e5eef432bd0


apps/extension/src/routes/popup/home/index.tsx
```diff
@@ -61,13 +62,11 @@ export const PopupIndex = () => {
...
-            <Button className='flex w-full items-center gap-2' variant='gradient'>
-              Manage portfolio <ExternalLink size={16} />
-            </Button>
-          </a>
-        )}
+        <a href={frontendUrl} target='_blank' rel='noreferrer'>
+          <Button className='flex w-full items-center gap-2' variant='gradient'>
+            Manage portfolio <ExternalLink size={16} />
+          </Button>
+        </a>
```
Valentine1898, line 69: For existing wallets, `frontendUrl` may not be set. In this case click on this button does not lead to anything

apps/extension/src/routes/popup/settings/settings-connected-sites/known-site.tsx
```diff
@@ -44,14 +59,23 @@ export const KnownSite = ({
...
+              onClick={onToggleDefault}
+            >
+              {isDefaultFrontend ? <StarFilledIcon /> : <StarIcon />}
+            </TooltipTrigger>
+            <TooltipContent>
+              {isDefaultFrontend ? 'Primary Penumbra frontend' : 'Set as primary Penumbra frontend'}
+            </TooltipContent>
+          </Tooltip>
+        </TooltipProvider>
+      )}
```
Valentine1898, line 76: Is this how it's intended that we can set the default frontend through both the` connected sites` settings and the `default frontend` setting?

apps/extension/src/routes/popup/home/index.tsx
VanishMax: Good catch, thanks! I added a redirect to the settings page on this button in case the frontend URL is not set

apps/extension/src/routes/popup/settings/settings-connected-sites/known-site.tsx
VanishMax: Misunderstood this part of the task. Actually, the stars were meant to be removed from this page, so the "Default frontend" settings page would be the only entry point into this

VanishMax:
Thanks @Valentine1898 and @grod220 for the comments. I improved the designs and logic of this feature according to your comments. Updated the video in the PR description, so you can check how it works now

apps/extension/.env.testnet
```diff
@@ -1,3 +1,2 @@
 IDB_VERSION=42
-MINIFRONT_URL=https://app.testnet.penumbra.zone
```
grod220, line 2: Really cool to finally get this .env file down to these two. Think we can rename to just `.env` and change the webpack config too:

```diff
- // Loads default vars from `.env.testnet` & `.env.mainnet` file in this directory.
- // Reference package.json build script.
- // TODO: add `.env.mainnet` when ready.
- //       const envPath = process.env['NODE_ENV'] === 'mainnet' ? '.env.mainnet' : '.env.testnet';
- //       dotenv.config({ path: envPath });
- dotenv.config({ path: '.env.testnet' });
+ // Loads default vars from `.env` file in this directory.
+ dotenv.config({ path: '.env' });

```

apps/extension/src/routes/popup/home/frontend-link.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { MouseEventHandler } from 'react';
+import { usePopupNav } from '../../../utils/navigate';
+import { PopupPath } from '../paths';
+
+export const FrontendLink = () => {
+  const frontendUrl = useStore(getDefaultFrontend);
+  const navigate = usePopupNav();
+
+  const onClick: MouseEventHandler = event => {
+    if (frontendUrl) return;
```
grod220, line 14: I think I'm confused by the `<a>` and `onClick` usage below. This is the case where a default frontend is not set right?

Would you be able to add some code comments so it's clearer for folks?

apps/extension/src/shared/components/default-frontend-form/new-frontend-input.tsx
```diff
@@ -0,0 +1,91 @@
...
+  useRef,
+  useState,
+} from 'react';
+import { SelectList } from '@penumbra-zone/ui/components/ui/select-list';
+import { cn } from '@penumbra-zone/ui/lib/utils';
+import { isValidUrl } from '../../utils/is-valid-url';
+
+const getUrlValidity = (url: string, initialUrl?: string): boolean => {
+  return isValidUrl(url) && url !== initialUrl;
+};
```
grod220, line 16: Can we rename `getUrlValidity` to something like `isValidAndNotInitial()`? So it makes the boolean return make more sense?

apps/extension/src/shared/components/default-frontend-form/new-frontend-input.tsx
```diff
@@ -0,0 +1,91 @@
...
+  };
+};
+
+interface NewFrontendInputProps {
+  defaultFrontend?: string;
+  selected: boolean;
+  onSelect: (url: string) => void;
+}
+
+export const NewFrontendInput = forwardRef<HTMLInputElement, NewFrontendInputProps>(
```
grod220, line 36: The RPC selector step has a need for a custom field as well. This feels like it would be a good candidate to extract the custom input out as a shared component. What do you think?

apps/extension/src/shared/components/default-frontend-form/use-is-focus.ts
```diff
@@ -0,0 +1,21 @@
+// Subscribe to the focus state of any given element
```
grod220, line 1: Can you move this comment down to right above `useInFocus`?

grod220 (approved):
Great job!

Valentine1898 (approved):

apps/extension/src/shared/components/default-frontend-form/new-frontend-input.tsx
VanishMax: I made the custom RPC input look the same as this one. Still, they have different logic for now, which we can possibly unify in the future

VanishMax:
@grod220 Thanks for the proposals, I implemented them all. Merging soon

PR #1271, opened 2024-06-10
Remove '0' from token swap input

jessepinho:
Numbers being used as booleans need to be coerced to a boolean in component markup or else they'll show up as `0` when they're equal to `0`. 

## Before
image

## After
image

### DIFF
--- a/apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
+++ b/apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
@@ -110,16 +110,18 @@ export const TokenSwapInput = () => {
             </div>
           </div>
         </div>
-        {priceHistory.startMetadata && priceHistory.endMetadata && priceHistory.candles.length && (
-          <CandlestickPlot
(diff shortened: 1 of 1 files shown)


apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -110,16 +110,18 @@ export const TokenSwapInput = () => {
...
-            candles={priceHistory.candles}
-            startMetadata={priceHistory.startMetadata}
-            endMetadata={priceHistory.endMetadata}
-            latestKnownBlockHeight={Number(latestKnownBlockHeight)}
-            getBlockDate={getBlockDate}
-          />
-        )}
+        {priceHistory.startMetadata &&
+          priceHistory.endMetadata &&
+          !!priceHistory.candles.length && (
```
jessepinho, line 115: This `!!` is the only meaningful change in this file; the rest is whitespace.

Valentine1898 (approved):

PR #1272, opened 2024-06-10
On hold: fix validator exchange rate

### LINKED ISSUE #1239: Expected shares discrepency bug 🐞
```
Error: [unknown] Tendermint error 1: failed to deliver transaction: executing transaction: 
given 1000000 unbonded stake, expected 795181 delegation tokens but description produces 795423
```

https://github.com/penumbra-zone/web/assets/16624263/8ed30a6d-3e5a-484c-84f9-8b356935d72e

Valentine1898:
close #1239 

context: 

> I think the problem is that we are taking `validatorExchangeRate` to create delegate transaction plan from `validatorInfo` instead of using `CurrentValidatorRate rpc `
> 
> There is a potential problem with `validatorInfo`, which is that we update the validator data in indexed-db when transitioning epochs, but we do it in the last block of the epoch that ends, not in the first block of the new epoch. 
> see: 
> https://github.com/penumbra-zone/web/blob/fe0d7ae1cdaa7a35bc833971d862c3ec4875ca0e/packages/query/src/block-processor.ts#L353-L356
> 
> I can confirm that `validatorExchangeRate` obtained from `validatorInfo` and `validatorExchangeRate` obtained by `CurrentValidatorRate rpc` have different values for the same validator
> 
> image
> 
> I also noticed that pcli uses `CurrentValidatorRate rpc` instead of using `validatorExchangeRate` from `validatorInfo`
> 
> https://github.com/penumbra-zone/penumbra/blob/da3ef0a312c193d74c55bfd0282dad2affdb93d0/crates/bin/pcli/src/command/tx.rs#L503-L508

### DIFF
--- a/.changeset/clever-months-learn.md
+++ b/.changeset/clever-months-learn.md
@@ -1,0 +1,6 @@
+---
+'@penumbra-zone/services': minor
+'minifront': minor
+---
+
+added currentValidatorRate rpc
--- a/.changeset/clever-months-learn.md
+++ b/.changeset/clever-months-learn.md
@@ -1,6 +1,0 @@
----
-'@penumbra-zone/services': minor
-'minifront': minor
----
-
-added currentValidatorRate rpc
--- a/apps/minifront/src/state/staking/index.ts
+++ b/apps/minifront/src/state/staking/index.ts
@@ -22,10 +22,7 @@ import {
   getDisplayDenomFromView,
   getValidatorInfoFromValueView,
 } from '@penumbra-zone/getters/value-view';
-import {
-  getRateData,
-  getVotingPowerFromValidatorInfo,
-} from '@penumbra-zone/getters/validator-info';
+import { getVotingPowerFromValidatorInfo } from '@penumbra-zone/getters/validator-info';
 import {
   getVotingPowerByValidatorInfo,
   isDelegationTokenForValidator,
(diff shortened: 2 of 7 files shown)


TalDerei:
should [updatePriceForValidatorDelegationToken](https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L586) also be updated to reflect using the `CurrentValidatorRate` rpc?

can we get ride of the [getRateData](https://github.com/penumbra-zone/web/blob/main/packages/getters/src/validator-info.ts#L11) getter?

TalDerei (commented):
should [updatePriceForValidatorDelegationToken](https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L586) also be updated to reflect using the `CurrentValidatorRate` rpc?

can we get ride of the [getRateData](https://github.com/penumbra-zone/web/blob/main/packages/getters/src/validator-info.ts#L11) getter?

Valentine1898:
> should [updatePriceForValidatorDelegationToken](https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L586) also be updated to reflect using the `CurrentValidatorRate` rpc?

Good question, we probably don't want to run a couple hundred extra queries while scanning blocks 
Actually I wonder why `validatorInfo` might contain `validatorExchangeRate` values that are not up to date. Maybe you know who from the core team can help us with this? 

> can we get ride of the [getRateData](https://github.com/penumbra-zone/web/blob/main/packages/getters/src/validator-info.ts#L11) getter?

I'm in no rush to remove getters that aren't being used, because in theory it could be useful to other consumers of the getters package

erwanor:
Wait, shouldn't the validator exchange rates be updated by the block processor? if they're not, what if made it do that instead of doing a leaky RPC on every delegation/undelegation request

Valentine1898:
> Wait, shouldn't the validator exchange rates be updated by the block processor? if they're not, what if made it do that instead of doing a leaky RPC on every delegation/undelegation request

block processor updates `validatorInfo`, and `validatorInfo` contains rate data. But the problem is that rate data retrieved from validatorInfo sometimes does not equal to rate data retrieved from `CurrentValidatorRate rpc `

Can rate data change during an epoch?

erwanor:
Generally speaking, rate data only ever change at epoch boundaries and any deviation from this behavior is a serious bug. I think what's more likely is that the calls happened in different epoch, and that the bug we're observing is due to lossy conversions when storing to IDB.

IMO it would be best to investigate this conjecture ("what happens when we record the exchange rate to IDB") before considering doing those RPC calls on each staking actions

Valentine1898:
> Generally speaking, rate data only ever change at epoch boundaries and any deviation from this behavior is a serious bug. I think what's more likely is that the calls happened in different epoch, and that the bug we're observing is due to lossy conversions when storing to IDB.
> 
> IMO it would be best to investigate this conjecture ("what happens when we record the exchange rate to IDB") before considering doing those RPC calls on each staking actions

Okay, that makes sense. 

I'll hold this PR until I investigate why rate data in indexed-db is outdated

Valentine1898:
closed as there is a better solution to this issue which will be fixed in core repo 

https://discord.com/channels/824484045370818580/1044457038886998137/1250096780880908298

PR #1274, opened 2024-06-11
Scope storage

turbocrime:
Moves all chrome extension storage access into the chrome extension.


packages/services-context/src/index.ts
```diff
@@ -10,9 +9,9 @@ import {
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
 import { ChainRegistryClient } from '@penumbra-labs/registry';
 import { AppParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/app/v1/app_pb';
-import { Jsonified } from '@penumbra-zone/types/jsonified';
 
 export interface ServicesConfig {
+  readonly chainId: string;
```
turbocrime, line 14: chainId is now a required part of ServicesConfig

apps/extension/src/wallet-services.ts
```diff
@@ -6,21 +6,49 @@ import {
...
+
+      if (!chainId) throw new Error('No chainId available');
+
       const services = new Services({
+        chainId,
         idbVersion: IDB_VERSION,
         grpcEndpoint,
         walletId: WalletId.fromJsonString(wallet.id),
         fullViewingKey: FullViewingKey.fromJsonString(wallet.fullViewingKey),
       });
```
turbocrime, line 37: chainId is retrieved from existing storage, or queried before launching services-context

it is also stored in idb, but we can't connect to idb without a chainid, as the database name includes chainid.

this could be simplified - we could store chainId as a string, instead of all params, and then we can retrieve params from idb. but that may require a storage migration at this point

packages/services/src/view-service/wallet-id.ts
turbocrime, line 1: this was the only rpc impl that changed. it just serves a context key directly, now. i deleted the tests

apps/extension/src/wallet-services.ts
```diff
@@ -6,21 +6,49 @@ import {
...
+
+      const chainId =
+        (await localExtStorage
+          .get('params')
+          .then(json => json && AppParameters.fromJson(json).chainId)) ??
+        (await createPromiseClient(AppService, createGrpcWebTransport({ baseUrl: grpcEndpoint }))
+          .appParameters({})
+          .then(({ appParameters }) => appParameters?.chainId));
+
+      if (!chainId) throw new Error('No chainId available');
```
grod220, line 29: This is challenging readability-wise, can you make this a helper function:

```ts
const getChainId = async (baseUrl: string) => {
  // Attempt from storage first
  const fromStorage = await localExtStorage
    .get('params')
    .then(json => json && AppParameters.fromJson(json).chainId);
  if (fromStorage) return fromStorage;

  // else make request from rpc
  const chainIdFromRpc = await createPromiseClient(AppService, createGrpcWebTransport({ baseUrl }))
    .appParameters({})
    .then(({ appParameters }) => appParameters?.chainId);
  if (!chainIdFromRpc) throw new Error('No chainId available');

  return chainIdFromRpc;
};


const chainId = await getChainId(grpcEndpoint);

```

apps/extension/src/wallet-services.ts
```diff
@@ -6,21 +6,49 @@ import {
...
         walletId: WalletId.fromJsonString(wallet.id),
         fullViewingKey: FullViewingKey.fromJsonString(wallet.fullViewingKey),
       });
 
-      // initialize
-      await services.getWalletServices();
+      const { blockProcessor, indexedDb } = await services.getWalletServices();
+
+      // sync the IndexedDb last block number with chrome.storage.local
+      // Later used in Zustand store
```
grod220, line 42: Probably nothing we should do now, but feels like a code-smell that we have three forms of storage for sync height that need to be in sync.

apps/extension/src/wallet-services.ts
```diff
@@ -29,7 +57,13 @@ export const startWalletServices = () =>
           case false:
             return false;
           case ServicesMessage.ClearCache:
-            void services.clearCache().finally(() => respond());
+            void (async () => {
+              blockProcessor.stop('clearCache');
+              await Promise.allSettled([localExtStorage.remove('params'), indexedDb.clear()]);
```
grod220, line 62: If this throws an error, it seems to be swallowed right? Is there a way to respond with an error?

packages/services-context/src/index.ts
```diff
@@ -45,12 +44,10 @@ export class Services implements ServicesInterface {
...
    */
-  private async getParams(querier: RootQuerier): Promise<AppParameters> {
-    // try to read params from storage
-    const storedParams = await localExtStorage
-      .get('params')
-      .then(j => j && AppParameters.fromJson(j))
-      .catch(() => undefined);
+  private async getParams(indexedDb: IndexedDb, querier: RootQuerier): Promise<AppParameters> {
+    // try to read params from idb
+    const storedParams = await indexedDb.getAppParams();
```
grod220, line 49: The function code comment seems to imply this is a fallback, probably needs updating:

```
   * Attempt to fetch the AppParameters from the remote fullnode, or fall back
   * to the local storage. Will throw to abort if the remote node reports an
```

grod220 (approved):

apps/extension/src/wallet-services.ts
turbocrime: an error here will throw to console, which is about all we can do anyway, as the calling location isn't set up to handle an error response. the extension will restart regardless. an error happening in this case is quite catastrophic and there is no clear way to handle it aside from reinstalling, so i think this is fine.

PR #1276, opened 2024-06-11
check height of incoming block

turbocrime:
prevents https://github.com/penumbra-zone/penumbra/issues/4577 in the web block processor

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -119,7 +119,6 @@ export class BlockProcessor implements BlockProcessorInterface {
   private async syncAndStore() {
     // start at next block, or genesis if height is undefined
     let currentHeight = (await this.indexedDb.getFullSyncHeight()) ?? -1n;
-    const startHeight = currentHeight + 1n;
 
     // this is the first network query of the block processor. use backoff to
     // delay until network is available
@@ -132,7 +131,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       { retry: () => true },
     );
 
-    if (startHeight === 0n) {
+    if (currentHeight === -1n) {
       // In the `for` loop below, we only update validator infos once we've
       // reached the latest known epoch. This means that, if a user is syncing
       // for the first time, they could experience a broken UI until the latest
@@ -146,13 +145,17 @@ export class BlockProcessor implements BlockProcessorInterface {
     // this is an indefinite stream of the (compact) chain from the network
     // intended to run continuously
     for await (const compactBlock of this.querier.compactBlock.compactBlockRange({
-      startHeight,
+      startHeight: currentHeight + 1n,
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
-      if (compactBlock.height !== ++currentHeight) {
-        throw new Error(`Unexpected block: Recieved ${compactBlock.height} at ${currentHeight}`);
+      // confirm block height to prevent corruption of local state
+      if (compactBlock.height === currentHeight + 1n) {
+        currentHeight = compactBlock.height;
+      } else {
+        throw new Error(`Unexpected block height: ${compactBlock.height} at ${currentHeight}`);
       }
+
       if (compactBlock.appParametersUpdated) {
         await this.indexedDb.saveAppParams(await this.querier.app.appParams());
       }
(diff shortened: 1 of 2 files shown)


packages/query/src/block-processor.ts
```diff
@@ -207,6 +150,9 @@ export class BlockProcessor implements BlockProcessorInterface {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
+      if (compactBlock.height !== ++currentHeight) {
+        throw new Error(`Unexpected block: Recieved ${compactBlock.height} at ${currentHeight}`);
+      }
```
grod220, line 155: What are the cases where this would happen? Why would startHeight ever be out of sync with what the response returns?

Is the bug that sometimes the query does not respect the startHeight we pass?

grod220: Does throwing an error correct the issue? Aka, after retry, would it be fixed?

hdevalence: Having the check is a good defense in depth. You are correct that the height should be set by the request. However, we built a machine that corrupts RPC responses and put it in front of the default testnet RPC.

None of the linked issues have investigated whether there is an actual bug in pd or not. This would be a good next step.

Alternatively, we could avoid corrupting RPC responses just by not choosing to load balance requests over nodes that have inconsistent views of the chain state.

packages/query/src/block-processor.ts
```diff
@@ -207,6 +150,9 @@ export class BlockProcessor implements BlockProcessorInterface {
       keepAlive: true,
       abortSignal: this.abortController.signal,
     })) {
+      if (compactBlock.height !== ++currentHeight) {
```
grod220, line 153: The ordering of `++currentHeight` and `currentHeight++` tends to be something that confuses developers. 

Think separate lines would be better:

```ts
currentHeight++;
if (compactBlock.height !== currentHeight) {
  ...
}
```

grod220: Can we add more details as a code comment above so later developers don't accidentally delete this thinking its unnecessary?

```ts
      // The compact block has a chance not to respect height ordering... (more details)
      // We manual track height to ensure it is valid... (more details)
```

packages/query/src/block-processor.ts
```diff
@@ -207,6 +150,9 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+        throw new Error(`Unexpected block: Recieved ${compactBlock.height} at ${currentHeight}`);
```
grod220, line 154:
```diff
- Recieved
+ Received
```

packages/query/src/block-processor.ts
```diff
@@ -116,67 +116,10 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-    }
-    return { relevantTx, recordsWithSources };
-  }
-
   private async syncAndStore() {
-    // start at next block, or 0 if height is undefined
-    const startHeight = ((await this.indexedDb.getFullSyncHeight()) ?? -1n) + 1n;
+    // start at next block, or genesis if height is undefined
+    let currentHeight = (await this.indexedDb.getFullSyncHeight()) ?? -1n;
+    const startHeight = currentHeight + 1n;
```
grod220, line 122: Let's combine this into one:

```ts
    let currentHeight = (await this.indexedDb.getFullSyncHeight()) ?? 0n;

    if (currentHeight === 0n) {
      // ...
    }

    for await (const compactBlock of this.querier.compactBlock.compactBlockRange({
      startHeight: currentHeight + 1n,
      keepAlive: true,
    })) {
      currentHeight++;
      if (compactBlock.height !== currentHeight) {
        throw new Error(`Unexpected block: Received ${compactBlock.height} at ${currentHeight}`);
      }
```

grod220 (approved):
A few comments, then good to merge.

packages/query/src/block-processor.ts
turbocrime: i and others have seen the issue in action.

after throwing, the automatic retry will reset the state and make another request. this will resolve the issue if the remote endpoint returns the expected start height.

if the remote endpoint does not return the expected start height, the user's local state will fail to sync.

PR #1292, opened 2024-06-13
fix: bech32m package exports

VanishMax:
Fixes the way packages are consumed: bech32m, protobuf, client, getters


VanishMax:
Weirdly, the linting and testing works locally. Any idea what causes it to fail in the CI?

jessepinho:
Hm, when I test it locally, `pnpm lint` and `pnpm test` both succeed. 

The lines that CI is complaining about are the ones where we return a call to `bech32mAssetId()`. When I "Go to definition" on those calls, it takes me to `packages/bech32m/dist/passet.d.ts`, which has a TypeScript error:

image

Perhaps tsconfig needs to be updated?

turbocrime:
we should talk about what's broken in the exports - it does work in example repos, and some consumers who had problems were actually using the incompatible node or stacks that did not support our compile target.

the way you're configuring exports here will not export types.

VanishMax:
@grod220 @turbocrime @jessepinho This PR is ready with 4 package fixes: bech32m, protobuf, client, getters. Tested the outputs in multiple ways, now it should be all correct.

Learned the difference between the `exports` field in package.json and `publishConfig.exports`. Apparently, pnpm workspace imports use the first field and therefore I configured it to point to `src/*.ts` files, while `publishConfig.exports` point to `dist/*.js` and `dist/*.d.ts` files.

FYI, the lints and tests were failing because CI doesn't do `pnpm build` before running checks. Changing the `exports` field fixed the issues

grod220 (approved):

jessepinho:
@VanishMax I'd love to hear more about what you've learned during one of our weekly syncs. (And @turbocrime you too, if you're interested.) I feel like these issues have been plaguing us for a while and I don't really understand what goes into fixing them.

Issue #1093, opened 2024-05-10
Reactivity of wallet state


grod220:
In the [client package](https://github.com/penumbra-zone/web/blob/main/packages/client/src/prax.ts) we expose methods that frontends will use to connect to our wallet. After working with the cosmos-kit, think we should consider some enhancements.

## cosmos-kit library

Cosmology's [cosmos-kit](https://github.com/cosmology-tech/cosmos-kit) is a standard in cosmos for connecting to wallets. They have a library managing the core logic and packages (such as [@cosmos-kit/react](https://github.com/cosmology-tech/cosmos-kit/tree/main/packages/react)) that provide helpers for various frontend frameworks. For our case, we are using the react package to access react hooks such as [useWallet](https://github.com/cosmology-tech/cosmos-kit/blob/87eb8b9f3b98fa8c5e2b4d46f6df5d70ce32692d/packages/react-lite/src/hooks/useWallet.ts#L11-L50) which gives us access to the user's wallet state:

```typescript
import { useWallet } from '@cosmos-kit/react';
const { status } = useWallet()

switch (status) {
  case WalletStatus.Disconnected:
    return <WalletButtonBase buttonText='Connect Wallet' onClick={onClickConnect} />;
  case WalletStatus.Connecting:
    return ...
  case WalletStatus.Connected:
   return ...
  case WalletStatus.Error:
   return ...
}
```

The hook is reactive. So whenever the user's wallet state changes, the UI automatically reflects the correct state.

Their API that they expose in [WalletContext](https://github.com/cosmology-tech/cosmos-kit/blob/87eb8b9f3b98fa8c5e2b4d46f6df5d70ce32692d/packages/core/src/types/hook.ts#L51-L130) is helpful to review as well. It has wallet and chain state:

```typescript
export interface ChainWalletContext {
  chain: Chain;
  assets: AssetList | undefined;
  wallet: Wallet;
  logoUrl: string | undefined;
   ...
}
```

Wallet connection status:
```typescript
export interface ChainWalletContext {
   ...
  isWalletDisconnected: boolean;
  isWalletConnecting: boolean;
  isWalletConnected: boolean;
  isWalletRejected: boolean;
  isWalletNotExist: boolean;
  isWalletError: boolean;
   ...
}
```

and actions the user can take

```typescript
export interface ChainWalletContext {
   ...
  connect: () => Promise<void>;
  disconnect: (options?: DisconnectOptions) => Promise<void>;
   ...
}
```

[...]

PR #1293, opened 2024-06-13
react wrapper for penumbra transport and client use

turbocrime:
current status:

provider and hooks work.

but the provider is complex. in order to expose a single 'state' that represents not only connection status but also any actions taken internally, or by a caller, and to represent any pending status of various promises, plus failures, it's necessary to track a lot of state. this could be refactored into a useReducer state machine.

minifront edits are mostly for testing use of the provider and hooks, and will be reverted in favor of a larger PR.

check out the current readme. https://github.com/penumbra-zone/web/blob/1cb96c4e7487269296022b296fd4d115f264422c/packages/react/README.md

much testing code is still present, will clean up soon.

### DIFF
--- a/.changeset/dirty-mice-help.md
+++ b/.changeset/dirty-mice-help.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/react': major
+---
+
+initial react wallet
--- a/packages/react/src/components/penumbra-provider.tsx
+++ b/packages/react/src/components/penumbra-provider.tsx
@@ -1,0 +1,188 @@
+import { PenumbraInjection } from '@penumbra-zone/client';
+import { useCallback, useEffect, useMemo, useState, type ReactNode } from 'react';
+import { PenumbraManifest } from '../manifest';
+import { PenumbraContext, penumbraContext, PenumbraProviderState } from '../penumbra-context';
+import { assertManifestOrigin, injectionOfKey, keyOfInjection } from '../util';
+
+type PenumbraProviderProps = {
+  children?: ReactNode;
+  injection?: PenumbraInjection;
+  origin?: string;
+  makeApprovalRequest?: boolean;
+} & ({ injection: PenumbraInjection } | { origin: string });
+
+export const PenumbraProvider = ({
+  children,
+  origin: providerOrigin,
+  injection: providerInjection,
+  makeApprovalRequest = false,
+}: PenumbraProviderProps) => {
+  providerOrigin ??= keyOfInjection(providerInjection);
+  providerInjection ??= injectionOfKey(providerOrigin);
+
+  const [providerState, setProviderState] = useState<PenumbraProviderState>();
+
+  const [failure, setFailureState] = useState<Error>();
+
+  const setFailure = useCallback(
+    (cause: unknown) => {
+      setFailureState(cause instanceof Error ? cause : new Error('Unknown failure', { cause }));
+      setProviderState(PenumbraProviderState.Failed);
+    },
+    [setFailureState, setProviderState],
+  );
+
+  const [port, setPort] = useState<MessagePort>();
+
+  const [manifest, setManifest] = useState<PenumbraManifest>();
+
+  const [providerConnected, setProviderConnected] = useState(providerInjection?.isConnected());
+
+  const [calledConnect, setCalledConnect] = useState(false);
+  const [calledRequest, setCalledRequest] = useState(false);
+  const [calledDisconnect, setCalledDisconnect] = useState(false);
+
+  const createdContext: PenumbraContext = useMemo(
+    () => ({
+      failure,
+      manifest,
+      origin: providerOrigin,
+      state: providerState,
+      ...(manifest && !failure
+        ? {
+            port,
+            connect:
+              providerInjection?.connect &&
+              (async () => {
+                setCalledConnect(true);
+                await providerInjection.connect();
+              }),
+            request:
+              providerInjection?.request &&
+              (async () => {
+                setCalledRequest(true);
+                await providerInjection.request();
+              }),
+            disconnect:
+              providerInjection?.disconnect &&
+              (async () => {
+                setCalledDisconnect(true);
+                await providerInjection.disconnect();
+              }),
+          }
+        : {}),
+    }),
+    [
+      failure,
+      manifest,
+      port,
+      providerInjection?.connect,
+      providerInjection?.connect,
+      providerInjection?.disconnect,
+      providerOrigin,
+      providerState,
+    ],
+  );
+
+  // track state including what methods have been called
+  useEffect(() => {
+    setProviderConnected(providerInjection?.isConnected());
+
+    if (failure) setProviderState(PenumbraProviderState.Failed);
+    else if (!providerOrigin || !providerInjection) setProviderState(PenumbraProviderState.Absent);
+    else if (!manifest) setProviderState(PenumbraProviderState.Injected);
+    else if (calledDisconnect) setProviderState(PenumbraProviderState.Disconnected);
+    else if (providerConnected) setProviderState(PenumbraProviderState.Connected);
+    else if (calledRequest) setProviderState(PenumbraProviderState.Requested);
+    else if (calledConnect) setProviderState(PenumbraProviderState.Pending);
+    else setProviderState(PenumbraProviderState.Present);
+  }, [
+    calledConnect,
+    calledDisconnect,
+    calledRequest,
+    failure,
+    manifest,
+    providerConnected,
+    providerInjection,
+    providerOrigin,
+  ]);
+
+  // fetch manifest to confirm presence of provider
+  useEffect(() => {
+    // don't repeat
+    if (failure ?? manifest) return;
+    // don't attempt with no provider
+    if (!providerOrigin || !providerInjection) return;
+
+    // sync assertion
+    try {
+      assertManifestOrigin(providerOrigin, providerInjection);
+    } catch (cause) {
+      setFailure(cause);
+      return;
+    }
+
+    // async fetch, returning abort function for cancel
+    const ac = new AbortController();
+    void fetch(providerInjection.manifest, { signal: ac.signal })
+      .then(
+        async res => {
+          // this cast is fairly safe coming from an extension manifest, where
+          // schema is enforced by chrome store.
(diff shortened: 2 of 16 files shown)


packages/react/src/components/provider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { PenumbraProviderNotAvailableError, PenumbraSymbol } from '@penumbra-zone/client';
+import { penumbraContext } from '../hooks/penumbra';
+
+export const PenumbraProvider = ({
+  children,
+  providerOrigin,
+  requestApproval = false,
+  required = false,
+}: {
+  children?: ReactNode | ReactNode[];
```
VanishMax, line 11: suggestion: leave only `children?: ReactNode`. It already includes `Iterable<ReactNode>` under the hood, so the array is not needed

packages/react/src/components/provider.tsx
```diff
@@ -0,0 +1,44 @@
...
+  required?: boolean;
+}) => {
+  console.log('PenumbraProvider', { children, providerOrigin, requestApproval, required });
+  const provider = useMemo(() => window[PenumbraSymbol]?.[providerOrigin], [providerOrigin]);
+
+  const [requested, setRequested] = useState(false);
+  const [providerConnected, setProviderConnected] = useState<boolean>();
+  const [port, setPort] = useState<MessagePort>();
+
+  setProviderConnected(provider?.isConnected());
```
VanishMax, line 23: suggestion: change `providerConnected` to be a memo instead of state since it depends on the `provider` value and doesn't run `setProviderConnected` elsewhere

.changeset/dirty-mice-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@penumbra-zone/react': major
+---
+
+initial react wallet
```
VanishMax, line 5: issue: trying to connect the updated minifront to the extension (i use the extension the branch `turbocrime/use-disconnect` of Prax) results in `too many rerenders in PenumbraProvider`

image

VanishMax: See the comment below about `providerConnected` variable – it solves the problem

apps/minifront/src/components/layout.tsx
```diff
@@ -5,19 +5,24 @@ import { Toaster } from '@repo/ui/components/ui/toaster';
...
-      <div className='flex min-h-screen w-full flex-col'>
-        <Header />
-        <main className='flex size-full flex-1 px-4'>
-          <Outlet />
-        </main>
-        <Footer />
-      </div>
-      <Toaster />
-    </MotionConfig>
+    <PenumbraProvider providerOrigin={praxOrigin}>
```
VanishMax, line 14: question: does setting `providerOrigin` to Prax mean users can't connect another wallet or that they have to create a state variable that's keeping an dynamic origin for that purpose?

apps/minifront/src/components/header/menu/provider.tsx
```diff
@@ -1,29 +1,27 @@
...
+      );
     void getChainId().then(setChainId);
-  }, []);
+  }, [penumbra?.manifest]);
 
-  if (!providerManifest) return null;
+  if (!penumbra || !providerManifest) return null;
 
   return (
     <NavigationMenu.Root className='max-w-96'>
```
VanishMax, line 27: issue (minor): something in this component triggers the html warning in the devtools: `validateDOMNesting(...): <li> cannot appear as a descendant of <li>`. Small but annoying issue

apps/minifront/src/components/header/menu/provider.tsx
```diff
@@ -60,7 +59,11 @@ export const ProviderMenu = () => {
             </NavigationMenu.Link>
           </NavigationMenu.Item>
           <NavigationMenu.Item className={cn(...itemStyle)}>
-            <NavigationMenu.Link hidden className={cn(...linkStyle)} onSelect={disconnect}>
+            <NavigationMenu.Link
+              hidden
+              className={cn(...linkStyle)}
+              onSelect={() => void penumbra.disconnect()}
```
VanishMax, line 65: question: is the disconnect button intentionally hidden?

also, calling the disconnect should probably redirect user to the connection page. Right now it simply displays an error in the block syncing

apps/minifront/src/components/header/menu/provider.tsx
```diff
@@ -1,29 +1,27 @@
...
-    //window[Symbol.for('penumbra')][providerOrigin].disconnect(),
-  }, [providerOrigin]);
-
   useEffect(() => {
-    void getPraxManifest().then(m => setProviderManifest(m as ProviderManifest));
+    if (penumbra?.manifest)
+      void fetch(penumbra.manifest).then(async m =>
+        setProviderManifest((await m.json()) as ProviderManifest),
+      );
     void getChainId().then(setChainId);
```
VanishMax, line 21: question: I see a similar check quite often in the code. Why is it needed?

packages/react/src/hooks/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './penumbra';
+export * from './service';
+export * from './transport';
```
VanishMax, line 3: suggestion: use defined exports, e.g. `export { usePenumbra } from './penumbra';`. 

I couldn't find the link as a proof but apparently Vite's HMR lags for the star-exports and performs better with defined exports. Yet, no proof from my side, so feel free to discard it 😁

packages/react/src/components/provider.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const PenumbraProvider = ({
+  children,
+  providerOrigin,
+  requestApproval = false,
+  required = false,
+}: {
+  children?: ReactNode | ReactNode[];
+  providerOrigin: string;
+  requestApproval?: boolean;
+  required?: boolean;
```
VanishMax, line 14: thought: i don't think the `required` property is needed here. If users want to, they can throw errors by themselves. And generally throwing errors within components is not a good practice

VanishMax (commented):
Huge and extremely important work!

FYI: the _question_ comments are written solely out of my curiosity. They don't necessary require changes but the honest reply would be greatly appreciated

packages/react/src/components/provider.tsx
turbocrime: the return value of isConnected will change, without any change in object ref of provider. wrapping in useMemo would fail to update after the first call

.changeset/dirty-mice-help.md
turbocrime: ah, ok, that was a late change i made and didn't test. maybe my cache thinking was incorrect

apps/minifront/src/components/layout.tsx
turbocrime: minifront should implement UI for the user to select a preferred wallet. currently minifront only supports prax and larger change is out of scope

apps/minifront/src/components/header/menu/provider.tsx
turbocrime: rebase error, will fix

apps/minifront/src/components/header/menu/provider.tsx
turbocrime: the provider manifest is used for a couple things

1. it's only possible to serve the manifest from an extension with the expected extension id. this means we can confirm the actual presence of the extension with a known id.
2. the manifest contains metadata about the provider. name, description, version, icons, etc. and this is useful to display to the user.

turbocrime:
among other edits, the README.md of the react package now contains a state flowchart of the provider.

this could ship but it might be better to refactor as FSM with useReducer.

```mermaid
stateDiagram-v2
    classDef GoodState fill:green
    classDef BadState fill:red
    state global_exists <<choice>>
    [*] --> global_exists: window[Symbol.for('penumbra')][validOrigin]
    global_exists --> Absent: undefined
    global_exists --> Injected: defined
    state manifest_present <<choice>>
    Injected --> manifest_present: fetch(injection.manifest)
    manifest_present --> Present: json
    state make_request <<choice>>
    Present --> make_request: makeApprovalRequest
    note left of Present
        Methods on the injection may
        be called after this point.
    end note
    Requested --> Pending
    make_request --> Requested: true
    make_request --> Pending: false
    Pending --> Connected
    note left of Connected
        Port is acquired and
        transports become active.
    end note
    Present --> Disconnected
    Requested --> Disconnected
    Pending --> Disconnected
    Connected --> Disconnected
    manifest_present --> Failed
    Requested --> Failed
    Pending --> Failed
    Disconnected --> Failed
    Absent:::BadState --> [*]
    Failed:::BadState --> [*]
    Disconnected --> [*]
    Connected:::GoodState --> [*]
```

packages/react/src/components/penumbra-provider.tsx
grod220, line 1: All of this seems to imply that:

1 - By having this hook in a component, it will attempt to initiate connections automatically
2 - The core logic of reactive state setting+updating takes place in react-land
3 - The web app is responsible for maintaining state (aka requires context)

Sort of in the spirit of https://github.com/penumbra-zone/web/issues/1093, for consumers, I think it would be far easier if all of this lived on the injection. For example:

```ts
enum WalletStatus {
  Disconnected,
  Connecting,
  Connected,
  Error,
}

type WalletStatusCallback = (status: WalletStatus) => void;
type Unsubscribe = () => void;

export interface PenumbraInjection {
  readonly connect: () => Promise<MessagePort>;
  readonly request: () => Promise<void>;
  readonly disconnect: () => Promise<void>;
  readonly isConnected: () => boolean | undefined;
  readonly manifest: string;
  readonly subscribeToWalletStatus: (callback: WalletStatusCallback) => Unsubscribe;
}
```

And then the exposed react hook could be simplified to something like this without the need for any context/provider:

```ts
export function useWalletStatus(): WalletStatus {
  const [status, setStatus] = useState<WalletStatus>(WalletStatus.Disconnected);

  useEffect(() => {
    const penumbra = getPenumbraInjection();
    
    const unsubscribe = penumbra.subscribeToWalletStatus((newStatus) => {
      setStatus(newStatus);
    });

    return () => {
      unsubscribe();
    };
  }, []);

  return status;
}

```

VanishMax: Agree on this one

turbocrime:
closed to be rewritten after #1420

PR #1298, opened 2024-06-14
Update auctions in place

### LINKED ISSUE #1090: Rework Dutch auction state so that auctions are updated in place, rather than loading entirely new objects
Right now, when an auction is ended or withdrawn from, we reload the auction list on the left of the screen. That can result in the auctions being reordered, as well as a brief flicker on the screen as the components are destroyed and re-mounted.

Instead, we should probably store the auctions as:
1. A `Map` of auction IDs to auctions, AND
2. An array of auction IDs in the sort order that we want to keep them in.

That way, auctions that update will be updated in place without the UI glitches.

jessepinho:
Previously, ZQuery provided no way to update items in place when a stream was restarted. Instead, when a stream was restarted, ZQuery would set the state to `undefined`, then start processing streaming responses again.

With this PR's change, we can now handle streams a bit more intelligently. ZQuery no longer sets the state to `undefined` when restarting a stream; instead, it calls a new `onStart()` reducer (if it's passed to `createZQuery()`). Similarly, when a stream ends, errors, or is aborted, it calls handlers for those. It also calls an `onValue()` reducer when a value is received from the stream.

All those reducers are now returned by a `stream` function that consumers pass to `createZQuery()`, which allows consumers to maintain a JavaScript scope over the life of a stream. This allows you to update streamed items in place, and — in combination with `onEnd()` — discard any items from a previous stream that aren't present in the current stream. (See the implementation and docs for more detail.)

## Before
https://github.com/penumbra-zone/web/assets/1121544/bc55c820-3cc9-4a9c-acdc-ba6ddea0fa75

## After
https://github.com/penumbra-zone/web/assets/1121544/d67538c4-4f05-4033-be28-1b26b60952ad

Closes #1090

### DIFF
--- a/apps/minifront/src/components/swap/auction-list/index.tsx
+++ b/apps/minifront/src/components/swap/auction-list/index.tsx
@@ -49,7 +49,7 @@ export const AuctionList = () => {
       [...getFilteredAuctionInfos(auctionInfos.data ?? [], filter, status?.fullSyncHeight)].sort(
         SORT_FUNCTIONS[filter],
       ),
-    [auctionInfos, filter, status?.fullSyncHeight],
+    [auctionInfos.data, filter, status?.fullSyncHeight],
   );
 
   return (
--- a/.changeset/four-forks-turn.md
+++ b/.changeset/four-forks-turn.md
@@ -1,0 +1,8 @@
+---
+'@penumbra-zone/zquery': major
+'minifront': minor
+---
+
+Beef up ZQuery's handling of streams; take advantage of it in minifront.
+
+BREAKING CHANGE: The `stream` property passed to `createZQuery()` should now return an object containing at least an `onValue` method. See the docs for the `stream` property for more info.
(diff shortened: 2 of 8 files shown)


apps/minifront/src/components/swap/auction-list/index.tsx
```diff
@@ -49,7 +49,7 @@ export const AuctionList = () => {
       [...getFilteredAuctionInfos(auctionInfos.data ?? [], filter, status?.fullSyncHeight)].sort(
         SORT_FUNCTIONS[filter],
       ),
-    [auctionInfos, filter, status?.fullSyncHeight],
+    [auctionInfos.data, filter, status?.fullSyncHeight],
```
jessepinho, line 52: Not strictly relevant to this PR, but should prevent unnecessary rerenders.

grod220 (approved):
LGTM

PR #1300, opened 2024-06-14
disintegration

turbocrime:
makes interfaces reasonable

removes services-context package, querier package, storage package

creates block processor package, idb package

most of this diff is changing tests, which for some reason were heavily dependent on interfaces


VanishMax (approved):
Major changes here, looking all good to me! It's an approve from me but I believe it would be beneficial for other team members to look at it too.

packages/storage/src/version.ts
```diff
@@ -0,0 +1 @@
+export default 42;
```
TalDerei: can we export a variable, something like

```
/**
 */
export const IDB_VERSION = 42;
```

TalDerei:
remind me why we removed the querier in the block processor?

grod220:
I'll take a look at this more later in the afternoon, but after looking through 500+ lines of code, this does not feel like a reasonable chunk of code we can ship at once. There are meaningful changes to logic and code organization that we need to break down into multiple PRs.

PR #1305, opened 2024-06-14
publish useful packages

turbocrime:
use tsup for package builds, instead of vite lib mode.

build is much faster.

all remaining packages are correctly dependent, minimally chunked/bundled/polyfilled for target, and should be usable outside the repository.


package.json
turbocrime, line 1: tsup moved to workspace dep, vite build plugins removed.

polyfills moved to workspace dep.

tsup inlines polyfills, and it's not present as a dependency in published packages.

chrome types removed from root repo, are now present in packages that use them.

packages/bech32m/package.json
```diff
@@ -4,16 +4,18 @@
   "license": "(MIT OR Apache-2.0)",
   "description": "Tools for manipulating Penumbra bech32m strings",
   "type": "module",
+  "engine": {
+    "node": ">=22"
+  },
```
turbocrime, line 9: engine set to latest node stable in published packages. older versions start to have trouble with typescript and package features that we use.

packages/bech32m/package.json
```diff
@@ -4,16 +4,18 @@
...
   "scripts": {
-    "build": "tsup src/*.ts --dts --minify --format esm",
-    "clean": "rm -rfv dist",
+    "build": "tsup --clean src/*.ts --dts --minify --keep-names --format esm",
+    "clean": "rm -rfv dist package penumbra-zone-*.tgz",
     "lint": "eslint src",
+    "prepack": "$npm_execpath run build",
     "test": "vitest run"
   },
```
turbocrime, line 16: this is fairly representative of the scripts in most packages now.

i added --keep-names which i prefer, i expect consumers will bundle if they want mangled names.

packages/bech32m/package.json
```diff
@@ -4,16 +4,18 @@
...
   "files": [
-    "./dist",
-    "README.md",
-    "CHANGELOG.md"
+    "dist"
   ],
```
turbocrime, line 19: there is some automatic behavior so listing README and some other things is unnecessary

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

packages/bech32m/package.json
```diff
@@ -28,16 +30,13 @@
       "./*": {
         "types": "./dist/*.d.ts",
         "default": "./dist/*.js"
-      }
+      },
+      "./chunk-*": null
```
turbocrime, line 34: don't export tsup's chunked files. they are still used internally

packages/bech32m/tsconfig.json
```diff
@@ -1,9 +1,5 @@
...
-  "exclude": ["node_modules", "dist"],
-  "compilerOptions": {
-    "outDir": "dist",
-    "noEmit": false,
-    "declarationMap": false
-  }
+  "extends": "tsconfig/tsup.json",
+  "include": ["src"],
+  "exclude": ["node_modules", "dist"]
 }
```
turbocrime, line 5: tsup builds use a tsup.json base and simply list include/exclude folders (mostly identical)

apps/minifront/package.json
```diff
@@ -66,7 +62,6 @@
     "autoprefixer": "^10.4.19",
     "firebase-tools": "^13.9.0",
     "postcss": "^8.4.38",
-    "promise.withresolvers": "^1.0.3",
```
turbocrime, line 69: with node 22 in ci we no longer require the polyfill for tests

packages/bech32m/package.json
turbocrime: specifically, older versions cause problems with importing the package - you may be running an older version internally and not experience the problem.

TalDerei:
high-level question, what's generally the difference between building with `turbo` versus `tsup`?

turbocrime:
turbo is the task runner.

the change is away from vite build lib mode.

vite is primarily a website a bundler and required a lot of configuration to build packages for publishing/consumption. lib mode is definitely an afterthought, and the plugins we needed to achieve things simply like external deps or typescript output were not well maintained - config didnt merge well, there were quirks to work around, etc

tsup was something we looked at initially but was decided against. @VanishMax demonstrated a pretty concise use of it that seemed much nicer than our vite build configs.

so this publishes the remainder of the packages, applies package config that solves issues experienced by consumers, and uses tsup for all package builds, except wasm. wasm might be done with a bit more effort.

turbocrime:
tsc was also used directly for some builds, but there are internal deps like polyfills that we *do* want to inline.

vite and tsup both use esbuild and rollup internally.

turbocrime:
this also avoids api-extractor which is very slow. api-extractor has benefits and features over rollup but we don't use them, and it lags behind typescript releases. we can enable it with a tsup flag if we choose.

packages/client/tsup.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
```
VanishMax, line 3: praise: thanks for extracting the config to the file!

question: how do you think, does it deserve its own package? It could export a function with `entry` argument, so the API would be `export default defineTsup('src/*.ts')`

apps/minifront/vite.config.ts
```diff
@@ -5,6 +5,7 @@ import { commitInfoPlugin } from './src/utils/commit-info-vite-plugin';
 import polyfillNode from 'vite-plugin-node-stdlib-browser';
 
 export default defineConfig({
+  define: { 'globalThis.__DEV__': 'import.meta.env.DEV' },
```
grod220, line 8: Instead of defining a custom DEV property, just using the globally available `import.meta.env.MODE` feels more standard

packages/query/src/block-processor.ts
```diff
@@ -308,8 +308,26 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
       }
 
-      if (globalThis.ASSERT_ROOT_VALID) {
-        await this.assertRootValid(compactBlock.height);
+      if (globalThis.__ASSERT_ROOT__) {
```
grod220, line 311: The body of this function is quite large as is. Think the previous helper function is more readable.

packages/ui/components/ui/box.tsx
```diff
@@ -51,12 +50,6 @@ export const Box = ({
       layout={layout ?? !!layoutId}
       layoutId={layoutId}
       className={cn('flex flex-col gap-4', variants({ spacing, state, overflow }))}
-      /**
-       * Set the border radius via the style prop so it doesn't get distorted by framer-motion.
-       *
-       * @see https://www.framer.com/motion/layout-animations/##scale-correction
-       */
-      style={{ borderRadius: RESOLVED_TAILWIND_CONFIG.theme.borderRadius.lg }}
```
grod220, line 59: Does this effect styling somehow?

packages/ui/package.json
```diff
@@ -1,24 +1,87 @@
 {
   "name": "@penumbra-zone/ui",
   "version": "3.5.0",
-  "private": true,
+  "private": false,
```
grod220, line 4: Can just remove this line

grod220 (approved):

packages/ui/package.json
```diff
@@ -1,24 +1,87 @@
...
   "license": "(MIT OR Apache-2.0)",
   "type": "module",
+  "engine": {
+    "node": ">=22"
+  },
   "scripts": {
+    "build": "tsup",
```
VanishMax, line 11: thought: interestingly, I wouldn't even think about using tsup for a UI build but it seems to be working just fine. Thanks for showing a way to do it!

packages/ui/package.json
```diff
@@ -1,24 +1,87 @@
...
     "build-storybook": "storybook build",
+    "clean": "rm -rf dist package penumbra-zone-*.tgz",
     "lint": "eslint components lib",
+    "prepack": "$npm_execpath run build",
```
VanishMax, line 15: question: do we need `prepack`? The `packages-release.yml` workflow defines a build script with Turbo, so the build of a package would probably happen twice increasing the CI pipeline time.

VanishMax:
@turbocrime It is great to see how you took over the packages build configuration and perfected the use of `tsup`. Awesome changes!

packages/client/package.json
```diff
@@ -4,17 +4,17 @@
...
   "description": "Package for connecting to any Penumbra extension, including Prax.",
   "type": "module",
+  "engine": {
+    "node": ">=22"
+  },
   "scripts": {
-    "build": "tsup src/*.ts --dts --minify --format esm",
-    "clean": "rm -rfv dist",
+    "build": "tsup",
+    "clean": "rm -rfv dist package penumbra-zone-*.tgz",
```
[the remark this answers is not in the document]
turbocrime: outputs of `pnpm pack`

packages/ui/package.json
turbocrime: it's nicer when using from cli

TalDerei (commented):
before merging, can we ensure that peer deps and tsup config are in-fact working? There were some issues last night that I'm not sure were resolved

apps/minifront/vite.config.ts
turbocrime: `import.meta.env.MODE` and `import.meta.env.DEV` are part of vite-env, and aren't generally available in places where we might like to have a dev mode switch, such as every package

it's typical to just refer to process.env.NODE_ENV but technically wrong, and since the packages target browser runtime instead of node runtime and don't import node types, it makes typescript upset

declaring a `__DEV__` global and then checking that, and requiring consumers to define it to indicate dev mode, is safer (won't error) and becoming a convention. it's used by apollo and react-native

https://www.apollographql.com/docs/react/development-testing/reducing-bundle-size/#why-use-globalthis__dev__-instead-of-processenvnode_env

packages/query/src/block-processor.ts
turbocrime: reverted

packages/ui/components/ui/box.tsx
turbocrime: this particular export was giving me trouble as i tried to package the tailwind config, i can revert this now that it remains private

packages/client/tsup.config.ts
turbocrime: i started writing a response this morning, but then noticed tsup was creating issues with duplicating exports and bundling in consumers

today i ended up removing tsup, and using bare tsc for packaging. the packages are now structured well enough that this works.

packages/ui/package.json
turbocrime: tsconfig now uses `incremental`, and without api-extractor running the slowdown is minimal

turbocrime: removed the `prebuild` script as suggested

turbocrime:
> before merging, can we ensure that peer deps and tsup config are in-fact working? There were some issues last night that I'm not sure were resolved

tsup was not in fact working. turns out it mangles types due to some underlying esbuild issue, and that was what was causing peer issues and prevented services init.

i ended up removing tsup. all package builds now use bare tsc.

TalDerei (approved):
testing against https://github.com/prax-wallet/web/pull/34 seems to work for me

PR #1306, opened 2024-06-14
Stop using loader functions, part 1

### LINKED ISSUE #1243: Stop using loader functions

jessepinho:
This PR goes through a bunch of routes where we're using react-router loaders, and replaces loader functionality with ZQuery. Note that this means some assumptions have changed about when data is available. react-router loaders wait to render their route components until data has loaded, which means your component can assume the data is already present. Since we're now effectively loading data from `useEffect()` (by way of ZQuery), we can't assume that data is present just because a component is rendered. So you'll see some instances of code that accounts for `undefined` where we didn't previously have to.

One of the benefits of this change is that route changes are instant. There's no lag time between when you click on a link to a route, and when the route changes, like there was before when route changes were blocked by loaders.

## Before (note the delayed route changes after I click)

https://github.com/penumbra-zone/web/assets/1121544/7fbb50a0-4acf-4789-b0d2-51d99c49fcef

## After (note the instant route changes after I click)

https://github.com/penumbra-zone/web/assets/1121544/22a7dba1-3ae8-4fef-a753-5e3f8479669e

## In this PR
- Remove the balances loader and create a `balances` slice with a `balancesByAccount` ZQuery object.
- Remove the Send form's loader and create a `transferableBalancesResponses` ZQuery object in the send state.
- Create a `shared` Zustand slice and added a `stakingTokenMetadata` ZQuery object to it, to be used by the send form and the IBC UIs.
- Remove the staking page's loader; in its place, use `state.shared.stakingTokenMetadata`, and combine the `accountSwitcherFilter` and `unstakedTokensByAccount` into a single ZQuery object, since they both depend on the same data.
- Remove the transactions' page's loader; in its place, convert its slice to use a ZQuery object.

## Future PRs
- Convert the `SwapLoader` and `IbcLoader` to ZQuery. Those look a bit involved and this PR is already touching a lot of files, so I haven't done those yet.
- Convert more of the staking slice to ZQuery. That would be a bit more involved as well, and doesn't involve getting rid of loaders, so I figured I'd punt on that for now.

Relates to #1243

### DIFF
--- a/apps/minifront/src/abort-loader.ts
+++ b/apps/minifront/src/abort-loader.ts
@@ -28,8 +28,12 @@ const retry = async (fn: () => boolean, ms = 500, rate = Math.max(ms / 10, 50)) 
  * timeout. This is a temporary solution until loaders properly await Prax
  * connection.
  */
-export const abortLoader = async (): Promise<void> => {
+export const abortLoader = async (): Promise<null> => {
   await throwIfPraxNotInstalled();
   await retry(() => isPraxConnected());
   throwIfPraxNotConnected();
+
+  // Loaders are required to return a value, even if it's null. By returning
+  // `null` here, we can use this loader directly in the router.
+  return null;
 };
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -1,7 +1,5 @@
-import { LoaderFunction, useLoaderData } from 'react-router-dom';
 import { AddressIcon } from '@repo/ui/components/ui/address-icon';
 import { AddressComponent } from '@repo/ui/components/ui/address-component';
-import { BalancesByAccount, getBalancesByAccount } from '../../../fetchers/balances/by-account';
 import {
   Table,
   TableBody,
@@ -11,7 +9,6 @@ import {
   TableRow,
 } from '@repo/ui/components/ui/table';
 import { ValueViewComponent } from '@repo/ui/components/ui/tx/view/value';
-import { abortLoader } from '../../../abort-loader';
 import { EquivalentValues } from './equivalent-values';
 import { Fragment } from 'react';
 import { shouldDisplay } from './helpers';
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -15,16 +15,12 @@ import { abortLoader } from '../../../abort-loader';
 import { EquivalentValues } from './equivalent-values';
 import { Fragment } from 'react';
 import { shouldDisplay } from './helpers';
-
-export const AssetsLoader: LoaderFunction = async (): Promise<BalancesByAccount[]> => {
-  await abortLoader();
-  return await getBalancesByAccount();
-};
+import { useBalancesByAccount } from '../../../state/balances';
 
 export default function AssetsTable() {
-  const balancesByAccount = useLoaderData() as BalancesByAccount[];
+  const balancesByAccount = useBalancesByAccount();
 
-  if (balancesByAccount.length === 0) {
+  if (balancesByAccount.data?.length === 0) {
     return (
       <div className='flex flex-col gap-6'>
         <p>
@@ -40,7 +36,7 @@ export default function AssetsTable() {
 
   return (
     <Table>
-      {balancesByAccount.map(account => (
+      {balancesByAccount.data?.map(account => (
         <Fragment key={account.account}>
           <TableHeader className='group'>
             <TableRow>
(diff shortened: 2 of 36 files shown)


apps/minifront/src/components/dashboard/assets-table/index.tsx
```diff
@@ -11,20 +9,15 @@ import {
...
-
-export const AssetsLoader: LoaderFunction = async (): Promise<BalancesByAccount[]> => {
-  await abortLoader();
-  return await getBalancesByAccount();
-};
+import { useBalancesByAccount } from '../../../state/balances';
 
 export default function AssetsTable() {
-  const balancesByAccount = useLoaderData() as BalancesByAccount[];
+  const balancesByAccount = useBalancesByAccount();
```
TalDerei, line 18: we were previously using `abortLoader` to sanity-check that certain preconditions are met before proceeding with a data-fetching operation. If any the checks failed, an exception was thrown, which aborted the loading process. What's the equivalent of this in `zquery` land?

apps/minifront/src/abort-loader.ts
```diff
@@ -32,8 +32,12 @@ const retry = async (fn: () => boolean, ms = 500, rate = Math.max(ms / 10, 50))
  * timeout. This is a temporary solution until loaders properly await Prax
  * connection.
  */
-export const abortLoader = async (): Promise<void> => {
+export const abortLoader = async (): Promise<null> => {
```
TalDerei, line 35: Is this the trick, ie. returning `null`, that enables instantaneous route changes?

apps/minifront/src/components/root-router.tsx
```diff
@@ -1,21 +1,23 @@
...
 import { SwapLoader } from './swap/swap-loader';
-import { StakingLayout, StakingLoader } from './staking/layout';
+import { StakingLayout } from './staking/layout';
 import { IbcLoader } from './ibc/ibc-loader';
 import { IbcLayout } from './ibc/layout';
+import { abortLoader } from '../abort-loader';
+import type { Router } from '@remix-run/router';
 
-export const rootRouter = createHashRouter([
+export const rootRouter: Router = createHashRouter([
```
TalDerei, line 20: can we add some comments somewhere on the high-level similarities and differences between react-router loaders and zquery?

From what I can tell, both **support**:
- async data fetching
- state access to access fetched data in components

In terms of their **differences**, this is where some more clarification would be helpful. Zquery anchors on the idea of global state management, where state can be updated and shared across an application, right? React-router on the other hand fetches data specifically for a _single_ route, and rather than being stored globally, it's only accessible by a specific component. I guess I'm wondering where component-level data fetching would generally be more preferred?

apps/minifront/src/state/staking/index.ts
```diff
@@ -170,15 +167,49 @@ const byBalanceAndVotingPower = (valueViewA: ValueView, valueViewB: ValueView):
   return byVotingPower;
 };
 
+export const { stakingTokensAndFilter, useStakingTokensAndFilter } = createZQuery({
+  name: 'stakingTokensAndFilter',
+  fetch: async () => {
+    const balancesByAccount = await getBalancesByAccount();
```
TalDerei, line 173: random question out of curiosity, can you somehow _chain_ zquery requests? for instance, instead of calling `getBalancesByAccount()`, something like

```
    // Access the state of balancesByAccount
    const balances = useStore.getState().balances.balancesByAccount;
```

TalDerei (approved):
this is really great work!

TalDerei:
good to merge barring the resolution of conflicts ~

apps/minifront/src/state/index.ts
```diff
@@ -18,9 +20,11 @@ import { createIbcInSlice, IbcInSlice } from './ibc-in';
 enableMapSet();
 
 export interface AllSlices {
```
TalDerei, line 22: can we add a comment on what _slices_ are how in the context of the `SliceCreator<SliceInterface>` custom type?

apps/minifront/src/components/dashboard/assets-table/index.tsx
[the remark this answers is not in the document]
turbocrime: it should be unnecessary - the loader check was only present because the loaders were suppressing the check done at page layout

TalDerei:
@jessepinho I still had a few questions I left as comments I was hoping you could answer, selfishly for my own curiosity?

jessepinho (commented):
whoops, forgot to submit my comments

apps/minifront/src/state/staking/index.ts
TalDerei: don't have specific use case in mind, was just curious if this was possible.

PR #1307, opened 2024-06-16
disconnect api

### LINKED ISSUE #1192: Disconnect wallet method
Across crypto, wallets expose a `disconnect()` method to allow the user the ability to revoke sharing data with a web app (from the context of the web app itself). This is also consistent with cosmos wallets (see cosmos-kit and graz).

Task: add a `disconnect()` method to the injected Penumbra provider:

https://github.com/penumbra-zone/web/blob/034da5c47e53b9b24d52c24315ffa5350d479ce4/packages/client/src/index.ts#L8-L13

This will allow for better integrations with external partners: https://github.com/penumbra-zone/web/issues/1190

### Approve vs connect

Also, in other wallets, these are two different states:
- approved: user has added website to connection whitelist 
- connected: user has granted permission to share data

I find this distinction a little confusing and largely unnecessary. For Prax, we basically have treated these as one. Relating to `disconnect()`, we can probably feel free to remove the domain from the whitelist as well.

turbocrime:
https://github.com/penumbra-zone/web/assets/134443988/00e35e04-1447-4ded-ad79-c86c1f0433a8

closes #1192 

adds disconnect method to page api

page injection is refactored as a class to more explicitly represent and control injection state

origin-agnostic connection init helpers are exported from client package

transport may convey a simple 'false' indicating termination

disconnect minifront ui ready in merged #1276 but commented out

still needs a type guard for the provider manifest

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -1,0 +1,111 @@
+# `@penumbra-zone/client`
+
+This package contains interfaces, types, and some helpers for using the page API to Penumbra providers.
+
+**To use this package, you need to [enable the Buf Schema Registry](https://buf.build/docs/bsr/generated-sdks/npm):**
+
+```sh
+echo "@buf:registry=https://buf.build/gen/npm/v1/" >> .npmrc
+```
+
+## A simple example
+
+```ts
+import { bech32mAddress } from '@penumbra-zone/bech32m';
+import { createPenumbraClient } from '@penumbra-zone/client/create';
+import { ViewService, SctService } from '@penumbra-zone/protobuf';
+
+// This may connect to any available injected provider.
+const viewClient = createPenumbraClient(ViewService);
+
+// Or, you might prefer a specific provider.
+const praxViewClient = createPenumbraClient(
+  ViewService,
+  'chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe',
+);
+
+const { address } = await praxViewClient.addressByIndex({});
+console.log(bech32mAddress(address));
+```
+
+## React use
+
+It's likely you want to use this client in your webapp, and there's a good
+chance you're using React. Penumbra providers use `@connectrpc` tooling, so
+these clients are supported by `@connectrpc/query` and `@tanstack/react-query`.
+
+After using `createPenumbraChannelTransport` from `@penumbra-zone/client/create`
+and `TransportProvider` from `@connectrpc/query` in a parent component, you can
+use convenient React idioms.
+
+You can see a full example of this at https://github.com/penumbra-zone/nextjs-penumbra-client-example
+
+### A parent component
+
+```ts
+"use client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { syncCreatePenumbraChannelTransport } from "@penumbra-zone/client/create";
+import { TransportProvider } from "@connectrpc/connect-query";
+import { useMemo } from "react";
+
+const queryClient = new QueryClient();
+
+export const PenumbraQueryProvider = ({
+  providerOrigin,
+  children,
+}: {
+  providerOrigin: string;
+  children: React.ReactNode;
+}) => {
+  const penumbraTransport = useMemo(
+    () => syncCreatePenumbraChannelTransport(providerOrigin),
+    [providerOrigin],
+  );
+  return (
+    <TransportProvider transport={penumbraTransport}>
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    </TransportProvider>
+  );
+};
+```
+
+### A querying component
+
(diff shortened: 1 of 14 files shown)


packages/client/src/prax.ts
grod220, line 1: Was this deletion intentional? Think we still need this no?

packages/transport-chrome/src/session-client.ts
```diff
@@ -59,33 +86,37 @@ export class CRSessionClient {
...
+  };
+
   private clientListener = (ev: MessageEvent<unknown>) => {
     try {
-      if (isTransportMessage(ev.data)) this.servicePort.postMessage(ev.data);
+      if (ev.data === false) this.disconnectService();
+      else if (isTransportMessage(ev.data)) this.servicePort.postMessage(ev.data);
       else if (isTransportStream(ev.data))
         this.servicePort.postMessage(this.requestChannelStream(ev.data));
       else console.warn('Unknown item from client', ev.data);
```
grod220, line 105: Please add brackets to these clauses as well as the ones below

packages/client/src/prax.ts
turbocrime: no, client lib is now agnostic. it just describes the shape of the api. consumers can identify the extension they prefer to connect to, or implement UI to choose. #1293 should eventually provide a simple react-compatible framework for making a choice

packages/transport-chrome/src/session-client.ts
grod220: I am confused why this comment has now been dismissed like three times without being addressed

turbocrime: out of scope addressed in #1349

packages/client/src/prax.ts
turbocrime: more generic methods now established in client/create.ts

packages/client/src/index.ts
```diff
@@ -2,15 +2,80 @@ export * from './error';
...
+  /** Should contain a URI located at the provider's origin, which returns a
+   * chrome extension manifest v3 describing this provider. */
   readonly manifest: string;
 }
 
 declare global {
   interface Window {
+    /** Records upon this global should identify themselves by a field name
+     * matching the origin of the provider. */
     readonly [PenumbraSymbol]?: undefined | Readonly<Record<string, PenumbraInjection>>;
```
VanishMax, line 79: question: can we export some function like `getInjectedWallet(walletOrigin)` instead of using `PenumbraSymbol`?

I imagine external users implement the same API for wallets, so a set of function could help them:
- `getInjectedWallet(walletOrigin)` – for dApp devs
- `getInjectedWallets()` – list all browser wallets
- `injectWallet({ connect, disconnect, onConnectionChange ... })` – for wallet developers
- `onWalletInjected(callback)` – a function that fires a callback each time a wallet is added (no need to wait for page reload). For this `window[PenumbraSymbol]` should probably be converted to a proxy object to control additions and deletions, idk.

`getInjectedWallet` could become the beginning of this improved DX

turbocrime: see utils in create.ts

other aspects of this should probably happen after externally_connectable

injection should generally happen before page scripts load unless not approved.

adding callback functionality seems less useful, as the wallet client and all requests will be entirely promise-based anyway

react wrapper in another PR implements context and hooks, for that style

VanishMax: > injection should generally happen before page scripts load unless not approved

agree except for the case when user installs the extension and the page is loaded. Real life scenario:
1. user enters the frontend
2. frontend says "to get started, install the extension"
3. they install the extension
4. page didn't react – for now

i know it's a small thing but, in my opinion, developers will find the `onWalletInjected(callback)`

turbocrime: the page must reload before a new content script can appear.

packages/client/README.md
turbocrime, line 1: readme added

packages/client/src/index.ts
turbocrime, line 1: elaborated comments

turbocrime (commented):
cc @grod220 can this merge this week?

packages/client/src/index.ts
turbocrime: react wrapper in https://github.com/penumbra-zone/web/pull/1293

and another style, connect/query use, is visible in https://github.com/penumbra-zone/nextjs-penumbra-client-example/commit/228b6eba37a2b2419e8e5e3d0b8510c20a3a9ede

and the new readme for this package https://github.com/penumbra-zone/web/pull/1307/files#diff-7b3f04edbf937e50b69972d84ca0e92c9078c962fb4965bb7647b7c6ba74c296

VanishMax (approved):
All good to go!

PR #1309, opened 2024-06-17
feat(minifront): #1051: ibc in account selector

### LINKED ISSUE #1051: IBC-in address selection
At the moment, for IBC-in, we are making deposits into Account #0 for users:

Screenshot 2024-05-07 at 4 51 37 PM

https://github.com/penumbra-zone/web/blob/bf28b67ed2b974fa4c513c7f0182d343a763f721/apps/minifront/src/components/ibc/ibc-in/destination-addr.tsx#L11-L16

We should A) evaluate this flow in general and B) consider giving users the choice of which account to deposit into.

====

1. Give users an account selector widget (do not show the addresses)
2. Instruct users to validate their address in the keplr popup in the extension (assumption this is built).

===

Update: account selector component is able to split up to just get the account input and not display the address. Should try to re-use this.

VanishMax:
Closes #1051 

Allows selecting user's account to IBC to.

image


grod220:
One thing, in the meeting where we discussed this, it was brought up that we don't want to show the addresses here as the only trustworthy address is the one they'd see in the Keplr dialog box and this text could be confusing. Along those lines, can you remove "Here is your normal address" and below?

VanishMax:
@grod220 Thanks for the noticing. I didn't read the issue correctly in the first place, just changed it

grod220 (approved):

PR #1316, opened 2024-06-18
Stop using loader functions for swap page

### LINKED ISSUE #1243: Stop using loader functions

jessepinho:
This PR removes the loader function from the swap route, and instead uses ZQuery, resulting in instant route changes when a user navigates to `/swap`.

Closes #1243


apps/minifront/src/components/send/send-form/index.tsx
jessepinho, line 1: Not strictly related to this PR, just something I noticed while working on it

apps/minifront/src/components/swap/swap-loader.ts
```diff
@@ -1,42 +0,0 @@
...
-export type SwapLoaderResponse = UnclaimedSwapsWithMetadata[];
-
-const getAndSetDefaultAssetBalances = async (swappableAssets: Metadata[]) => {
-  const balancesResponses = await getSwappableBalancesResponses();
-  const stakingTokenAssetMetadata = await getStakingTokenMetadata();
-
-  // set initial denom in if there is an available balance
-  if (balancesResponses[0]) {
-    useStore.getState().swap.setAssetIn(balancesResponses[0]);
-    useStore.getState().swap.setAssetOut(swappableAssets[0]!);
```
jessepinho, line 25: These two `set*` calls are now made in the ZQuery object's `fetch` function.

apps/minifront/src/components/swap/swap-loader.ts
```diff
@@ -1,42 +0,0 @@
...
-    useStore.getState().swap.setStakingAssetMetadata(stakingTokenAssetMetadata);
```
jessepinho, line 26: This is now accessed via `useStakingTokenMetadata`

apps/minifront/src/components/swap/swap-loader.ts
```diff
@@ -1,42 +0,0 @@
...
-};
-
-export const SwapLoader: LoaderFunction = async (): Promise<null> => {
-  await abortLoader();
-  const assets = await getAllAssets();
-  const swappableAssets = assets.filter(isSwappable);
-
-  const balancesResponses = await getAndSetDefaultAssetBalances(swappableAssets);
-  useStore.getState().swap.setBalancesResponses(balancesResponses);
-  useStore.getState().swap.setSwappableAssets(swappableAssets);
```
jessepinho, line 39: These are now ZQuery objects

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -143,21 +138,41 @@ export const TokenSwapInput = () => {
...
-
-          <div className='flex flex-col gap-2 pt-2'>
-            <ArrowRight size={16} className='text-muted-foreground' />
-          </div>
-
-          <div className='flex h-full flex-col gap-2'>
-            <AssetSelector assets={swappableAssets} value={assetOut} onChange={setAssetOut} />
-            {assetOut && <BalanceValueView valueView={assetOutBalance} />}
-          </div>
+          <FadeIn condition={!!balancesResponses.data && !!swappableAssets.data}>
```
jessepinho, line 141: This fades in the input/output selectors once the necessary data has loaded:

screencap

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -76,24 +79,16 @@ const tokenSwapInputSelector = (state: AllSlices) => ({
...
-    hasStakingTokenMeta,
-  } = useStoreShallow(tokenSwapInputSelector);
+  const stakingTokenMetadata = useStakingTokenMetadata();
+  const balancesResponses = useBalancesResponses();
+  const swappableAssets = useSwappableAssets();
+  const { amount, setAmount, assetIn, setAssetIn, assetOut, setAssetOut, priceHistory } =
+    useStoreShallow(tokenSwapInputSelector);
   // State to manage privacy warning display
-  const [showNonNativeFeeWarning, setshowNonNativeFeeWarning] = useState(false);
+  const [showNonNativeFeeWarning, setShowNonNativeFeeWarning] = useState(false);
```
jessepinho, line 88: typo fix

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
grod220: Can we handle the error case as well? Maybe some red text with a loading error message?

grod220 (approved):
LGTM

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
jessepinho: done

jessepinho:
@TalDerei 
> @jessepinho can you fold in this [comment](https://github.com/penumbra-zone/web/pull/1268#discussion_r1643130786) to reduce code duplication and have the _warning_ use our shiny new zquery loaders?

Ah, sure — but I'm going to do that as a separate super-small PR, since as part of that, I want to calculate `showNonNativeFeeWarning` via a selector too, and that's out of scope for this PR.

PR #1320, opened 2024-06-18
Update changeset package name

grod220:
Fixes changeset publishing block on main


VanishMax (approved):

turbocrime:
lol i thought it might be something like that. should issue changeset

PR #1322, opened 2024-06-18
encryption: avoid throwing error

TalDerei:
References https://github.com/penumbra-zone/web/pull/1140#discussion_r1609735586

### DIFF
--- a/packages/crypto/src/encryption.ts
+++ b/packages/crypto/src/encryption.ts
@@ -140,19 +140,16 @@ export class Key {
   // Attempts to decrypt Box into message. If failure, returns `null`.
   async unseal(box: Box): Promise<string | null> {
     try {
+      // Decrypt the ciphertext using the nonce and key
       return await decrypt(box.cipherText, box.nonce, this.key);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/crypto/src/encryption.ts
```diff
@@ -140,19 +140,16 @@ export class Key {
...
-      if (e instanceof DOMException) {
-        if (e.name === 'OperationError') {
-          return null;
-        }
-      }
-
-      throw e;
+      // Log a generic error message
+      console.error('Decryption failed due to an unexpected error.');
+      return null;
```
grod220, line 152: What do you think about this?

```ts
    } catch (e) {
      // Logging a generic unknown error to avoid exposing sensitive information via throwing
      if (!(e instanceof TypeError || (e instanceof DOMException && e.name === 'OperationError'))) {
        console.error('Decryption failed due to an unexpected error.');
      }
      return null;
    }

```

grod220 (approved):

packages/crypto/src/encryption.ts
TalDerei: changed 👍

PR #1331, opened 2024-06-18
properly output esm

turbocrime:
uses `tsc-alias` to generate correct import paths for esm support

this may have been related to issues @VanishMax was experiencing

this should permit esm resolution to work properly for the packages' internal modules

edit: okay, tsc-alias doesn't completely solve the issue - 

bundlers are fine now, but anything that needs to use node module resolution (vitest, many consumers) will encounter problems because our packages currently use nonstandard module resolution to identify imports from external dependencies.

the tsc-alias tool is only indended for internal paths.

probably we should just comply with esm path spec internally, but that would require changing nearly every import in the codebase, so this PR now combines tsc-alias and changes imports to name external 'js' files that are not resolved via package.json exports.

this mainly just means pb types imports now require a 'js' extension, because those generated packages don't actually define any exports or main configuration.

there's hundreds of pb type imports, so it's a large diff. but this means we don't have to configure a bundler for every package, and the packages should have much better support for many consumers.

edit again:

instead of literally modifying imports, now mutating in a more custom way

this solution is bad. eventually a larger linter/bundler reconfig should happen

### DIFF
--- a/.changeset/funny-schools-kick.md
+++ b/.changeset/funny-schools-kick.md
@@ -1,0 +1,18 @@
+---
+'@penumbra-zone/transport-chrome': minor
+'@penumbra-zone/transport-dom': minor
+'@penumbra-zone/perspective': minor
+'@penumbra-zone/protobuf': minor
+'@penumbra-zone/services': minor
+'@repo/tsconfig': minor
+'@penumbra-zone/bech32m': minor
+'@penumbra-zone/getters': minor
+'@penumbra-zone/storage': minor
+'@penumbra-zone/client': minor
+'@penumbra-zone/crypto-web': minor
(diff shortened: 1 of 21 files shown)


TalDerei (approved):
so `tsc-alias` ensures that module paths are correctly resolved to relative paths? assuming this is related to the export name / path mangling that was happening?

packages/eslint-config/eslint.config.mjs
turbocrime, line 1: enforcing resolveable paths with eslint config

TalDerei (approved):

packages/eslint-config/eslint.config.mjs
turbocrime: reverted

Issue #1313, opened 2024-06-17
fix broken "max" button 


TalDerei:
References https://github.com/penumbra-zone/web/pull/1268#issuecomment-2164609477. Currently, the "max" functionality implemented in https://github.com/penumbra-zone/web/pull/1151 is broken in the context of non-zero gas fees. During transaction creation, the planner specifies outputs and fills in spends to balance the transaction, subtracting fees from the _input_. If the user attempts to spend their _entire_ balance, they don't have any additional funds to pay for fees. Instead, we want to modify the planner request to spend all the available notes, and instead subtract fees from the _output_. 

https://github.com/penumbra-zone/penumbra/pull/4615 was a necessary proto change towards that. 

cc @grod220

grod220:
This is only an issue if the user is trying to spend their max amount of UM and they have no other token right?

TalDerei:
![Screenshot 2024-06-13 at 12 37 10 PM](https://github.com/penumbra-zone/web/assets/70081547/f6677d73-1ff9-40d9-855c-98835fb83daf)

Valentine1898:
> This is only an issue if the user is trying to spend their max amount of UM
Not just UM, but other tokens as well if used as an alternative fee token

> and they have no other token
It doesn't matter, if the user has UM, then the alternative fee token will not be used

hdevalence:
We want to avoid a situation where a user has to manually decrease the amount slightly to leave a bit left over for gas, like on ethereum, we have the technology to deliver a better UX here so we should try to do so. It will require tracking some extra state in the send dialog (since we effectively want to secretly track an `isSendingMax` bool and branch on it).

conorsch:
The related Planner changes in https://github.com/penumbra-zone/penumbra/pull/4615 shipped as part of `v0.77.4`, so this work should now be unblocked.

TalDerei:
are we intending to extend the max functionality to swaps and delegations as well?

PR #1333, opened 2024-06-19
fees: support max functionality for sends with multi-asset fees

TalDerei:
References https://github.com/penumbra-zone/web/issues/1313. 

max functionality is only supported for general `sends`, and `swaps` / `delegations` were decided as out of scope for now.


apps/minifront/src/state/send/index.ts
```diff
@@ -36,6 +36,8 @@ export interface SendSlice {
   setFeeTier: (feeTier: FeeTier_Tier) => void;
   sendTx: () => Promise<void>;
   txInProgress: boolean;
+  isSendingMax: boolean;
+  SetIsSendingMax: (isSendingMax: boolean) => void;
```
jessepinho, line 40:
```suggestion
  setIsSendingMax: (isSendingMax: boolean) => void;
```

(and update throughout codebase)

packages/services/src/view-service/transaction-planner/fees.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (outputAsset) return outputAsset;
+
+  const spendAsset = request.spends.map(o => o.value?.assetId).find(Boolean);
+  if (spendAsset) return spendAsset;
+
+  const swapAsset = request.swaps.map(assetIn => assetIn.value?.assetId).find(Boolean);
+  if (swapAsset) return swapAsset;
+
+  const auctionScheduleAsset = request.dutchAuctionScheduleActions
+    .map(a => a.description?.outputId)
```
jessepinho, line 17: The output of an auction is the asset the user is trying to buy, so the user may not have any of it yet. I think you want the input asset ID, right?

packages/services/src/view-service/transaction-planner/fees.ts
```diff
@@ -0,0 +1,34 @@
...
+    .find(Boolean);
+  if (auctionScheduleAsset) return auctionScheduleAsset;
+
+  const auctionEndAsset = request.dutchAuctionEndActions.map(a => a.auctionId?.inner).find(Boolean);
+  if (auctionEndAsset) {
+    return new AssetId({ inner: auctionEndAsset });
+  }
```
jessepinho, line 24: Hm... Auction IDs are not asset IDs. Are you thinking of an auction NFT? (Even if so, though, I can't imagine that gas fees could be paid with auction NFTs...?)

packages/services/src/view-service/transaction-planner/fees.ts
```diff
@@ -0,0 +1,34 @@
+import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+
+// Attempts to extract a fee token from the assets used in the actions of the planner request
+// Priority in descending order
+export const extractAltFee = (request: TransactionPlannerRequest): AssetId => {
+  const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
+  if (outputAsset) return outputAsset;
```
jessepinho, line 8: Performance note: this will loop over the array twice, first to get the asset IDs, and second to find a truthy one. It'd be more efficient to rewrite it like this:

```suggestion
  const outputWithAssetId = request.outputs.find(o => o.value?.assetId);
  if (outputWithAssetId) return outputWithAssetId.value!.assetId;
```

(Same note for the instances below this as well.)

packages/services/src/view-service/transaction-planner/fees.ts
```diff
@@ -0,0 +1,34 @@
...
+  const auctionEndAsset = request.dutchAuctionEndActions.map(a => a.auctionId?.inner).find(Boolean);
+  if (auctionEndAsset) {
+    return new AssetId({ inner: auctionEndAsset });
+  }
+
+  const auctionWithdrawAsset = request.dutchAuctionWithdrawActions
+    .map(a => a.auctionId?.inner)
+    .find(Boolean);
+  if (auctionWithdrawAsset) {
+    return new AssetId({ inner: auctionWithdrawAsset });
```
jessepinho, line 30: Same note here

packages/services/src/view-service/transaction-planner/fees.ts
jessepinho, line 1: Looks like this was copied from `../fees.ts`, but is missing its tests. Can you please delete the original file, and move the tests down to this directory and update them?

apps/minifront/src/state/send/index.ts
```diff
@@ -111,7 +119,42 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
...
+      feeMode:
+        typeof feeTier === 'undefined'
+          ? { case: undefined }
+          : {
+              case: 'autoFee',
+              value: { feeTier },
+            },
+    });
+  }
+
```
jessepinho, line 157: Looks like the `spends` vs. `outputs` key is the only thing that's different in the two cases code below. Could we DRY up the code here a bit? e.g., 

```suggestion
  const spendOrOutput:
    | PartialMessage<TransactionPlannerRequest_Spend>
    | PartialMessage<TransactionPlannerRequest_Output> = {
    address: { altBech32m: recipient },
    value: {
      amount: toBaseUnit(
        BigNumber(amount),
        getDisplayDenomExponentFromValueView.optional()(selection?.balanceView),
      ),
      assetId: getAssetIdFromValueView(selection?.balanceView),
    },
  };

  return new TransactionPlannerRequest({
    ...(isSendingMax ? { spends: [spendOrOutput] } : { outputs: [spendOrOutput] }),
    source: getAddressIndex(selection?.accountAddress),

    // Note: we currently don't provide a UI for setting the fee manually. Thus,
    // a `feeMode` of `manualFee` is not supported here.
    feeMode:
      typeof feeTier === 'undefined'
        ? { case: undefined }
        : {
            case: 'autoFee',
            value: { feeTier },
          },

    memo: new MemoPlaintext({
      returnAddress: getAddress(selection?.accountAddress),
      text: memo,
    }),
  });

```

jessepinho (changes requested):
This is great, but I have a few notes about the logic where I don't think it's working the way you expect it to. I've made a number of comments where I see potential problems — please re-request review after addressing them.

jessepinho:
Actually, sorry, one big-picture question: is it actually _correct_ to set the `spends` instead of the `outputs` for a send transaction? I'm not sure what that would do... is that being handled by your code changes to `planner.rs`?

packages/services/src/view-service/transaction-planner/fees.ts
jessepinho: Also, I'm noticing that all of the issues I've raised with this file were already present before this PR — you just moved the file in this PR. So, feel free to address those issues in a separate PR.

apps/minifront/src/state/send/index.ts
TalDerei: changed naming convention

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: nice catch I switched those in my head, this should be `input?.assetId`

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: that is indeed the unique id of the auction to close, which has me questioning what asset id would be relevant for the fee? think we'd need a database lookup here

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: the signature of the return type will `AssetId | undefined` in that case, and we shouldn't assume the requestor is guaranteed to send an `assetId` along with the request.

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: slipped in after merging, removed the duplicate and kept original

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: I think we should actually just not include them as targets for multi-fees in this pr, and handle them separately after more thought.

apps/minifront/src/state/send/index.ts
TalDerei: good candidate for DRY, changed

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: see above comment

TalDerei:
> Actually, sorry, one big-picture question: is it actually _correct_ to set the `spends` instead of the `outputs` for a send transaction? I'm not sure what that would do... is that being handled by your code changes to `planner.rs`?

We modified the [protobuf definition](https://github.com/penumbra-zone/penumbra/pull/4615) to add spend actions as a special user intent to signify max functionality. In a regular send transaction, the planner constructs an output request. However, for send transactions where the user attempts to spend their entire balance, we track additional internal state to identify this intent and then branch on that logic to construct a spend request. The spend action is handled differently by the [planner](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/planner.rs#L244).

packages/services/src/view-service/transaction-planner/fees.ts
TalDerei: see https://github.com/penumbra-zone/web/issues/1375 and https://github.com/penumbra-zone/web/issues/1414

packages/services/src/view-service/fees.test.ts
```diff
@@ -92,48 +70,6 @@ describe('extractAltFee', () => {
     expect(result.equals(swapAssetId)).toBeTruthy();
   });
 
-  it('extracts the fee from dutchAuctionScheduleActions', () => {
```
VanishMax, line 95: question: why are the tests removed?

VanishMax (approved):
Only one question, so feel free to answer it and merge the PR 🚢

packages/services/src/view-service/fees.test.ts
TalDerei: dutch auctions for multi-fees are more complex than I thought, and will be handled in a separate PR. As such, the unit tests should be removed since the corresponding logic in the `extractAltFees` method was removed. We can fold the unit tests back in during https://github.com/penumbra-zone/web/issues/1414

VanishMax: in this case, it would probably be better to use the `it.skip()` along with the `todo: fix tests` comment surely not forget about them

TalDerei: sure good suggestion 👍

TalDerei:
@jessepinho switched auctions to default to using the native staking token rather than supporting multi-asset fees, which will implemented separately to reduce scope.

packages/services/src/view-service/fees.ts
```diff
@@ -1,30 +1,39 @@
...
+    return stakingTokenAssetId;
   }
 
   const auctionWithdrawAsset = request.dutchAuctionWithdrawActions
     .map(a => a.auctionId?.inner)
     .find(Boolean);
   if (auctionWithdrawAsset) {
-    return new AssetId({ inner: auctionWithdrawAsset });
+    return stakingTokenAssetId;
   }
```
grod220, line 37: I made a comment about these here: https://github.com/penumbra-zone/web/pull/1422/files#r1666473204

TalDerei: removed auctions from scope, and will properly handle in separate pr

packages/wasm/crate/src/planner.rs
```diff
@@ -482,6 +484,40 @@ pub async fn plan_transaction(
         ));
     }
 
+    for tpr::Spend { value, address } in request.spends {
```
grod220, line 487: Do I understand this correctly: if a tx planner request has a `spend`, their intention is to take every note they have and put it in the ownership of another address?

TalDerei: correct, we spend the entire SNR set, and set the change address to the receiver.

jessepinho (approved):

PR #1337, opened 2024-06-19
feat(minifront): #1211: remove swap restrictions

### LINKED ISSUE #1211: Remove restriction input asset

VanishMax:
Closes #1211 

What's done:
- extend `BalanceSelector` to show not only assets with balances but all available assets 
- fix the issues with empty wallets not rendering a swap block correctly
- reduce the height of `BalanceSelecor` and `AssetSelector` to `90dvh`
- auto focus the search inputs in `BalanceSelecor` and `AssetSelector`
- change validations of the swap input to allow entering any possible values

The balance selector now looks like this:

image


packages/ui/components/ui/icon-input.tsx
```diff
@@ -1,21 +1,18 @@
...
- * Use this to render an input with an icon to its left, such as a search field
- * with a magnifying glass.
- */
-export const IconInput = ({
-  value,
-  onChange,
-  icon,
-  placeholder,
-}: {
+interface IconInputProps extends Omit<InputProps, 'onChange'> {
```
jessepinho, line 4: issue: we shouldn't be allowing all props to pass through to underlying components in our component design system

This is part of a bigger discussion*, but in short, we've had issues with our component system becoming quite messy because any props can be passed through. For example, this change to the props interface now allows `children`, `className`, etc. to be passed to `<IconInput />`. `children` doesn't make sense to pass to it, and `className` allows consumers to drastically change the appearance of `<IconInput />` in ways they shouldn't be able to.

Instead, please just add `autoFocus` as a prop; we can keep adding more props as we need them, rather than allowing all props by default.

\* I'm soon going to be converting #432 to a tracking issue so that we can gradually shift our design system components away from our current model.

apps/minifront/src/state/swap/instant-swap.ts
```diff
@@ -31,6 +34,7 @@ import { divideAmounts } from '@penumbra-zone/types/amount';
 import { bech32mAssetId } from '@penumbra-zone/bech32m/passet';
 import { SwapSlice } from '.';
 import { sendSimulateTradeRequest } from './helpers';
+import { amountMoreThanBalance } from '../send';
```
jessepinho, line 37: Since `amountMoreThanBalance` is now used by multiple slices, please extract it to the `./helpers.ts` file, so that slices aren't reaching into each other's code.

apps/minifront/src/state/swap/instant-swap.ts
```diff
@@ -163,12 +167,17 @@ export const createInstantSwapSlice = (): SliceCreator<InstantSwapSlice> => (set
...
+
 const assembleSwapRequest = async ({
   assetIn,
   amount,
   assetOut,
 }: Pick<SwapSlice, 'assetIn' | 'assetOut' | 'amount'>) => {
-  if (!assetIn) throw new Error('`assetIn` was undefined');
+  if (!assetIn) throw new Error('`assetIn` is undefined');
+  if (!assetOut) throw new Error('`assetOut` is undefined');
+  if (!isValidAmount(amount, assetIn)) throw new Error('Invalid amount');
```
jessepinho, line 180: Good catch 👍🏻

apps/minifront/src/state/swap/index.ts
```diff
@@ -15,6 +15,7 @@ import {
   InstantSwapSlice,
   createInstantSwapSlice,
   instantSwapSubmitButtonDisabledSelector,
+  isValidAmount,
```
jessepinho, line 18: Please extract `isValidAmount` to `./helpers.ts`, so that slices aren't reaching into each other's code.

apps/minifront/src/state/swap/index.ts
```diff
@@ -31,9 +32,9 @@ interface Actions {
   setBalancesResponses: (balancesResponses: BalancesResponse[]) => void;
   setStakingAssetMetadata: (metadata: Metadata) => void;
   setSwappableAssets: (assets: Metadata[]) => void;
-  setAssetIn: (asset: BalancesResponse) => void;
+  setAssetIn: (asset?: BalancesResponse) => void;
   setAmount: (amount: string) => void;
-  setAssetOut: (metadata: Metadata) => void;
+  setAssetOut: (metadata?: Metadata) => void;
```
jessepinho, line 37: Are these changes necessary? Won't they always be defined if these methods are being called?

jessepinho: Ah, I see why they're needed from the swap loader.

I have a PR (#1316) to remove this loader, so I guess it's up to you whether to rebase on that once it's merged and then change this requirement, or just change this requirement now and tweak the loader which is going to be removed soon anyway... if that makes sense 😆

jessepinho (approved):
Great work! A lot of little notes here and there, but they're all relatively small things, so feel free to merge after addressing.

apps/minifront/src/components/shared/balance-selector.tsx
```diff
@@ -9,19 +9,101 @@ import { getAddressIndex } from '@penumbra-zone/getters/address-view';
...
 import { motion } from 'framer-motion';
+import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { filterMetadataBySearch } from './asset-selector';
+import { getMetadataFromBalancesResponseOptional } from '@penumbra-zone/getters/balances-response';
+import { AssetIcon } from '@repo/ui/components/ui/tx/view/asset-icon';
+import { emptyBalanceResponse } from '../../utils/empty-balance-response';
 
-const bySearch = (search: string) => (balancesResponse: BalancesResponse) =>
+const isMetadata = (asset: BalancesResponse | Metadata): asset is Metadata => {
+  return Boolean('symbol' in asset && 'name' in asset && 'display' in asset);
```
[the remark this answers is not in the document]
VanishMax: Thanks, didn't know about this! Got to modify the suggestion a little to `return asset.getType().typeName === Metadata.typeName;` to make it work

apps/minifront/src/state/swap/index.ts
VanishMax: Noticed the removal of swap loader. I added some changes to the zquery fetchers of swappableAssets and balancesResponses, so this code with optional parameters is irrelevant now

packages/ui/components/ui/icon-input.tsx
VanishMax: Thanks for the detailed explanation, it really helped me understand the idea! Changed this to only add `autoFocus`

PR #1339, opened 2024-06-19
Stop using loader functions for IBC out; add selector capabilities to ZQuery

### LINKED ISSUE #1243: Stop using loader functions

jessepinho:
This PR gets rid of our final loader: the IBC Out loader.

The IBC Out loader was responsible for loading and returning balances, asset metadatas, and the staking token metadata. It also was responsible for setting the initial balance selection and chain selection in the IBC Out form's Zustand state.

This PR moves things around a bit so that the IBC Out form can use Zustand entirely:
1. Add selector functionality to ZQuery's `use[Name]()` hooks.
    - This mimics React Query's `select` property that can be passed to [`useQuery`](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery#usequery). Now, multiple parts of the app can access the same data in state, but reformat it into whatever shape is useful for them using a selector. Of course, whenever the data is updated in _any_ part of the app, the update gets propagated to _every_ part of the app.
2. Add `assets` and `balancesResponses` to the `shared` Zustand slice.
    - Previously, the `swap` slice had `swappableAssets` and `balancesResponses`. Similarly, the `send` slice had `transferableBalancesResponses`. Now that the `ibcOut` slice also needs to load and use balance responses, all three features (swap, send, and IBC) all use the `shared` slice's instance of that data, and use selectors to filter/map/reduce/etc. as needed.
    - Furthermore, the staking slice had a `stakingTokensAndFilter` ZQuery object. But the data for that object can now be calculated using the `balancesResponses` and `stakingTokenMetadata` ZQuery objects in the `shared` slice. So I removed the `stakingTokensAndFilter` ZQuery object from the `staking` slice, and instead created a `useStakingTokensAndFilter()` hook that uses both `shared.balancesResponses` and `shared.stakingTokenMetadata` to create the necessary data.
3. Introduce Zustand middlewares.
    - There are a number of places where we want to mutate the state once e.g., an API call has returned. Previously, we were doing that in the fetcher. For example, in the `send` slice, we'd fetch the balances responses, then [set `state.send.selection` to the first balances response](https://github.com/penumbra-zone/web/pull/1339/files#diff-b771e10f5243fc1cb84e351a8420cc7585eef71a057e4cf373871a8e59ec435aL30-L34) before returning the balances responses. Now that multiple minifront features are using data from a single `shared` slice (e.g., the send/swap/ibcOut slices all use `shared.balancesResponses`), we needed a way to set certain state values in those other slices in response to the `shared` slice updating.
    
      In Redux, this would be trivial — we'd just have a reducer that responds to the `SET_BALANCES_RESPONSE` action and updates the state accordingly. But Zustand doesn't have a notion of unique action names, so instead, we have to use middlewares that check the before/after states and determine whether a particular mutation that we care about has taken place.

      [Per Zustand's author](https://github.com/pmndrs/zustand/discussions/1770#discussioncomment-5714662), there is no "official" way to create middleware in Zustand. Instead you just literally replace the function passed to `create()` with one that mutates `get`/`set`/`store`. So that's the approach I took here. I created several middlewares that update other pieces of state once a given piece of state updates. (For example, the [`sendSelectionMiddleware`](https://github.com/penumbra-zone/web/pull/1339/files#diff-b771e10f5243fc1cb84e351a8420cc7585eef71a057e4cf373871a8e59ec435aR171) updates `state.send.selection` to the first balances response when it detects that `state.shared.balancesResponses` has been updated.)
4. 🚨 New addition since I had to rebase on `main`: #1353 introduced the use of query params to control the swap asset in/out. Unfortunately, rebasing on that PR introduced some serious complications to this PR, so I more or less had to rewrite the `setInitialAssets()` function that was introduced there and use it as a middleware. That added a bit of heft to this PR, which you can see in the `swapBalancesMiddleware`.

## For future PRs, perhaps
- Right now, a component that uses a `use[Name]()` hook from ZQuery triggers a re-fetch, even if the data has already previously been fetched. It'd be great to have a smarter caching/revalidation strategy in the future (probably taking cues from TanStack Query), but that's out of scope for this PR.

Closes #1243


apps/minifront/src/components/swap/helpers.ts
jessepinho, line 1: The helpers in this file were moved to [`state/swap/helpers.ts`](https://github.com/penumbra-zone/web/pull/1339/files#diff-39d3f3b8e678d5d10fc3aec43e080473b9aa17dc436e2a159566437971afdfb5)

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,3 +71,47 @@ export const sendCandlestickDataRequest = async (
...
+    ?.filter(isKnown)
+    .filter(balance => isSwappable(getMetadata(balance.balanceView)))
+    .sort(byBalanceDescending),
+});
+
+export const swappableAssetsSelector = (zQueryState: ZQueryState<Metadata[]>) => ({
+  loading: zQueryState.loading,
+  error: zQueryState.error,
+  data: zQueryState.data?.filter(isSwappable),
+});
```
jessepinho, line 117: The helpers above these two selectors are simply moved over from `components/swap/helpers.ts`, but these two selectors are new. They allow us to use the `shared.balancesResponses` and `shared.assets` properties and filter/sort them for the swap feature.

apps/minifront/src/state/staking/index.ts
```diff
@@ -129,20 +125,6 @@ export interface StakingSlice {
...
-    unstakedTokensByAccount: Map<number, ValueView | undefined>;
-    /**
-     * All accounts for which staking is relevant, to be passed to
-     * `<AccountSwitcher />` as the `filter` prop. This includes accounts with:
-     * - delegation tokens
-     * - unbonding tokens
-     * - staking (UM) tokens (since they can be delegated)
-     */
-    accountSwitcherFilter: number[];
-  }>;
```
jessepinho, line 145: Now moved to the custom [`useStakingTokensAndFilter()` hook](https://github.com/penumbra-zone/web/pull/1339/files#diff-37aed1ca05297402c813cd650aab4035f8348b168d481cfa5333139abe42d0f7)

apps/minifront/src/state/staking/index.ts
```diff
@@ -455,49 +399,6 @@ const assembleUndelegateRequest = ({
   });
 };
 
-/**
```
jessepinho, line 458: These two reducer functions below are moved to the new hook as well

apps/minifront/src/state/staking/index.test.ts
```diff
@@ -66,7 +66,8 @@ const validatorInfo4 = new ValidatorInfo({
   },
 });
 
-vi.mock('../../fetchers/registry', () => ({
+vi.mock('../../fetchers/registry', async () => ({
+  ...(await vi.importActual('../../fetchers/registry')),
```
jessepinho, line 70: It was complaining about a missing property otherwise

apps/minifront/src/components/ibc/ibc-out/chain-selector.tsx
```diff
@@ -7,23 +7,27 @@ import {
...
+const chainSelectorSelector = (state: AllSlices) => ({
+  chain: state.ibcOut.chain,
+  setChain: state.ibcOut.setChain,
+});
 
 export const ChainSelector = () => {
-  const { chain, setChain } = useStore(ibcOutSelector);
-  const { chains: ibcConnections } = useLoaderData() as IbcLoaderResponse;
+  const { chain, setChain } = useStoreShallow(chainSelectorSelector);
+  const chains = useChains();
```
jessepinho, line 22: This is generated by the new ZQuery `chains` object in the `ibcOut` slice

jessepinho: Also, I found it confusing that we're calling them `ibcConnections` in some places and `chains` in others, so I normalized on `chains`

apps/minifront/src/components/send/helpers.ts
```diff
@@ -15,19 +8,3 @@ export const penumbraAddrValidation = (): Validation => {
...
-
-export const isTransferable = (metadata: Metadata) =>
-  nonTransferableAssetPatterns.every(pattern => !pattern.matches(getDisplay(metadata)));
-
-export const getTransferableBalancesResponses = async (): Promise<BalancesResponse[]> => {
-  const balancesResponses = await getBalances();
-  return balancesResponses.filter(
-    balance => isUnknown(balance) || isTransferable(getMetadata(balance.balanceView)),
-  );
-};
```
jessepinho, line 33: Moved [here](https://github.com/penumbra-zone/web/pull/1339/files#diff-f91054ffb202d0abdebea8130d45bc1e9b6af6dc75f0a893b6b0ebbab8670332)

apps/minifront/src/components/staking/account/delegation-value-view/staking-actions/form-dialog.tsx
```diff
@@ -42,7 +42,7 @@ export const FormDialog = ({
    * A `ValueView` representing the address's balance of staking (UM) tokens.
    * Used to show the user how many tokens they have available to delegate.
    */
-  unstakedTokens?: ValueView;
+  stakingTokens?: ValueView;
```
jessepinho, line 45: Not strictly part of this PR, but we had a mix of the terms "unstaked tokens" and "staking tokens," and both refer to the same thing, so I figured "staking tokens" is clearer

apps/minifront/src/components/staking/account/delegation-value-view/staking-actions/index.test.tsx
```diff
@@ -26,17 +26,15 @@ const zeroBalance = new ValueView({
 const validatorInfo = new ValidatorInfo({ validator: {} });
 
 let MOCK_STAKING_TOKENS_AND_FILTER: {
-  unstakedTokensByAccount: Map<number, ValueView | undefined>;
+  stakingTokens: ValueView | undefined;
```
jessepinho, line 29: `useStakingTokensAndFilter()` now returns an object with `stakingTokens` and `accountSwitcherFilter` properties, whereas before the properties were `unstakedTokensByAccount` and `accountSwitcherFilter`. Aside from the rename of `unstakedTokens` to `stakingTokens`, that property is also now a single `ValueView` representing the staking tokens for a given account (which is passed into the `useStakingTokensAndFilter()` hook as an argument), rather than being a map of account numbers to staking token `ValueView`s

apps/minifront/src/components/staking/account/delegation-value-view/staking-actions/index.tsx
```diff
@@ -41,17 +41,11 @@ export const StakingActions = ({
   const state = useStoreShallow(stakingActionsSelector);
   const validator = getValidator(validatorInfo);
 
-  const stakingTokensAndFilter = useStakingTokensAndFilter();
-  const unstakedTokensForThisAccount = stakingTokensAndFilter.data?.unstakedTokensByAccount.get(
-    state.account,
-  );
+  const { stakingTokens } = useStakingTokensAndFilter(state.account);
```
jessepinho, line 44: See my note in the test file above — `stakingTokens` already represents the staking tokens for this account, rather than a map.

apps/minifront/src/components/staking/account/use-staking-tokens-and-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+  ValueView,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
+import { getDisplayDenomFromView } from '@penumbra-zone/getters/value-view';
+import { assetPatterns } from '@penumbra-zone/types/assets';
+
+/**
+ * Function to use with `reduce()` over an array of `BalancesByAccount` objects.
+ * Returns a map of accounts to `ValueView`s of the staking token.
+ */
+const toStakingTokensByAccount = (
```
jessepinho, line 19: This reducer and the one below it are moved over, as-is, [from the `staking` slice](https://github.com/penumbra-zone/web/pull/1339/files#diff-7c5ec26a35f650875c894552352cecdc5e1042768e81d382bffa45e0a4dc0c1e)

apps/minifront/src/components/staking/account/use-staking-tokens-and-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+      assetPatterns.unbondingToken.matches(displayDenom) ||
+      displayDenom === stakingTokenMetadata.display
+    );
+  });
+
+  if (isRelevantAccount) return [...accountSwitcherFilter, curr.account];
+  return accountSwitcherFilter;
+};
+
+export const useStakingTokensAndFilter = (
```
jessepinho, line 58: This hook is basically doing what the `stakingTokensAndFilter` ZQuery object used to be doing. But since that object was just using balances and the staking token metadata, which are now both in the `shared` slice, this hook uses the `shared` slice as its source of data, and massages the data into the shape expected by the consumers of the hook.

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -63,11 +67,11 @@ const tokenSwapInputSelector = (state: AllSlices) => ({
 export const TokenSwapInput = () => {
   const status = useStatus();
   const latestKnownBlockHeight = status.data?.latestKnownBlockHeight ?? 0n;
-  const balancesResponses = useBalancesResponses();
```
jessepinho, line 66: This version of `useBalancesResponses()` was only fetching swappable balances responses. The new `useBalancesResponses()` fetches _all_ balances responses, so we now use the `swappableBalancesResponsesSelector` to filter for just the balances we want.

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -63,11 +67,11 @@ const tokenSwapInputSelector = (state: AllSlices) => ({
...
-  const swappableAssets = useSwappableAssets();
```
jessepinho, line 67: Same here

apps/minifront/src/components/tx-details/index.tsx
```diff
@@ -16,7 +16,7 @@ export const TxDetailsErrorBoundary = () => {
 export const TxDetails = () => {
   const { hash } = useParams<{ hash: string }>();
 
-  const txInfo = useTransactionInfo(hash!);
+  const txInfo = useTransactionInfo(undefined, hash!);
```
jessepinho, line 19: `use[Name]()` hooks take an optional `options` object as their first argument. We'll leave it undefined here since we only want to pass arguments (in this case, `hash!`) directly to the `fetch` function.

Unfortunately, I couldn't figure out how to get TypeScript to play nice with adding a `fetchArgs` property to the options object, which would have been a nicer/more consistent DX. So instead, the arguments to the `fetch` arguments are any arguments that come after the options object.

apps/minifront/src/state/balances.ts
jessepinho, line 1: The whole purpose of this slice was to store `balancesByAccount`. Since selectors can now be cached with ZQuery, we instead use a `balancesByAccountSelector` for the `<AssetsTable />` component, and cache it so it only runs when the underlying data has changed.

apps/minifront/src/state/ibc-out.ts
```diff
@@ -23,13 +23,29 @@ import { Chain } from '@penumbra-labs/registry';
 import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { Channel } from '@buf/cosmos_ibc.bufbuild_es/ibc/core/channel/v1/channel_pb';
 import { BLOCKS_PER_HOUR } from './constants';
+import { ZQueryState, createZQuery } from '@penumbra-zone/zquery';
+import { getChains } from '../fetchers/registry';
+
+export const { chains, useChains } = createZQuery({
```
jessepinho, line 29: This used to be in the React Router loader, so we needed to move it to ZQuery

apps/minifront/src/state/send/index.ts
```diff
@@ -19,34 +19,9 @@ import {
 import { getAddress, getAddressIndex } from '@penumbra-zone/getters/address-view';
 import { toBaseUnit } from '@penumbra-zone/types/lo-hi';
 import { isAddress } from '@penumbra-zone/bech32m/penumbra';
-import { ZQueryState, createZQuery } from '@penumbra-zone/zquery';
-import { getTransferableBalancesResponses } from '../components/send/helpers';
-
-export const { transferableBalancesResponses, useTransferableBalancesResponses } = createZQuery({
```
jessepinho, line 25: We now use `shared.balancesResponses` for the Send form, plus a `transferableBalancesResponsesSelector` to select only transferable balances responses.

Then we use middleware (farther down in this file) to check for when `shared.balancesResponses` has become populated, and update `state.send.selection` if it's not already set.

apps/minifront/src/state/shared.ts
```diff
@@ -16,10 +22,70 @@ export const { stakingTokenMetadata, useStakingTokenMetadata } = createZQuery({
...
+      account: index.account,
+      address: getAddress(curr.accountAddress),
+      balances: [curr],
+    });
+  }
+
+  return acc;
+};
+
+export const balancesByAccountSelector = (
```
jessepinho, line 89: Normally, I'd colocate a use case-specific selector with the component that uses it. But this is used by multiple components, so I figured it belongs colocated with the source of the data.

apps/minifront/src/components/swap/helpers.ts
jessepinho, line 1: This is all moved to [`state/swap/helpers.ts`](https://github.com/penumbra-zone/web/pull/1339/files#diff-39d3f3b8e678d5d10fc3aec43e080473b9aa17dc436e2a159566437971afdfb5)

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,3 +71,47 @@ export const sendCandlestickDataRequest = async (
...
+  assetPatterns.unbondingToken,
+];
+
+const isSwappable = (metadata: Metadata) =>
+  nonSwappableAssetPatterns.every(pattern => !pattern.matches(getDisplay(metadata)));
+
+const isKnown = (balancesResponse: BalancesResponse) =>
+  getValueViewCaseFromBalancesResponse.optional()(balancesResponse) === 'knownAssetId';
+
+export const swappableBalancesResponsesSelector = (
```
jessepinho, line 102: The helpers above these two selectors are simply moved over from `components/swap/helpers.ts`, but these two selectors are new. They allow us to use the `shared.balancesResponses` and `shared.assets` properties and filter/sort them for the swap feature.

apps/minifront/src/state/swap/index.ts
```diff
@@ -18,58 +18,10 @@ import {
...
 import { createPriceHistorySlice, PriceHistorySlice } from './price-history';
 import { getMetadata } from '@penumbra-zone/getters/value-view';
-import { createZQuery, ZQueryState } from '@penumbra-zone/zquery';
-import { getSwappableBalancesResponses, isSwappable } from '../../components/swap/helpers';
-import { getAllAssets } from '../../fetchers/assets';
-import { emptyBalanceResponse } from '../../utils/empty-balance-response';
 import { isValidAmount } from '../helpers';
 
-// When both `balancesResponses` and `swappableAssets` are loaded, set initial assetIn and assetOut
-const setInitialAssets = (state: SwapSlice) => {
```
jessepinho, line 28: The `swap` slice no longer has its own `balancesResponses` and `swappableAssets` ZQuery objects, since those have been moved to the `shared` slice. We don't want the `shared` slice to have to worry about setting the asset in/out for the `swap` slice, since that would create a circular dependency of sorts between the `shared` and `swap` slices. Instead, the functionality of `setInitialAssets` has been moved to the middlewares at the bottom of this file.

packages/zquery/src/types.ts
```diff
@@ -12,6 +15,58 @@ export interface ZQueryState<DataType, FetchArgs extends unknown[] = []> {
   };
 }
 
+export type UseHookOptions<
+  ResolvedDataType,
+  SelectorType extends
+    | ((zQueryState: AbridgedZQueryState<ResolvedDataType>) => unknown)
+    | undefined = undefined,
+> = SelectorType extends undefined
+  ?
```
jessepinho, line 24: Unfortunately, TypeScript (or at least, VSCode) requires that the doc blocks above these properties be repeated for each branch of this ternary operator — that is, once for when they're undefined, and once for when they're defined. 🤷🏻

jessepinho:
Moving back to draft because of merge conflicts that will require a bit of manual resolution

apps/minifront/src/state/swap/set-initial-assets.test.ts
jessepinho, line 1: See my PR description — all of this logic had to be moved into the `swapBalancesMiddleware`

apps/minifront/src/state/swap/set-initial-assets.ts
jessepinho, line 1: same here

jessepinho:
OK @penumbra-zone/web-reviewers — could you please review this now? It got hung up on a rebase that was pretty hefty, and since this has gotten pretty big, it'd be super helpful to get a review soon so I can merge it ASAP. Thanks!

VanishMax (approved):
@jessepinho amazing! Thanks for the detailed explanation of this PR idea problems that you faced. 

The introduced concept of middleware is unusual, it feels like the data flow gets harder to control and predict now. Anyway, I see the point and agree that this solution works nicely.

I see you did extra work for balance selectors and variables renaming. Even though I'm not against it, it would probably be better to put in another PRs. Just so the team doesn't collide with what you are doing.

Approving the PR and leaving some questions for discussions

apps/minifront/src/state/swap/swap-balances-middleware.test.ts
```diff
@@ -0,0 +1,293 @@
...
+    // `initializeStore()` initializes the store with all middlewares already
+    // attached, so there's no need to set up a test rig for
+    // `swapBalancesMiddleware` specifically
+    useStore = create<AllSlices>()(initializeStore());
+    vi.restoreAllMocks();
+    vi.stubGlobal('location', { hash: '#swap' });
+  });
+
+  describe('when setting `shared.balancesResponses.data`', () => {
+    describe('when the `from` query param is set', () => {
```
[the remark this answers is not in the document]
jessepinho: I might have over-used one-off `describe`s in this test file 😂 but I do like to use nested describes in the following cases:
1. When there are more than one instance of a given use case, and you don't want to have to repeat the use case in the test names. This also makes for cleaner test output:
    image
2. When you want to share test setup in a `beforeEach` — e.g., [here](https://github.com/penumbra-zone/web/blob/jessepinho/asset-selector-loading-indicator-web-1386/apps/minifront/src/state/swap/swap-balances-middleware.test.ts#L187-L190), where we want to stub the query parameter for a whole "sub-suite" of tests.
3. When test names would get really long because they're describing a very specific use case — e.g., `it('shows an error message when the array is empty and there is no filter')` could be written with a couple nested `describe`s for `when the array is empty` and `when there is no filter`, especially if we think there may be more tests at any of the nesting levels. (I'm very open to debate on this point, though — that might be overkill.)

Definitely let's discuss at a retro!

apps/minifront/src/state/swap/index.test.ts
```diff
@@ -95,22 +119,25 @@ describe('Swap Slice', () => {
...
+            _zQueryInternal: {
+              fetch: vi.fn(),
+              referenceCount: 0,
+            },
+          };
+          state.shared.balancesResponses = {
+            data: [balancesResponseWithMetadata1, balancesResponseWithMetadata2],
+            loading: false,
+            revalidate: vi.fn(),
+            _zQueryInternal: {
```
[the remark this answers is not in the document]
jessepinho: Hm, you mean for mocking for tests? 

Hm... marking it as optional would mess up the typings in a number of places. However, I could take a cue from your test code and use `as ZQueryState` with an incomplete ZQuery object, for testing purposes...?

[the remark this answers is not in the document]
VanishMax: Yeah, this could help a little

apps/minifront/src/state/swap/swap-balances-middleware.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * This middleware sets `swap.assetIn` and `swap.assetOut` based on the relevant
+ * query params as well as the balances responses and assets in the `shared`
+ * slice.
+ */
+export const swapBalancesMiddleware: Middleware = f => (set, get, store) => {
+  const modifiedSetter: typeof set = (...args) => {
+    const before = get();
+    set(...args);
+    const after = get();
```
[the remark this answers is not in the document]
jessepinho: Good question — this is something I should bring up at Thursday's web weekly. I've made a point of `return`ing early from middlewares if they have no extra work to do, which means their impact on memory/performance should be negligible, and they shouldn't cause unnecessary rerenders. But if we're going to be using middlewares from now on, we need to be mindful of this concern. Let's discuss Thursday!

PR #1341, opened 2024-06-20
switch to using IDB_VERSION defined inside the package

### LINKED ISSUE #1332: needs idb version bump
idb schema changes require an idb version bump

https://github.com/penumbra-zone/web/pull/1268

https://github.com/penumbra-zone/web/blame/f5c511e08b5fbf614def6fa499a14c37e769c2e8/packages/storage/src/indexed-db/index.ts#L105-L109

probably need to start writing migrations as well

Valentine1898:
closes #1332

### DIFF
--- a/packages/storage/src/indexed-db/indexed-db.test.ts
+++ b/packages/storage/src/indexed-db/indexed-db.test.ts
@@ -97,6 +97,8 @@ describe('IndexedDb', () => {
       expect((await dbB.getAssetsMetadata(metadataA.penumbraAssetId!))?.name).toBe(metadataA.name);
     });
 
+    // TODO: Do not skip this test after vitest has been updated to v2.0.0.
+    // use vi.mock to override the IDB_VERSION value (vi.mock is not available in browser mode for vitest 1.6.0).
     it.skip('increasing version should re-create object stores', async () => {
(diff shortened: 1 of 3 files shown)


packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -98,15 +97,14 @@ describe('IndexedDb', () => {
       expect((await dbB.getAssetsMetadata(metadataA.penumbraAssetId!))?.name).toBe(metadataA.name);
     });
 
-    it('increasing version should re-create object stores', async () => {
+    it.skip('increasing version should re-create object stores', async () => {
```
Valentine1898, line 100: https://github.com/vitest-dev/vitest/pull/5765
`vi.mock` are still not available in browser mode, so you should disable this test until the next stable version of vitest is published.

Valentine1898:
> Can you also just confirm that `DB_VERSION = 43` is the correct version or whether we should bump to 44 🙏

The current version, 43, is correct. The latest version defined in the extension's `.env` file is 42.

https://github.com/prax-wallet/web/blob/e9b0d0d3431f3b7756a19ff1c9c47dbc157d4bc6/apps/extension/.env#L1

PR #1346, opened 2024-06-24
fix(minifront): #1326: add `useWheelPrevent` hook

### LINKED ISSUE #1326: Disable default input scroll behavior
We've gotten reports of users getting confused as input amounts change on scroll:

https://github.com/penumbra-zone/web/assets/16624263/bfdb0642-cc38-48d5-bf90-16adc950fe33

This appears to be a default behavior of html input elements. If you hover your cursor above the input field and scroll, it increases/decreases the amount. Given these inputs are quite sensitive, we should override this behavior and prevent it.

This is happening on swap, send, and ibc forms.

VanishMax:
Closes #1326 

Prevents HTML input field from changing the value on `wheel` event


grod220:
Nice, can you confirm this is fixed for the swap, send, staking, and ibc forms?

grod220 (approved):

VanishMax:
@grod220 right, I didn't apply it elsewhere. Created a NumberInput component with built-in behavior of preventing the scroll

PR #1353, opened 2024-06-26
feat: #1169: add swap query params

### LINKED ISSUE #1169: Add URL parameters to swap page to prepopulate token types
It would be useful to be able to link to the swap page from other parts of the UI with either (or both) of the token types pre filled. 

For instance, on the dashboard we could have a Trade button at the end of each row that would link to the swap page with that asset selected as the input token. To do this we would need a way to fill in that info in URL parameters.

VanishMax:
Closes #1169 

Adds support for `from`, `to`, and `account` query parameters on the Swap page. Also, adds a link from the assets table to trade each available asset.

Query params save `asset.symbol` string and compute the matches by it.

https://github.com/penumbra-zone/web/assets/29180358/ef89dba7-f00f-4d8b-b41d-88431b6a90a3

### DIFF
--- a/.changeset/lovely-teachers-relax.md
+++ b/.changeset/lovely-teachers-relax.md
@@ -1,0 +1,5 @@
+---
+'minifront': minor
+---
+
+Add query parameters to the swap page to allow pre-setting swap assets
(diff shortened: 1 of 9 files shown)


grod220 (commented):
Can we use a more subtle design aesthetic to the trade button? Having a primary button color repeated is a bit jarring. Perhaps also it should only show on hover so there isn't so much repetition? 

Also, noticed that when you change assets in the selector, it does not update the query params. Can we have these synced?

VanishMax:
@grod220 thanks for noticing!  I addressed both the design and syncing the query params within the page. Also, added more tests for the swap state

grod220 (approved):
Nice tests 👍

PR #1354, opened 2024-06-26
finalize ceremony keys and update to v.78.0 deps

TalDerei:


grod220 (approved):
Didn't realize there was a new chain up already 😅

PR #1355, opened 2024-06-26
Shim Promise.withResolvers() for ZQuery and minifront tests

jessepinho:
Not sure how this got removed. This PR makes sure tests pass when running with vitest — e.g., `pnpm vitest packages/zquery`

## Before
image

## After

image

### DIFF
--- a/apps/minifront/package.json
+++ b/apps/minifront/package.json
@@ -64,6 +64,7 @@
     "@types/react": "^18.3.2",
     "@types/react-dom": "^18.3.0",
     "@types/react-helmet": "^6.1.11",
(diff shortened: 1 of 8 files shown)


turbocrime (changes requested):
make sure you're using the correct node and typescript - Promise.withResolvers exists and doesn't need to be shimmed

https://github.com/penumbra-zone/web/assets/134443988/5c44840e-7dce-4e78-9c7a-a7c3fbf27113

vacekj (approved):

jessepinho:
@turbocrime Ah, good catch! Thanks for calling that out. That was the issue. Closing this PR.

PR #1365, opened 2024-06-28
Update revision number parsing

grod220:
Our current parsing is inconsistent with the way the Rust core code is doing this. This PR aligns with the core implementation: https://github.com/penumbra-zone/ibc-types/blob/73da5ee5d06b5f62c186b31ce37f7669edc8bbf2/crates/ibc-types-core-connection/src/identifier.rs#L76-L87 so as to prevent bugs with potential chain ids.

### DIFF
--- a/.changeset/tasty-pugs-develop.md
+++ b/.changeset/tasty-pugs-develop.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
+Update revision number parsing logic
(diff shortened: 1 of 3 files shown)


VanishMax (approved):
Awesome? Does it mean that IBC form will now work for custom chains?

grod220:
Actually, this more pertains to IBC-in and ensuring that transactions are properly sending the right revision number (which is required on the cosmos message)

PR #1367, opened 2024-06-28
bump registry

Valentine1898:
related to https://github.com/prax-wallet/web/pull/66

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -142,7 +142,7 @@ export class IndexedDb implements IndexedDbInterface {
       new IbdUpdater(db),
       constants,
       chainId,
-      registryClient.get(chainId).stakingAssetId,
+      registryClient.globals().stakingAssetId,
     );
     await instance.saveRegistryAssets(registryClient, chainId); // Pre-load asset metadata from registry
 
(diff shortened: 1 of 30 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -812,11 +816,7 @@ export class IndexedDb implements IndexedDbInterface {
   }
 
   fetchStakingTokenId(): AssetId {
-    const registryClient = new ChainRegistryClient();
-    const registry = registryClient.get(this.chainId);
-    const stakingToken = registry.stakingAssetId;
-
-    return stakingToken;
+    return this.stakingTokenAssetId;
```
grod220, line 819: Probably `fetch` is the wrong name for this. Why don't we just make this a public variable so folks can grab it from the instance: https://github.com/penumbra-zone/web/pull/1367/files#diff-8298d111bc94531ec943d5f024a9ab57657ad28670caea51e3aa9eda88bf3954R83

grod220: Also, if this is done, we should make storage a major version bump given we change the API in a breaking way

grod220 (approved):

packages/storage/src/indexed-db/index.ts
Valentine1898: @grod220 will you be able to publish packages to npm?

PR #1369, opened 2024-06-28
Remove staking token metadata for block processor

Valentine1898:


packages/query/src/block-processor.ts
```diff
@@ -77,21 +77,15 @@ export class BlockProcessor implements BlockProcessorInterface {
   private readonly viewServer: ViewServerInterface;
   private readonly abortController: AbortController = new AbortController();
   private numeraires: AssetId[];
-  private readonly stakingTokenMetadata: Metadata;
```
Valentine1898, line 80: The block processor does not need `Metadata`, only the `AssetId`

grod220 (approved):
LGTM

PR #1392, opened 2024-07-02
don't get ratelimited

### LINKED ISSUE #1401: Genesis sync takes a long time
Genesis block sync takes an unreasonable amount of time because validator delegation token metadata is requested before the genesis block, and the client becomes ratelimited by the RPC endpoint.

Delegation token metadata should not block genesis sync, nor make an inappropriate number of requests.

turbocrime:
ratelimit validator assets init.

these changes should prevent genesis sync from freezing.

fixes #1401


packages/query/src/queriers/compact-block.ts
```diff
@@ -17,16 +11,10 @@ export class CompactBlockQuerier implements CompactBlockQuerierInterface {
...
-    abortSignal,
-  }: CompactBlockRangeParams): AsyncIterable<CompactBlock> {
-    const req = new CompactBlockRangeRequest({ keepAlive, startHeight });
-    const iterable = this.client.compactBlockRange(req, { signal: abortSignal });
-    for await (const res of iterable) {
-      if (!res.compactBlock) throw new Error('No block in response');
-      yield res.compactBlock;
-    }
+  compactBlockRange(
+    req: { startHeight?: bigint; keepAlive?: boolean },
```
TalDerei, line 15: what's generally the distinction here between the former which was an async generator function that yielded `CompactBlock` objects from a `CompactBlockRangeRequest`, versus this construction that directly returns an `AsyncIterable` of `CompactBlockRangeResponse` objects. Aren't they functionally the same?

packages/query/src/block-processor.ts
```diff
@@ -128,29 +128,17 @@ export class BlockProcessor implements BlockProcessorInterface {
...
 
-    if (currentHeight === -1n) {
-      // In the `for` loop below, we only update validator infos once we've
-      // reached the latest known epoch. This means that, if a user is syncing
-      // for the first time, they could experience a broken UI until the latest
-      // known epoch is reached, since they may have delegation tokens but no
-      // validator info to go with them. So we'll update validator infos at the
-      // beginning of sync as well, and force the rest of sync to wait until
-      // it's done.
-      await this.updateValidatorInfos(0n);
```
TalDerei, line 139: nice, not sure why we were even doing this

TalDerei (approved):

packages/query/src/block-processor.ts
turbocrime: @jessepinho would have more idea on how critical this is

if this promise was voided it might also be fine? at least it may be possible to grab compactblocks before the limit kicks in

unfortunately the batch metadata request does not exist. it may also be possible to just ratelimit our requests locally.

packages/query/src/queriers/compact-block.ts
turbocrime: i'm not sure, but it the former call seemed to delay processing of the stream for some reason.

i'm skeptical of the need for these queriers in general - and mostly they just unwrap one layer of an object, or catch errors. some of them are very thin like this.

packages/query/src/block-processor.ts
grod220: Don't think we should delete this as then we wouldn't have validator info until the latest epoch is synced. But a `void` sounds reasonable.

> they could experience a broken UI until the latest known epoch is reached

This sounds like we should make our UI in the staking page more robust so it doesn't throw if there is missing validator info.

turbocrime: after examining, it seems like nothing is broken - they just don't render with icon etc. so i have no concerns voiding this

packages/storage/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "tsc --build && tsc-alias",
-    "clean": "rm -rfv dist package penumbra-zone-transport-dom-*.tgz",
+    "clean": "rm -rfv dist package penumbra-zone-*.tgz",
```
turbocrime, line 11: a few of these were misconfigured so they're fixed

PR #1396, opened 2024-07-02
Fee extraction tests + refactor

grod220:
Some follow up after https://github.com/penumbra-zone/web/issues/1165

After pairing with @vacekj, noticed a few areas of improvement with type checking + test coverage


packages/services/src/view-service/fees.ts
grod220, line 1: For this file, focused on trying to accomplish the equivalent logic without manual type casting. Also, used map+find to cover cases where asset ids may be on some notes and not others of the same category.

TalDerei (approved):

PR #1398, opened 2024-07-02
Fix auction pre-filled amounts

### LINKED ISSUE #1387: Auction pre-filled amounts
Screenshot 2024-07-01 at 3 29 10 PM
 
A few things with these inputs:

- They should have a better visual indicator that they are indeed inputs (at the moment, it doesn't look editable)
- We should not pre-fill amounts here. Only when estimating should we do that.
- We should allow an empty state. This also means we likely need to add some validation logic to ensure folks add values here. At the moment, a minimum of 1 is required and won't allow for an empty input.

jessepinho:
- Start with empty max/min output amounts
- Allow users to empty the fields as well
- Only show an error when invalid values have been entered in the fields (not when the fields are empty)
- Show a `0` placeholder to indicate that it's a text field

image

Closes #1387

### DIFF
--- a/.changeset/nine-students-fry.md
+++ b/.changeset/nine-students-fry.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
+Tweak max/min outputs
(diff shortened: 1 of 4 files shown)


vacekj (approved):
lgtm

PR #1399, opened 2024-07-02
Combine upcoming + active auctions

### LINKED ISSUE #1388: Combine upcoming + active
Screenshot 2024-07-01 at 3 33 13 PM

Right now there are three choices. We should only have two: All & Active. That means, we should combine upcoming & active into the active state.

jessepinho:
See inline comments.

image

Closes #1388

### DIFF
--- a/apps/minifront/src/components/swap/auction-list/helpers.ts
+++ b/apps/minifront/src/components/swap/auction-list/helpers.ts
@@ -1,17 +1,6 @@
 import { AuctionInfo } from '../../../fetchers/auction-infos';
-import { Filter } from '../../../state/swap/dutch-auction';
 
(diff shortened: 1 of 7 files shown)


apps/minifront/src/components/swap/auction-list/helpers.ts
jessepinho, line 1: We no longer need to use different sorting directions for the various tabs, since everything should be in ascending order. (Previously, we were sorting active auctions by start height in _descending_ order, so that the most recently started ones would sort the top — which wasn't really a necessity, but is now actually bad UX since that would mean that the upcoming auctions would sort before the active ones.)

grod220 (approved):

PR #1400, opened 2024-07-02
Move the abortLoader to the root route

### LINKED ISSUE #1344: Minifront sometimes fails to show frontend onboarding page
if the transactions path is navigated to on an unconnected minifront, instead of showing an onboarding page to the user, minifront shows a broken page, and eventually a transport timeout

https://localhost:5173/#/dashboard/transactions

Screen Shot 2024-06-21 at 17 55 08

i believe this is related to loader work @jessepinho because that's where connection checks were performed

jessepinho:
This fixes an issue where some routes were missing the loader, so the landing page would not show up correctly when the extension was not yet connected.

Closes #1344

Note that #1344 only mentions the transactions page, but the same issue was also present on the `/send/receive` page, since it was also missing the `abortLoader`.

### DIFF
--- a/.changeset/light-clouds-look.md
+++ b/.changeset/light-clouds-look.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
+Fix issue where the connect page wasn't rendered right on some routes.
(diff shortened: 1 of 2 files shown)


grod220 (approved):
Nice, far simpler this way

PR #1403, opened 2024-07-02
feat(minifront): #1149: sort assets table by priority score

### LINKED ISSUE #1149: Assets table ordering
After https://github.com/penumbra-zone/penumbra/pull/4441 is deployed, the rpcs will have an ability to assign priority scores before returning metadata to the frontend. This is something that the dashboard view can use. At the moment the ordering is a bit arbitrary:

Screenshot 2024-05-23 at 1 19 37 PM

We can consider adding scores like so:

| Asset          | Priority Score |
|----------------|-------|
| UM             | 0     |
| Normal denoms  | 10    |
| Auctions       | 20    |
| Delegations    | 30    |
| Unbonding      | 40    |
| Unknown assets | 50    |

That way, the frontend can simply sort on priority score

====

Alternatively, the registry can manually add priority scores for each asset: https://github.com/prax-wallet/registry/blob/main/registry/penumbra-testnet-deimos-8.json

VanishMax:
Closes #1149 

Idea of this PR - add "smart" sorting to the assets table. Here are some important points on how it's done:
- All assets filtering, sorting and grouping is done in the `filteredBalancesByAccountSelector` function
- Sorting multiplies the "priority score" by the token balance amount and compares this value to another token. 
- Priority scores are defined in this ascending order: unknown token -> unbonding -> delegation -> auctions -> tokens from the registry -> UM

Example: user has UM token (priority 50) with balance 100 and delUm (priority 20) with balance 200 but UM will go before delUm token even though second's balance is bigger.

Before making this solution, I tried to integrate the priority scores into the `services` package and failed – the scoring system surely needs access to the registry, and registry requires knowing `chainId`. That's why I moved it to the minifront.


jessepinho:
Hey @VanishMax — looks like there are some conflicts, which may be due to PRs I recently merged (sorry 😬 ) . Could you rebase before I review?

VanishMax:
@jessepinho sure, just finished rebasing

apps/minifront/src/state/shared.ts
```diff
@@ -86,3 +88,11 @@ const groupByAccount = (acc: BalancesByAccount[], curr: BalancesResponse): Balan
 export const balancesByAccountSelector = (
   zQueryState: AbridgedZQueryState<BalancesResponse[]>,
 ): BalancesByAccount[] => zQueryState.data?.reduce(groupByAccount, []) ?? [];
+
+export const filteredBalancesByAccountSelector =
```
jessepinho, line 92: Hm... is this meant to be used throughout the app? At the moment, it's only being used in the dashboard, so I'd personally prefer to colocate this with the Dashboard component (or at least in a helpers file next to it).

My concern is that, if this is going to be used in multiple places, it's a bit specific — it filters out certain asset types that we might not want to filter out everywhere, and it sorts by priority, which may (?) not be needed everywhere. Thoughts?

jessepinho (approved):
Overall LGTM, although I left a note about where the selector should be located. Also, I know there is now a `priorityScore` in the `Metadata` type, so I wonder if `getAssetPriorityScore` should actually be located in the chain registry repo (perhaps the `priorityScore` field could get auto-populated in the [`getMetadata`](https://github.com/prax-wallet/registry/blob/main/npm/src/registry.ts#L52-L59) method of the registry?).

Just a thought, but we can also merge this as-is and iterate.

grod220 (commented):
There are some challenges for priority scoring on the frontend side:
- Every frontend would be responsible for its own scoring method
- The frontend does not have access to the pricing data (which the mininode does). Eventually, we'll likely want to move to a model where highest marketcap tokens get a priority score bump. 

Think that means the most appropriate spot for scoring is in:
```
packages/services/src/view-service/balances.ts
```

VanishMax:
@grod220 how ok is it for the `services` package to depend on the Prax registry? Additionally, there's a problem that, even if it's ok to depend on the registry, `services` don't know the `chainId` to query the registry.

I think Jesse's idea of moving the scoring to the registry is valid – it has all the needed data to score the assets. Only that the non-registry assets will always have 0 score. 

> Every frontend would be responsible for its own scoring method
If that's correct, can't the scoring live in minifront?

turbocrime:
> @grod220 how ok is it for the `services` package to depend on the Prax registry? Additionally, there's a problem that, even if it's ok to depend on the registry, `services` don't know the `chainId` to query the registry.

generally, if services need to know something, it should be provided or accessed through a context key. see the ctx dir in services.

services can query idb, which should contain all registry asset data.

there should probably be a key for the native token id or other special token ids.

services can probably operate without chain id, but could add a key for chain id if necessary, or query app params from idb.

grod220:
> Every frontend would be responsible for its own scoring method

This was more of a comment on a downside and something we should try to avoid. 

> services can query idb, which should contain all registry asset data.

Given the chain registry syncs with the local assets table, the local db will always be better to query metadata. From within the balances rpc method, it could even just call the `assetMetadataById` rpc method to get what it needs without needing knowledge of chain id. 

> there should probably be a key for the native token id or other special token ids.

Yes, we should either add a special table for the native staking token id or expose it in a ctx (which would use the registry to find it).

VanishMax:
@grod220 @turbocrime @jessepinho Thank you for the feedback and ideas on how to improve it. It took me some learning and figuring out how to debug the services package but I found a way to simplify the scoring. IndexedDB truly had all data I needed to fetch the assets and apply the manipulations.

jessepinho (approved):
Looks great! 🚢

PR #1406, opened 2024-07-03
RFC: styling the UI library

jessepinho:
This PR is a request for comments on the ADR I've written. Please read the ADR ([Markdown preview here](https://github.com/penumbra-zone/web/blob/04319e9a79347203f9eea19a5f7d0493dbe54878/docs/adrs/005-ui-library-css.md)), and make inline comments on the file or general comments here in the PR thread.

I've written the ADR as though it's a finished decision, but obviously it's not — I just wrote what my decision would be, as a starting point for the rest of the team to comment on it.

Since our web weekly tomorrow is canceled for July 4th, we'll need to talk about this issue asynchronously here in the PR. Perhaps we can come to a final decision at next Thursday's web weekly, if we haven't already resolved the issue here.

Related to https://github.com/prax-wallet/web/issues/81

### DIFF
--- a/docs/adrs/005-ui-library-css.md
+++ b/docs/adrs/005-ui-library-css.md
@@ -25,6 +25,7 @@ The new UI library will not only be consumed by minifront (and possibly Prax); i
   - We _definitely_ don't want that. For example, they could redefine the `spacing` values so that `px-1` creates horizontal padding of `6px` instead of `4px`, which would screw up our designs throughout the library.
 - Consumers will expect to be able to pass Tailwind classes to our UI components.
   - This relates to the first point: if consumers are required to _install and configure_ Tailwind just to use our library, they won't be happy about having to install a dependency they can't actually use.
+- Consumers have to import `.css` files alongside our component files.
 - Even if we figured out a way to compile our Tailwind CSS with our components so that consumers didn't need to install Tailwind, the resultant CSS would be inefficiently big.
   - For example, if three different Penumbra UI library components use `bg-card-radial`, then either A) for each component that uses that class, the class will need to be defined in that component's CSS file (resulting in code duplication), or B) we'll export a single kitchen-sink CSS file containing all the styles for _all_ the Penumbra UI library components, including all of those that the consumer doesn't use. Both of these result in overhead.
   - Counterpoint: The same will be true even if we use CSS modules or native CSS.
--- a/docs/adrs/005-ui-library-css.md
+++ b/docs/adrs/005-ui-library-css.md
@@ -1,0 +1,69 @@
+# ADR 005: Styling the UI library
+
+As we prepare to build out our new library of UI components, we need to decide from the get-go how we'll style our components with CSS -- specifically, whether to use Tailwind (which is used throughout Prax and minifront) vs. some other system like CSS modules or CSS-in-JS.
+
+The new UI library will not only be consumed by minifront (and possibly Prax); it will also be consumed by other clients in the Penumbra ecosystem as they build UIs for Penumbra.
+
+## Pros of using Tailwind
+
+- Tailwind integrates nicely with the code we already have.
+- We could export a Penumbra Tailwind config file that others could use throughout their app, not just for the components we define.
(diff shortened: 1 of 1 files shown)


grod220:
It's really too bad there isn't some build tool that allows non-tailwind projects to use tailwind components. Given that's the case, I think the reasoning here makes sense.

The question then becomes, which css-in-js framework/method should we use? Ideas:

- Plain objects passed to `className` prop
- [styled-components](https://styled-components.com/)
- [emotion](https://emotion.sh/docs/introduction)
- [styled-jsx](https://github.com/vercel/styled-jsx)
- [linaria](https://github.com/callstack/linaria#documentation)
- [vanilla-extract](https://vanilla-extract.style/)

VanishMax:
@grod220 @jessepinho Tailwind can be built just like any other solutions – by extracting CSS from JS, and consumers won't need to know about Tailwind existence. Take a look at the [vite-react-component-library-starter](https://github.com/rayyamhk/vite-react-component-library-starter) template, it builds into `index.js`, `index.d.ts`, and `index.css`. Theoretically, we could reconfigure it to build different components separately (not into one `index.js`) and have CSS being exported in chunks to reduce the styling load

jessepinho:
@VanishMax we'd still run into the issues I've outlined [here](https://github.com/penumbra-zone/web/blob/04319e9a79347203f9eea19a5f7d0493dbe54878/docs/adrs/005-ui-library-css.md#cons-of-using-tailwind) — particularly the last one, where we'd either have to duplicate CSS classes across component CSS files, or we'd have to export a massive kitchen-sink CSS with a bunch of code that consumers don't need.

VanishMax:
@jessepinho CSS chunks might solve the issue if configured correctly. Referencing your example, if three different Penumbra UI library components use `bg-card-radial`, then a chunk `abc1123xyz.chunk.css` file would be created and referenced by `1.css`, `2.css`, etc.

This is still a theory since I've done the same with JS chunks but not CSS but might be considered

jessepinho:
@grod220 I could make a separate ADR/RFC for which framework to use, but I was personally leaning toward styled-components. It's the most popular one, and I've used it extensively in the past. However, `emotion` looks equally suited to the task.

Notes on the others:
- `styled-jsx` [requires a Babel plugin](https://github.com/vercel/styled-jsx?tab=readme-ov-file#getting-started). I'm a little unclear on whether that would be required for consumers as well, but I suspect it would, since it appears to modify the base definition of the `<style />` JSX component.
- `linaria` and `vanilla-extract` (clever name!) both compile CSS files, so they would still require consumers to import CSS files alongside our component JS files.
- Plain `style`* objects would give us the flexibility we need, but that basically means we're doing CSS-in-JS anyway, so we may as well use one of the established libraries like `emotion` or `styled-components`, IMO. 

\* You wrote "Plain objects passed to `className` prop" — I assume you meant the `style` prop?

grod220 (approved):
I believe styled-components had some issues with the new server components? Not sure, but we should ensure there's compatibility there with whatever we go with. It's a 👍 from me, but @VanishMax any further thoughts on this?

jessepinho:
@grod220 hm, I'm not familiar with those issues, but I'm seeing support in [Remix](https://github.com/remix-run/examples/tree/main/styled-components) and [NextJS](https://nextjs.org/docs/app/building-your-application/styling/css-in-js#styled-components). Theoretically, we should probably discourage Penumbra UI consumers from doing server-side work anyway for privacy reasons, but 🤷🏻‍♂️

VanishMax:
One important aspect I forgot to mention previously about the UI kit is the **accessibility**. Using Radix-ui helps us maintain quality across UI components in terms of the full keyboard support but we should not forget about the color contrast, HTML semantics and screen reader accessibility. 

Saying that, I propose we keep working with radix and possibly use some components from [react-aria](https://react-spectrum.adobe.com/react-aria/) (e.g. Calendar or DatePicker, which is not supported in Radix). 

Anyway, it doesn't relate to this RFC, just a thought. An approve from my side:)

VanishMax (approved):

jessepinho:
Excellent call-out, @VanishMax — I definitely will be leaning on existing libraries more complex components like date pickers, calendars, etc. For simpler components, I'll likely roll them myself to avoid messy base style overrides, or I'll use headless UI libraries (like Radix, as you suggested) when needed.

PR #1408, opened 2024-07-03
dev-pack script

turbocrime:
improves local tarball workflow.

run `pnpm dev:pack` in this repository, to watch code changes and repack tarballs. 
run `pnpm dev` in another terminal window if you also want to start the webapp


TalDerei (changes requested):
can we also restore [pack-public.sh](https://github.com/penumbra-zone/web/pull/1408/files#diff-1ba47a3a840b2bc749829ef093364aa9769889e33eaa496ed1443a29f54fb26a) and move it to the scripts folder

package.json
```diff
@@ -13,7 +13,10 @@
     "clean:modules": "rm -rf node_modules apps/*/node_modules packages/*/node_modules pnpm-lock.yaml",
     "clean:vitest-mjs": "find . -type f -name 'vite*.config.ts.timestamp-*-*.mjs' -ls -delete",
     "compile": "turbo compile",
-    "dev": "turbo dev",
+    "dev": "turbo dev:compile dev:app --concurrency 18",
+    "dev:app": "turbo dev:compile dev:app",
+    "dev:compile": "turbo dev:compile",
+    "dev:pack": "turbo dev:pack --concurrency 16",
```
TalDerei, line 19: is there difference between difference concurrency numbers?

TalDerei (approved):
confirmed offline that script properly propagates changes

package.json
turbocrime: concurrency limit is 10 unless set. these are precisely the necessary value. if more packages are added, or something else changes, they may need to be updated.

PR #1411, opened 2024-07-03
README for new UI library

jessepinho:
This sets out some coding standards for the new UI library to ensure file structure, naming, etc. are consistent.

Only requesting a review from Gabe, but others should feel free to review as well.

Closes https://github.com/prax-wallet/web/issues/82

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -28,7 +28,7 @@
     "./lib/utils": "./lib/utils.ts",
     "./postcss.config.js": "./postcss.config.js",
     "./styles/*": "./styles/*",
-    "./*": "./components/v2/*/index.tsx"
+    "./*": "./src/*/index.tsx"
   },
   "publishConfig": {
     "exports": {
@@ -53,6 +53,9 @@
       },
       "./styles/*": {
         "default": "./styles/*"
+      },
+      "./*": {
+        "default": "./src/*/index.tsx"
       }
     }
   },
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -27,7 +27,8 @@
     "./lib/toast/transaction-toast": "./lib/toast/transaction-toast.tsx",
     "./lib/utils": "./lib/utils.ts",
     "./postcss.config.js": "./postcss.config.js",
-    "./styles/*": "./styles/*"
+    "./styles/*": "./styles/*",
+    "./*": "./components/v2/*/index.tsx"
   },
   "publishConfig": {
     "exports": {
--- a/packages/ui/readme.md
+++ b/packages/ui/readme.md
@@ -207,7 +207,9 @@ Storybook stories should be located next to the component they apply to, and hav
 
 When writing stories, make sure to tag your stories with [`autodocs`](https://storybook.js.org/docs/react/writing-docs/autodocs). This is a Storybook feature that analyzes your component code to auto-generate documentation for your component, including a code sample, controls for each of the props, etc.
 
-Also, be sure to add [JSDoc](https://jsdoc.app/about-getting-started.html#adding-documentation-comments-to-your-code)-style comments (`/** ... */`) before any props that need to be documented. Storybook will automatically pull these into the Storybook UI as documentation.
+### Documentation of component props should be written with JSDoc syntax.
+
+[JSDoc-style comments](https://jsdoc.app/about-getting-started.html#adding-documentation-comments-to-your-code) (`/** ... */`) should be written before any props that need to be documented so that A) IDEs can pull them in for tooltips, and B) Storybook can use them in the Storybook UI.
 
 ### Components built for Protobuf types must be suffixed with `Component` to avoid naming collisions.
 
--- a/packages/ui/readme.md
+++ b/packages/ui/readme.md
@@ -1,28 +1,210 @@
 # Penumbra UI
 
-This package contains the UI components for the Penumbra web.
+The Penumbra UI library is a set of UI components purpose-built for the Penumbra ecosystem. Use these components to get rendering of various Penumbra data types out of the box, and to create a UI that is consistent with other Penumbra UIs' look and feel.
 
 ## Set up
 
 First, install the library:
 
 ```bash
-npm i @penumbra-zone/ui
+npm install @penumbra-zone/ui
 ```
 
-Then, configure the Tailwind in your project. Edit `tailwind.config.js`:
+Then, use components by importing them from their specific files:
 
-```js
-export default {
-  content: [
-    // Parses the classes of the UI components
-    './node_modules/@penumbra-zone/ui/**/*.js',
-  ],
-};
+```tsx
+import ValueViewComponent from '@penumbra-zone/ui/ValueViewComponent';
 ```
 
-Finally, import the library CSS to the entry point of your app:
+Deprecated components can be imported from `@penumbra-zone/ui/components/ui/<component-name>`, where `<component-name>` should be replaced with the kebab-cased component name.
 
-```js
-import '@repo/ui/styles/globals.css';
+## Development
+
+These guidelines are for maintainers of the components in the Penumbra UI library.
+
+### Components must be located at `./components/v2/<ComponentName>/index.tsx`.
+
+This ensures that the Penumbra UI `package.json` `exports` field works as intended, so that components can be imported via `@penumbra-zone/ui/<ComponentName>`.
+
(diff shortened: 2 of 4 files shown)


packages/ui/package.json
```diff
@@ -27,7 +27,8 @@
     "./lib/toast/transaction-toast": "./lib/toast/transaction-toast.tsx",
     "./lib/utils": "./lib/utils.ts",
     "./postcss.config.js": "./postcss.config.js",
-    "./styles/*": "./styles/*"
+    "./styles/*": "./styles/*",
+    "./*": "./components/v2/*/index.tsx"
```
jessepinho, line 31: This will allow us to import v2 components using shorter syntax: `import ValueViewComponent from `@penumbra-zone/ui/ValueViewComponent`

VanishMax: this should also be added to the `publishConfig.exports` below

VanishMax: suggestion: create an `src` dir and put the components there to not nest it from the `components/v2` – I see that eventually we'll move everything to the v2 style

packages/ui/readme.md
```diff
@@ -1,28 +1,230 @@
...
+
+```tsx
+export default function MyComponent({ color }: { color: Color }) {
+  // ...
+}
+```
+
+This makes component code easy to read. An exception to this may be made when props are especially complex, such as when they are conditional.
+
+There is no need to export the props interface, since consumers can simply use `React.ComponentProps<typeof ComponentName>` to extract the props interface from the component.
```
VanishMax, line 97: thoughts: disagree on this. I believe the props must always be exported. It's not hard to do, and it is usually expected from the UI libraries

packages/ui/readme.md
```diff
@@ -1,28 +1,230 @@
...
+
+### Components must not accept `className` or `style` props.
+
+This ensures that each component is internally responsible for its own styling.
+
+Any variations of the component's appearance must be controlled via props like `variant`, `state`, etc. — not by kitchen-sink props like `className` and `style`, which allow arbitrary changes to be made that could interfere with the internal structure of the component and cause visual inconsistencies between instances of Penumbra UI components.
```
VanishMax, line 103: question: even though I agree on this idea, there are cases where `className` existence is important. For example, take a look at the [AssetsTable Row implementation in minifront](https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/components/dashboard/assets-table/index.tsx#L79). When the TableRow with class 'group' is hovered, the inner Link changes its opacity. How can we handle such case?

packages/ui/readme.md
```diff
@@ -1,28 +1,230 @@
...
+      - index.test.tsx
+      - Dropdown.tsx ❌ Wrong - Menu shouldn't be importing a child of HeaderWithDropdown
+    - Menu/
+      - index.tsx
+      - index.test.tsx
+```
+
+### Components must be the default export of the component file.
+
+This allows them to be imported as the default, rather than destructuring an object. It also encourages UI library maintainers to only export one component per file.
```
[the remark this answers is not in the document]
jessepinho: Hm, a few thoughts:

- I don't have a strong opinion about this one way or the other, so I'm happy to change it.
- It seems that a lot of the discussions of this issue are claiming to be about named vs. default exports, when they _should_ be about named vs. unnamed exports. See Addy Osmani's tweet [here](https://x.com/addyosmani/status/1411233253948747777), which encourages people to name their exports (whether they are default exports or not). 
- I haven't had the experience that others are describing, where IDEs don't work well with default exports. In my experience with VSCode, if I want to import a default-exported `MyComponent`, I type `<MyComponent` in my JSX markup, and it automatically knows to import it. If I later rename `MyComponent`, it also updates everywhere in the codebase that imports `MyComponent`, even though it's the default export (since, after all, it is still a named function). It could just be that VSCode is especially smart about this, but I doubt it — I think it's a language-level thing, since `MyComponent` is actually a named function that happens to be exported as the default.

If you feel strongly about this, though, I can definitely remove this guideline.

packages/ui/readme.md
```diff
@@ -1,28 +1,230 @@
...
+  return (
+    <Pill>
+      <AssetIcon display={metadata.display} src='./icon.png' />
+      {metadata.display}
+    </Pill>
+  );
+}
+```
+
+### Components should include Storybook stories.
```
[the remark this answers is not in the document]
jessepinho: Ah, let me write something up about testing.

Storybook does have integrations with screenshot (not snapshot) testing, but the only one I know of is Chromatic, which is [a bit pricey](https://www.chromatic.com/pricing). We could consider that in the future, but for now, I should at least include a point about unit tests.

packages/ui/readme.md
```diff
@@ -1,28 +1,230 @@
...
-import '@repo/ui/styles/globals.css';
+## Development
+
+These guidelines are for maintainers of the components in the Penumbra UI library.
+
+### Components must be located at `./components/v2/<ComponentName>/index.tsx`.
+
+This ensures that the Penumbra UI `package.json` `exports` field works as intended, so that components can be imported via `@penumbra-zone/ui/<ComponentName>`.
+
+Note that `<ComponentName>` should be replaced with the `UpperCamelCase` component name — e.g., `./components/v2/LoadingIndicator/index.tsx`.
```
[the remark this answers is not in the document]
jessepinho: As long as people import it using the same capitalization at the file name, it should work across all systems, right? (I wanted to do this change both A) to make imports a little less annoying to type, and B) because it seems to be the norm across other libraries like MUI — [example](https://mui.com/system/react-box/#basics).)

packages/ui/readme.md
jessepinho: Oh man, good find of an edge case... Hm.

Went down a rabbit hole trying to solve this problem, and here's what I came up with:

```tsx
            <TableBody>
              {account.balances.map((assetBalance, index) => (
                <TableRow key={index}>
                  <TableCell>
                    <ValueViewComponent view={assetBalance.balanceView} />
                  </TableCell>
                  <TableCell>
                    <EquivalentValues valueView={assetBalance.balanceView} />
                  </TableCell>
                  <TableCell>
                    <Link
                      className='transition md:opacity-0 [tr:hover>td>&]:opacity-100'
                      to={getTradeLink(assetBalance)}
                    >
                      <Button variant='secondary' size='md'>
                        Trade
                      </Button>
                    </Link>
                  </TableCell>
                </TableRow>
              ))}
            </TableBody>
```

I've removed `group` entirely, and used nesting selectors to get the same effect. (I tried using `group` in combination with `&` ([docs](https://tailwindcss.com/docs/hover-focus-and-other-states#arbitrary-groups)) to get the same effect, but it didn't work. I suspect I'm doing something wrong, and could fix this with a bit more work, but it didn't seem worth the time right now.)

This isn't a great solution, so I'll keep playing with it once I cross this particular bridge. But for now, I think we can still avoid `className`/`style` props.

packages/ui/readme.md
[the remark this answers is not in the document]
jessepinho: I've pushed https://github.com/penumbra-zone/web/pull/1411/commits/f01b8935ccb1fdfa695b4d38b999f7c4a259b8af to address this; but pending your response to my comment above, I can revert it back to my original guideline.

packages/ui/readme.md
jessepinho: OK, fair enough. Done: https://github.com/penumbra-zone/web/pull/1411/commits/3785649d314e2a54918f43e9f56577d8763da244

packages/ui/package.json
jessepinho: Great idea — this is much cleaner. I've made this change (as well as adding it to `publishConfig`), and added a deprecation README for the v1 components: https://github.com/penumbra-zone/web/pull/1411/commits/f0550479f7fa4853ce16c9a65df6d33571aa0008

packages/ui/readme.md
VanishMax: Wow, didn't know it's possible with Tailwind! Thanks for showing this. It probably worth being documented

packages/ui/readme.md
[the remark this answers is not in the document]
jessepinho: Done: https://github.com/penumbra-zone/web/pull/1411/commits/6bffda89a769e3ddada4990801bb1ee02db15d64

VanishMax (approved):
This version is well-polished! Approving since all my concerns are noted but it's probably worth to wait for @grod220 to approve too

jessepinho:
Thanks @VanishMax ! @grod220 please weigh in when you have a chance

grod220 (approved):
Ship it 🚢

PR #1412, opened 2024-07-03
1310 modify gasprices storage to support multi asset fees

### LINKED ISSUE #1310: Modify `GasPrices` storage to support multi-asset fees
You are right, `GasPrices` does contain an `asset_id`, I missed that because `asset_id` is never defined. 
But the problem is that our `GasPrices` storage is deprecated and not suitable for multi-asset fees

Now we store `GasPrices` assuming that we can only have one record 
```
GAS_PRICES: {
    key: 'gas_prices';
    value: Jsonified<GasPrices>;
  };
``` 
Instead, we should use the `assetId` as the key
```
GAS_PRICES: {
    key: Jsonified<Required<GasPrices>['assetId']['inner']>;
    value: Jsonified<GasPrices>;
  };
``` 
But in that case, I expect we'll run into an error, because pd purposely omits the assetId for native `GasPrices`, but that's bad for us, because undefined can't be a key in indexed-db

We will have to manually add the `assetId` for native `GasPrices` or come up with another trick

Also, note that the getGasPrices function will also have to be modified 
```typescript
async getGasPrices(): Promise<GasPrices | undefined> {
    const jsonGasPrices = await this.db.get('GAS_PRICES', 'gas_prices');
    if (!jsonGasPrices) return undefined;
    return GasPrices.fromJson(jsonGasPrices);
  }
```  

Also need to change the **_gasPrices rpc_** for the view service so that it retrieves `alt_gas_prices`

And the block processor logic that should save `alt_gas_prices` to indexed-db when needed

_Originally posted by @Valentine1898 in https://github.com/penumbra-zone/web/pull/1268#discussion_r1642984352_

### LINKED ISSUE #1397: multi-asset fee bug
for swaps / auctions, figure out why the multi-asset fees (enabled when the user does not have native **UM** tokens) sometimes works for swapping one asset into another, but not vice versa

https://github.com/penumbra-zone/web/assets/70081547/eef7ed41-ec24-4e31-8df7-7cb1f2164358

https://github.com/penumbra-zone/web/assets/70081547/32428704-5aaf-443a-a780-d0d20b7a8a2b

### LINKED ISSUE #1404: more multi-asset fee bugs
- [ ] investigate why sending some tokens (for instance pizza) display insufficient note balance. 

Screenshot 2024-07-02 at 2 46 44 PM

-----------------------------------------------------------------------------------------------------------------------------------

- [ ] investigate why testUSD shows a zero fee balance when other fee tokens display a non-zero fee amount

Screenshot 2024-07-02 at 2 44 37 PM

-----------------------------------------------------------------------------------------------------------------------------------

- [ ] handle cases where some tokens (delegation tokens) have no asset metadata attached to them, which breaks the extension

You can provide a way better UX than this when your app throws

Valentine1898:
Closes #1310
Fixes #1404
Fixes #1397


packages/query/src/block-processor.ts
```diff
@@ -162,12 +163,18 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+          new GasPrices({
+            assetId: this.stakingAssetId,
+            ...compactBlock.gasPrices,
+          }),
+        );
+      }
+      if (compactBlock.altGasPrices.length) {
+        for (const gasPrice of compactBlock.altGasPrices) {
+          await this.indexedDb.saveGasPrices(gasPrice);
+        }
```
TalDerei, line 176: this suggests that the RPC node is returning information in compact blocks about _two_ alternative gas fees, but with zero gas prices. Manual testing confirmed that all _other_ token denominations result in a "GasPrices not available" error.

Screenshot 2024-07-04 at 1 15 13 AM

packages/storage/src/indexed-db/index.ts
```diff
@@ -824,7 +829,11 @@ export class IndexedDb implements IndexedDbInterface {
 
     return spendableUMNotes.some(note => {
       const umNote = SpendableNoteRecord.fromJson(note);
-      return umNote.heightSpent === 0n && !isZero(getAmountFromRecord(umNote));
+      return (
+        umNote.heightSpent === 0n &&
+        !isZero(getAmountFromRecord(umNote)) &&
+        umNote.addressIndex?.equals(addressIndex)
```
TalDerei, line 835: nice filtering for address index here

packages/wasm/crate/src/planner.rs
```diff
@@ -202,28 +206,12 @@ pub async fn plan_transaction(
...
-    // Check if the decoded gas fee token is different from the staking token asset ID.
-    // If the gas fee token is different, use the alternative gas fee token with a 10x
-    // multiplier.
-    if alt_gas != *STAKING_TOKEN_ASSET_ID {
-        gas_prices = GasPrices {
-            asset_id: alt_gas,
-            block_space_price: gas_prices.block_space_price * 10,
-            compact_block_space_price: gas_prices.compact_block_space_price * 10,
-            verification_price: gas_prices.verification_price * 10,
-            execution_price: gas_prices.execution_price * 10,
```
TalDerei, line 217: gas fee token price multiplier will instead be determined by governance proposals, rather than arbitrarily setting a 10x multiplier.

TalDerei (approved):
this seems to properly implement alternative gas prices retrieved from the compact block! 

tagging https://github.com/penumbra-zone/web/issues/1413 as relevant follow-up work.

PR #1420, opened 2024-07-05
injection state and event subscription

turbocrime:
Describe interfaces for injection state.

as suggested in https://github.com/penumbra-zone/web/pull/1293#discussion_r1660844191

uses an EventTarget to manage subscribers, and an event type extending CustomEvent

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -1,4 +1,5 @@
 export * from './error';
+export * from './event';
 
 export const PenumbraSymbol = Symbol.for('penumbra');
 
@@ -67,12 +68,6 @@ export interface PenumbraInjection {
    * `PenumbraInjectionState` value. */
   readonly addEventListener: EventTarget['addEventListener'];
   readonly removeEventListener: EventTarget['removeEventListener'];
-}
-
-export class PenumbraInjectionStateEvent extends CustomEvent<PenumbraInjectionState> {
-  constructor(override readonly detail: PenumbraInjectionState) {
-    super('penumbrastate', { detail });
-  }
 }
 
 export enum PenumbraInjectionState {
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -2,21 +2,28 @@ export * from './error';
 
 export const PenumbraSymbol = Symbol.for('penumbra');
 
-/** This interface describes the simple API to request, connect, or disconnect a provider.
+/**
+ * This interface describes the simple API to request, connect, or disconnect a
+ * provider. These methods allow a page to acquire permission to connect, and
+ * obtain a `MessagePort` to be used for client creation.
  *
- * There are three states for each provider, which may be identified by calling
- * the synchronous method `isConnected()`:
+ * There are three connection states for each provider, which may be identified
+ * by calling the synchronous method `isConnected()`:
  * - `true`: a connection is available, and a call to `connect` should resolve
  * - `false`: no connection available. calls to `connect` or `request` will fail
- * - `undefined`: a `request` may be pending, or no `request` has been made
+ * - `undefined`: a call may be pending, or no call has been made
+ *
+ * Each injection may should also track state-changing actions, so calling
+ * `.state()` should provide more detail including currently pending state,
+ * enumerated by `PenumbraInjectionState`.
  *
  * Any script in page scope may create an object like this, so clients should
- * confirm a provider is actually present by confirming provider origin and
- * fetching the provider manifest. Provider details such as name, version,
- * website, brief descriptive text, and icons should be available in the
- * manifest.
+ * confirm a provider is actually present. Presence can be securely verified by
+ * fetching the identified provider manifest from the provider's origin.
  *
- * Presently clients may expect the manifest is a chrome extension manifest v3.
+ * Presently clients can expect the manifest is a chrome extension manifest v3.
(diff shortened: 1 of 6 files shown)


packages/client/src/index.ts
```diff
@@ -31,34 +38,68 @@ export const PenumbraSymbol = Symbol.for('penumbra');
...
+  /* request for approval satisfied */
+  'Requested' = 'Requested',
+
+  /* connection attempt pending */
+  'ConnectPending' = 'ConnectPending',
+  /* connection successful and active */
+  'Connected' = 'Connected',
+
+  /* disconnect was called to release approval */
+  'Disconnected' = 'Disconnected',
```
grod220, line 96: Instead of emitting strings, what if we just emit a snapshot of the state? Aka:

```ts
enum LifecycleState {
  NOT_REQUESTED,
  PENDING,
  SUCCESS,
  ERROR,
}

interface PenumbraInjectionState {
  connect: LifecycleState;
  request: LifecycleState;
  disconnect: LifecycleState;
}
```

VanishMax: Is this question about the return types for injected functions? Right now, they return `Promise<void>` or throw error if something happens.

As for emitting strings vs enum numbers, I personally lean towards strings as the more readable option

packages/client/src/index.ts
```diff
@@ -2,21 +2,28 @@ export * from './error';
...
- * There are three states for each provider, which may be identified by calling
- * the synchronous method `isConnected()`:
+ * There are three connection states for each provider, which may be identified
+ * by calling the synchronous method `isConnected()`:
  * - `true`: a connection is available, and a call to `connect` should resolve
  * - `false`: no connection available. calls to `connect` or `request` will fail
- * - `undefined`: a `request` may be pending, or no `request` has been made
+ * - `undefined`: a call may be pending, or no call has been made
+ *
+ * Each injection may should also track state-changing actions, so calling
```
VanishMax, line 16:
```suggestion
 * Each injection may also track state-changing actions, so calling
```

VanishMax (approved):
Great changes! I completely agree with the proposal and can't wait to integrate it further

packages/client/src/index.ts
grod220: It's not related to the return types, but a way for someone to know in a subscription-like way what the state of all the possible requests are. And given each method has 4 states, in theory this would be 12 different strings possibly emitted. And then, the consumer of this would need to have their own state object or useState() variables where they take these 12 different strings and have to update each useState() individually.

turbocrime: > As for emitting strings vs enum numbers, I personally lean towards strings as the more readable option

this is a string enum - it does not become numbers, it essentially compiles to strings. returned values and parameter values exist in JS as strings and in typescript it provides some convenient collection testing possibilities like `if (someString in PenumbraInjectionState)`

turbocrime: >Instead of emitting strings, what if we just emit a snapshot of the state? Aka:

generally i think it's bad to track and expose the internal state of this injection in an explicit or detailed way. doing this restricts implementation possibilities, and expecting a consumer to inspect and interpret some internal state of the injection which may subtly change version-to-version or differ implementation-to-implementation seems unreliable.

most state a caller is interested in is evident by the return values or rejections of the methods they are using. i think it is reasonable to expect the caller to independently understand what actions they have taken, and to use return values and thrown errors.

any state tracking method implemented here is, from the caller's perspective, mostly just answering "what have i already done?" and "did it work?", both of which a caller knows if they are using return values. for synchronous inspection the original boolean-or-undefined state would have satisfied this situation. 

>And given each method has 4 states, in theory this would be 12 different strings possibly emitted. And then, the consumer of this would need to have their own state object or useState() variables where they take these 12 different strings and have to update each useState() individually.

there are three effective methods (request, connect, disconnect) with four effective call states (uncalled, pending, resolved, rejected) meaning there are 64 possible combinations, most of which are invalid and unreachable. if we expect the caller does not know what actions their own code has taken, more detailed state inquiry is more helpful to them, but representing invalid or impossible states seems not helpful.

if the objective here is to simplify use of this interface, expanding from a simple truthy/falsy state to a matrix of 64 mostly invalid states seems like a step backwards. so in the interest of simplification the enum only identifies valid states, and nonsensical states are represented by `Failed` and rejected promises.

valid transitions are like

```mermaid
stateDiagram-v2
direction LR
    Present
    Present --> ConnectionPending
    ConnectionPending --> Connected
    Connected --> Disconnected
    Present --> RequestPending
    RequestPending --> Requested
    Requested --> ConnectionPending
    Connected --> [*]
    Disconnected --> [*]
    note right of Connected
        isConnected(): true
    end note
    note left of Disconnected
        isConnected(): false
    end note
```

and there is not really room for deviation from this path. and there are already meaningful and specific error types for the reachable failure states.

turbocrime: > a way for someone to know in a subscription-like way what the state of all the possible requests are.

this is satisfied by the `addEventListener` method now exposed

turbocrime:
i'm going to add an 'origin' field to the emitted events, so that event handlers may easily identify the emitting injection

packages/client/src/index.ts
VanishMax: @turbocrime this flow diagram should definitely go to the client package readme or docs. It's great 👍

Issue #1407, opened 2024-07-03
swap-claims fail for multi-asset fees 


TalDerei:
In the presence of multi-asset fees, the swap claims fail because it fails to extract a valid assetId since https://github.com/penumbra-zone/web/pull/1396/files#diff-caad9b17fa66f6badd2a44a36749fc51c4452f35e34f7fe38905467885f60f0bR6 doesn't check for `SwapClaim` actions . It was previously working, but there was a regression somewhere in the chain of changes that now causes it to fail. We need to check if the transaction planner request in the view service implementation is a `SwapClaim` action, and use the correct corresponding assetId.

<img width="1728" alt="Screenshot 2024-07-02 at 3 56 09 PM" src="https://github.com/penumbra-zone/web/assets/70081547/b1066fe0-7a75-4eeb-8a3f-7476a88b7526">

PR #1421, opened 2024-07-05
fix swap claim bug for multi-fees

TalDerei:
references https://github.com/penumbra-zone/web/issues/1407

### DIFF
--- a/packages/services/src/view-service/fees.test.ts
+++ b/packages/services/src/view-service/fees.test.ts
@@ -8,7 +8,6 @@ import {
   TransactionPlannerRequest_Output,
   TransactionPlannerRequest_Swap,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
-import { extractAltFee } from './fees';
 import { AuctionId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1/auction_pb';
 
 describe('extractAltFee', () => {
@@ -21,8 +20,8 @@ describe('extractAltFee', () => {
         }),
       ],
     });
-    const result = extractAltFee(request);
-    expect(result.equals(umAssetId)).toBeTruthy();
+    const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
+    expect(outputAsset!.equals(umAssetId)).toBeTruthy();
   });
 
   it('skips over outputs that do not have assetIds', () => {
@@ -35,8 +34,8 @@ describe('extractAltFee', () => {
         }),
       ],
     });
-    const result = extractAltFee(request);
-    expect(result.equals(umAssetId)).toBeTruthy();
+    const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
+    expect(outputAsset!.equals(umAssetId)).toBeTruthy();
   });
 
   it('prioritizes outputs over all else', () => {
@@ -74,8 +73,8 @@ describe('extractAltFee', () => {
       ],
     });
 
-    const result = extractAltFee(request);
-    expect(result.equals(outputAssetId)).toBeTruthy();
+    const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
+    expect(outputAsset!.equals(outputAssetId)).toBeTruthy();
   });
 
   it('extracts the fee from swaps', () => {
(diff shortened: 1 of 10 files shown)


packages/services/src/view-service/fees.ts
```diff
@@ -1,15 +1,27 @@
 import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+import { IndexedDbInterface } from '@penumbra-zone/types/indexed-db';
 
-// Attempts to extract a fee token from the assets used in the actions of the planner request
-// Priority in descending order
```
grod220, line 5: Why are we removing the code comment?

packages/services/src/view-service/fees.test.ts
```diff
@@ -88,57 +87,7 @@ describe('extractAltFee', () => {
...
-    const result = extractAltFee(request);
-    expect(result.inner).toEqual(auctionWithdrawAuctiontId.inner);
-  });
-
-  it('throws an error when no asset ID is found', () => {
-    const request = new TransactionPlannerRequest({});
-
-    expect(() => extractAltFee(request)).toThrow(
-      'Could not extract alternative fee assetId from TransactionPlannerRequest',
-    );
```
grod220, line 142: Were these deleted by mistake?

packages/services/src/view-service/fees.ts
```diff
@@ -1,15 +1,27 @@
...
+): Promise<AssetId> => {
   const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
   if (outputAsset) return outputAsset;
 
   const swapAsset = request.swaps.map(assetIn => assetIn.value?.assetId).find(Boolean);
   if (swapAsset) return swapAsset;
 
+  const swapCommitment = request.swapClaims
+    .map(swapClaim => swapClaim.swapCommitment)
+    .find(commitment => commitment !== undefined);
```
grod220, line 17: You can simplify this like the others to:

```diff
- .find(commitment => commitment !== undefined);
+ .find(Boolean)
```

packages/services/src/view-service/fees.test.ts
grod220, line 1: Can you write a test for swap claims as well?

packages/services/src/view-service/fees.ts
```diff
@@ -1,15 +1,27 @@
...
+  if (swapCommitment) {
+    const swaps = await indexedDb.getSwapByCommitment(swapCommitment);
+    if (swaps?.swap?.tradingPair?.asset1) {
+      return swaps.swap.tradingPair.asset1;
+    }
+  }
```
grod220, line 23: I believe this is saying "If a user is making a swap claim, attempt to find the assetIn they originally swapped with for an alt fee".

But given swap claims happen in different blocks than swaps, isn't it possible the user doesn't have that original asset anymore? Like what if they swapped the entire balance of it out? Is this logic we can't really depend on?

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -14,11 +14,13 @@ export const transactionPlanner: Impl['transactionPlanner'] = async (req, ctx) =
   const { indexedDb } = await services.getWalletServices();
 
   // Query IndexedDB directly to check for the existence of staking token
-  const nativeToken = await indexedDb.hasStakingAssetBalance(req.source);
+  const nativeToken = await indexedDb.hasStakingAssetBalance();
```
grod220, line 17: What's the thinking around removing source here? Is the expectation that paying fees should not be scoped to one account and can be paid by any of your accounts? Is this consistent with the core repo? At the moment, in minifront, we've kinda been scoping all activity to a single account in a tx.

packages/services/src/view-service/fees.test.ts
TalDerei: will fold them back in when we implement auctions seperately in pr. by passing in the idb handle to `extractAltFees`, we're changing the method behavior so it's expected to necessarily change the tests

packages/services/src/view-service/fees.ts
TalDerei: simplified

packages/services/src/view-service/fees.ts
TalDerei: as discussed, the original asset value balances are irrelevant since the swap pre-pays the swap claim. The assetId however must be consistent between the swap / swap claim and can be retrieved by querying the swap table in the database or by calling the `SwapByCommitment` service implementation using the swap commitment as the key.

TalDerei:
edge case discovered during pairing:
  
1. Swap GM --> GN
2. acquire UM
3. swap claim never hits extractAltFee

screenshot_2024-07-05_at_4 53 17___pm

packages/services/src/view-service/fees.test.ts
```diff
@@ -21,8 +20,8 @@ describe('extractAltFee', () => {
         }),
       ],
     });
-    const result = extractAltFee(request);
-    expect(result.equals(umAssetId)).toBeTruthy();
+    const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
+    expect(outputAsset!.equals(umAssetId)).toBeTruthy();
```
jessepinho, line 24: 🚨 This (and the three changed tests below) are no longer testing `extractAltFee()`, since they don't call it. They should be changed to actually call `extractAltFee()`, or else they're essentially useless tests.

TalDerei: we're passing in an indexDB now, not sure how we'd simulate this in the test since MockIndexDB has different interface?

TalDerei:
[swapClaims](https://github.com/penumbra-zone/web/blob/swap-claim-bug/packages/services/src/view-service/fees.ts#L15) inside `extractAltFee` now properly retrieve the asset id from the claim fee rather than the pair inputs. This solves a host of canonical ordering bugs.

packages/services/src/view-service/fees.test.ts
TalDerei: logic has changed with the inclusion of indexDB, so we should change these tests accordingly.

packages/services/src/view-service/transaction-planner/index.ts
Valentine1898: I think the issue is that `source` should always be set, and we should not create transaction plans that are not linked to any account (address index)

Valentine1898: The fact that `hasStakingAssetBalance()` now doesn't take AddressIndex into account leads to a bug where the planner thinks he has a native token for the commission but can't spend it. This bug will be reproduced quite often

grod220 (approved):

packages/services/src/view-service/fees.test.ts
jessepinho: Un-resolving this because this is a pretty important issue: the tests here are literally not testing anything. It's the equivalent of writing an `add()` function, then writing a test like this:

```ts
it('adds', () => {
  const result = 2 + 2;
  expect(result).toBe(4);
})
```

In this test, we haven't called the `add()` function, so if we break the `add()` function later, the tests won't catch it.

Happy to pair on updating `MockedIndexedDB` if that would help!

Valentine1898 (approved):
LGTM!

Just one suggestion, I think planner should throw an error if the client has not passed `source` to `TransactionPlannerRequest`

packages/services/src/view-service/fees.test.ts
grod220: Talked offline. Given this blocks deployment, we want to move forward. However, we should absolutely follow up with this with a separate PR.

grod220:
Pairing together w/ @TalDerei, merging!

Issue #1375, opened 2024-07-01
extend multi-asset fees to ICS20 withdrawals


TalDerei:
functionality for multi-asset fees is currently supported for sends and swaps, but should be extended to support IBC-related actions like ICS20 withdrawals.

<img width="677" alt="Screenshot 2024-07-04 at 7 12 53 PM" src="https://github.com/penumbra-zone/web/assets/70081547/1a4747a8-5b63-42af-a16c-57beab7c1b8c">

PR #1422, opened 2024-07-05
fix ics20 ibc withdrawals

TalDerei:
references https://github.com/penumbra-zone/web/issues/1375

@turbocrime finally using your `assetIdFromBaseDenom` method :)

### DIFF
--- a/packages/services/src/view-service/fees.ts
+++ b/packages/services/src/view-service/fees.ts
@@ -1,20 +1,18 @@
 import { AssetId } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb';
 import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb';
+import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
 
 // Attempts to extract a fee token from the assets used in the actions of the planner request
 // Priority in descending order
-export const extractAltFee = (
-  request: TransactionPlannerRequest,
-  stakingTokenAssetId: AssetId,
-): AssetId => {
+export const extractAltFee = (request: TransactionPlannerRequest): AssetId => {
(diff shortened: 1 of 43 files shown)


packages/services/src/view-service/fees.ts
```diff
@@ -3,13 +3,19 @@ import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_
...
+  stakingTokenAssetId: AssetId,
+): AssetId => {
   const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
   if (outputAsset) return outputAsset;
 
   const swapAsset = request.swaps.map(assetIn => assetIn.value?.assetId).find(Boolean);
   if (swapAsset) return swapAsset;
 
+  const ics20Withdrawl = request.ics20Withdrawals.find(Boolean);
+  if (ics20Withdrawl) return stakingTokenAssetId;
```
grod220, line 17: I'm a little confused about the logic here. If a request has any ics20Withdrawal, it should use the staking token instead of an alternative fee? But if they ended up in `extractAltFee()` it guarantees they don't have the native token. So their transaction will error.

grod220: == Pairing discussion ==

```ts
  const ics20WithdrawAsset = request.ics20Withdrawals.map(w => w.denom).find(Boolean);
  if (ics20WithdrawAsset) return assetIdFromBaseDenom(ics20WithdrawAsset.denom);
```

jessepinho:
@grod220 can you approve? I don't have enough context on this problem (or on ICS 20 withdrawals) to know whether this change should work

jessepinho:
(also, a test for the ICS 20 withdrawals should be added to the `extractAltFee` tests, no?)

grod220 (approved):

grod220:
Todo for tests later plz

turbocrime (approved):
xD

PR #1426, opened 2024-07-05
Support delegator voting in planner

grod220:
With delegator voting now supported in the TxPlanner (https://github.com/penumbra-zone/penumbra/pull/4689) we can extend the wasm planner code to handle these incoming requests. Unlocks governance in the extension. 

Code modeled off of:
- Core: https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pcli/src/command/tx.rs#L883-L948
- Web view service impl: https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/notes-for-voting.ts

Note, this is part 1 as a new testing framework PR is coming soon to add test coverage for these new methods.

@plaidfinch fyi


grod220:
@Valentine1898 can you review as you implemented this in typescript already?

packages/wasm/crate/src/storage.rs
```diff
@@ -386,6 +389,63 @@ impl IndexedDBStorage {
...
+            let Some(delegation_token) = delegation_assets.get(&record.note.asset_id()) else {
+                continue;
+            };
+
+            // Ensure if it's been spent, it is later than votable vote height
+            if let Some(height_spent) = record.height_spent {
+                if height_spent < votable_at_height {
+                    continue;
+                }
+            }
```
Valentine1898, line 442: Are you sure this will work for unspent notes? 
If height_spent is not set, it is serialized as 0 (maybe this is only true for typescript)

packages/wasm/crate/src/storage.rs
```diff
@@ -386,6 +389,63 @@ impl IndexedDBStorage {
...
+
```
Valentine1898, line 443: You should also filter out notes that were created after the voting started 

```typescript
// Note must be created at a height lower than the height of the vote
      const noteIsCreatedBeforeVote = note.heightCreated < votableAtHeight;
```

packages/wasm/crate/src/storage.rs
```diff
@@ -386,6 +389,63 @@ impl IndexedDBStorage {
...
+        let delegation_assets = self.get_delegation_assets().await?;
+
+        let all_notes = self
+            .get_notes(NotesRequest {
+                include_spent: true,
+                address_index,
+                asset_id: None,
+                amount_to_spend: None,
+            })
+            .await?;
```
Valentine1898, line 428: You are loading all notes into memory, this could be an issue in the future given that WASM has memory limitations

Valentine1898 (approved):

packages/wasm/crate/src/storage.rs
Valentine1898: Perhaps it would be better if we could pass `votable_notes` to `TransactionPlannerRequest.DelegatorVote`

packages/wasm/crate/src/storage.rs
grod220: This is relying upon the existing NotesRequest handler (https://github.com/penumbra-zone/web/blob/08ba3d65a0fc8f3a6ba2e4e5beafa2b183178ac3/packages/wasm/crate/src/storage.rs#L145-L147) which seems to indicate an unspent note is deserialized as a None.

packages/wasm/crate/src/storage.rs
grod220: The memory note is definitely valid. It's something we'll need some time to think about. Multiple parts of the planner rely upon having all the notes in memory.

packages/wasm/crate/src/storage.rs
grod220: Good catch! Will add.

TalDerei:
will do another round of async review today

TalDerei (commented):
The wasm implementation looks functionally equivalent to the rust code.

PR #1443, opened 2024-07-08
feat: #1383: add shielded address info

### LINKED ISSUE #1383: Shield funds text info
On the IBC page, we should have text that informs the user: “The deposit will be made to a freshly generated address belonging to you. You can inspect what account it points to inside of Prax.”

Also, there should be a way to copy the address next to the address selector component.

Screenshot 2024-07-01 at 2 35 54 PM

VanishMax:
Closes #1383 

image


TalDerei (commented):
If the text is getting cut off like that, we should put the sentences on separate lines.

VanishMax:
@TalDerei Right, forgot to remove the redundant css class. Now it looks better:

image

TalDerei (approved):

PR #1444, opened 2024-07-08
fix: #657: stop truncating metadata symbols

### LINKED ISSUE #657: Use style for truncation insted of string manipulation
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

`text-overflow: ellipsis;` provides automatic truncation that is dynamic to the bounding box, but several places in the repo we implement this behavior in javascript and create strings with ellipses.

in some cases we even have both active on the same element

Screenshot 2024-03-01 at 20 33 41

instead of doing things the hard way, we should just use the style behavior in most cases.

if we have a *specific format* we want to display, in that case we *should not truncate*

VanishMax:
Closes #657 

Removed metadata symbol trimming from wasm and added max-width to the `ValueComponent` with the native tooltip, so users can always access the asset symbol.

image


Valentine1898:
This affects some other ui components
image

It is also necessary to bump indexed-db version, otherwise metadata that is already stored in the DB will not be updated

VanishMax:
@Valentine1898 thanks for noting the issues. Fixed multiple places with incorrect truncation + improved the Dialog component for mobile screens

packages/storage/src/indexed-db/config.ts
```diff
@@ -2,4 +2,4 @@
  * The version number for the IndexedDB schema. This version number is used to manage
  * database upgrades and ensure that the correct schema version is applied.
  */
-export const IDB_VERSION = 44;
+export const IDB_VERSION = 45;
```
Valentine1898, line 5: To clarify, this will cause mainnet users must sync again after updating the extension in chrome store. Should we already consider indexed-db migrations without data loss?

Valentine1898 (approved):
LGTM

packages/storage/src/indexed-db/config.ts
grod220: Think we are ok for now, syncing is very fast with the low block height at the moment

PR #1446, opened 2024-07-08
1441 use the epochduration based price relevance thresholds for delegation token prices

### LINKED ISSUE #1441: Use the epochDuration-based PRICE_RELEVANCE_THRESHOLDS for delegation token prices
https://github.com/penumbra-zone/web/blob/704607a12fb41ef93b69f466f2fae68ee6c6e862/packages/types/src/assets.ts#L3-L8

The issue is that `epochDuration` in mainnet (34560n) is different from epochDuration in testnet (719n), and this causes delegation token prices to be considered outdated before the epoch ends

Valentine1898:
Closes #1441 

image

### DIFF
--- a/packages/services/src/view-service/balances.ts
+++ b/packages/services/src/view-service/balances.ts
@@ -46,12 +46,12 @@ export const balances: Impl['balances'] = async function* (req, ctx) {
 
   const appParameters = await indexedDb.getAppParams();
 
-  const aggregator = new BalancesAggregator(
+  const aggregator = new BalancesAggregator({
     ctx,
     indexedDb,
     latestKnownBlockHeight,
-    appParameters?.sctParams?.epochDuration,
-  );
+    epochDuration: appParameters?.sctParams?.epochDuration,
+  });
 
   for await (const noteRecord of indexedDb.iterateSpendableNotes()) {
     if (noteRecord.heightSpent !== 0n) continue;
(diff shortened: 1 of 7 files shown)


packages/services/src/view-service/balances.ts
```diff
@@ -44,7 +44,14 @@ export const balances: Impl['balances'] = async function* (req, ctx) {
...
 
-  const aggregator = new BalancesAggregator(ctx, indexedDb, latestKnownBlockHeight);
+  const appParameters = await indexedDb.getAppParams();
+
+  const aggregator = new BalancesAggregator(
+    ctx,
+    indexedDb,
+    latestKnownBlockHeight,
+    appParameters?.sctParams?.epochDuration,
+  );
```
grod220, line 54: can you convert this to a object param vs an additional ordered param?

packages/storage/src/indexed-db/index.ts
```diff
@@ -678,9 +682,12 @@ export class IndexedDb implements IndexedDbInterface {
...
   }
 
-  private determinePriceRelevanceThresholdForAsset(assetMetadata: Metadata): number {
+  private determinePriceRelevanceThresholdForAsset(
+    assetMetadata: Metadata,
+    epochDuration: bigint,
+  ): bigint {
     if (assetPatterns.delegationToken.capture(assetMetadata.display)) {
-      return PRICE_RELEVANCE_THRESHOLDS.delegationToken;
+      return epochDuration;
```
grod220, line 690: Can we grab this from the `APP_PARAMETERS` table instead of passing it as an argument down?

Valentine1898: > Can we grab this from the `APP_PARAMETERS` table instead of passing it as an argument down?

We can, but that means we will read it from indexed-db for each asset , instead of reading it once

grod220: Ah true

grod220 (approved):

TalDerei:
@Valentine1898 I imagine we'll have some other conflicting app parameters, some of which were specifically modified for mainnet. I will dm you the list of changes

PR #1450, opened 2024-07-08
react provider 3

turbocrime:
implements

- `<PenumbraContextProvider>` wrapper component
- sync hook `usePenumbra` which acquires context from component
- hooks `usePenumbraTransport` and `usePenumbraService`

`PenumbraContextProvider` verifies presence of an extension before forwarding context, and provides manifest data synchronously.

uses listener interface, so requires https://github.com/prax-wallet/web/pull/92

check out the readme https://github.com/penumbra-zone/web/blob/09d052686de21d1301c8eca3b712aa2517a0be14/packages/react/README.md


turbocrime:
cc @grod220

packages/react/src/hooks/use-penumbra-transport.ts
```diff
@@ -0,0 +1,75 @@
+import {
+  type ChannelTransportOptions,
+  createChannelTransport,
+} from '@penumbra-zone/transport-dom/create';
+import { useEffect, useMemo, useState } from 'react';
+import { usePenumbra } from './use-penumbra.js';
+import { PenumbraInjectionState } from '@penumbra-zone/client';
+
+/** Unconditionally returns a Transport to the provided Penumbra context. This
```
jessepinho, line 9:
```suggestion
/**
 * Unconditionally returns a Transport to the provided Penumbra context. This
```

packages/react/src/hooks/use-penumbra-transport.ts
```diff
@@ -0,0 +1,75 @@
...
+      resolvePort(port);
+    } else if (failure) {
+      rejectPort(failure);
+    }
+  }, [failure, penumbra, port, resolvePort, rejectPort, state]);
+
+  return transport;
+};
+
+/** Promises a Transport to the provided Penumbra context.  Awaits confirmation
```
jessepinho, line 40:
```suggestion
/**
 * Promises a Transport to the provided Penumbra context.  Awaits confirmation
```

packages/react/src/hooks/use-penumbra-transport.ts
```diff
@@ -0,0 +1,75 @@
...
+  // handle context updates
+  useEffect(() => {
+    if (port) {
+      resolvePort(port);
+    } else if (failure) {
+      rejectPort(failure);
+    }
+  }, [failure, penumbra, port, resolvePort, rejectPort, state]);
+
+  return transport;
```
jessepinho, line 37: Does the transport automatically queue up requests until the connection is ready? Or how is the queueing handled?

jessepinho (approved):
Fantastic work! This is really great. I'm excited to see it in use.

I left a lot of comments, but they're almost all very minor things. Feel free to self-merge after addressing.

packages/react/src/util.ts
```diff
@@ -0,0 +1,27 @@
...
+  keyOrigin ? window[PenumbraSymbol]?.[keyOrigin] : undefined;
+
+export const assertStringIsOrigin = (s?: string) => {
+  if (!s || new URL(s).origin !== s) {
+    throw new TypeError('Invalid origin');
+  }
+  return s;
+};
+
+export const assertManifestOrigin = (s?: string, injection?: PenumbraInjection) => {
```
VanishMax, line 21: question: can this be moved to the `@penumra-zone/client/create`?

packages/react/src/manifest.ts
```diff
@@ -0,0 +1,40 @@
+/** Currently, Penumbra manifests are chrome extension manifest v3. There's no type
+ * guard because manifest format is enforced by chrome. This type only describes
+ * fields we're interested in.
+ *
+ * @see https://developer.chrome.com/docs/extensions/reference/manifest#keys
+ */
+export interface PenumbraManifest {
```
VanishMax, line 7: question: can this also be moved to the client package?

packages/react/src/components/penumbra-provider.tsx
```diff
@@ -0,0 +1,204 @@
...
+          ?.connect()
+          .then(p => setProviderPort(p))
+          .catch(setFailureUnknown);
+        break;
+      default:
+        break;
+    }
+  }, [makeApprovalRequest, providerState, providerInjection?.connect, manifest, failure]);
+
+  const createdContext: PenumbraContext = useMemo(
```
grod220, line 172: Do we need to have context? It would be nice if the consumer wouldn't need to wrap their app with a provider. Especially since it appears that this initiates actions automatically. Doesn't all state live within the injection now? Aka, can't we just rely upon a `usePenumbra()` hook without a provider?

packages/react/README.md
```diff
@@ -0,0 +1,207 @@
...
+import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js';
+import { usePenumbraServiceSync } from '@penumbra-zone/react/hooks/use-penumbra-service';
+import { ViewService } from '@penumbra-zone/protobuf';
+import { useQuery } from '@tanstack/react-query';
+import { AccountBalancesTable } from './imaginary-components';
+
+export default function AssetBalancesByAccount({ assetIdFilter }: { assetIdFilter: AssetId }) {
+  const viewClient = usePenumbraServiceSync(ViewService);
+
+  const { isPending, data: groupedBalances } = useQuery({
```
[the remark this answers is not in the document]
turbocrime: this is a naming decision by connectrpc, and we can't really avoid it - i'll add a clarifying comment

packages/react/src/components/penumbra-provider.tsx
```diff
@@ -0,0 +1,204 @@
+import {
+  isPenumbraInjectionStateEvent,
+  PenumbraInjection,
+  PenumbraInjectionState,
+} from '@penumbra-zone/client';
+import { useCallback, useEffect, useMemo, useState, type ReactNode } from 'react';
+import { PenumbraManifest } from '../manifest.js';
+import { PenumbraContext, penumbraContext } from '../penumbra-context.js';
+import { assertManifestOrigin, injectionOfKey, keyOfInjection } from '../util.js';
```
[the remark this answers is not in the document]
turbocrime: this package is using node module resolution, for portability. it requires `.js` extensions on imports.

packages/react/src/components/penumbra-provider.tsx
```diff
@@ -0,0 +1,204 @@
...
+  children?: ReactNode;
+  injection?: PenumbraInjection;
+  origin?: string;
+  makeApprovalRequest?: boolean;
+} & ({ injection: PenumbraInjection } | { origin: string });
+
+export const PenumbraProvider = ({
+  children,
+  origin: providerOrigin,
+  injection: providerInjection,
```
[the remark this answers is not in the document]
turbocrime: i don't like to overload `origin` as it is a global in some environments

packages/react/src/util.ts
```diff
@@ -0,0 +1,27 @@
...
+      keyOrigin &&
+      // matching injection
+      valueInjection === injection,
+  )?.[0];
+
+export const injectionOfKey = (keyOrigin?: string) =>
+  keyOrigin ? window[PenumbraSymbol]?.[keyOrigin] : undefined;
+
+export const assertStringIsOrigin = (s?: string) => {
+  if (!s || new URL(s).origin !== s) {
```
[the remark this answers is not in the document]
turbocrime: technically, a provider could identify itself as any origin hosting a manifest. there is nothing, technically, preventing a chrome extension or even another embedded script in the dapp itself from creating a provider record that uses any protocol and any transport.

in the interest of versatility and forward compatibility, especially with other browsers that may implement provider extensions, i haven't restricted this.

there's no security downside, because anything capable of injecting a provider has already pwned the page.

packages/react/src/util.ts
```diff
@@ -0,0 +1,27 @@
...
+  return s;
+};
+
+export const assertManifestOrigin = (s?: string, injection?: PenumbraInjection) => {
+  const originString = assertStringIsOrigin(s);
+  if (!injection?.manifest || new URL(injection.manifest).origin !== originString) {
+    throw new TypeError('Invalid manifest origin');
+  }
+  return [originString, injection] satisfies [string, PenumbraInjection];
+};
```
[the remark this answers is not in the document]
turbocrime: realized there was some duplication here so these are eliminated, will just call methods from the client package now

packages/react/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { usePenumbra } from './hooks/use-penumbra.js';
+export { PenumbraProvider } from './components/penumbra-provider.js';
```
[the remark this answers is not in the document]
turbocrime: oops, dropped it in a rebase

packages/react/src/util.ts
turbocrime: there's already methods there for this, it was duplicated and now deleted

packages/react/src/hooks/use-penumbra-transport.ts
turbocrime: behavior is in https://github.com/penumbra-zone/web/blob/main/packages/transport-dom/src/create.ts

each request awaits a shared connection attempt as part of its lifecycle, involving the `getPort` option. the connection respects the timeout configuration passed into the transport.

the difference between the two methods is that

- `usePenumbraTransportSync` will create a transport, and the transport will await a port with its timeout clock beginning at the first request.
- `usePenumbraTransport` will promise a transport, but won't init the transport until a port appears in context. the transport will spend no timeout clock.

packages/react/src/components/penumbra-provider.tsx
turbocrime: if the developer just wants to grab a client wherever they are, `@penumbra-zone/client` already provides methods for that.

```ts
import { createPenumbraClient } from '@penumbra-zone/client/create';

// this goes through the same connection process
const viewClient = createPenumbraClient(ViewService, someOptionalOrigin);

export const ChainIdComponent = () => {
    const [chainId, setChainId] = useState<string>();
    useEffect(() => { // or useQuery or whatever
        const { parameters } = await viewClient.appParameters({});
        return setChainId(parameters?.chainId);
    });
    return <div>{chainId}</div>
};
```

i agree that a dedicated react package isn't much of a value-add. at best this may help prevent a naive developer from constructing/deconstructing a transport and client for every component on every rerender. 

yes the injection tracks its own detailed connection state including called methods, so now the caller doesn't have to remember what they've done. but all that it delivers is a `MessagePort`. the page still needs to construct its own transport etc.

and if the goal is to share state among components, context providers are the way to achieve shared state in react, unless we expect users to prop-drill the clients everywhere. and it's convenient to throw in things like prefetched manifest data at this level.

PR #1452, opened 2024-07-08
Start building out Storybook for the Penumbra UI library

jessepinho:
image

This PR is focused on laying the foundation and setting up Storybook to display our UI components. There are a couple basic implementations included here (using styled-components and creating an Icon component) that may be revised or removed later, but I needed something to put into Storybook :)

## In this PR
- CI pipelines (written by Conor) to autodeploy A) tagged releases to `ui.penumbra.zone`, B) merges to `main` to `preview.ui.penumbra.zone`, and C) PR updates to ephemeral URLs.
- Mark all old components as "Deprecated" in Storybook
- Create a Penumbra-branded theme for Storybook
- For now, start using styled-components, although this isn't yet set in stone.
- Create an initial (not yet finished) `Icon` component to test Storybook with.


packages/ui/.gitignore
jessepinho, line 1: This is the directory that Storybook will build to, and will be deployed to `ui.penumbra.zone`

conorsch:
@jessepinho I pushed a new commit for CI workflows for storybook. You can see the bot is already helpfully adding ephemeral URLs. As the commit text describes, there are workflows for preview & stable channels, as well.

packages/ui/src/Icon/index.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+    IconComponent: {
+      options: ['ArrowRightLeft', 'Send', 'Wallet'],
+      mapping: { ArrowRightLeft, Send, Wallet },
+    },
+  },
+};
+
+export default meta;
+
+export const Basic: StoryObj<typeof Icon> = {
```
jessepinho, line 19: image

packages/ui/src/styled-components.d.ts
jessepinho, line 1: Will be built out over time, if we stick with styled-components

packages/ui/package.json
jessepinho, line 1: The reordering of this file is due to `syncpack lint`

jessepinho: only new elements in this file are the dependencies on `styled-components`, `@storybook/manager-api`, and `@storybook/theming`

TalDerei:
@jessepinho maybe we can get a demo of this during next weekly sync?

packages/ui/.firebaserc
turbocrime, line 1: thought: per firebase docs, it seems like firebaserc is supposed to be per-developer config. there is another firebaserc in the repo - should these be present?

packages/ui/.gitignore
turbocrime, line 1: suggestion: we could add this to the top-level gitignore - there may eventually be other packages containing `storybook-static`

also it would prevent the prettierignore diff, as prettier docs say it will respect the gitignore adjacent to prettierignore

packages/ui/.storybook/main.js
turbocrime, line 1: suggestion: i realize the `.storybook` dir already exists, but i think it would be nice to make it non-hidden. docs say it only takes a `-c` cli flag to change this

turbocrime (approved):

packages/ui/.gitignore
jessepinho: good call — done

packages/ui/.storybook/main.js
jessepinho: Hm... I'll punt on this for now, as I'm not too crazy about bucking the Storybook convention. But we can definitely revisit if you feel strongly about it.

PR #1461, opened 2024-07-10
fix eslint

turbocrime:
correcting module resolution generates a large diff, and breaks some of the more strangely configured internal packages.

so this would be a good opportunity to refactor eslint.

eslint config is moved to root and cleaned up.

benefits

1. only one eslint config, and it's no longer broken
2. plugins now working: react, react-hooks, react-refresh, storybook
3. javascript lint now works instead of always failing
5. the fixed eslint is finding bugs
7. added pretty-quick for diff-only formatting

some minor code changes apply fixes.

there is additional configuration in this PR that is currently disabled. it would be good to

1. enable import ordering https://github.com/penumbra-zone/web/blob/83a78ae70a84b8ae132b4be0522eefec20f15ddf/eslint.config.js#L71-L86
2. enable previously-used rules that broke https://github.com/penumbra-zone/web/blob/83a78ae70a84b8ae132b4be0522eefec20f15ddf/eslint.config.js#L149-L151
4. enable extra rules  https://github.com/penumbra-zone/web/blob/83a78ae70a84b8ae132b4be0522eefec20f15ddf/eslint.config.js#L155-L160


package.json
```diff
@@ -44,29 +45,47 @@
...
+    "eslint": "^9.6.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-import-resolver-typescript": "^3.6.1",
+    "eslint-plugin-import-x": "^0.5.3",
+    "eslint-plugin-oxlint": "^0.5.0",
+    "eslint-plugin-prettier": "^5.1.3",
+    "eslint-plugin-react": "7.8.0-rc.0",
+    "eslint-plugin-react-hooks": "5.1.0-rc-c3cdbec0a7-20240708",
+    "eslint-plugin-react-refresh": "^0.4.8",
+    "eslint-plugin-storybook": "0.9.0--canary.156.ed236ca.0",
```
turbocrime, line 73: using canary/next versions of react plugins

vacekj:
Thanks for the PR, much needed changes - props on enabling oxlint, but wondering if replacing both oxlint and eslint with biome would be feasible and desirable? Personally have a lot of good experience there, but wondering if it has everything we need (100% feature parity with eslint). Let me know your thoughts.

turbocrime:
i'd prefer that, but biome rules are named completely differently, and definitely don't cover everything currently used

turbocrime:
removing the oxlint config, as it seems like oxlint isn't actually compatible with eslint flat config yet and there's no easy way to achieve that. the plugin despite documenting flat-config use doesn't provide any supporting functionality - it simply disables rules.

package.json
```diff
@@ -19,10 +20,12 @@
     "dev:pack": "turbo dev:pack --concurrency 16",
     "format": "turbo format",
     "format:prettier": "prettier --write .",
+    "format:pretty-quick": "pretty-quick",
     "format:syncpack": "syncpack format",
-    "lint": "turbo lint -- --max-warnings 0",
```
TalDerei, line 23: per your suggestion, should we apply `--max-warnings 0` to the ci workflows?

TalDerei (approved):

package.json
turbocrime: yes that's preferable, i'll make ci use a per-package `lint:strict` so that

1. acceptable warnings can be configured per-package (nice for new rule introductions)
2. a typical manual call to `pnpm lint` can still show warnings to a developer
3. `lint:strict` can include additional checks like adding a `tsc --noEmit` call

Issue #1459, opened 2024-07-10
displaying improper fee metadata


TalDerei:
In the process of https://github.com/prax-wallet/web/pull/85 and https://github.com/penumbra-zone/web/pull/1419, which specifically addressed the rendering of fees in the transaction approval dialogue and transaction details page for swaps, there have been several stacked changes that seemed to have caused a regression. As a result, the proper fee denomination is no longer being rendered correctly in the approval dialogue and transaction details page.

Additionally, the fee estimation in minifront needs to display the proper fee denom. cc @vacekj

TalDerei:
_update:_ Currently, the FeeParameters are set to zero in the testnet, which means this issue accurately reflects the regressions we're seeing on the testnet. On the mainnet, these parameters are correctly set, so this is not an issue there. However, the fee estimation in minifront using the fixed staking asset denomination remains incorrect.

PR #1468, opened 2024-07-11
fix asset metadata for fee estimation 

TalDerei:
references https://github.com/penumbra-zone/web/issues/1459

currently, the fee estimation in minifront is hardcoded to use the staking token. Instead, we reconstruct the metadata from the `assetId` to display the proper token asset. cc @vacekj 

If the user **has** UM, the will display the native staking token asset

Screenshot 2024-07-11 at 7 49 36 AM

If the user **doesn't have** UM (simulated by setting the `gasFeeToken` to `extractAltFee` in the transaction planner service implementation), it will reconstruct the alternative fee metadata from the assetId.

Screenshot 2024-07-11 at 7 50 43 AM

Currently, the estimated fee is displaying zero since the gas prices are set to zero for that assetId. 

Screenshot 2024-07-11 at 7 50 03 AM

If gas prices aren't set for a particular assetId, it will error with "`GasPrices not available`"

Screenshot 2024-07-11 at 7 50 35 AM


apps/minifront/src/state/send/index.ts
```diff
@@ -84,18 +88,23 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
...
       const fee = txPlan.transactionParameters?.fee;
-      if (!fee?.amount) {
-        return;
-      }
+
+      const feeAssetMetadata = fee?.assetId
+        ? await getAssetTokenMetadata(fee.assetId)
+        : await getStakingTokenMetadata();
+
+      if (!fee?.amount) return;
```
TalDerei, line 103: `fee?.assetId` will return `undefined` when the user has UM, since we opt-out of storing the native staking token assetId in the naive case. In that context, we need to query the staking token metadata from the registry.

apps/minifront/src/fetchers/registry.ts
```diff
@@ -34,6 +35,20 @@ export const getStakingTokenMetadata = async () => {
   return stakingAssetsMetadata;
 };
 
+export const getAssetTokenMetadata = async (assetId: AssetId) => {
```
TalDerei, line 38: will need to be modified after https://github.com/penumbra-zone/web/pull/1472 is merged

vacekj (approved):
lgtm

TalDerei:
requesting additional reviews for assurance after the remote registry work was merged

PR #1469, opened 2024-07-11
Use the generated metadata for delegation tokens

### LINKED ISSUE #887: Extension makes a lot of requests during sync
While attempting to sync on a slow network, I opened the network inspector on the popup to see how it was making progress. I noticed two surprises:

- There were dozens to hundreds of `AssetMetadataById` requests at the outset. Looking at them, they seemed mostly requesting info on delegation tokens. Could these be coalesced into a single request?
- During sync, there were hundreds of `AppParameters` requests. I'm not sure why these are required during sync?

### LINKED ISSUE #1189: Avoid 'AssetMetadataById' queries, use prax registry and wasm metadata generation instead
We have the issue that we create a lot of requests during sync #887
Most of these requests are `AssetMetadataById` for delegation tokens 

https://github.com/penumbra-zone/web/blob/777b68c45eb06fc0d8a4080e9fd2439aff1d1500/packages/query/src/block-processor.ts#L366-L371

There is also an issue for adding an `AssetMetadataByIds` query https://github.com/penumbra-zone/penumbra/issues/4162 

The suggestion is to simply not use `AssetMetadataById` queries, instead use only asset data from https://github.com/prax-wallet/registry and for delegation tokens we can simply generate metadata with wasm using `validator_identity` (we already do that for `lpnft` and `auctionnft`) 
```rust
DelegationToken::from(&validator_identity).denom();
``` 
This should solve #887 and we'll get rid of a hundred requests during sync
Plus, from a privacy standpoint, we should avoid network requests whenever possible

Valentine1898:
closes #1189
closes #887


turbocrime (approved):
nitpick: `staking` => `stake`

i think 'stake' as a noun is a more correct name than 'staking' as a verb. otherwise looks good to me.

Valentine1898:
> nitpick: `staking` => `stake`

Good point, going to rename it

Issue #1359, opened 2024-06-27
Add block height to tx details


grod220:
Transaction details page should have blockHeight prominently at the top. At the moment, it isn't present on the page:

<img width="877" alt="Screenshot 2024-06-27 at 5 20 20 PM" src="https://github.com/penumbra-zone/web/assets/16624263/b111ddb2-8f15-4d9c-af70-8a2f5651b59c">

PR #1470, opened 2024-07-11
feat: display block height in tx detail

vacekj:
fixes https://github.com/penumbra-zone/web/issues/1359

image


turbocrime (approved):
suggestion: it would be nice if the style matched (font style and color)

turbocrime:
i wonder if we should make a practice of consistently displaying 'epoch' along with block height, maybe with a short notation.

TalDerei:
we should keep the style consistent with the transaction hash. I'd also remove the 'in', and maybe you can experiment with encapsulating the block height in a small pill in the top right, like the memo field?

TalDerei (changes requested):
we should keep the style consistent with the transaction hash. I'd also remove the 'in', and maybe you can experiment with encapsulating the block height in a small pill in the top right, similar to the memo field?

vacekj:
@TalDerei addressed your feedback and used the pill.

TalDerei (approved):

PR #1472, opened 2024-07-12
Use remote registry API

grod220:
Since https://github.com/prax-wallet/registry/pull/29 minifront has been using the bundled registry in the npm package. While this has given us consistency of schema and non-reliance upon github, users are feeling the pains of not having a dynamically updated registry source (users not seeing updated icons, using ibc connections would have to wait on chrome store review, etc).

This PR utilizes v10 of the registry which has remote methods that call github for the latest registry: https://github.com/prax-wallet/registry/pull/57.


apps/minifront/src/components/tx-details/tx-viewer.tsx
```diff
@@ -10,19 +10,28 @@ import { useQuery } from '@tanstack/react-query';
...
 
 const OPTIONS = [
   { label: 'Your View', value: TxDetailsTab.PRIVATE },
   { label: 'Public View', value: TxDetailsTab.PUBLIC },
-  { label: 'Reciever View', value: TxDetailsTab.RECIEVER },
+  { label: 'Receiver View', value: TxDetailsTab.RECEIVER },
 ];
 
+const getMetadata: MetadataFetchFn = async ({ assetId }) => {
+  const feeAssetId = assetId ? assetId : new ChainRegistryClient().bundled.globals().stakingAssetId;
```
grod220, line 29: Given `stakingAssetId` should never change, in this PR I often grab this from the bundle

apps/minifront/src/components/tx-details/tx-viewer.tsx
```diff
@@ -10,19 +10,28 @@ import { useQuery } from '@tanstack/react-query';
...
 import { classifyTransaction } from '@penumbra-zone/perspective/transaction/classify';
 import { uint8ArrayToHex } from '@penumbra-zone/types/hex';
+import { viewClient } from '../../clients';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
 
 export enum TxDetailsTab {
   PUBLIC = 'public',
   PRIVATE = 'private',
-  RECIEVER = 'reciever',
+  RECEIVER = 'receiver',
```
grod220, line 19: Typo fix

packages/ui/components/ui/tx/index.tsx
```diff
@@ -2,60 +2,92 @@ import { TransactionView } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbr
...
+        {txv.bodyView?.actionViews.map((av, i) => (
           <ActionViewComponent av={av} feeValueView={feeValueView} key={i} />
         ))}
       </ViewSection>
       <ViewSection heading='Parameters'>
         <ViewBox
           label='Transaction Fee'
           visibleContent={
-            <div className='font-mono'>
+            <div className='flex items-center gap-2'>
```
TalDerei, line 78: think this fixes https://github.com/penumbra-zone/web/issues/1474?

TalDerei (approved):
few questions here:

1. by making live https calls, if the registry becomes unaccessible for some reason, would this block transaction creation?
2. what's the frequency and expected latency of these remote registry requests?

packages/ui/components/ui/tx/index.tsx
grod220: Indeed it does 👍

grod220:
> by making live https calls, if the registry becomes unaccessible for some reason, would this block transaction creation?

No tx blocking would occur. During say a tx approval, it requests metadata from the local view service. If for some reason the service worker wasn't able to reach github, the registry may not be populated and the fee token would simply display as an unknown asset.

The ibc page, however, does rely upon the remote registry to know what ibc channels are available.

> what's the frequency of these remote registry network requests?

Once during service worker instantiation and once on the ibc page

Issue #1347, opened 2024-06-24
Candlestick chart should use symmetric data


turbocrime:
after #1266 it is clear that candlestick chart needs to make a symmetric request to have complete data

PR #1475, opened 2024-07-13
use inverse candlestick data in chart

turbocrime:
fixes https://github.com/penumbra-zone/web/issues/1347 #1183

due to the way price data is recorded, symmetric comparisons do not return symmetric data. to get the complete picture, the client must combine both datasets.

1. query the intended comparison direction (start token -> end token)
2. query the opposite comparison direction (end token -> start token)
3. invert the opposite data (reciprocal values, high becomes low)
4. combine the data (blocks with duplicate records use the highest high, lowest low, sum volumes)

you'll need to run a local devnet and conduct trading activity to view a chart.

### DIFF
--- a/apps/minifront/src/state/swap/helpers.ts
+++ b/apps/minifront/src/state/swap/helpers.ts
@@ -24,7 +24,6 @@ import { fromBaseUnitAmount } from '@penumbra-zone/types/amount';
 import { BalancesResponse } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js';
 import { isKnown } from '../helpers';
 import { AbridgedZQueryState } from '@penumbra-zone/zquery/src/types';
-import { toPlainMessage } from '@bufbuild/protobuf';
 
 export const sendSimulateTradeRequest = ({
   assetIn,
@@ -51,6 +50,15 @@ export const sendSimulateTradeRequest = ({
   return simulationClient.simulateTrade(req);
 };
 
+/**
+ * Due to the way price data is recorded, symmetric comparisons do not return
+ * symmetric data. to get the complete picture, a client must combine both
+ * datasets.
+ *  1. query the intended comparison direction (start token -> end token)
+ *  2. query the inverse comparison direction (end token -> start token)
+ *  3. flip the inverse data (reciprocal values, high becomes low)
+ *  4. combine the data (use the highest high, lowest low, sum volumes)
+ */
 export const sendCandlestickDataRequests = async (
   { startMetadata, endMetadata }: Pick<PriceHistorySlice, 'startMetadata' | 'endMetadata'>,
   limit: bigint,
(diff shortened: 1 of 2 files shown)


apps/minifront/src/state/swap/price-history.ts
```diff
@@ -26,20 +27,18 @@ export const createPriceHistorySlice = (): SliceCreator<PriceHistorySlice> => (s
...
-          swap.priceHistory.endMetadata = endMetadata;
-          swap.priceHistory.candles = data;
-        });
-      }
+    ).then(candles => {
+      set(({ swap }) => {
+        swap.priceHistory.startMetadata = startMetadata;
+        swap.priceHistory.endMetadata = endMetadata;
+        swap.priceHistory.candles = candles;
+      });
```
grod220, line 41: I wonder if zQuery could assist in the fetch lifecycle of this slice & the useEffects in the component

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,22 +64,69 @@ export const sendCandlestickDataRequest = async (
...
+        pair: { start: end, end: start },
+        limit,
+      },
+      { signal },
+    )
+    .catch(suppressAbort);
+
+  const candlestickData = (await directReq)?.data ?? [];
+
+  const inverseCandlestickData = (await inverseReq)?.data ?? [];
```
grod220, line 95: Can you add documentation about the purpose of querying the inverse? It's not clear from this code why it's necessary.

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,22 +64,69 @@ export const sendCandlestickDataRequest = async (
...
+    [
+      ...candlestickData,
+      ...inverseCandlestickData.map(
+        cd =>
+          new CandlestickData({
+            ...toPlainMessage(cd),
+            close: 1 / cd.close,
+            high: 1 / cd.low,
+            low: 1 / cd.high,
+            open: 1 / cd.open,
```
grod220, line 107: What does this accomplish? Can you add a code comment?

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,22 +64,69 @@ export const sendCandlestickDataRequest = async (
...
+      },
+      { signal },
+    )
+    .catch(suppressAbort);
+
+  const candlestickData = (await directReq)?.data ?? [];
+
+  const inverseCandlestickData = (await inverseReq)?.data ?? [];
+
+  const collated = Map.groupBy(
```
grod220, line 97: What's the purpose of a map if it later gets converted into an array and the keys are thrown out? If there is a specific reason, please document. Also, think it would be easier to save it as an object `{ direct: ..., inverse: ...}` than an ordered array. That way you wouldn't have to use the non-null assertion below.

apps/minifront/src/state/swap/helpers.ts
grod220: Is this saying that given swaps:
- A -> B
- B -> A

These are the same category?
- A.close
- 1 / B.close

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -63,22 +64,69 @@ export const sendCandlestickDataRequest = async (
...
+            high: 1 / cd.low,
+            low: 1 / cd.high,
+            open: 1 / cd.open,
+          }),
+      ),
+    ],
+    ({ height }) => height,
+  );
+
+  const combined = Array.from(collated.values()).map(([first, extra]) => {
```
grod220, line 114: think it would be clearer to stick with `direct` & `inverse` versus `first` & `extra`

grod220 (approved):
Mostly just documentation feedback. After addressing, feel free to merge.

apps/minifront/src/state/swap/helpers.ts
turbocrime: commented

apps/minifront/src/state/swap/helpers.ts
turbocrime: order and presence of either are not guaranteed. this was refactored to use 'reduce'

apps/minifront/src/state/swap/price-history.ts
turbocrime: will make zquery adaption a separate issue/pr

PR #1486, opened 2024-07-16
export better manifest type

turbocrime:
`PenumbraManifest` type now refers to `@types/chrome` instead of creating a type from scratch

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -5,20 +5,20 @@ import { PenumbraSymbol } from './symbol.js';
 
 declare global {
   interface Window {
-    /** Records injected upon this global should identify themselves by a field
-     * name matching the origin of the provider. */
+    /** Records injected upon this global should be identified by a name matching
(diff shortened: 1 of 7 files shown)


packages/client/src/index.ts
```diff
@@ -30,14 +30,12 @@ export const getPenumbraManifest = async (
   return manifestJson;
 };
 
-export const getAllPenumbraManifests = (): Record<
-  keyof (typeof window)[typeof PenumbraSymbol],
-  Promise<PenumbraManifest>
-> =>
+/** Fetch all manifests for all providers available on the page. */
+export const getAllPenumbraManifests = (signal?: AbortSignal) =>
```
VanishMax, line 34: suggestion: add function return types. When working on Nextjs example, my IDE had troubles collecting the types correctly, so the explicit typing might help

VanishMax (approved):
Good improvement! Only one suggestion

PR #1464, opened 2024-07-10
WIP: UI library part 2

jessepinho:

PR #1488, opened 2024-07-17
Build out Penumbra UI theme

jessepinho:
This PR extracts part of https://github.com/penumbra-zone/web/pull/1464, so that it can be reviewed in smaller chunks.

## In this PR
- Tweak the Storybook theme
- Build out a `theme` object, and provide it via `ThemeProvider`, based on [Sam's Figma specs](https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=1-112&m=dev). Penumbra UI components will consume this implicitly by using styled-components: when defining CSS-in-JS styles, they can consume the theme like this:
```ts
const Div = styled.div`
  color: ${props => props.theme.color.text.primary};
`
```


packages/ui/src/ThemeProvider/FontFaces.tsx
```diff
@@ -0,0 +1,116 @@
...
+import poppinsItalicLatinExt from './fonts/Poppins-Italic-LatinExt.woff2?url';
+import poppinsMedium from './fonts/Poppins-Medium.woff2?url';
+import poppinsMediumItalic from './fonts/Poppins-MediumItalic.woff2?url';
+import poppinsMediumItalicLatinExt from './fonts/Poppins-MediumItalic-LatinExt.woff2?url';
+import poppinsMediumLatinExt from './fonts/Poppins-Medium-LatinExt.woff2?url';
+import poppinsRegular from './fonts/Poppins-Regular.woff2?url';
+import poppinsRegularLatinExt from './fonts/Poppins-Regular-LatinExt.woff2?url';
+import workSansMedium from './fonts/WorkSans-Medium.woff2?url';
+import workSansMediumLatinExt from './fonts/WorkSans-Medium-LatinExt.woff2?url';
+import workSansMediumVietnamese from './fonts/WorkSans-Medium-Vietnamese.woff2?url';
```
jessepinho, line 13: Downloaded these font files individually from Google Fonts, since Google Fonts somehow has SUPER compressed versions of them.

packages/ui/src/ThemeProvider/fonts.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*?url' {
```
jessepinho, line 1: For getting paths to fonts for `@font-face` declarations

packages/ui/src/ThemeProvider/fonts/IosevkaTerm-Regular.woff2
jessepinho, line 1: Eventually, all these font files will replace the ones we have now

packages/ui/src/tailwindConfig.ts
jessepinho, line 1: Probably still WIP — haven't had a chance to test it yet.

packages/ui/src/ThemeProvider/theme.ts
jessepinho, line 1: These values are based on names/etc. from Sam's designs [here](https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=1-112&m=dev).

VanishMax (approved):
Awesome!

I'm a little afraid of all the fonts loading only after JS is loaded but I we can always tweak the performance later if there are any issues

TalDerei:
> I'm a little afraid of all the fonts loading only after JS is loaded but I we can always tweak the performance later if there are any issues

hmm are the fonts usually relatively heavy enough to impact performance?

VanishMax:
> hmm are the fonts usually relatively heavy enough to impact performance?

The `fonts` dir in this PR is 206KB. This might be more than the sum of all JS packages we're installing

PR #1489, opened 2024-07-17
 Build out new UI components

jessepinho:
## In this PR
- `<Grid />` component, and a Storybook story for it
- `<Text />` component along with corresponding style utilities, and a Storybook story showing how it's used
- `asTransientProps` helper that makes it easy to prefix props with `$`, for use in styled-components


packages/ui/src/Grid/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1489-jessepinho-new-compo-1q7wf18u.web.app/?path=/docs/ui-library-grid--docs

packages/ui/src/Typography.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1489-jessepinho-new-compo-1q7wf18u.web.app/?path=/docs/ui-library-typography--docs

packages/ui/src/Body/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Body text used throughout most of our UIs.
+ *
+ * Renders a `<span />` by default; pass the `as` prop to use a different HTML
+ * element with the same styling.
+ */
+export const Body = styled.span`
+  ${body}
+`;
```
VanishMax, line 12: thought: it was unexpected to see components like `Body`, `Technical` and others being simple texts. The DX of this approach is ultimately not the best. Imagine a common routine of changing one text type to another: users will have to import a new text type, change the usage, and remove the old import.

main suggestion of this PR: simply create a `Typography` component with `type: 'body' | 'h1' | 'h2' | 'large' | etc` property. It will still be lightweight – there's not much logic or styles applied

packages/ui/src/Typography.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { Detail } from './Detail';
+import { H1 } from './H1';
+import { H2 } from './H2';
+import { H3 } from './H3';
+import { H4 } from './H4';
+import { Large } from './Large';
+import { P } from './P';
+import { Small } from './Small';
+import { Strong } from './Strong';
+import { Technical } from './Technical';
```
VanishMax, line 12: nitpick: the directory structure is unnecessary huge. Can we make it all exported from one?

packages/ui/src/utils/typography.ts
```diff
@@ -0,0 +1,126 @@
...
+`;
+
+export const h1 = css`
+  ${base}
+
+  font-family: ${props => props.theme.font.heading};
+  font-size: ${props => props.theme.fontSize.text6xl};
+  font-weight: 500;
+  line-height: ${props => props.theme.lineHeight.text6xl};
+`;
```
VanishMax, line 20: suggestion: rename the components (or Typography's type prop) to align with the design. For example, make `type: 'text-6xl'` be rendered as h1.

I imagine developers working with the designs will see the 'text-6xl' property of some text in the wireframes, they simply put 'text-6xl' in the code and expect the component to render everything correctly

packages/ui/src/utils/typography.ts
```diff
@@ -0,0 +1,126 @@
...
+  line-height: ${props => props.theme.lineHeight.textBase};
+`;
+
+export const tableHeading = css`
+  ${base}
+
+  font-family: ${props => props.theme.font.default};
+  font-size: ${props => props.theme.fontSize.textBase};
+  font-weight: 500;
+  line-height: ${props => props.theme.lineHeight.textBase};
```
VanishMax, line 109: suggestion: might be a one-liner:

```css
font: 500 normal 36px/44px "Poppins", sans-serif
```

packages/ui/src/Body/index.tsx
jessepinho: Ugh so, that's actually what I was doing initially, for the same reason. Then I split them out (https://github.com/penumbra-zone/web/pull/1489/commits/7740d3f7ffc0fdb484d74fafe55120d77676ebb5) into separate components because I realized it would get super verbose to have to wrap a little bit of text in `<Typography type='technical'>{text}</Typography>` instead of just `<Technical>{text}</Technical>`.

I'm not 100% married to the approach that I have here, but I do feel like it's easier to use (and read) than the more verbose approach.

Although... maybe this could be mitigated by just using a more abbreviated component name? e.g., `Text` (like in React Native). And instead of a `type` prop, it would allow you to set a single boolean prop for the type of text you want to render, and it'd use TypeScript to enforce that you'd only pass one, e.g.:

```tsx
<Text h1>This is an H1</Text>
<Text body>This is body text</Text
<Text h2 technical>This will throw a TypeScript error</Text>
```

This would also fix the issue you mentioned of importing a new type and removing the old import.

I think I'm liking this solution! I'm going to implement it and request re-review.

packages/ui/src/Typography.stories.tsx
jessepinho: Will fix per comment above

packages/ui/src/utils/typography.ts
jessepinho: I would do that if we didn't have interpolate variables, but since we do, I think that'd be pretty hard to read — e.g.:

```css
font: ${props => `500 normal ${props.theme.fontSize.textBase}/${props.theme.lineHeight.textBase} ${props.theme.font.default}`};
```

packages/ui/src/utils/typography.ts
jessepinho: So, these are named to match the [typography _semantics_](https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=37-7542&m=dev), rather than the [typography _primitives_](https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=1-112&m=dev). The semantics are designed to group the primitives (such as the 6xl font size, 6xl line height, 500 weight, and heading font) into semantically meaningful groups. 

Figma shows users both the semantics and primitives in dev mode. In this screenshot, the semantic name is the CSS comment (`tab`), while the primitives are the individual rules:
image


Also, more than one semantic grouping share the same design

packages/ui/src/Body/index.tsx
jessepinho: @VanishMax OK, the new `<Text />` component is ready for your review. Please give this PR the 👍🏻 if it looks good to you.

packages/ui/src/utils/typography.ts
VanishMax: got it, thanks for explaining

VanishMax (approved):
Looks really good!

Issue #1454, opened 2024-07-09
display warning message for disabled DEX / IBC functionality 


TalDerei:
The app parameter, `is_enabled`, will be modified upon governance proposals to enable IBC / DEX functionality. Currently, however, the web code doesn't use the parameter in any capacity. This may present issues if the registry contains assets or IBC channels, but the IBC / DEX hasn't been enabled. 

Attempting to submit transactions will fail since the DEX is disabled on-chain. Rather than allowing transactions to fail, we should display a modal with warning message `"DEX / IBC has not been enabled yet"`, effectively disabling the functionality in the UI.

TalDerei:
closing as out of scope for now. If this becomes a desirable feature in the future, we can re-open. cc @hdevalence

PR #1491, opened 2024-07-17
disable the dex and ibc in minifront

TalDerei:
references https://github.com/penumbra-zone/web/issues/1454

cc @hdevalence the warning messages can probably be improved. 

Screenshot 2024-07-17 at 12 05 01 AM

### DIFF
--- a/apps/minifront/src/fetchers/app-params.ts
+++ b/apps/minifront/src/fetchers/app-params.ts
@@ -1,0 +1,7 @@
+import { AppParameters } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/app/v1/app_pb';
+import { viewClient } from '../clients';
+
+export const getAppParameters = async (): Promise<AppParameters | undefined> => {
+  const { parameters } = await viewClient.appParameters({});
+  return parameters;
+};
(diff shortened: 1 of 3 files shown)


grod220:
Instead of a toast that shows up after they attempt an action, how about we just have a little banner at the top of the page with a message? Sort of like the red:
662957b844850e404d74288a_apple-website-notification-bar-1160x467

TalDerei:
> Instead of a toast that shows up after they attempt an action, how about we just have a little banner at the top of the page with a message?

that's a good idea. I think we should also still display a toast with the friendlier error message, effectively disabling the functionality in the UI. Otherwise, the alternative is allowing transaction planning to occur and letting the transactions fail.

the only thing I was worried about is the banner possibly displaying when it shouldn't, given our weird sync behaviors?

Valentine1898:
> the only thing I was worried about is the banner possibly displaying when it shouldn't, given our weird sync behaviors?

I think this can be avoided if we show banner only when `AppParameters` are available and only if `dexEnabled` is false

TalDerei:
closing as out of scope for now, since changes happening in a week should capture this. If this is a desirable feature in the future, we can re-open.

Issue #1437, opened 2024-07-06
prevent multiple claim attempts for swap claims


TalDerei:
The user should _not_ be able to click the claim button multiple times. If a swap claim error occurs, it currently displays multiple error messages that need to be manually cleared. 

https://github.com/penumbra-zone/web/assets/70081547/2c8afdb1-f2f1-497a-bbe1-0fb2fde2d036

PR #1497, opened 2024-07-17
swaps: prevent multiple swap claims

TalDerei:
references https://github.com/penumbra-zone/web/issues/1437

https://github.com/user-attachments/assets/25119b6a-6b4f-4664-bb89-e6523cbd6e07

### DIFF
--- a/apps/minifront/src/components/swap/unclaimed-swaps.tsx
+++ b/apps/minifront/src/components/swap/unclaimed-swaps.tsx
@@ -7,6 +7,7 @@ import { getSwapRecordCommitment } from '@penumbra-zone/getters/swap-record';
 import { uint8ArrayToBase64 } from '@penumbra-zone/types/base64';
 import { GradientHeader } from '@repo/ui/components/ui/gradient-header';
 import { useStoreShallow } from '../../utils/use-store-shallow';
+import { useState } from 'react';
 
 const unclaimedSwapsSelector = (state: AllSlices) => ({
   claimSwap: state.unclaimedSwaps.claimSwap,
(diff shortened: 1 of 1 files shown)


VanishMax:
@TalDerei interestingly, I couldn't reproduce this issue when I tried it. Why doesn't the `isInProgress` function work for you?

TalDerei:
> @TalDerei interestingly, I couldn't reproduce this issue when I tried it. Why doesn't the `isInProgress` function work for you?

not sure, but did you see the screen recording in https://github.com/penumbra-zone/web/issues/1437?

TalDerei:
@VanishMax one explanation might be that `isInProgress` is asynchronous, so there might be a delay in reflecting the state change?

VanishMax:
Yes, I saw the recording. Then tried to reproduce it and failed because the `isInProgress` function worked correctly for me. Can't check right now – don't have unclaimed swaps in the account.

I think there are two options: fix `isInProgress` and remove `isClaiming`, or remove `isInProgress`. Two equal functions at the same place seem like too much

apps/minifront/src/components/swap/unclaimed-swaps.tsx
```diff
@@ -7,15 +7,23 @@ import { getSwapRecordCommitment } from '@penumbra-zone/getters/swap-record';
...
 export const UnclaimedSwaps = () => {
   const unclaimedSwaps = useUnclaimedSwaps();
-  const { claimSwap, isInProgress } = useStoreShallow(unclaimedSwapsSelector);
+  const { claimSwap } = useStoreShallow(unclaimedSwapsSelector);
+  const [claim, setClaim] = useState<string[]>([]);
+
+  // Internal state management for tracking the IDs of the swaps that are currently being claimed.
+  const handleClaim = async (id: string, swap: SwapRecord) => {
+    setClaim(previous => [...previous, id]);
+    await claimSwap(id, swap);
```
VanishMax, line 25: question: would it be useful to remove the `id` from `claim` in case the error happens in `claimSwap` function?

TalDerei: this is a good point, I modified it that if the `claimSwap` throws an error, we remove the id from the claim array, allowing the button to be clicked again for claiming the swap.

VanishMax (approved):

Issue #1433, opened 2024-07-06
check for source in transaction planner request


TalDerei:
Following up from #1397. 

Ensure that a `TransactionPlannerRequest` requires a source. The planner should throw an error and reject the request if the client has not provided a source in the `TransactionPlannerRequest`. This means that transaction plans must be linked to a valid account (address index).

Issue #1494, opened 2024-07-17
manually ending and withdrawing dutch auctions fail


TalDerei:
failure due to not passing in the source in the transaction planner request for auction end and withdraw actions.

<img width="1728" alt="Screenshot 2024-07-17 at 1 48 45 AM" src="https://github.com/user-attachments/assets/c5b3c9d3-b824-4236-8ed0-84f87de68227">

PR #1499, opened 2024-07-17
check for source in transaction planner request

TalDerei:
references https://github.com/penumbra-zone/web/issues/1433


Valentine1898 (approved):
LGTM! 

Does Minifront pass source in all cases?

Consider changing the buf documentation, as it may seem that source is not required
https://buf.build/penumbra-zone/penumbra/file/main:penumbra/view/v1/view.proto#L246

TalDerei:
> Does Minifront pass source in all cases?

it does not, opened https://github.com/penumbra-zone/web/issues/1494

PR #1504, opened 2024-07-17
fix format ci

turbocrime:
eslint reconfig included introduction of pretty-quick for a small speed benefit

doing some local work, a prettier minor bump changed some specific parenthesis formatting around nullish coalescing operators within ternary conditional statements, and it broke pretty-quick, which assumed an unedited file would not need reformatting

this just reverts to using prettier for all formatting and format-checking, since the speed benefit was small


TalDerei (approved):
good to merge after green ci

PR #1510, opened 2024-07-18
re-enable react hooks lint

turbocrime:
and fix

### DIFF
--- a/apps/minifront/src/components/shared/number-input/use-wheel-prevent.ts
+++ b/apps/minifront/src/components/shared/number-input/use-wheel-prevent.ts
@@ -18,7 +18,7 @@ export const useWheelPrevent = () => {
     const ac = new AbortController();
     inputRef.current?.addEventListener('wheel', onWheel, { signal: ac.signal });
     return () => ac.abort();
-  }, [inputRef]);
+  }, []);
 
   return inputRef;
 };
--- a/apps/minifront/src/components/shared/number-input/use-wheel-prevent.ts
+++ b/apps/minifront/src/components/shared/number-input/use-wheel-prevent.ts
@@ -16,7 +16,7 @@ export const useWheelPrevent = () => {
 
   useEffect(() => {
     const ac = new AbortController();
-    inputRef.current?.addEventListener('wheel', onWheel, { signal: ac.signal });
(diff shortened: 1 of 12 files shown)


packages/zquery/src/get-use-hook.ts
turbocrime, line 1: @jessepinho it seems like following the 'rules of hooks' as defined by the linter causes tests to fail in zquery. any comments, refactor suggestions?

turbocrime: there's also a lexical block here that should probably have a comment describing its purpose

apps/minifront/src/components/shared/number-input/use-wheel-prevent.ts
```diff
@@ -15,12 +15,10 @@ export const useWheelPrevent = () => {
...
-    inputRef.current?.addEventListener('wheel', onWheel);
-
-    return () => {
-      inputRef.current?.removeEventListener('wheel', onWheel);
-    };
-  }, []);
+    const ac = new AbortController();
+    inputRef.current?.addEventListener('wheel', onWheel, { signal: ac.signal });
+    return () => ac.abort();
+  }, [inputRef]);
```
jessepinho, line 21:
```suggestion
  }, []);
```

Not needed, since `inputRef` never changes. (And ESLint doesn't complain.)

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -85,7 +85,7 @@ export const TokenSwapInput = () => {
     } else {
       return priceHistory.load();
     }
-  }, [assetIn, assetOut]);
+  }, [assetIn, assetOut, priceHistory]);
```
jessepinho, line 88:
```suggestion
  }, [assetIn, assetOut, priceHistory.load]);
```

(Otherwise, this effect will run any time _any_ property of the `priceHistory` slice changes.)

jessepinho (approved):
Couple small comments, then ship it!

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
turbocrime: ah, this might have been related to some of the state churn i mentioned. thanks!

PR #1512, opened 2024-07-18
re-enable react refresh lint

turbocrime:
and fix

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -101,7 +101,7 @@ export default tseslint.config(
           allowExpressions: true,
         },
       ],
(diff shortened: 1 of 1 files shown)


VanishMax:
@turbocrime @grod220 do we need this after #1721 and #1510 ?

turbocrime:
react-refresh is distinct from react-hooks

it does look like we missed it in #1721 https://github.com/penumbra-zone/web/blob/b61060401c99fb9a13b4fb39c6f5a133515b6b97/eslint.config.js#L104

turbocrime:
#1750 to remove the unused config

PR #1516, opened 2024-07-18
Create `<SegmentedPicker />`, `<Button />`, `<ButtonGroup />` components

jessepinho:
Originally was going to split out `<SegmentedPicker />` and `<Button* />` to separate PRs, but it turned out that they used shared styles, so I'm PRing them together.

## In this PR
- Create a `<SegmentedPicker />` component
- Create a `<Button />` component
- Create a `<ButtonGroup />` component
- Create a set of shared style utilities that takes care of hover/active/etc. states
- Create Storybook stories for them all, and write some tests

Closes #115


packages/ui/src/Button/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1516-jessepinho-segmented-d5zkeqe2.web.app/?path=/docs/ui-library-button--docs

packages/ui/src/ButtonGroup/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1516-jessepinho-segmented-d5zkeqe2.web.app/?path=/docs/ui-library-buttongroup--docs

packages/ui/src/SegmentedPicker/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1516-jessepinho-segmented-d5zkeqe2.web.app/?path=/docs/ui-library-segmentedpicker--docs

jessepinho:
Moving back to draft to address some changes that came up in my recent 1:1 with Sam

jessepinho:
Just was one change after all. This PR is ready for review again 👍🏻

packages/ui/src/Button/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+   * ```tsx
+   * import { ChevronRight } from 'lucide-react';
+   * <Button icon={ChevronRight}>Label</Button>
+   * <Button icon={ChevronRight} iconOnly>Label</Button>
+   * ```
+   */
+  icon?: LucideIcon;
+}
+
+export type ButtonProps = BaseButtonProps & (IconOnlyProps | RegularProps);
```
VanishMax, line 147: todo: add html-native `type` attribute and assign it to `button` as default. Otherwise it will always trigger a submit and reload if used within a form

packages/ui/src/Button/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  actionType = 'default',
+}: ButtonProps) => {
+  const priority = useContext(ButtonPriorityContext);
+
+  return (
+    <StyledButton
+      {...asTransientProps({ iconOnly, size, actionType, priority })}
+      disabled={disabled}
+      onClick={onClick}
+      aria-label={iconOnly ? children : undefined}
```
VanishMax, line 166: suggestion (a11y): do the same as `aria-label` with the `title` attribute, so the text description of the icon button will be visible on long hover

packages/ui/src/SegmentedPicker/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {options.map(option => (
+        <SegmentButton
+          key={option.value.toString()}
+          onClick={() => onChange(option.value)}
+          $actionType={actionType}
+          disabled={option.disabled}
+          $getFocusOutlineColor={theme => theme.color.action[outlineColorByActionType[actionType]]}
+          $getBorderRadius={theme => theme.borderRadius.xs}
+        >
+          {value === option.value && <SelectedIndicator layout layoutId={layoutId} />}
```
VanishMax, line 131: praise: great use of animations!

packages/ui/src/SegmentedPicker/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+}: SegmentedPickerProps<ValueType>) => {
+  const layoutId = useId();
+
+  return (
+    <Root>
+      {options.map(option => (
+        <SegmentButton
+          key={option.value.toString()}
+          onClick={() => onChange(option.value)}
+          $actionType={actionType}
```
VanishMax, line 126: question: what does `$` mean here?

packages/ui/src/SegmentedPicker/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  onChange,
+  options,
+  actionType = 'default',
+}: SegmentedPickerProps<ValueType>) => {
+  const layoutId = useId();
+
+  return (
+    <Root>
+      {options.map(option => (
+        <SegmentButton
```
VanishMax, line 123: thought/issue (a11y): semantically, `SegmentedPicker` is used as either a radio button group, or as tabs. Each of these usages have their W3C standards for accessibility. For now, the component doesn't fill the standards fully:
1. Tabs: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/. `SegmentedPicker` doesn't support arrow navigation and aria attributes like `aria-selected` for the selected button, etc.
2. Radio group: https://www.w3.org/WAI/ARIA/apg/patterns/radio/. `SegmentedPicker` doesn't use the required roles and keyboard navigation.

I don't mind having this version of a component for now – it looks good and works correctly, at least partially accessible with the keyboard. Yet, I believe that in the future `SegmentedPicker` should be split into `RadioGroup` and `Tabs` components correspondingly.

packages/ui/src/ButtonPriorityContext.ts
```diff
@@ -0,0 +1,14 @@
...
+ * For internal Penumbra UI library use only. `<Button />`s should only ever be
+ * of the `primary` priority, unless they are a secondary button in a
+ * `<ButtonGroup />`.
+ *
+ * We don't want to expose a `priority` property on `<Button />`, because we
+ * don't want engineers to ever manually assign `priority='secondary'`. So
+ * instead, we'll create this context, which we'll set from inside `<ButtonGroup
+ * />` and consume from inside `<Button />`.
+ */
+export const ButtonPriorityContext = createContext<Priority>('primary');
```
VanishMax, line 14: suggestion: move the file to `utils` or somewhere else from the root `src` folder.

packages/ui/src/utils/button.ts
```diff
@@ -0,0 +1,84 @@
+import { css, DefaultTheme } from 'styled-components';
+
+export type ActionType = 'default' | 'accent' | 'unshield' | 'destructive';
+
+export type Priority = 'primary' | 'secondary';
```
VanishMax, line 5: question: why is this called `priority`? And why is it controlled by `ButtonPriorityContext` and not just set by user?

packages/ui/src/Button/index.tsx
jessepinho: Done!

packages/ui/src/Button/index.tsx
jessepinho: Great idea. [Done](https://github.com/penumbra-zone/web/pull/1516/commits/777c06abd4970a4369dd972fbffc8bf07bce9950).

packages/ui/src/SegmentedPicker/index.tsx
jessepinho: It's a styled-components convention for [transient props](https://styled-components.com/docs/api#transient-props) — props that are not passed to the underlying DOM elements.

packages/ui/src/ButtonPriorityContext.ts
jessepinho: done

packages/ui/src/utils/button.ts
jessepinho: I need to write some docs for some of the choices Sam and I are making regarding naming in the UI library; one of them is that `priority` represents the choice of a `primary` vs `secondary` item in a set (rather than using a generic term like `variant`). The idea is that our props should indicate to devs how the components are _used_, rather than how they _look_.

Re: the context, see [here](https://github.com/penumbra-zone/web/pull/1516/files#diff-d116d5fdfb5c55efb62165b5508b1e0edb0f3cb3c0204ad804215e9295b815ba).

packages/ui/src/SegmentedPicker/index.tsx
jessepinho: Ah, this is a good call-out — this seems like a great candidate for using an already existing component (like Radix UI's [Tabs](https://www.radix-ui.com/primitives/docs/components/tabs)). Since we're going to be using `<SegmentedPicker />` for tabs anyway, that seems like the right fit. I'll move this PR back to draft until this is addressed.

jessepinho: Actually, on second thought, I'll do this as a separate PR to keep PRs smaller.

PR #1520, opened 2024-07-19
don't miss validator state transitions

### LINKED ISSUE #1453: Jailed validators are not correctly removed from the set of active validators
Not all validators are returned in a `ValidatorInfoResponse`, even if the flag `showInactive` is `true`.

This means that some classes of validator, including jailed validators, will fail to transition out of 'active' state.

The block processor should either

1. clear the validator table every time it updates validator info, instead of upserting.
2. or, compare the list of returned validators to the table, to identify validators missing from the response.

If (2) then the block processor could specifically query validator info for specific validators that the user may be interested in.

turbocrime:
fixes #1453

if you have a wallet that will witness the specific state transition, please test


TalDerei:
closed by https://github.com/penumbra-zone/web/issues/1453#issuecomment-2270836323

PR #1526, opened 2024-07-22
Create `<Pill />`/`<ValueViewComponent />` components

jessepinho:
## In this PR
- Create a `<Pill />` component
- Create a `<ValueViewComponent />` component
- Create a `<ConditionalWrap />` utility component for conditionally wrapping markup in other markup


packages/ui/src/Pill/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1526-jessepinho-pill-vj0bxyyq.web.app/?path=/docs/ui-library-pill--docs

packages/ui/src/ValueViewComponent/AssetIcon/DelegationTokenIcon.tsx
jessepinho, line 1: This file is copied over almost verbatim from [here](https://github.com/penumbra-zone/web/blob/main/packages/ui/components/ui/asset-icon/delegation-token-icon.tsx). It just has some tweaks to how it uses the `size` prop, and it now uses styled-components.

packages/ui/src/ValueViewComponent/AssetIcon/UnbondingTokenIcon.tsx
jessepinho, line 1: Like `DelegationTokenIcon`, this file is copied over almost verbatim from [here](https://github.com/penumbra-zone/web/blob/main/packages/ui/components/ui/asset-icon/unbonding-token-icon.tsx). It just has some tweaks to how it uses the `size` prop, and it now uses styled-components.

packages/ui/src/ValueViewComponent/AssetIcon/index.tsx
jessepinho, line 1: Like `DelegationTokenIcon`, this file is copied over almost verbatim from [here](https://github.com/penumbra-zone/web/blob/main/packages/ui/components/ui/asset-icon/index.tsx). It just has some tweaks to how it uses the `size` prop, and it now uses styled-components.

packages/ui/src/ValueViewComponent/AssetIcon/Identicon/index.tsx
jessepinho, line 1: Copied almost verbatim from [here](https://github.com/penumbra-zone/web/blob/main/packages/ui/components/ui/identicon/index.tsx)

packages/ui/src/ValueViewComponent/AssetIcon/Identicon/generate.ts
jessepinho, line 1: Copied verbatim from [here](https://github.com/penumbra-zone/web/blob/main/packages/ui/components/ui/identicon/generate.ts)

packages/ui/src/ValueViewComponent/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1526-jessepinho-pill-vj0bxyyq.web.app/?path=/docs/ui-library-valueviewcomponent--docs

jessepinho:
sorry about the changes y'all — got some tweaks in Figma that I had to update here

packages/ui/src/ValueViewComponent/sampleValueViews.ts
```diff
@@ -0,0 +1,111 @@
...
+const uunbondString = 'uunbonding_start_at_123_' + validatorIkString;
+const unbondAsset = { inner: u8(32) };
+
+const DELEGATION_TOKEN_METADATA = new Metadata({
+  display: delString,
+  base: udelString,
+  denomUnits: [{ denom: udelString }, { denom: delString, exponent: 6 }],
+  name: 'Delegation token',
+  penumbraAssetId: delAsset,
+  symbol: 'delUM(abc...xyz)',
```
VanishMax, line 23: issue: the symbol should be truncated

image

previously, we removed any programmatic truncation from the token symbols, and they theoretically can be long enough to not fit into the view. This way, CSS truncation is needed.

and from a11y point, when truncation happens, the `title` attribute is the best to show the full symbol in the tooltip

packages/ui/src/ConditionalWrap.tsx
```diff
@@ -0,0 +1,60 @@
...
+ *     <NonTooltipWrapper>{children}</NonTooltipWrapper>
+ *   )}
+ * >
+ *   Here is the content that may or may not need a tooltip.
+ * </ConditionalWrap>
+ * ```
+ *
+ * @see https://stackoverflow.com/a/56870316/974981
+ */
+export const ConditionalWrap = ({
```
VanishMax, line 52: suggestion: could you move the file somewhere from the `src`? I feel like this directory should only have directories inside. As for shared components or utils, they could go into `shared` or `utils` without `index.tsx` in them, so it won't be exported outside the library

VanishMax (changes requested):
only one small bug 🐛

packages/ui/src/ConditionalWrap.tsx
jessepinho: Done! 

FWIW, components in src will not get exported if they're not structured like `ComponentName/index.tsx` (see our [`exports`](https://github.com/penumbra-zone/web/blob/main/packages/ui/package.json#L24)).

That said, I agree with your suggestion to move it to avoid confusion.

packages/ui/src/ValueViewComponent/sampleValueViews.ts
jessepinho: Hm... [we’re still programmatically truncating _symbols_](https://github.com/penumbra-zone/web/blob/cbc2419/packages/wasm/crate/src/metadata.rs#L171), although perhaps not _display names_, of assets. And it’s not just at runtime — the programmatically truncated symbols are stored in the database that way as well.

We could still put a max-width + text-overflow on these, but that has security implications: if two symbols start with the same text, and we don't show the complete text, users could think they're e.g., sending one token when they're actually sending a different one. If we show the complete symbol, that can't happen. (Granted, we're truncating the symbol when it includes e.g., a validator ID, but those shouldn't have any collisions.)

I'm tempted to leave it as-is, to avoid this security issue. Then, consumers of the component would be forced to accommodate a variable width in their layout (possibly by adding a horizontal scrollbar if needed).

Your thoughts?

VanishMax: Interestingly, the tests in Wasm package fail. They are not integrated into the CI, so we didn't even know about it (created issue #1546). I forgot to change them in [this commit](https://github.com/penumbra-zone/web/commit/cbc241931c45651ffbc31d5bd4f8db2dec4f0164) that removed the truncation. 

Agree on the risk of showing the equal first park of two symbols. Maybe we could programmatically truncate the inner part of the braces on the client and, for example, show it on hover?

```html
<p>
  <span>unboundUMat123(abc</span>
  <span title="unboundUMat123(abcdefghklmnxyz)">...</span>  <!-- Notice the title that could be a tooltip  -->
  <span>xyz)</span>  
</p>
```

jessepinho: Oh! Didn't realize the implementation had changed.

OK, hm. Since we don't know that all symbols will have braces like the unbonding/delegation tokens, I'd rather not programmatically truncate them. Instead, I've just pushed up [a commit](https://github.com/penumbra-zone/web/pull/1526/commits/35ad53e71cf5277451b8c72dd3ddd32fc87a8716) that truncates it with CSS and then adds a `title` with the full symbol. 

There is [this weird way](https://stackoverflow.com/a/10017343/974981) to detect if the CSS ellipsis has been applied. I'd love to do that, and show a tooltip in that case, but I'd prefer to do that in a separate PR.

(Re: the build failures — there appears to be a CI issue. Resolving that separately with L...)

jessepinho: Screenshot:

image

Issue #1530, opened 2024-07-22
Price history slice should use ZQuery for candlestick request


turbocrime:
some initial work on this

https://github.com/penumbra-zone/web/commit/b09c0831af895a2b20c37ba519ba5b44e4aab45e

PR #1533, opened 2024-07-23
candlestick chart updates

### LINKED ISSUE #1529: Swap form inputs should be located together at the top
Currently, the swap token input component includes the candlestick graph. This pushes the other form input components down the page. All of the form inputs should be located together at the top.

The candlestick graph should be located below, with the visualization components. The candlestick graph should also be vertically smaller.

turbocrime:
fixes https://github.com/penumbra-zone/web/issues/1530
fixes #1529 

- move candlestick chart from token input to independent component
- use zquery for fetching candlestick data
- some minor changes to graph rendering

Screenshot 2024-07-23 at 04 24 10

### DIFF
--- a/apps/minifront/src/components/swap/swap-form/price-history.tsx
+++ b/apps/minifront/src/components/swap/swap-form/price-history.tsx
@@ -11,7 +11,7 @@ import { useCandles, useRevalidateCandles } from '../../../state/swap/price-hist
 import { useStoreShallow } from '../../../utils/use-store-shallow';
 import { Button } from '@repo/ui/components/ui/button';
 
-const priceHistoryRequestSelector = (state: AllSlices) => ({
+const priceHistorySelector = (state: AllSlices) => ({
   startMetadata: getMetadataFromBalancesResponseOptional(state.swap.assetIn),
   endMetadata: state.swap.assetOut,
   historyLimit: state.swap.priceHistory.historyLimit,
@@ -26,14 +26,13 @@ const latestKnownBlockHeightSelector = (
 ) => state.data?.latestKnownBlockHeight;
 
 export const PriceHistory = () => {
-  const { startMetadata, endMetadata, historyLimit, historyStart } = useStoreShallow(
-    priceHistoryRequestSelector,
-  );
+  const { startMetadata, endMetadata, historyLimit, historyStart } =
+    useStoreShallow(priceHistorySelector);
   const latestKnownBlockHeight = useStatus({ select: latestKnownBlockHeightSelector });
 
   const [lastCandlesUpdate, setLastCandlesUpdate] = useState<bigint>();
   const [staleCandles, setStaleCandles] = useState<boolean>();
-  const [pause, setPause] = useState<bigint>();
+  const [pauseAtHeight, setPauseAtHeight] = useState<bigint>();
 
   const candles = useCandles(
     { select: combinedCandlestickDataSelector },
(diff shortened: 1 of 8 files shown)


apps/minifront/src/components/swap/swap-form/price-history.tsx
```diff
@@ -0,0 +1,111 @@
...
+    if (!staleCandles && latestKnownBlockHeight && lastCandlesUpdate) {
+      setStaleCandles(latestKnownBlockHeight - lastCandlesUpdate > 10n);
+    }
+  }, [lastCandlesUpdate, latestKnownBlockHeight, pause, staleCandles]);
+
+  useEffect(() => {
+    if (staleCandles && !pause) {
+      setStaleCandles(false);
+      setLastCandlesUpdate(latestKnownBlockHeight);
+      refetchCandles(...candleRefetchArgs);
```
jessepinho, line 65: Hm, I should probably have changed the `revalidate` function to return a promise, to allow you to do things like `setLastCandlesUpdate()` _after_ revalidation completes, but 🤷🏻‍♂️ 

More of a note to myself — nothing needs to change here

apps/minifront/src/state/swap/helpers.ts
```diff
@@ -74,73 +75,71 @@ export const sendCandlestickDataRequests = async (
...
+          // highest high, lowest low
+          acc.high = Math.max(acc.high, cur.high);
+          acc.low = Math.min(acc.low, cur.low);
+
+          // these accumulate to be averaged
+          acc.open += cur.open;
+          acc.close += cur.close;
+          return acc;
+        },
+        new CandlestickData({ height, low: Infinity, high: -Infinity }),
```
jessepinho, line 128: This is probably my own ignorance, but shouldn't `low` and `high` be swapped? i.e., 
```suggestion
        new CandlestickData({ height, low: -Infinity, high: Infinity }),
```

apps/minifront/src/state/swap/price-history.ts
```diff
@@ -1,47 +1,73 @@
...
+  /**
+   * History limit becomes the maximum width of the chart domain (block height).
+   */
+  setHistoryLimit: (limit: bigint) => void;
+  /**
+   * Setting history start will cause the chart domain to begin at the specified
+   * block height and extend towards the present. Setting history start to
+   * `undefined` or `0n` will cause the chart domain to end at the present block
+   * height and extend towards the past.
+   */
```
jessepinho, line 18: We love a good doc block 👏🏻

apps/minifront/src/state/swap/price-history.ts
```diff
@@ -1,47 +1,73 @@
...
+  fetch: (
+    startMetadata: Metadata | undefined,
+    endMetadata: Metadata | undefined,
+    historyLimit: bigint | undefined,
+    historyStart: bigint | undefined,
+  ) =>
+    Promise.all([
+      sendCandlestickDataRequest(startMetadata, endMetadata, historyLimit, historyStart),
+      sendCandlestickDataRequest(endMetadata, startMetadata, historyLimit, historyStart),
+    ]).then(([direct, inverse]) => ({ direct, inverse })),
```
jessepinho, line 33: suggestion: make this a named function import

Not a requirement, but it's a little cleaner to have something like:
```suggestion
  fetch: getDirectAndInverseCandlestickData,
```

...and thus keep logic out of the ZQuery object definition.

apps/minifront/src/state/swap/price-history.ts
```diff
@@ -1,47 +1,73 @@
...
-      PRICE_RELEVANCE_THRESHOLDS.default * 2n,
-      ac.signal,
-    ).then(candles => {
-      set(({ swap }) => {
-        swap.priceHistory.startMetadata = startMetadata;
-        swap.priceHistory.endMetadata = endMetadata;
-        swap.priceHistory.candles = candles;
-      });
+  setHistoryLimit: blocks => {
+    useStore.setState(state => {
```
jessepinho, line 63:
```suggestion
    set(state => {
```

That's the pattern we're using throughout the rest of Zustand (except in ZQuery, which _has_ to use `useStore`).

jessepinho (approved):
LGTM after addressing the `issue` comment I left. Everything else is non-blocking. Let me know if you want me to re-review. Nice work!

apps/minifront/src/components/swap/swap-form/simulate-swap-result/index.tsx
```diff
@@ -18,40 +18,50 @@ import { useStoreShallow } from '../../../../utils/use-store-shallow';
...
-          getDisplayDenomExponentFromValueView.optional()(state.swap.assetIn?.balanceView),
+          getDisplayDenomExponentFromValueView(state.swap.assetIn.balanceView),
         ),
-        metadata: getMetadata.optional()(state.swap.assetIn?.balanceView),
+        metadata: getMetadata(state.swap.assetIn.balanceView),
       },
     },
-  }),
-});
+  });
```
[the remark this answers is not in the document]
turbocrime: this was occasionally causing the simulate swap result component to attempt an undefined access, and it would error the page.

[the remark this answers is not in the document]
turbocrime: i can revert it and make a separate issue. sorry, should have documented this.

apps/minifront/src/state/swap/helpers.ts
turbocrime: no, this is the 'initial' value of the reduce, which are intended to be clobbered by the first arriving values.

`Math.min(Infinity, x)` always returns x, `Math.max(-Infinity, y)` always returns y. but `Math.min(undefined, x)` returns `NaN`.

apps/minifront/src/components/swap/swap-form/simulate-swap-result/index.tsx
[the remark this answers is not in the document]
turbocrime: wasn't able to replicate this so maybe it was some side-effect of my devnet

PR #1419, opened 2024-07-04
improve fee rendering in transaction details page

TalDerei:
corresponding prax PR: https://github.com/prax-wallet/web/pull/85. fixes regressions from https://github.com/penumbra-zone/web/issues/1265

**before**

Screenshot 2024-07-04 at 2 16 57 PM

----------------------------------------------------------------------------------------------------------------------------

**after**

Screenshot 2024-07-04 at 2 12 52 PM

PR #1539, opened 2024-07-23
Delegator voting action view

grod220:
Adds action view support to the tx-details page. Paired w/ @plaidfinch to get current design.

Visible:

Screenshot 2024-07-23 at 10 07 43 PM

Opaque:
Screenshot 2024-07-23 at 10 07 48 PM

### DIFF
--- a/packages/ui/components/ui/tx/actions-views/delegator-vote.tsx
+++ b/packages/ui/components/ui/tx/actions-views/delegator-vote.tsx
@@ -1,5 +1,4 @@
 import { ViewBox } from '../viewbox';
-import { ValueWithAddress } from './value-with-address';
 import { ValueViewComponent } from '../../value';
 import { getAddress } from '@penumbra-zone/getters/note-view';
 import { ActionDetails } from './action-details';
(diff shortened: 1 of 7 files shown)


Valentine1898 (approved):
LGTM 

Unbonded UM is the same as UM?

grod220:
It's the equivalent value of your delegation token in UM

packages/ui/components/ui/tx/actions-views/delegator-vote.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {
+  Metadata,
+  ValueView,
+} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb.js';
+import { AddressViewComponent } from '../../address-view';
+import { base64ToUint8Array } from '@penumbra-zone/types/base64';
+
+// TODO: This is sad, but at the moment, we aren't provided the metadata to have a rich display.
+//       Given the high-priority of getting action view support, this is added.
+//       We should properly implement ValueViews into the protos for DelegatorVote action view and delete this code.
```
TalDerei, line 22: @grod220 retro reviewing.

regarding the to-do, can we retrieve this information from the `TransactionViewComponent` similarly to how we handle it for the `SwapViewComponent`? refer to this [example](https://github.com/penumbra-zone/web/pull/1419/files#diff-c4407fc8843a60c40b2a2bb104fad82b6a50e11ef1807380dff00f9fc5e6c4e6R16-R37).

grod220: I'm not quite seeing what you mean. The issue is that DelegatorVoteView is missing ValueView's. Until we update those protos, we won't have what we need for this view.

TalDerei:
retro review LGTM

PR #1543, opened 2024-07-23
`<SegmentedPicker />`: Use Radix UI's Tabs component under the hood, and rename to `<Tabs />`

jessepinho:
This PR addresses @VanishMax 's [comment](https://github.com/penumbra-zone/web/pull/1516#discussion_r1687901787) about accessibility issues with the `<SegmentedPicker />` component. To address the issues, I refactored `<SegmentedPicker />` to use Radix UI's fully accessible `<Tabs />` component under the hood. Per Sam's agreement, I also renamed this component to `<Tabs />`.

This component may be modified further once it's being used in actual pages, so this is just a gradual improvement for now.

### DIFF
--- a/packages/ui/src/Tabs/index.stories.tsx
+++ b/packages/ui/src/Tabs/index.stories.tsx
@@ -1,11 +1,10 @@
 import type { Meta, StoryObj } from '@storybook/react';
 import { useArgs } from '@storybook/preview-api';
 
-import { SegmentedPicker } from '.';
+import { Tabs } from '.';
 
-const meta: Meta<typeof SegmentedPicker> = {
-  component: SegmentedPicker,
-  title: 'SegmentedPicker',
+const meta: Meta<typeof Tabs> = {
+  component: Tabs,
   tags: ['autodocs', '!dev'],
   argTypes: {
     value: { control: false },
@@ -15,7 +14,7 @@ const meta: Meta<typeof SegmentedPicker> = {
 };
 export default meta;
 
-type Story = StoryObj<typeof SegmentedPicker>;
+type Story = StoryObj<typeof Tabs>;
 
 export const Basic: Story = {
   args: {
@@ -34,6 +33,6 @@ export const Basic: Story = {
 
     const onChange = (value: { toString: () => string }) => updateArgs({ value });
 
-    return <SegmentedPicker {...props} onChange={onChange} />;
+    return <Tabs {...props} onChange={onChange} />;
   },
 };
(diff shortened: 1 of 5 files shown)


packages/ui/src/utils/button.ts
```diff
@@ -31,10 +31,15 @@ const focusOutline = css<{
    * disabled button, the overlay of the disabled button would be above the
    * outline, making the outline appear to be partly cut off.
    */
-  &:focus::after {
+  &:focus-within {
+    outline: none;
+  }
+
+  &:focus-within::after {
```
jessepinho, line 38: Radix UI's `<Tabs />` uses `focus-within` rather than `focus`, so this addresses that change.

packages/ui/src/Tabs/index.tsx
```diff
@@ -68,37 +69,27 @@ const SelectedIndicator = styled(motion.div)`
...
 
-export interface SegmentedPickerOption<ValueType> {
-  /**
-   * The value to pass to the `onChange` handler when clicked. Must be unique
-   * across all segments, and must be either a string, number, or an object with
-   * a `.toString()` method so that it can be used as a React key.
-   */
-  value: ValueType;
+export interface TabsTab {
+  value: string;
```
jessepinho, line 73: Radix UI's `<Tabs />` only supports `string`s as the value type.

packages/ui/src/Tabs/index.stories.tsx
```diff
@@ -34,6 +33,6 @@ export const Basic: Story = {
 
     const onChange = (value: { toString: () => string }) => updateArgs({ value });
 
-    return <SegmentedPicker {...props} onChange={onChange} />;
+    return <Tabs {...props} onChange={onChange} />;
```
VanishMax, line 36: suggestion (non-blocking, worth implementing): expose `TabsContent` component from Radix and somehow allow the `children` for `Tabs`.

If you inspect the markup of the [Radix Tabs example](https://www.radix-ui.com/primitives/docs/components/tabs), you can see that `Tabs.Content` components are nested inside the `Tabs.Root`. It allows Radix to set useful attributes like `data-state="active"` and `role="tabpanel"`. Plus, there is a rendering logic involved that all `[role="tabpanel"]` divs are rendered but inactive divs have empty contents. All of this corresponds to W3C standards.

Developers might and probably will forget to implement this, so it would be useful to help them (and ourselves)

VanishMax (approved):

packages/ui/src/Tabs/index.stories.tsx
jessepinho: Good call — will look into it. I'm not sure that `<TabsContent />` will be appropriate in our case, since we won't have multiple tab contents rendered on the page at the same time — `<Tabs />` will usually be used for routing, I believe. But I created [a ticket to look into it](https://github.com/prax-wallet/web/issues/131) nonetheless.

PR #1545, opened 2024-07-24
Create `<Table />` component; introduce the `<Density />` context/component

jessepinho:
image

This PR creates the `<Table />` component, and also introduces a `<Density />` component and context. 

Previously, a lot of the new Penumbra UI components were using `size` props. This wasn't the best name for a prop with a type of `'sparse' | 'dense'`, so after some discussion, Sam and I decided to rename the concept from `size` to `density`, and rename the values to `'sparse' | 'compact'`. 

This PR also changes `density` (née `size`) to no longer be a prop, but a context. That's because it'd be impossible to have consistent layouts if all components were manually passing `density` to each other; instead, we use a context so that an entire region of layout is consistent.

The `<Table />` component was the perfect component to start using `density` as a context, rather than a prop, because there are so many nested components inside a `<Table />` that can take advantage of the context and avoid prop drilling.

I wanted to go through all the _other_ components that currently use the `size` prop and refactor them to use `density`, but I resisted the temptation in favor of smaller PR scope. Once this PR is approved, I'll go back and change the rest of them.


packages/ui/src/Table/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1545-jessepinho-table-xq38icu4.web.app/?path=/docs/ui-library-table--docs

packages/ui/src/Table/index.tsx
jessepinho, line 1: Note that, unlike what I did with the `<Text />` component, with `<Table />` I decided to make each sub-component a property on `Table` — e.g., `<Table.Thead />`, `<Table.Td />`, etc. — rather than a React prop — e.g., `<Table thead />`, `<Table td />`, etc. 

This is because table elements are always deeply nested, and the closing tags would be hard to read if they're all the same component:

```tsx
<Table>
  <Table tbody>
    <Table tr>
      <Table td>
      </Table>
    </Table> {/* ("Wait, which element does this correlate to?!") */}
  </Table>
</Table>
```

VanishMax (approved):
love the changes!

praise: great use of context for the `Density` component 🔥

PR #1549, opened 2024-07-24
Fix remaining __DEV__ globals

grod220:
Found more after https://github.com/penumbra-zone/web/pull/1547

### DIFF
--- a/apps/minifront/vite.config.ts
+++ b/apps/minifront/vite.config.ts
@@ -8,7 +8,7 @@ import polyfillNode from 'vite-plugin-node-stdlib-browser';
 
 export default defineConfig(({ mode }) => {
   return {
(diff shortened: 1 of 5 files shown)


apps/minifront/vite.config.ts
```diff
@@ -8,7 +8,7 @@ import polyfillNode from 'vite-plugin-node-stdlib-browser';
 
 export default defineConfig(({ mode }) => {
   return {
-    define: { 'globalThis.__DEV__': mode === 'development' },
+    define: { 'globalThis.__DEV__': mode !== 'production' },
```
grod220, line 11: During a test run, the mode is apparently `test`. So inverting the conditional so it can catch this mode as well. Otherwise, some tests fail.

TalDerei (approved):

Issue #1546, opened 2024-07-24
Run `cargo test` in Wasm package in GitHub actions


VanishMax:
Running `cargo test` within `packages/wasm/crate` fails with errors – existing tests are not passing, they were broken with recent changes. Wouldn't have happened if we had tests integrated into the CI

vacekj:
AFAIK, Rust tests are already runnig in the CI - see `turbo-ci.yml:145`

https://github.com/penumbra-zone/web/blob/161ed5c2589059c3b34fada4bf61bd6f6a0bf550/.github/workflows/turbo-ci.yml#L145

Running this command locally runs the tests and they pass, they are also passing in CI: https://github.com/penumbra-zone/web/actions/runs/10198450011/job/28213555671

@jessepinho @VanishMax asking permission to close.

VanishMax:
@vacekj Please check #1616 PR – I broke the local tests in wasm package but CI runs successfully.

Looks like `turbo test:rust` doesn't work correctly

vacekj:
I see - it's because test:rust doesn't run all of the tests - it specifically doesn't run tests that are in source files and are included using #[cfg(test)] - looking more into this.

vacekj:
Okay, so the reason why wasm-pack test doesn't run the tests, is because it uses `cargo build --test`, which doesn't compile (and later run) tests outside of the `tests` directory. This is why `cargo test` picks them up and fails, but `wasm-pack test` doesn't.

wasm-pack looks like it has pretty poor maintenance, so instead of trying to fix this upstream, I will move the tests from the source directory into the tests directory.

PR #1552, opened 2024-07-24
Customize vote symbol

grod220:
Vote receipt tokens now have proper symbols:

Before:
Screenshot 2024-07-24 at 7 55 44 PM

After:
Screenshot 2024-07-24 at 7 17 38 PM

Also related to https://github.com/penumbra-zone/web/issues/1546, fixed a bunch of tests

### DIFF
--- a/.changeset/old-rocks-count.md
+++ b/.changeset/old-rocks-count.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/wasm': minor
+---
+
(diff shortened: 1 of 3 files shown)


Valentine1898 (approved):
LGTM! 

Perhaps it's better to display such assets as just “Voting receipt” (so it doesn't look like other assets)?

grod220:
Let's stick with this for now and we can iterate later if it's not clear to voters 👍

PR #1553, opened 2024-07-24
Use murmurhash for identicon colors

grod220:
We are currently using `djb2a` to hash the symbol (string -> number) to generate an identicon color background. The issue is that hashing algorithm produces similar outputs for similar inputs. While working on VotingReceiptToken displays, it became obvious this made it harder to distinguish proposal tokens:

Proposal 61
Screenshot 2024-07-24 at 8 19 11 PM
Proposal 62
Screenshot 2024-07-24 at 8 19 26 PM

This PR implements murmurhash. This makes like strings generate completely different numbers (which in turn generates unique colors):

Proposal 61
Screenshot 2024-07-24 at 8 17 56 PM
Proposal 62
Screenshot 2024-07-24 at 8 18 12 PM

The only downside to this is that is kinda "resets" the identities of folks address identicons. Given how early we are in the chain launch, I find it acceptable.


Valentine1898 (approved):
> The only downside to this is that is kinda "resets" the identities of folks address identicons. Given how early we are in the chain launch, I find it acceptable

As far as I understand, in the current phase of mainnet the user had no way to get an asset with a generated icon, so this should definitely not be an issue

Interestingly, I originally thought the red icon corresponded to a “No” vote and the green one to a “Yes” vote

Issue #1558, opened 2024-07-25
remove source requirement account from transaction request


TalDerei:
In order to support backwards compatibility with the existing frontend code already deployed, we should remove the assertion check (https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/transaction-planner/index.ts#L70) and switch to using the default zero source in the case of a missing source in a transaction planner request. 

cc @Valentine1898

TalDerei:
closing as not currently planned

PR #1559, opened 2024-07-25
Default planner source

grod220:
Closes https://github.com/penumbra-zone/web/issues/1558


TalDerei (approved):
removing my review until further discussion takes place.

Valentine1898:
I believe that using the default source may not be obvious to developers, and may lead to a case where funds are mistakenly spent from the wrong account.

In addition, the RPC documentation does not mention that if source is omitted, the default value will be used
```
// If present, only spends funds from the given account.
penumbra.core.keys.v1.AddressIndex source = 4;
``` 

But at the same time, the rust planner uses the default value for source (and it's not `unwrap_or_default`), and we probably want the rust planner and extension planner to be consistent
https://github.com/penumbra-zone/penumbra/blob/08f3b2b0e97509169601f35b22160500932162e8/crates/view/src/service.rs#L763-L768

PR #1560, opened 2024-07-25
modify description for shielded swaps

TalDerei:
references https://github.com/penumbra-zone/web/issues/1548

open to suggestions here

PR #1561, opened 2024-07-25
1555 cannot end and withdraw auction from accounts other than 0 account

### LINKED ISSUE #1555: Cannot end and withdraw auction from accounts other than 0 account
**Describe the bug**
If you create an auction from any other account other than account 0, you will not be able to end that auction and will get an error 
image

https://github.com/penumbra-zone/web/blob/5cac34fbb65b69569c694a268d0066fac6a94912/apps/minifront/src/state/swap/dutch-auction/index.ts#L179-L180

This is because getting `source` from `swap.assetIn` doesn't work and `source=0` will always be used

**Screenshots**
image

Valentine1898:
closes #1555

### DIFF
--- a/apps/minifront/src/state/swap/dutch-auction/index.ts
+++ b/apps/minifront/src/state/swap/dutch-auction/index.ts
@@ -12,7 +12,7 @@ import { ZQueryState, createZQuery } from '@penumbra-zone/zquery';
 import { AuctionInfo, getAuctionInfos } from '../../../fetchers/auction-infos';
 import { Metadata } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb.js';
 import { bech32mAuctionId } from '@penumbra-zone/bech32m/pauctid';
-import { getAddressIndex } from '@penumbra-zone/getters/balances-response';
+import { AddressIndex } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js';
 
 /**
  * Multipliers to use with the output of the swap simulation, to determine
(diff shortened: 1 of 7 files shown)


packages/ui/components/ui/dutch-auction-component/expanded-details/index.tsx
```diff
@@ -104,6 +107,12 @@ export const ExpandedDetails = ({
           <AuctionIdComponent auctionId={auctionId} />
         </Row>
       )}
+
+      {addressIndex && (
+        <Row label='Account'>
+          <span className='font-mono text-xs'>{addressIndex.account}</span>
+        </Row>
+      )}
```
Valentine1898, line 115: Before: 
image
After:
image

apps/minifront/src/state/swap/dutch-auction/index.ts
```diff
@@ -175,18 +179,19 @@ export const createDutchAuctionSlice = (): SliceCreator<DutchAuctionSlice> => (s
     }
   },
 
-  endAuction: async auctionId => {
-    const source = getAddressIndex.optional()(get().swap.assetIn);
-    const req = new TransactionPlannerRequest({ dutchAuctionEndActions: [{ auctionId }], source });
+  endAuction: async (auctionId, addressIndex) => {
+    const req = new TransactionPlannerRequest({
+      dutchAuctionEndActions: [{ auctionId }],
+      source: addressIndex,
```
TalDerei, line 185: ack. passing in the `addressIndex` rather than fetching it from the swap information makes sense.

Valentine1898:
I have no idea why CI is broken

TalDerei:
> I have no idea why CI is broken

same with #1560

grod220:
A new version of Rust was shipped. Do `rustup update` locally. It looks like we have a dep with a build issue. We should create a new PR to fix this to unblock work.

grod220 (approved):
Nice catch!

TalDerei (approved):

PR #1563, opened 2024-07-25
rm -fv tgz before dev:pack

turbocrime:
prevents accidentally referring to old tarballs

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "dev": "turbo dev:compile dev:app",
     "dev:app": "turbo dev:compile dev:app",
     "dev:compile": "turbo dev:compile",
-    "dev:pack": "turbo dev:pack --concurrency 16",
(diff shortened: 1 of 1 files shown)


grod220 (approved):

package.json
```diff
@@ -17,7 +17,7 @@
     "dev": "turbo dev:compile dev:app",
     "dev:app": "turbo dev:compile dev:app",
     "dev:compile": "turbo dev:compile",
-    "dev:pack": "turbo dev:pack --concurrency 16",
+    "dev:pack": "rm -rfv ./packages/*/penumbra-zone-*.tgz && turbo dev:pack --concurrency 16",
```
conorsch, line 20: Nit: you don't need the `-r` option, because the shell globs are well specified, and will always result in a list of filenames.

PR #1566, opened 2024-07-25
use `Any.pack` instead of manually creating `Any` message

turbocrime:
this is the correct way to create a packed `Any` message


grod220 (approved):

jessepinho (approved):
Nice! Didn't know about this.

Issue #1523, opened 2024-07-22
Remove Provider in favor of hook


grod220:
The [PenumbraContextProvider](https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L19) is a react component that consumers are required to wrap around their app.

It should be possible to convert this to a hook so consumers aren't required to wrap their app in it.

PenumbraState has a failure state, so this shouldn't be necessary:
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L32-L49

`assertGlobalPresent()` in the client package performs this:
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L51-L66

This will have to remain, but can live in a hook and attach itself on mount:
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L68-L94

Should not exist (https://github.com/penumbra-zone/web/issues/1522):
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L96-L117

Can be moved to hook:
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L119-L149

Can be exposed by hook:
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L151-L196

All told, I don't see a reason to require consumers to have a global context for this when a hook would be easier.

turbocrime:
it's a scoped context. a provider doesn't have to be used at the top level of the page.

the provider can be used as a top level component to provide a scope that encompasses the page.

or it can be used in a smaller component, and will then be scoped only part of the page.

the context provider and hook model is common among other react wallet packages and APIs.

i've done some work to make a hook executing without a provider use a default global scope. i'll push that branch up soon as a PR.

PR #1567, opened 2024-07-26
fulfill adr 006

### LINKED ISSUE #1521: Client instantiation should not trigger approval
Currently, the [createPromiseClient](https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/client/src/create.ts#L77-L80) function triggers an approval dialog:

https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/client/src/create.ts#L26-L28

This has causes consumers to use some work-around patterns to have lazy initialization so it only does this during a request. Example:

```ts
const viewClientGetter = async () => {
    return await createPenumbraClient(ViewService, PRAX_ORIGIN);
  }
```

And then they would do something like 
```ts
const viewClent = await viewClientGetter()
const result = viewClient.addressById({ .. })
```

In our own example repo, we do the same thing: https://github.com/penumbra-zone/nextjs-penumbra-client-example/blob/main/app/fetchers.ts#L8-L10

**TODO:** Make client instantiation lazy. Do not invoke any kind of connection or approval requests until a request is made. Clients are responsible for acquiring connection state separately via a connect button.

### LINKED ISSUE #1522: Remove `makeApprovalRequest` from provider
https://github.com/penumbra-zone/web/blob/f515113fac3b4e3a4d7a2561c20ac0f06d535abd/packages/react/src/components/penumbra-context-provider.tsx#L22

Consumers should thoughtfully acquire approval requests via a connect button. Auto-approval requests are an anti-pattern in crypto and have been known to annoy users.

Todo: Remove `makeApprovalRequest` from hook + logic for auto approval requests.

turbocrime:
deletes react package fixes #1522 #1523

fixes #1521

prax-wallet/web#120 will need work in prax but this interface is compatible 

matching PR https://github.com/prax-wallet/web/pull/134

cc @VanishMax review

interfaces very close to meeting ADR as written.

1. request and connect are no longer distinct methods
2. `isConnected` returns a boolean only
3. PenumbraClient wraps the page API
4. PenumbraClient is created with `createPenumbraClient`
5. PenumbraClient is connected with an instance method which returns manifest URL
6. `createServiceClient` is available

Minifront is updated to use `createPenumbraClient` to init a transport. Larger refactor to support `createServiceClient` is out of scope.


packages/client/src/index.ts
```diff
@@ -1,46 +1,68 @@
...
-): Promise<PenumbraManifest> => {
-  const manifestJson = await assertProviderManifest(penumbraOrigin, signal);
-  if (!isPenumbraManifest(manifestJson)) {
-    throw new TypeError('Invalid manifest');
+  private penumbraStateEventListener = (evt: Event) => {
+    if (isPenumbraStateEvent(evt)) {
+      for (const callback of this.callbacks) {
+        void callback(evt.detail);
+      }
+    }
```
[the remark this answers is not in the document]
turbocrime: it seems awkward to re-specify this type without failing to respect some feature of the DOM event listener api, so i haven't re-specified the type any narrower. specifying a narrower type would also risk consumers writing handlers that aren't capable of dealing with the flexibility of the interface.

i think with the client wrapper providing the simpler callback interface, and documentation of what types the event listeners must support, this is adequate.

welcome to suggestions that don't break anything.

VanishMax (approved):
One small suggestion and questions. The rest is all good 👍

apps/minifront/src/components/extension-not-connected.tsx
```diff
@@ -34,11 +35,12 @@ const handleErr = (e: unknown) => {
 
 const useExtConnector = () => {
   const [result, setResult] = useState<boolean>();
+  const navigate = useNavigate();
 
   const request = async () => {
     try {
       await requestPraxAccess();
```
VanishMax, line 42: issue: after the update, reloading Minifront results in displaying the connection screen.

suggestion: add `isConnected()` check and, if connected, `navigate('/')`

packages/client/src/get.ts
```diff
@@ -0,0 +1,58 @@
+import { assertPenumbra, assertProviderManifest, assertProviderRecord } from './assert.js';
+import { PenumbraManifest, isPenumbraManifest } from './manifest.js';
+import type { PenumbraProvider } from './provider.js';
+import { PenumbraSymbol } from './symbol.js';
+
+import './global.js';
+
+/** Return the specified provider, without verifying anything. */
+export const getPenumbraUnsafe = (penumbraOrigin: string) =>
```
grod220, line 9: issue: not sure this naming makes sense if it can safely return undefined

grod220 (approved):
What's the plan with the react-package? Will there be a follow up to update this?

packages/client/src/client.ts
```diff
@@ -0,0 +1,98 @@
...
+import { PenumbraProviderNotConnectedError } from './error.js';
+import { PenumbraStateEventDetail, PenumbraStateEvent, isPenumbraStateEvent } from './event.js';
+import { availableOrigin } from './get.js';
+import { PenumbraProvider } from './provider.js';
+import { PenumbraState } from './state.js';
+
+export interface IPenumbraClient {
+  /**
+   * Asks users to approve the connection to a specific browser manifest URL.
+   * If `manifest` argument is not provided, tries to connect to the first injected provider.
```
VanishMax, line 11: suggestion (idea): 

```suggestion
   * If `manifest` argument is not provided, tries to connect to the previously-connected or the first injected provider.
```

from the implementation point, it's going to look over the `isConnected()` of all injected providers and connect to the first one with `true`

turbocrime:
closing in favor of #1647 which will see more pushes with implementation of updated interface

PR #1570, opened 2024-07-26
prevent swap NaN

turbocrime:
sometimes, the swap page will throw and error the page after completing a swap. the conditions for triggering this are unclear.
 
Screenshot 2024-07-26 at 4 31 32 PM

### DIFF
--- a/apps/minifront/src/components/swap/swap-form/simulate-swap-result/index.tsx
+++ b/apps/minifront/src/components/swap/swap-form/simulate-swap-result/index.tsx
@@ -18,40 +18,50 @@ import { useStoreShallow } from '../../../../utils/use-store-shallow';
 const HIDE = { clipPath: 'polygon(0 0, 100% 0, 100% 0, 0 0)' };
 const SHOW = { clipPath: 'polygon(0 0, 100% 0, 100% 100%, 0 100%)' };
 
-const simulateSwapResultSelector = (state: AllSlices) => ({
-  input: new ValueView({
+const simulateSwapResultSelector = (state: AllSlices) => {
+  if (!state.swap.amount || !state.swap.assetIn?.balanceView) {
+    return undefined;
+  }
+  return new ValueView({
     valueView: {
       case: 'knownAssetId',
       value: {
         amount: toBaseUnit(
           new BigNumber(state.swap.amount),
-          getDisplayDenomExponentFromValueView.optional()(state.swap.assetIn?.balanceView),
+          getDisplayDenomExponentFromValueView(state.swap.assetIn.balanceView),
         ),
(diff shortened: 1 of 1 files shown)


apps/minifront/src/components/swap/swap-form/simulate-swap-result/index.tsx
```diff
@@ -18,40 +18,50 @@ import { useStoreShallow } from '../../../../utils/use-store-shallow';
...
     valueView: {
       case: 'knownAssetId',
       value: {
         amount: toBaseUnit(
           new BigNumber(state.swap.amount),
-          getDisplayDenomExponentFromValueView.optional()(state.swap.assetIn?.balanceView),
+          getDisplayDenomExponentFromValueView(state.swap.assetIn.balanceView),
         ),
-        metadata: getMetadata.optional()(state.swap.assetIn?.balanceView),
+        metadata: getMetadata(state.swap.assetIn.balanceView),
```
jessepinho, line 33: Same issue as before — why are these no longer optional? No PR description so I have 0 context on this change

turbocrime: i'm trying to prevent throwing a NaN in some unidentified conditions. it seems like maybe getters are asserting types to aggressively. what is bad about these not being optional?

turbocrime: it seems like `getDesplayDenomExponentFromValueView` includes an optional inside, as well, so changing this syntax might not have any effect

jessepinho: But how would the metadata be related to a NaN? Again, this stuff needs to be explained in the PR description. And in your updated description, it looks like you're not sure if this will actually fix it, right? And yet it could be introducing a bug where the getter would throw if metadata is missing, which seems like incorrect behavior.



> it seems like `getDesplayDenomExponentFromValueView` includes an optional inside, as well, so changing this syntax might not have any effect

Ah, so the way getters work is that they should never throw from inside the getter function. Instead, they should just return `undefined` if the value they are trying to get can't be returned. Then, the _user_ of the getter can determine whether or not it should throw in case of `undefined` (by leaving off, vs. calling, `.optional()` on the getter). The reason `getDisplayDenomExponentFromValueView` calls `.optional()` is because the getter _itself_ is OK with returning `undefined`. But then a consumer can call `getDisplayDenomExponentFromValueView.optional()(valueView)` or just `getDisplayDenomExponentFromValueView(valueView)` if they want it to not throw vs. throw, respectively. Make sense?

PR #1574, opened 2024-07-27
improve getter tools, don't export optional getters

turbocrime:
improves legibility of computed result of the `Getter` type
- third 'optional' type parameter removed - now simply shows source and target
- type no longer contains verbose constant conditions
- now shows `undefined` as part of the target type, when the target type includes undefined
- now shows `NonNullable` as part of the target type, when the target type does not include undefined

getters package no longer exports optional getters. exporting optional getters has the potential to cause confusion. a consumer who wishes to use an optional getter can use the `.optional()` utility provided on every getter.

`.optional()` can now be applied at any point in a getter pipe, which affects the pipe in an intuitive way
- the pipe before the `.optional()` call will continue to assert results
- the pipe after the `.optional()` call may return undefined


TalDerei (approved):

TalDerei (approved):

TalDerei:
other reviews would be helpful to sanity check

packages/ui/tsconfig.json
turbocrime, line 1: `GetterMissingValueError` uses [es2022 'error cause'](https://tc39.es/proposal-error-cause/) which was previously permitted by this tsconfig, but for some reason downgraded to es2020 at some point.

the many other fields added by a8a5f41f7d791059e47f873c6948a26a057e674b were already present in the configs which are extended, so all of those added fields had no effect.

so it's reverted essentially to abcaab6, plus a newer lib. i can't identify a practical difference aside from the lib option.

https://github.com/penumbra-zone/web/blob/abcaab6fa139ebcdb84a0eb91726c91c355cdf66/packages/ui/tsconfig.json#L1-L10

turbocrime (commented):
rebased

VanishMax (approved):
great DX improvement 👍

Issue #1569, opened 2024-07-26
Undelegation fails in a confusing way if account does not contain token to pay fee


turbocrime:
**Describe the bug**
The error a user sees when attempting to 'Undelegate' from an account without an asset capable of paying a transaction fee is confusing

**To Reproduce**

1. Delegate 100% of tokens from an account, or, delegate some tokens from an account and then transfer all assets out of the account.
2. Go to the 'Stake' page
3. Click 'Undelegate'
4. See error 'Could not extract alternative fee assetId from TransactionPlannerRequest'

**Expected behavior**
User should see an error that explains the account needs some other token to pay undelegation transaction fees. Or, if fee is 0, the user should not see any error and the transaction should succeed.

**Screenshots**
<img width="821" alt="Screenshot 2024-07-26 at 12 56 44" src="https://github.com/user-attachments/assets/42617507-809d-4d64-bb1c-f7ead10cd21e">

**Additional context**
I examined this issue on a local devnet with 0 fees.

TalDerei:
> The error a user sees when attempting to 'Undelegate' from an account without an asset capable of paying a transaction fee is confusing

https://github.com/penumbra-zone/web/pull/1577/commits/fd52094fac7737681827fc89c7d0b12118654ecf

PR #1577, opened 2024-07-29
modify extractAltFee to be smarter

TalDerei:
see code comments for motivation behind this work. additionally, references https://github.com/penumbra-zone/web/issues/1569 bug

--------------------------------------------------------------------------------------------------------------------------
**high-level flow:**

1. Fetch all alternative gas prices from the GasPrices table.

2. Extract assetId from the transaction request and check that:
     - The balance for the asset is positive
     - The asset is a valid fee token with non-zero gas prices
     
3. If the above checks fail, fall back to using another asset from the GasPrices table.

### DIFF
--- a/apps/minifront/vitest.config.ts
+++ b/apps/minifront/vitest.config.ts
@@ -5,7 +5,5 @@ export default defineConfig({
     snapshotSerializers: [],
     environment: 'jsdom',
     setupFiles: ['./tests-setup.ts'],
-    silent: false,
-    reporters: 'default',
   },
 });
--- a/apps/minifront/vitest.config.ts
+++ b/apps/minifront/vitest.config.ts
@@ -5,5 +5,7 @@ export default defineConfig({
     snapshotSerializers: [],
     environment: 'jsdom',
     setupFiles: ['./tests-setup.ts'],
+    silent: false,
+    reporters: 'default',
   },
 });
--- a/packages/services/src/view-service/fees.ts
+++ b/packages/services/src/view-service/fees.ts
@@ -105,20 +105,22 @@ export const extractAltFee = async (
 export const getAssetFromGasPriceTable = async (
   request: TransactionPlannerRequest,
   indexedDb: IndexedDbInterface,
-  specificAssetId?: AssetId,
+  assetId?: AssetId,
 ): Promise<AssetId | undefined> => {
+  // Fetch alternative gas prices, excluding those with a zero price.
   const altGasPrices = await indexedDb.getAltGasPrices();
 
-  // If a specific asset ID is provided, check its balance is positive and GasPrices for that asset are set.
-  if (specificAssetId) {
-    const balance = await indexedDb.hasTokenBalance(request.source!, specificAssetId);
-    const filteredGasPrices = altGasPrices.filter(gp => gp.assetId?.equals(specificAssetId));
-    if (balance && filteredGasPrices.length > 0) {
-      return specificAssetId;
+  // If a specific asset ID is provided, extracted from the transaction request, check its balance is
+  // positive and GasPrices for that asset exist.
+  if (assetId) {
+    const balance = await indexedDb.hasTokenBalance(request.source!, assetId);
+    if (balance) {
+      return assetId;
     }
   }
 
-  // If no specific asset ID is provided or if the specific asset ID has no balance, check assets in GasPrices table.
+  // If no specific asset ID is provided or if the specific asset ID has no balance, check assets in the
+  // GasPrices table as a fallback case.
   for (const gasPrice of altGasPrices) {
     if (gasPrice.assetId) {
       const balance = await indexedDb.hasTokenBalance(request.source!, gasPrice.assetId);
--- a/packages/services/src/view-service/fees.ts
+++ b/packages/services/src/view-service/fees.ts
@@ -4,45 +4,43 @@ import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
 import { IndexedDbInterface } from '@penumbra-zone/types/indexed-db';
 
 // Attempts to extract a fee token, with priority in descending order, from the assets used
-// in the actions of the transaction planner request (TPR).
-//
-// (1) For outputs, spends, swaps, scheduling auctions, and ICS withdrawals, the naive approach
-// of "using the first alternative token for which the user has a balance" is similar to "extracting
-// the input asset from the transaction request". While both methods aim to retrieve an alternative
-// fee token, they may result in selecting different tokens.
-//
-// (2) Swap claims are a special case where the claim must use the swap's fee token.
-//
-// (3) Ending and withdrawing auctions are another special case that requires a database query to
-// retrieve the relevant auction and the corresponding asset ID used in that auction. For example,
-// if a Dutch auction is scheduled with a fee paid using GM, ending and withdrawing the auction will
-// also attempt to use GM. However, if the user spends their entire GM balance after scheduling the
-// auction, they won't be able to end or withdraw it because it expects GM to pay for fees. In this
-// scenario, the fee system will use the first available alternative token balance from the
-// GAS_PRICES IndexedDB table. This allows the user to end and withdraw auctions using another asset,
-// such as GN, if GM is unavailable.
+// in the actions of the transaction planner request (TPR). If no fee token is found from the
+// specified assets, it falls back to checking the gas prices table for an asset with a positive balance.
 export const extractAltFee = async (
   request: TransactionPlannerRequest,
   indexedDb: IndexedDbInterface,
 ): Promise<AssetId> => {
   const outputAsset = request.outputs.map(o => o.value?.assetId).find(Boolean);
   if (outputAsset) {
-    return outputAsset;
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb, outputAsset);
+    if (assetId) {
+      return assetId;
+    }
   }
 
   const spendAsset = request.spends.map(o => o.value?.assetId).find(Boolean);
   if (spendAsset) {
-    return spendAsset;
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb, spendAsset);
+    if (assetId) {
+      return assetId;
+    }
   }
 
   const swapAsset = request.swaps.map(assetIn => assetIn.value?.assetId).find(Boolean);
   if (swapAsset) {
-    return swapAsset;
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb, swapAsset);
+    if (assetId) {
+      return assetId;
+    }
   }
 
   const ics20WithdrawAsset = request.ics20Withdrawals.map(w => w.denom).find(Boolean);
   if (ics20WithdrawAsset) {
-    return assetIdFromBaseDenom(ics20WithdrawAsset.denom);
+    const withdrawAsset = assetIdFromBaseDenom(ics20WithdrawAsset.denom);
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb, withdrawAsset);
+    if (assetId) {
+      return assetId;
+    }
   }
 
   const swapCommitment = request.swapClaims
@@ -50,7 +48,7 @@ export const extractAltFee = async (
     .find(Boolean);
   if (swapCommitment) {
     const swaps = await indexedDb.getSwapByCommitment(swapCommitment);
-    // If the claimFee assetId is undefined, it means the swap was made with the stakingTokenAsset
+    // If the claimFee assetId is undefined, it means the swap was made with the stakingTokenAsset.
     return swaps?.swap?.claimFee?.assetId ?? indexedDb.stakingTokenAssetId;
   }
 
@@ -58,24 +56,21 @@ export const extractAltFee = async (
     .map(a => a.description?.input)
     .find(Boolean);
   if (auctionScheduleAsset?.assetId) {
-    return auctionScheduleAsset.assetId;
+    const inputAssetId = auctionScheduleAsset.assetId;
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb, inputAssetId);
+    if (assetId) {
+      return assetId;
+    }
   }
 
   const auctionEndAsset = request.dutchAuctionEndActions.map(a => a.auctionId).find(Boolean);
   if (auctionEndAsset) {
     const endAuction = await indexedDb.getAuction(auctionEndAsset);
     if (endAuction.auction?.input?.assetId) {
-      const checkAssetBalance = await indexedDb.hasTokenBalance(
-        request.source,
-        endAuction.auction.input.assetId,
-      );
-      if (checkAssetBalance) {
-        return endAuction.auction.input.assetId;
-      } else {
-        const assetId = await getAssetFromGasPriceTable(request, indexedDb);
-        if (assetId) {
-          return assetId;
-        }
+      const inputAssetId = endAuction.auction.input.assetId;
+      const assetId = await getAssetFromGasPriceTable(request, indexedDb, inputAssetId);
+      if (assetId) {
+        return assetId;
       }
     }
   }
@@ -86,17 +81,10 @@ export const extractAltFee = async (
   if (auctionWithdrawAsset) {
     const withdrawAuction = await indexedDb.getAuction(auctionWithdrawAsset);
     if (withdrawAuction.auction?.input?.assetId) {
-      const checkAssetBalance = await indexedDb.hasTokenBalance(
-        request.source,
-        withdrawAuction.auction.input.assetId,
-      );
-      if (checkAssetBalance) {
-        return withdrawAuction.auction.input.assetId;
-      } else {
-        const assetId = await getAssetFromGasPriceTable(request, indexedDb);
-        if (assetId) {
-          return assetId;
-        }
+      const inputAssetId = withdrawAuction.auction.input.assetId;
+      const assetId = await getAssetFromGasPriceTable(request, indexedDb, inputAssetId);
+      if (assetId) {
+        return assetId;
       }
     }
   }
@@ -108,7 +96,7 @@ export const extractAltFee = async (
     }
   }
 
-  throw new Error('Could not extract alternative fee assetId from TransactionPlannerRequest');
+  throw new Error('Not able to pay fee for transaction');
 };
 
 // Attempts to find an alternative fee token for a transaction by checking the user's balance of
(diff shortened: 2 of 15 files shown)


apps/minifront/vitest.config.ts
```diff
@@ -5,5 +5,7 @@ export default defineConfig({
     snapshotSerializers: [],
     environment: 'jsdom',
     setupFiles: ['./tests-setup.ts'],
+    silent: false,
+    reporters: 'default',
```
grod220, line 9: Both of these are default values, can remove

packages/services/src/view-service/fees.ts
```diff
@@ -4,7 +4,22 @@ import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
 import { IndexedDbInterface } from '@penumbra-zone/types/indexed-db';
 
 // Attempts to extract a fee token, with priority in descending order, from the assets used
-// in the actions of the planner request.
+// in the actions of the transaction planner request (TPR).
+//
+// (1) For outputs, spends, swaps, scheduling auctions, and ICS withdrawals, the naive approach
+// of "using the first alternative token for which the user has a balance" is semantically
+// equivalent to "extract the input asset from the transaction request".
```
grod220, line 11: Are these actually the equivalent? They sound different. One sounds like it takes from balances and the other takes from the input, lessening the effect of say a withdraw.

packages/services/src/view-service/fees.ts
```diff
@@ -4,7 +4,22 @@ import { assetIdFromBaseDenom } from '@penumbra-zone/wasm/asset';
...
+//
+// (2) Swap claims are a special case where the claim must use the swap's fee token.
+//
+// (3) Ending and withdrawing auctions are another special case that requires a database query to
+// retrieve the relevant auction and the corresponding asset ID used in that auction. For example,
+// if a Dutch auction is scheduled with a fee paid using GM, ending and withdrawing the auction will
+// also attempt to use GM. However, if the user spends their entire GM balance after scheduling the
+// auction, they won't be able to end or withdraw it because it expects GM to pay for fees. In this
+// scenario, the fee system will use the first available alternative token balance from the
+// GAS_PRICES IndexedDB table. This allows the user to end and withdraw auctions using another asset,
```
grod220, line 21: Why in this case are the alt fees limited to what's in the GAS_PRICES table and in the other cases, it's not?

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +85,46 @@ export const extractAltFee = async (
...
+      if (checkAssetBalance) {
+        return withdrawAuction.auction.input.assetId;
+      } else {
+        const assetId = await retrieveAltAsset(request, indexedDb);
+        if (assetId) {
+          return assetId;
+        }
+      }
+    }
+  }
```
grod220, line 101: This looks like the equivalent to above. Maybe abstract to a helper function instead?

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +85,46 @@ export const extractAltFee = async (
...
+
+  if (request.undelegations.length > 0) {
+    const assetId = await retrieveAltAsset(request, indexedDb);
+    if (assetId) {
+      return assetId;
```
grod220, line 106: What's the deal with this case?

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +85,46 @@ export const extractAltFee = async (
...
     }
   }
 
   throw new Error('Could not extract alternative fee assetId from TransactionPlannerRequest');
 };
+
+// Attempts to find an alternative fee token for a transaction by checking the user's balance of
+// various gas price assets. It queries the IndexedDb for alternative gas prices and iterates
+// through them, returning the asset ID of the first asset for which the user has a balance.
+export const retrieveAltAsset = async (
```
grod220, line 116: The name feels a bit too generic. How about `getAssetFromGasPriceTable`?

packages/storage/src/indexed-db/index.ts
```diff
@@ -885,4 +885,24 @@ export class IndexedDb implements IndexedDbInterface {
       );
     });
   }
+
+  async getAssetTokenMetadata(
```
grod220, line 889: Don't think this name is accurate. This should be something like: `hasTokenBalance`.

packages/storage/src/indexed-db/index.ts
```diff
@@ -885,4 +885,24 @@ export class IndexedDb implements IndexedDbInterface {
...
+    addressIndex: AddressIndex | undefined,
```
grod220, line 890: If it's undefined, this function will always fail right? Is this intended?

apps/minifront/vitest.config.ts
TalDerei: removed

packages/services/src/view-service/fees.ts
TalDerei: realize that comment needs some clarity:

"For outputs, spends, swaps, scheduling auctions, and ICS withdrawals, the naive approach of "using the first alternative token for which the user has a balance" is similar to "extracting the input asset from the transaction request". While both methods aim to retrieve an alternative fee token, they may result in selecting different tokens."

Requesting an alternative token can be queried from the incoming request or the balance. The former provides more consistency, taking the input asset as the alternative fee (eg. transfer of GM will pay for fees in GM). The latter may result in some intuitive fee selection (eg. transfer of GM might pay for fees in GN).

packages/services/src/view-service/fees.ts
TalDerei: Consider case **(1)**: Performing outputs, spends, swaps, auction scheduling, or ICS withdrawals. These actions cannot be executed without an available balance of the input asset being traded. In that context, you can use the input asset specified in the transaction request because there is a _guarantee_ that the user has a balance for that asset ID. Therefore, the information encoded in the request is sufficient.

Consider case **(3)**: Performing auction ends, auction withdrawals, or undelegations. These are _secondary_ actions, as they require a prior action to occur, such as scheduling for auctions or delegating for undelegations. If the original assets used in those initial transactions have been spent or transferred out, retrieving the asset ID from the `GAS_PRICES` table for which the user has an available balance acts as a fallback for covering the fee. The downside is that you might have to use a different asset ID (e.g., if you auction GM --> UM, you might pay fees in GN when ending or withdrawing the auction because you've run out of GM).

packages/storage/src/indexed-db/index.ts
TalDerei: nice catch, changed

packages/storage/src/indexed-db/index.ts
TalDerei: actually AddressIndex should never be undefined, since it will fail the transaction source [check](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/transaction-planner/index.ts#L70)?

packages/services/src/view-service/fees.ts
Valentine1898: > Consider case (1): Performing outputs, spends, swaps, auction scheduling, or ICS withdrawals. These actions cannot be executed without an available balance of the input asset being traded.

This is true, but we may not have GasPrices for the input asset, in which case we would also probably want to use a fallback option

packages/services/src/view-service/fees.ts
TalDerei: https://github.com/penumbra-zone/web/issues/1569

packages/services/src/view-service/fees.ts
TalDerei: Interesting, why would that be the case?

packages/services/src/view-service/fees.ts
TalDerei: removed `stakingTokenAssetId` and made `hasTokenBalance` a helper function.

packages/services/src/view-service/fees.ts
TalDerei: @Valentine1898 specifically [getAltGasPrices](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L421) should be modified to only retrieve assets for which `GasPrices` are non-zero.

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +86,46 @@ export const extractAltFee = async (
...
+  }
+
+  if (request.undelegations.length > 0) {
+    const assetId = await getAssetFromGasPriceTable(request, indexedDb);
+    if (assetId) {
+      return assetId;
     }
   }
 
   throw new Error('Could not extract alternative fee assetId from TransactionPlannerRequest');
```
grod220, line 111: Can we give alternative error, something user-readable like:
```
Not able to pay fee for transaction
```

grod220 (commented):
can we get changesets for this too plz

packages/storage/src/indexed-db/index.ts
TalDerei: in that case, can we remove `undefined` from the type signature and use the non-null assertion operator on source?

packages/services/src/view-service/fees.ts
TalDerei: changed

packages/services/src/view-service/fees.ts
TalDerei: simplified entire description:
```
// Attempts to extract a fee token, with priority in descending order, from the assets used
// in the actions of the transaction planner request (TPR). If no fee token is found from the
// specified assets, it falls back to checking the gas prices table for an asset with a positive balance.
```

packages/services/src/view-service/fees.ts
TalDerei: @turbocrime

packages/services/src/view-service/fees.ts
TalDerei: for https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L421-L426, we need to do something like this:

```
  async getAltGasPrices(): Promise<GasPrices[]> {
    const allGasPrices = await this.db.getAll('GAS_PRICES');
    return allGasPrices
      .map(gp => GasPrices.fromJson(gp))
      .filter(gp => 
        !gp.assetId?.equals(this.stakingTokenAssetId)
        // (gp.blockSpacePrice > 0n || gp.compactBlockSpacePrice > 0n || gp.verificationPrice > 0n || gp.executionPrice > 0n)
      ); 
  }
  ```

the issue is once we make this change and bump the idb version, nothing will work for us locally because all our alternative fee tokens has zero gas prices.

File: packages/services/src/view-service/fees.ts [hunk 10]
```diff
@@ -60,9 +81,52 @@ export const extractAltFee = async (
...
+  // If a specific asset ID is provided, check its balance is positive and GasPrices for that asset are set.
+  if (specificAssetId) {
+    const balance = await indexedDb.hasTokenBalance(request.source!, specificAssetId);
+    const filteredGasPrices = altGasPrices.filter(gp => gp.assetId?.equals(specificAssetId));
+    if (balance && filteredGasPrices.length > 0) {
+      return specificAssetId;
+    }
+  }
+
+  // If no specific asset ID is provided or if the specific asset ID has no balance, check assets in GasPrices table.
```
TalDerei, line 121: this is a fallback

TalDerei:
still need to add changeset, bump idb version, and reach consensus on https://github.com/penumbra-zone/web/pull/1577#discussion_r1696602433.

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,52 @@ export const extractAltFee = async (
...
+    if (gasPrice.assetId) {
+      const balance = await indexedDb.hasTokenBalance(request.source!, gasPrice.assetId);
+      if (balance) {
+        return gasPrice.assetId;
+      }
     }
   }
 
-  throw new Error('Could not extract alternative fee assetId from TransactionPlannerRequest');
+  return undefined;
```
TalDerei, line 131: this means user has no alternative assets to pay for fees

packages/types/src/indexed-db.ts
```diff
@@ -150,7 +150,7 @@ export interface IndexedDbInterface {
     auctionId: AuctionId,
   ): Promise<{ input: Value; output: Value } | undefined>;
 
-  hasStakingAssetBalance(addressIndex: AddressIndex | undefined): Promise<boolean>;
+  hasTokenBalance(addressIndex: AddressIndex, assetId: AssetId): Promise<boolean>;
```
TalDerei, line 153: helper function for checking if user has balance for any asset in any account

packages/services/src/view-service/fees.ts
TalDerei: if we don't check `(gp.blockSpacePrice > 0n || gp.compactBlockSpacePrice > 0n || gp.verificationPrice > 0n || gp.executionPrice > 0n)`, alternative fees can be paid with assets with zero gas prices.

TalDerei: I moved the filtering out of the `getAltGasPrices` db call into `getAssetFromGasPriceTable`, so we don't have to perform a db change if we want to modify that function.

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,65 @@ export const extractAltFee = async (
...
+): Promise<AssetId | undefined> => {
+  // Fetch alternative gas prices, excluding those with a zero price.
+  const altGasPrices = await indexedDb.getAltGasPrices();
+  const filteredAltGasPrices = altGasPrices.filter(
+    gp =>
+      gp.blockSpacePrice > 0n ||
+      gp.compactBlockSpacePrice > 0n ||
+      gp.verificationPrice > 0n ||
+      gp.executionPrice > 0n,
+  );
```
TalDerei, line 118: filtering for only positive gas prices

TalDerei:
still need to add changeset, bump idb version, and reach consensus on https://github.com/penumbra-zone/web/pull/1577#discussion_r1696602433.

TalDerei:
high-level flow:

**1.** Fetch all alternative gas prices from the `GasPrices` table.

**2.** Extract assetId from the transaction request and check that:
- The balance for the asset is positive.
- The asset is a valid fee token.

**3.** If the above checks fail, fall back to using another asset from the `GasPrices` table.

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,56 @@ export const extractAltFee = async (
...
+    }
+  }
+
+  throw new Error('Not able to pay fee for transaction');
+};
+
+// Attempts to find an alternative fee token for a transaction by checking the user's balance of
+// various gas price assets. It queries the IndexedDb for alternative gas prices and iterates
+// through them, returning the asset ID of the first asset for which the user has a balance.
+export const getAssetFromGasPriceTable = async (
```
TalDerei, line 105: extracted all the complicated logic into a unified `getAssetFromGasPriceTable` function

packages/services/vitest.config.ts
```diff
@@ -5,5 +5,7 @@ export default defineConfig({
   plugins: [wasm()],
   test: {
     setupFiles: ['./tests-setup.js'],
+    silent: false,
+    reporters: 'default',
```
grod220, line 9: todo: remove

packages/storage/src/indexed-db/index.ts
```diff
@@ -422,7 +422,14 @@ export class IndexedDb implements IndexedDbInterface {
...
     return allGasPrices
       .map(gp => GasPrices.fromJson(gp))
-      .filter(gp => !gp.assetId?.equals(this.stakingTokenAssetId));
+      .filter(
+        gp =>
+          !gp.assetId?.equals(this.stakingTokenAssetId) &&
+          (gp.blockSpacePrice > 0n ||
+            gp.compactBlockSpacePrice > 0n ||
+            gp.verificationPrice > 0n ||
+            gp.executionPrice > 0n),
```
grod220, line 431: question: what does this mean? We should not accept genesis gas prices?

grod220 (approved):

packages/zquery/vitest.config.ts
```diff
@@ -4,5 +4,7 @@ export default defineConfig({
   test: {
     environment: 'jsdom',
     setupFiles: ['./tests-setup.ts'],
+    silent: false,
+    reporters: 'default',
```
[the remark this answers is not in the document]
TalDerei: removed

packages/storage/vitest.config.ts
```diff
@@ -12,5 +12,7 @@ export default defineConfig({
       enabled: true,
       headless: true,
     },
+    silent: false,
+    reporters: 'default',
```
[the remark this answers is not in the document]
TalDerei: removed

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,52 @@ export const extractAltFee = async (
...
+  request: TransactionPlannerRequest,
+  indexedDb: IndexedDbInterface,
+  specificAssetId?: AssetId,
+): Promise<AssetId | undefined> => {
+  const altGasPrices = await indexedDb.getAltGasPrices();
+
+  // If a specific asset ID is provided, check its balance is positive and GasPrices for that asset are set.
+  if (specificAssetId) {
+    const balance = await indexedDb.hasTokenBalance(request.source!, specificAssetId);
```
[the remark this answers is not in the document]
TalDerei: modified to use type gaurd

packages/services/src/view-service/fees.ts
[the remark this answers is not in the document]
TalDerei: yes that's right 👍

packages/services/src/view-service/fees.ts
Valentine1898: That's nice, I think we can make the choice of alternative fee assets even smarter if we filter out small balances when it's obvious that it won't be enough to pay the base tx fee (without considering the cost of actions).

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,56 @@ export const extractAltFee = async (
...
+  }
+
+  // If no specific asset ID is provided or if the specific asset ID has no balance, check assets in the
+  // GasPrices table as a fallback case.
+  for (const gasPrice of altGasPrices) {
+    if (gasPrice.assetId) {
+      const balance = await indexedDb.hasTokenBalance(request.source!, gasPrice.assetId);
+      if (balance) {
+        return gasPrice.assetId;
+      }
```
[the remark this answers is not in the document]
TalDerei: good call, will create a separate issue for this

packages/services/vitest.config.ts
TalDerei: removed

packages/storage/src/indexed-db/index.ts
TalDerei: I was concerned about a scenario where a governance proposal for `feeParams` could pass that adds an asset with zero gas prices. In that context, wouldn't users be able to transact with zero gas fees for that asset?

not sure what you mean by rejecting genesis gas prices?

Valentine1898: > I was concerned about a scenario where a governance proposal for feeParams could pass that adds an asset with zero gas prices. In that context, wouldn't users be able to transact with zero gas fees for that asset?

Even if that happens, we probably shouldn't override the community's election

TalDerei: sure good call, removed

packages/services/src/view-service/fees.ts
TalDerei: good point, thought about that early on, but wanted to limit the scope

packages/services/src/view-service/fees.ts
```diff
@@ -60,9 +81,60 @@ export const extractAltFee = async (
...
+  }
+
+  // If no specific asset ID is provided or if the specific asset ID has no balance, check assets in the
+  // GasPrices table as a fallback case.
+  for (const gasPrice of altGasPrices) {
+    if (gasPrice.assetId) {
+      const balance = await indexedDb.hasTokenBalance(request.source, gasPrice.assetId);
+      if (balance) {
+        return gasPrice.assetId;
+      }
```
grod220, line 135: The concurrent pattern would look something like this:

```ts
// in func
 try {
    return await Promise.any(
      altGasPrices.map(g => checkBalance(indexedDb, request.source, g.assetId)),
    );
  } catch (error) {
    return undefined;
  }
};

// outside of func
const checkBalance = async (
  indexedDb: IndexedDbInterface,
  source?: AddressIndex,
  gasPriceAsset?: AssetId,
) => {
  if (!source) {
    throw new Error('No source passed');
  }

  if (gasPriceAsset) {
    const balance = await indexedDb.hasTokenBalance(source, gasPriceAsset);
    if (balance) {
      return gasPriceAsset;
    }
  }

  throw new Error('No balance for asset');
};

```

packages/services/src/view-service/fees.ts
TalDerei: references https://github.com/penumbra-zone/web/issues/1595

Issue #1571, opened 2024-07-27
All delegator vote views returned by transactions RPC are opaque


plaidfinch:
When asking Prax for the user's own transactions, all delegator vote views returned are opaque rather than visible. This means that it is impossible to tell which account was used for voting, and how the vote was cast, inhibiting frontends from providing meaningful feedback to users about their already-used voting power (i.e. which proposals they have already voted on) on a per-account basis, as well as how they voted previously.

Proposed fix: the delegator vote action views returned should instead be visible in all cases, because there is no reason the user should not know the contents of their own delegator vote actions.

PR #1579, opened 2024-07-29
Perspective -> Delegator view 

grod220:
Fixes: https://github.com/penumbra-zone/web/issues/1571


apps/minifront/src/fetchers/transactions.ts
grod220, line 1: unused fetcher

Valentine1898 (approved):

PR #1580, opened 2024-07-29
fix(transport-chrome): allow http for localhost

VanishMax:
Relates to https://github.com/prax-wallet/web/pull/129#issuecomment-2253255191

### DIFF
--- a/packages/transport-chrome/src/session-manager.ts
+++ b/packages/transport-chrome/src/session-manager.ts
@@ -96,16 +96,14 @@ export class CRSessionManager {
       return;
     }
 
-    // origin restrictions
     const fromThisExtension = sender.id === chrome.runtime.id;
-    // frameId == 0 for top-level documents
-    const fromPageHttps =
-      !sender.frameId &&
-      sender.tab?.id &&
-      (sender.origin.startsWith('https://') ||
-        sender.origin.startsWith('http://localhost:') ||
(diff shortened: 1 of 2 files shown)


packages/transport-chrome/src/session-manager.ts
```diff
@@ -99,7 +99,12 @@ export class CRSessionManager {
     // origin restrictions
     const fromThisExtension = sender.id === chrome.runtime.id;
     // frameId == 0 for top-level documents
-    const fromPageHttps = !sender.frameId && sender.tab?.id && sender.origin.startsWith('https://');
+    const fromPageHttps =
+      !sender.frameId &&
+      sender.tab?.id &&
+      (sender.origin.startsWith('https://') ||
+        sender.origin.startsWith('http://localhost:') ||
+        sender.origin === 'http://localhost');
```
grod220, line 107: Think local host should be broken out. I find this easier to read:

```ts
const fromThisExtension = sender.id === chrome.runtime.id;
const fromPageHttps = !sender.frameId && sender.tab?.id && sender.origin.startsWith('https://');
const isLocalhost =
      sender.origin.startsWith('http://localhost:') || sender.origin === 'http://localhost';
const validOrigin = isLocalhost || fromPageHttps || fromThisExtension;
if (!validOrigin) {
  return;
}
```

grod220 (approved):

PR #1582, opened 2024-07-29
feat(protobuf): #1483: bundle buf into protobuf

VanishMax:
Closes #1483 

Closes #1503 

Re-exports all message types from `@penumbra-zone/protobuf/types`.

It doesn't, for now, bundle those types into the `protobuf` package. Users still need to install peer dependencies from `@buf`.

Issue #1483, opened 2024-07-16
Bundle @buf interfaces into Protobuf


VanishMax:
It is hard to track the `@buf` package versions. If not the IDE import substitutions, I would never learn the path from which I should import common things like `Metadata`, `Address`, `BalancesResponse`, etc.

It would be way easier to have only one entry point for such interfaces and `@penumbra-zone/protobuf` package is the great fit for it. We could improve two things at once:
1. Bundle the specific version of the `@buf/...` into a specific version of `@penumbra-zone/protobuf`. There won't be a need to separately install `@buf` packages by consumers and no more need to configure the `@buf` registry
2. Make the imports straightforward and only from one place

I don't have the exact solution on how to structure the `protobuf` package. One option is to put the interfaces next to the services it is used within. For example, `Metadata` is used in the `ViewService`, so it could be placed in `@penumbra-zone/protobuf/view-service`

PR #1583, opened 2024-07-29
feat: #1483: apply new protobuf imports

VanishMax:
Relates to #1582 

This is a large PR made separately from #1582 but is dependent from it. Updates all imports from `@buf/...` packages to `@penumbra-zone/protobuf/types`


VanishMax:
closed in favor of #1624

Issue #1377, opened 2024-07-01
Sync modal 


TalDerei:
Sync status is critical to the user's experience with Penumbra & minifront. When the sync is not current, their balances could be incorrect & there could be problems issuing a transaction. Minifront should implement a full-screen modal to communicate sync status in a more in-your-face way. Details:

- If sync is up-to-date, show the UI as expected
- If sync is not up-to-date, have a full-screen modal (blur background?), that gives messaging for current status.
  - Should be dismissible to allow user to continue if they choose
  - Should have messaging for what syncing is doing
  - Should give a % status, "50.2% synced" w/ blockheight
  - Should re-use airdrop network+computer graphics to communicate data is flowing. Ask @smmhrdmn (Sam) for this asset.

smmhrdmn:
Here is the [file](https://www.figma.com/design/H5TPKVz1Um7TzVefwYent4/Mini-Front----Miscellaneous?node-id=201-1914&t=Lhrgc4tDtCfKRAnb-1), [prototype](https://www.figma.com/proto/H5TPKVz1Um7TzVefwYent4/Mini-Front----Miscellaneous?page-id=201%3A1914&node-id=4002-1035&viewport=-18123%2C-27973%2C0.64&t=st1RXeLcnZO5uuEQ-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=4002%3A1035&show-proto-sidebar=1), and [animation](https://app.lottiefiles.com/share/2cef754c-3fdc-421a-b8eb-406eaecff5d0)! Lmk if you have any comments or suggestions for copy.

PR #1584, opened 2024-07-29
Create `<SyncingModal />` component, along with a `<Dialog />` Penumbra UI component

jessepinho:
🚨 23 of the changed files are just renaming the Penumbra UI library's `<ThemeProvider />` component to the more aptly named `<PenumbraUIProvider />`.

https://github.com/user-attachments/assets/4b503949-be7f-4eec-a1ed-88ad805167a2

## In this PR
- Create a `<Dialog />` component in the Penumbra UI library
- Create a `<SyncingDialog />` component in minifront — **this is the first component in Minifront that uses the new Penumbra UI library!**
- Rename `<ThemeProvider />` to `<PenumbraUIProvider />`
- Add a few new values to theme (`blur` and `color.base`)
- Add `<PenumbraUIProvider />` to the root of minifront, so that minifront can successfully use Penumbra UI components
- Delete the `styled-components.d.ts` declaration merge we were previously using, and move it into the `theme.ts` file under `<PenumbraUIProvider />`. This was necessary because minifront didn't recognize the declaration merging we were doing, since minifront wasn't importing `packages/ui/src/styled-components.d.ts` (and we don't want to force consumers to import an ambient declaration file). Now that it's moved into `theme.ts`, the types automatically get merged

Closes https://github.com/penumbra-zone/web/issues/1377


packages/ui/src/Dialog/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1584-jessepinho-card-vf8z1vua.web.app/?path=/docs/ui-library-dialog--docs

packages/ui/src/Text/index.tsx
```diff
@@ -178,6 +203,19 @@ export type TextProps = TextType & {
...
+ */
+const omit = <ObjectType extends Record<string, unknown>>(
+  object: ObjectType,
+  key: keyof ObjectType,
+) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- we're discarding the unused key
+  const { [key]: _, ...rest } = object;
+  return rest;
+};
+
```
jessepinho, line 218: Created this to fix a previous issue where React was passing `h1`, `technical`, `large`, etc. as attributes to the DOM.

apps/minifront/src/components/layout.tsx
```diff
@@ -3,21 +3,24 @@ import { HeadTag } from './metadata/head-tag';
 import { Header } from './header/header';
 import { Toaster } from '@repo/ui/components/ui/toaster';
 import { Footer } from './footer/footer';
-import { MotionConfig } from 'framer-motion';
 import '@repo/ui/styles/globals.css';
+import { SyncingModal } from './syncing-modal';
 
 export const Layout = () => {
   return (
-    <MotionConfig transition={{ duration: 0.1 }}>
```
jessepinho, line 11: This is already included in `PenumbraUIProvider`, which is now at the top of the component tree in `main.tsx`

packages/ui/src/ButtonGroup/index.tsx
```diff
@@ -48,6 +50,12 @@ export interface ButtonGroupProps<IconOnly extends boolean> {
    * Will be used for all buttons in the group.
    */
   iconOnly?: IconOnly;
+  /**
+   * In some cases, you will want to disable responsiveness and enforce that
+   * buttons are always rendered in a column, rather than a row (such as in a
+   * narrow layout width, such as a dialog). In those cases, set this to `true`.
+   */
+  column?: boolean;
```
jessepinho, line 58: Added this for the purpose of the `Dialog` component, which will always display buttons in a column.

jessepinho:
sorry all, I accidentally pushed up a commit to this branch that was supposed to be on a diff branch. Just fixed it

VanishMax (approved):
It's nice to see some new designs being applied. And a little sad that' after users will see the beautiful 'syncing' screen, it will close and show the old designs. 

Also, `Dialog` with Radix is extremely neat. Support for such UI library is going to be easy

PR #1585, opened 2024-07-29
Customize symbol on denom query

grod220:
Allows for receipt tokens not seeing previously to be rendered correctly


packages/storage/src/indexed-db/config.ts
```diff
@@ -2,4 +2,4 @@
  * The version number for the IndexedDB schema. This version number is used to manage
  * database upgrades and ensure that the correct schema version is applied.
  */
-export const IDB_VERSION = 45;
+export const IDB_VERSION = 46;
```
grod220, line 5: Bumping so that user's saved receipt tokens are wiped

Valentine1898 (approved):

PR #1586, opened 2024-07-29
Use the new `<Density />` component in place of the `size` prop throughout the new UI components

jessepinho:
After my previous PR for [introducing the `<Density />` component](https://github.com/penumbra-zone/web/pull/1545) was merged, we now have a `<Density />` component that sets a context variable for components to use to determine whether to render with `compact` or `sparse` density.

Previously, a bunch of the new UI library components were using a `size` prop to control density. In that previous PR, I mentioned that I would create a follow-up PR to remove their `size` props and instead use the new `useDensity()` hook, which uses the context value set by the `<Density />` component.

This PR is that follow-up PR. It makes that change for the `<Button />`, `<ButtonGroup />`, `<Pill />`, and `<ValueViewComponent />` components.

It also introduces a `density` tag for Storybook stories, which adds a density control to stories of components that have density variants.

### DIFF
--- a/packages/ui/src/ButtonGroup/index.tsx
+++ b/packages/ui/src/ButtonGroup/index.tsx
@@ -55,7 +55,7 @@ const isIconOnly = (props: ButtonGroupProps<boolean>): props is ButtonGroupProps
 
 /**
  * Use a `<ButtonGroup />` to render multiple buttons in a group with the same
- * `actionType` and `size`.
+ * `actionType`.
  *
  * When rendering multiple Penumbra UI buttons together, always use a
  * `<ButtonGroup />` rather than individual `<Button />`s. This ensures that
--- a/packages/ui/src/ButtonGroup/index.tsx
+++ b/packages/ui/src/ButtonGroup/index.tsx
@@ -1,25 +1,27 @@
 import { LucideIcon } from 'lucide-react';
 import { MouseEventHandler } from 'react';
-import { ActionType, Size } from '../utils/button';
+import { ActionType } from '../utils/button';
 import { Button } from '../Button';
 import styled from 'styled-components';
 import { media } from '../utils/media';
 import { ButtonPriorityContext } from '../utils/ButtonPriorityContext';
+import { Density } from '../types/Density';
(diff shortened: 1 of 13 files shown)


packages/ui/src/ButtonGroup/index.tsx
```diff
@@ -64,43 +64,37 @@ const isIconOnly = (props: ButtonGroupProps<boolean>): props is ButtonGroupProps
...
  * priority.)
  */
-export const ButtonGroup = ({
-  actionType = 'default',
-  size = 'sparse',
-  ...props
-}: ButtonGroupProps<boolean>) => (
-  <Root $size={size}>
-    {/* Annoying TypeScript workaround — we need to explicitly delineate the
+export const ButtonGroup = ({ actionType = 'default', ...props }: ButtonGroupProps<boolean>) => {
```
jessepinho, line 67: You might want to review this with "Hide whitespace" enabled — most of the changes in this markup are indentation changes, due to the fact that the component now is wrapped in curly braces due to the `useDensity` hook

packages/ui/src/Table/index.stories.tsx
```diff
@@ -34,46 +30,7 @@ const DATA = [
   [1024, 'Delegate', '9d80ffa9113f7eed74ddeff8eddfda6a89106c6cdf336565f9cbaf90810396bf'],
 ];
 
-const Column = styled.div`
```
jessepinho, line 37: This wrapper code is now moved to `.storybook/preview.jsx`

packages/ui/src/ValueViewComponent/AssetIcon/index.tsx
```diff
@@ -15,7 +15,6 @@ const IconImg = styled.img`
 
 export interface AssetIcon {
   metadata?: Metadata;
-  size?: 'sparse' | 'dense';
```
jessepinho, line 18: This actually wasn't even being used in this component — it was a leftover that I forgot to remove.

VanishMax (approved):

PR #1596, opened 2024-07-30
fix send form memory bug

### LINKED ISSUE #1592: Error when trying to spend max and alternative fees
**Describe the bug**
You will get an error if you try to spend the maximum balance using an alternative fee asset
At the same time using native fee asset you can spend all your native tokens

**Expected behavior**
It is expected that the user can max spend when using alternative fee asset 

**Screenshots**
image

Additional context
The error “ConnectError: [unknown] Failed to retrieve or ran out of notes for asset” is not meaningful to users

### LINKED ISSUE #1594: The transaction plan does not match the user's intention
**Describe the bug**
Send form has a memory effect. If you try to send max amount first, but then enter any other amount minifront will still try to spend the max value 

**To Reproduce**

1. Go to Send
2. Try to send the maximum amount 
3. Cancel transaction
4. Enter any amount 
5. Submit the transaction, and notice that the transaction plan is still trying to spend the maximum 

**Screenshots**

https://github.com/user-attachments/assets/f4991087-8cf4-4752-9185-9ebde3441d1d

Valentine1898:
closes #1594 
closes #1592

### DIFF
--- a/apps/minifront/src/state/send/index.ts
+++ b/apps/minifront/src/state/send/index.ts
@@ -15,6 +15,7 @@ import {
   FeeTier_Tier,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1/fee_pb';
 import {
+  getAmount,
   getAssetIdFromValueView,
   getDisplayDenomExponentFromValueView,
 } from '@penumbra-zone/getters/value-view';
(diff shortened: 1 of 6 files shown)


apps/minifront/src/state/send/index.ts
```diff
@@ -155,6 +143,9 @@ const assembleRequest = ({
       assetId: getAssetIdFromValueView(selection?.balanceView),
     },
   };
+  const isSendingMax = getAmount(selection?.balanceView).equals(
+    spendOrOutput.value?.amount as Amount,
+  );
```
Valentine1898, line 148: Determining `isSendingMax` at the time of `TrasactionPlannerRequest` creation seems like a simpler solution than managing state

TalDerei (approved):

PR #1600, opened 2024-07-30
NaN bug with simluate selector

grod220:
After a swap claim is successful, if you have the swap estimate component active, it will throw this error:

Screenshot 2024-07-30 at 11 25 13 PM

The reason is because the swap amount in the slice is reset on success and leaves the valueview component with bad inputs (an empty string versus something that can be parsed into a number).


Valentine1898 (approved):

turbocrime:
cc @jessepinho

PR #1603, opened 2024-07-31
Fix idb account contains asset, fix wasm gas prices

### LINKED ISSUE #1601: hasTokenBalance does not correctly detect token balances
hasTokenBalance contains incorrect logic

https://github.com/penumbra-zone/web/blob/d18f2d0cfbe8977dece1db5babe0b5616dc6eb9c/packages/storage/src/indexed-db/index.ts#L872-L887

### does not query notes table by valid index

https://github.com/penumbra-zone/web/blob/d18f2d0cfbe8977dece1db5babe0b5616dc6eb9c/packages/storage/src/indexed-db/index.ts#L876

an incoming asset id may not possess a binary 'inner' field, but may contain a `altBech32m` or `altBaseDenom` field. idb is indexed by the binary inner, but `hasTokenBalance` does not ensure the presence of this field before using an asset id to query.

### performs invalid comparison of address indicies

https://github.com/penumbra-zone/web/blob/d18f2d0cfbe8977dece1db5babe0b5616dc6eb9c/packages/storage/src/indexed-db/index.ts#L884

an address index contains an integer field that may be undefined, and a randomizer that may not be relevant for certain comparison purposes. `hasTokenBalance` fails to compare zero/undefined equally, and will reject mismatched randomizers

## effects

this results in hasTokenBalance failing to correctly determine the presence of a token in an account when it queries idb, depending on the input.

on conditions that are present in the view service now: for some transactions, the view service planner will incorrectly determine that the planner-specified (or native) fee is not present. so,

### transactions may break

the planner will enter alt fee selection logic which uses this same broken method to identify presence of tokens. the planner may fail to discover an alternate fee token due to the same issue, and throw the rpc call.

### the planner will not always respect the transaction plan

the planner will enter alt fee selection logic, and may be able to discover an alternate fee that disrespects the user's transaction plan.

### wasm building may fail

or, the planner may provide uncomparable asset ids or asset ids missing a binary inner to the wasm planner, which seems to have some similar or related problems querying idb.

## suggested correction

require input parameters of `hasTokenBalance` to contain a more restricted message type

### `addressIndex` parameter should be `{ account: number, randomizer: never }`

caller will be responsible for constructing correct input

i believe randomizer is not relevant for this specific comparison. please correct me if this is wrong.

### `assetId` parameter should be ``passet1${string}`` or `{ inner: Uint8Array }` and method should assert length

caller will be responsible for constructing correct input

### wasm planner cannot always identify correct gas fee

- wasm planner cannot identify gas prices if gas prices are not present in idb. this is incorrect, wasm planner should default to zero gas prices.
- wasm planner may be incorrectly expecting a binary-inner field of assetid input to be present.
- wasm planner may be failing to confirm it is querying idb with valid assetId.inner

### LINKED ISSUE #1602: Wasm planner throws on undefined gas fees
**Describe the bug**
Wasm planner throws undefined gas fees as an error

https://github.com/penumbra-zone/web/blob/d18f2d0cfbe8977dece1db5babe0b5616dc6eb9c/packages/wasm/crate/src/planner.rs#L204-L208

**To Reproduce**

1. Use a chain with no gas fees
2. Plan a transaction with wasm that executes plan_transaction_inner
3. Wasm throws because gas prices are not available

**Expected behavior**
Undefined means zero. Wasm should use the default gas fee (zero) and successfully plan the transaction.

turbocrime:
fixes #1601
fixes #1602 

this PR corrects idb method used to identify tokens present in an account by correctly comparing address index and asset id. the method will assert the validity of incoming arguments.

this PR corrects wasm planner to use default-zero gas fees if none are present in idb query.

some minifront ui may still be submitting plans with asset id that are not valid.

it should be decided if:

1. incoming rpc with asset id that do not contain a proper inner binary field should be accepted
2. incoming rpc with asset id that do not contain a proper inner binary field should be rejected

tested on deimos (zero chain gas fees) successfully with ibc-osmo gda to um and ibc-osmo out.

tested on phobos (chain gas fees) successfully with ibc-osmo gda to um and ibc-osmo out.

## ci is failing because tests are using incorrect data.

### DIFF
--- a/packages/services/src/view-service/fees.test.ts
+++ b/packages/services/src/view-service/fees.test.ts
@@ -34,7 +34,7 @@ describe('extractAltFee', () => {
       getAuction: vi.fn(),
       saveGasPrices: vi.fn(),
       getAltGasPrices: vi.fn(),
-      hasTokenBalance: vi.fn(),
+      accountHasSpendableAsset: vi.fn(),
     };
   });
 
@@ -66,7 +66,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
@@ -101,7 +101,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
@@ -155,7 +155,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(outputAssetId)).toBeTruthy();
@@ -191,7 +191,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
@@ -227,7 +227,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
@@ -313,7 +313,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
@@ -339,7 +339,7 @@ describe('extractAltFee', () => {
       seqNum: 0n,
     });
 
-    mockIndexedDb.hasTokenBalance?.mockResolvedValueOnce(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValueOnce(true);
 
     const request = new TransactionPlannerRequest({
       dutchAuctionEndActions: [
@@ -362,7 +362,7 @@ describe('extractAltFee', () => {
 
     mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
     mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.accountHasSpendableAsset?.mockResolvedValue(true);
 
     const result = await extractAltFee(request, mockIndexedDb as unknown as IndexedDbInterface);
     expect(result.equals(inputAssetId)).toBeTruthy();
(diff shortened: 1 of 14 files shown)


packages/services/src/view-service/fees.ts
```diff
@@ -117,9 +118,12 @@ export const getAssetFromGasPriceTable = async (
...
+    const balance = await indexedDb.accountHasSpendableAsset(request.source, assetId);
     // This check ensures that the alternative fee token is a valid fee token, for example, TestUSD is not.
-    const isInGasTable = altGasPrices.find(gp => gp.assetId?.equals(assetId));
+    const isInGasTable = altGasPrices.find(
+      // TODO: assert this more thoroughly
+      gp => bech32mAssetId(gp.assetId!) === bech32mAssetId(assetId),
+    );
     if (balance && isInGasTable) {
       return assetId;
     }
```
turbocrime, line 129: this part could be improved but i'm going to bed

packages/storage/src/indexed-db/index.ts
```diff
@@ -869,20 +869,26 @@ export class IndexedDb implements IndexedDbInterface {
...
-        !isZero(getAmountFromRecord(spendableNote)) &&
-        spendableNote.addressIndex?.equals(addressIndex)
-      );
+    const assetNotes = (
+      await this.db.getAllFromIndex('SPENDABLE_NOTES', 'assetId', uint8ArrayToBase64(inner))
+    ).map(note => SpendableNoteRecord.fromJson(note));
+
+    const spendableNotes = assetNotes.filter(({ note, addressIndex, heightSpent }) => {
+      const noteAddressIndex = toPlainMessage(addressIndex ?? new AddressIndex());
+      const accountMatch = noteAddressIndex.account === account;
```
turbocrime, line 885: is it correct to ignore account randomizer for this comparison?

packages/services/src/view-service/fees.test.ts
turbocrime, line 1: someone should fix these tests

TalDerei:
There are a few points to note about https://github.com/penumbra-zone/web/issues/1602.

For context, `GasPrices not Available` essentially means that "the fee we're trying to use for the transaction hasn't been defined on-chain and therefore isn't in the GasPrices table, making it unusable for the transaction". That's sounds reasonable to me. 

**1.** This situation isn't representative of mainnet conditions because the mainnet chain _does_ have gas prices.

**2.** If alternative fees in the GasPrices table are _undefined_, it seems they already default to 0. For example:

Screenshot 2024-07-31 at 1 44 32 AM

https://github.com/user-attachments/assets/4de9737d-db88-4a2e-9188-f11e931681bf

**3.** The consequence of the refactor is that it wouldn't even reach that point because it will fail early and be caught by minfront if the user doesn't have a balance for any of the alternative fees defined in the GasPrices table. In the example below, notice that the account doesn't have any `UM, GM, or GN`, which are the only assets defined in the GasPrices table. `Not able to pay fee for transaction` is the error we should expect to see. 

https://github.com/user-attachments/assets/166ae1d4-e6cc-48f0-805c-e069c94ee273

packages/services/src/view-service/fees.ts
```diff
@@ -117,9 +118,12 @@ export const getAssetFromGasPriceTable = async (
   // If a specific asset ID is provided, extracted from the transaction request, check its balance is
   // positive and GasPrices for that asset exist.
   if (assetId) {
-    const balance = await indexedDb.hasTokenBalance(request.source, assetId);
+    const balance = await indexedDb.accountHasSpendableAsset(request.source, assetId);
     // This check ensures that the alternative fee token is a valid fee token, for example, TestUSD is not.
-    const isInGasTable = altGasPrices.find(gp => gp.assetId?.equals(assetId));
```
TalDerei, line 122: this checks if the `assetId` passed in exists in the `GasPrices` table. since we're comparing assetId objects directly, why isn't the equality method sufficient here? can you explain when we should use Bech32m encoded string representations instead?

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -7,22 +7,29 @@ import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_
...
 
   const services = await ctx.values.get(servicesCtx)();
   const { indexedDb } = await services.getWalletServices();
 
-  // Query IndexedDB directly to check for the existence of staking token
-  const nativeToken = await indexedDb.hasTokenBalance(req.source!, indexedDb.stakingTokenAssetId);
+  const noNativeTokenAvailable = !(await indexedDb.accountHasSpendableAsset(
+    toPlainMessage(req.source ?? new AddressIndex()),
+    indexedDb.stakingTokenAssetId,
+  ));
```
TalDerei, line 22: actually shouldn't `AddressIndex` never be undefined, since it will fail the transaction source [check](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/transaction-planner/index.ts#L70). In that case, why can't we use the non-null assertion operator on source?

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -7,22 +7,29 @@ import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_
...
 
+  const planningSwapClaim = req.swapClaims.length > 0;
   // Check if we should use the native token or extract an alternate gas fee token.
   // Special cased for swap claims as gas fee needs to match the claimFee on the corresponding swap.
-  const needsAltFeeToken = !nativeToken || req.swapClaims.length > 0;
-  const gasFeeToken = needsAltFeeToken
-    ? await extractAltFee(req, indexedDb)
-    : indexedDb.stakingTokenAssetId;
+  console.log('alt fee reasons', noNativeTokenAvailable, planningSwapClaim);
```
TalDerei, line 27: remove log

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -7,22 +7,29 @@ import { TransactionPlannerRequest } from '@buf/penumbra-zone_penumbra.bufbuild_
...
+
+  const gasFeeToken =
+    noNativeTokenAvailable || planningSwapClaim
```
TalDerei, line 30: I actually prefer keeping `needsAltFeeToken`, so we can tack on more invariants later.

packages/wasm/crate/src/planner.rs
```diff
@@ -205,7 +205,7 @@ pub async fn plan_transaction_inner<Db: Database>(
     let gas_prices = storage
         .get_gas_prices_by_asset_id(&fee_asset_id)
         .await?
-        .ok_or_else(|| anyhow!("GasPrices not available"))?;
+        .unwrap_or_default();
```
TalDerei, line 208: see comment below addressing https://github.com/penumbra-zone/web/issues/1602

packages/storage/src/indexed-db/index.ts
TalDerei: I don't think the randomizer is relevant here? https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/planner.rs#L177-L180.

cc @Valentine1898

packages/services/src/view-service/fees.ts
TalDerei: this checks if the assetId passed in exists in the GasPrices table. since we're comparing assetId objects directly, why isn't the equality method sufficient here? can you explain when we should use Bech32m encoded string representations instead?

packages/services/src/view-service/fees.test.ts
TalDerei: https://github.com/penumbra-zone/web/pull/1603/commits/af28b4d49eb000e11d82939bc79bfc3cac7676c5

TalDerei:
> 1. incoming rpc with asset id that do not contain a proper inner binary field should be accepted

remind me why an asset id wouldn't have inner binary field? do Noble USDC, ATOM, and OSMO?

packages/storage/src/indexed-db/index.ts
Valentine1898: the randomizer should be ignored 
https://discord.com/channels/824484045370818580/1044457038886998137/1220144758970781707

packages/wasm/crate/src/planner.rs
Valentine1898: I think it's a mistake, we shouldn't use default gas prices
And use only those gas prices that have been published in compact blocks

Valentine1898 (changes requested):

packages/wasm/crate/src/planner.rs
TalDerei: I agree with this sentiment. For instance, this would allow paying for fees with non-supported gas tokens, for instance `TestUSD`.

grod220:
Taking this over to expedite bug fixes

TalDerei:
prioritizing https://github.com/penumbra-zone/web/pull/1605, this pr can be useful follow-up work.

turbocrime:
update: address-index comparison is corrected by #1605 but remaining issues persist

TalDerei:
closing this per the comments https://github.com/penumbra-zone/web/pull/1603#issuecomment-2259848470 and https://github.com/penumbra-zone/web/pull/1603#discussion_r1698050976. This changeset is outdated and introduces unnecessary changes to what appear to be non-issues.

I believe that any incoming rpc with an asset id that lacks a proper inner binary field should be rejected. However, if we decide otherwise, we can address this in a separate PR and reduce the scope of the changes.

PR #1606, opened 2024-07-31
`<Card />` component

jessepinho:
Creates a simple `<Card />` component, which will likely grow in complexity over time as we start using it in various screens.

[Demo here](https://penumbra-ui-preview--pr1606-jessepinho-card-2-rz1wxq2j.web.app/?path=/docs/ui-library-card--docs)


packages/ui/src/Card/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1606-jessepinho-card-2-rz1wxq2j.web.app/?path=/docs/ui-library-card--docs

packages/ui/src/Card/storiesBg.png
VanishMax, line 1: a wish: can this image be compressed a little? When I first entered the Card's storybook, it was empty for like 5 seconds until the image loaded

VanishMax (approved):

packages/ui/src/Card/storiesBg.png
jessepinho: Ah, yes — good call. It was 8MB! I've compressed it to a 100kb JPEG.

PR #1609, opened 2024-08-01
Bug fix: detect ibc UM

grod220:
Currently, minifront does not allow to ibc in non-native assets. That means, at the moment, this includes IBC'ing UM to penumbra from on a counter-party chain.

This PR detects addresses for UM on counterparty chains and allows users to transfer it in.


apps/minifront/src/components/ibc/ibc-in/hooks.ts
```diff
@@ -102,10 +107,63 @@ interface UseCosmosChainBalancesRes {
...
+  if (logoUri) {
+    return logoUri;
+  }
+
+  if (asset.images?.length) {
+    const first = asset.images[0]!;
+    return first.svg || first.svg;
+  }
+  return undefined;
+};
```
grod220, line 121: There are two places where icons can live. On osmosis, penumbra does not have `logo_URIs` so going to pull from both fields to see if an image is available.

VanishMax (approved):
Looks good

PR #1616, opened 2024-08-01
fix: test wasm crate in ci correctly

VanishMax:
fixes https://github.com/penumbra-zone/web/issues/1546

This moves the tests from the src directory of the wasm/crate package into its tests directory, so that wasm-pack correctly picks them up.


packages/wasm/crate/tests/test_metadata.rs
```diff
@@ -0,0 +1,164 @@
...
+    };
+    let customized_metadata = customize_symbol_inner(metadata.clone()).unwrap();
+
+    assert_eq!(
+        customized_metadata.symbol,
+        "auction@123(jqyupqnzznyfpq940mv0ac33pyx77s7af3kgdw4nstjmp3567dks8n5amh)"
+    );
+}
+
+#[wasm_bindgen_test]
```
grod220, line 130: todo: this one doesn't need the `#[wasm_bindgen_test]` attribute right? It can just be `#[test]`. Can you confirm with the other tests as well and if the containing functions don't use wasm bindgen, let's do the normal test attribute.

grod220 (approved):
I know lots of other rusty folks like to colocate their tests with the implementation code, but I personally prefer it separated into different files like this

packages/wasm/crate/tests/test_metadata.rs
vacekj: I'm all for colocation, but in this case it's the straightforward way to run the tests with wasm-pack.

Unfortunately without the `#[wasm_bindgen_test]` attribute, the test won't run, so I have to leave it there.

grod220: Oh wait a sec, think there is an issue with this resolution. We have two kinds of tests within the crate:
- Unit tests. Normal run-of-the-mill tests that run in a normal rust env via `cargo test`. Has the #[test] attribute.
- Headless browser wasm tests. These are run in a headless browser environment by wasm pack and necessary for functionality that specifically uses wasm_bindgen. Has the #[wasm_bindgen_test] attribute.

We should be able to run both of these tests. It will require two separate commands and it looks like our CI is only running the wasm-pack ones. It shouldn't be necessary to add #[wasm_bindgen_test] to every Rust test as this slows down local development.

vacekj: I see - I will revert this and make sure the CI runs both the unit and wasm-pack tests

Issue #1503, opened 2024-07-17
investigate re-exporting individual protobuf message types from `@penumbra-zone/protobuf`


turbocrime:
here's the current frequency of each individual message type import

data collected with

`rg --multiline "import \{([\n\s\w,]+)\} from ('@buf/.+');" --replace '$2$1'` and then a bunch of iterative manipulation, sorry

```
  83 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb.js' Metadata
  51 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb.js' ValueView
  40 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js' BalancesResponse
  30 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/asset/v1/asset_pb.js' AssetId
  26 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js' Address
  22 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/num/v1/num_pb.js' Amount
  21 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js' FullViewingKey
  20 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1/auction_pb.js' AuctionId
  17 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js' SpendableNoteRecord
  17 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/stake/v1/stake_pb.js' ValidatorInfo
  16 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js' AddressView
  15 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js' TransactionPlannerRequest
  15 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/view/v1/view_pb.js' SwapRecord
  15 '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js' AddressIndex
[...]

PR #1624, opened 2024-08-02
use local codegen for protobuf message types

### LINKED ISSUE #1483: Bundle @buf interfaces into Protobuf

turbocrime:
fixes #1483 #1503

replaces #1582 #1583

it's simplest to use `buf generate` to create the definitions


package.json
```diff
@@ -32,19 +32,21 @@
...
-    "@buf/cosmos_ibc.connectrpc_es": "1.4.0-20240703151202-d7191877c701.3",
-    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.10.0-20240703080008-312294d02bf9.1",
-    "@buf/penumbra-zone_penumbra.connectrpc_es": "1.4.0-20240703080008-312294d02bf9.3",
     "@buf/tendermint_tendermint.bufbuild_es": "1.10.0-20231117195010-33ed361a9051.1",
     "@bufbuild/protobuf": "^1.10.0",
     "@connectrpc/connect": "^1.4.0",
     "@connectrpc/connect-web": "^1.4.0"
   },
   "devDependencies": {
+    "@buf/penumbra-zone_penumbra.bufbuild_es": "1.10.0-20240703080008-312294d02bf9.1",
```
turbocrime, line 49: this one @buf dep is still necessary because the registry incorrectly fails to declare a dependency

packages/protobuf/package.json
```diff
@@ -11,41 +11,54 @@
     "build": "tsc --build --verbose",
     "clean": "rm -rfv dist *.tsbuildinfo package penumbra-zone-*.tgz",
     "dev:pack": "tsc-watch --onSuccess \"$npm_execpath pack\"",
+    "gen:cosmos-sdk": "buf generate buf.build/cosmos/cosmos-sdk:e7a85cef453e4b999ad9aff8714ae05f",
+    "gen:ibc": "buf generate buf.build/cosmos/ibc:7ab44ae956a0488ea04e04511efa5f70",
+    "gen:ics23": "buf generate buf.build/cosmos/ics23:55085f7c710a45f58fa09947208eb70b",
+    "gen:penumbra": "buf generate buf.build/penumbra-zone/penumbra:312294d02bf945ffa4c1b1dd7cd91328",
```
turbocrime, line 17: versions used in core

https://github.com/penumbra-zone/penumbra/blob/main/proto/buf.lock

packages/protobuf/src/gen/.npmignore
turbocrime, line 1: empty npmignore will allow this dir to be packaged

packages/protobuf/package.json
```diff
@@ -11,41 +11,52 @@
...
     "lint": "eslint src",
     "lint:fix": "eslint src --fix",
-    "lint:strict": "tsc --noEmit && eslint src --max-warnings 0"
+    "lint:strict": "tsc --noEmit && eslint src --max-warnings 0",
+    "prebuild": "pnpm gen:cosmos-sdk --clean && pnpm gen:ibc && pnpm gen:ics23 && pnpm gen:penumbra && touch src/gen/.npmignore",
```
VanishMax, line 21: praise: thanks for this – will way easier for the rest of us to apply new bufs

packages/perspective/src/translators/swap-claim-view.ts
```diff
@@ -2,7 +2,7 @@ import { Translator } from './types.js';
 import {
   SwapClaimView,
   SwapClaimView_Opaque,
-} from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/dex/v1/dex_pb.js';
+} from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
```
VanishMax, line 5: question: any chance in the future we could simplify the imports? Maybe after applying v2 of the generator we could export types as barrels?

VanishMax (approved):
Great improvement!

This is ultimately a better way than how it was done in #1582. Was there a specific reason to not continue working or commenting on #1582 ?

turbocrime:
going to wait until recent core message type changes are merged for ibc compat before trying to merge this

packages/perspective/src/translators/swap-claim-view.ts
turbocrime: the service definitions have a barrel export defined, but i think trying to collect other types into barrels is likely to be very complex and not very useful.

apps/minifront/src/state/helpers.ts
```diff
@@ -155,18 +155,7 @@ const broadcast = async (
...
-        : never)
-    : (getTxId(t as PartialMessage<Transaction>).then(({ inner }) =>
-        uint8ArrayToHex(inner),
-      ) as T extends Required<PartialMessage<TransactionId>> ? never : Promise<string>);
-
-const getTxId = (tx: Transaction | PartialMessage<Transaction>) =>
+const txSha256 = (tx: Transaction | PartialMessage<Transaction>) =>
   sha256Hash(tx instanceof Transaction ? tx.toBinary() : new Transaction(tx).toBinary()).then(
     inner => new TransactionId({ inner }),
   );
```
turbocrime, line 161: this caused problems at some point during this work, and the types were technically incorrect, so it's now simplified

PR #1625, opened 2024-08-02
feat: loading spinner and streaming for assets table

### LINKED ISSUE #1381: Assets table status + streaming
Upon loading the assets table, two things are noticeable: 

1. There is not an indicator of loading status 
    - We should give visual indications of the loading (our classic spinner) & error statuses
2. All assets load at once
    - The `Balances` rpc method is a streaming one. This means, the client should receive assets as they come. It will also assist in the perceived responsiveness of the page if we didn't wait until all of them were loaded.

https://github.com/penumbra-zone/web/assets/16624263/1c21c7ef-55d4-46ac-b3b2-9f99825d40bd

vacekj:
fixes #1381

### DIFF
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -20,6 +20,7 @@ import { BalancesByAccount, groupByAccount, useBalancesResponses } from '../../.
 import { AbridgedZQueryState } from '@penumbra-zone/zquery/src/types';
 import { shouldDisplay } from '../../../fetchers/balances/should-display';
 import { sortByPriorityScore } from '../../../fetchers/balances/by-priority-score';
+import { Oval } from 'react-loader-spinner';
 
 const getTradeLink = (balance: BalancesResponse): string => {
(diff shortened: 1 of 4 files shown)


TalDerei:
this is great!

https://github.com/user-attachments/assets/d8af408d-3ce0-4a3f-8448-ddf8b6ae28ba

Additionally, assets don't stream in during sync and require a manual refresh. I'm wondering if this is something we could incorporate into the current changeset, or if it would be better handled in a separate PR?

**after sync has completed, and before refreshing:**
Screenshot 2024-08-02 at 8 34 44 AM

**after sync has completed, and after refreshing:**
Screenshot 2024-08-02 at 8 36 10 AM

TalDerei (approved):

vacekj:
@TalDerei I wonder what's causing them to not stream in during syncing - is it perhaps because the fetch method is not refetching automatically?

TalDerei:
> @TalDerei I wonder what's causing them to not stream in during syncing - is it perhaps because the fetch method is not refetching automatically?

I think this might be more involved, so will open separate issue for now.

PR #1614, opened 2024-08-01
fix(minifront): #1598: display non-native-fee-warning per account

VanishMax:
Closes #1598

Checks if user has UM token in the same account as the selected source token

https://github.com/user-attachments/assets/7933599f-c555-4673-97c9-805404334fab

PR #1626, opened 2024-08-02
Fix gas prices RPC

Valentine1898:


packages/services/src/view-service/gas-prices.ts
```diff
@@ -24,13 +24,13 @@ export const gasPrices: Impl['gasPrices'] = async (_, ctx) => {
...
-  const altGasPRices = await indexedDb.getAltGasPrices();
+  const altGasPrices = await indexedDb.getAltGasPrices();
   if (!gasPrices) {
     throw new ConnectError('Gas prices is not available', Code.NotFound);
   }
 
   return {
     gasPrices,
-    altGasPRices,
+    altGasPrices,
```
Valentine1898, line 34: Shouldn't that have resulted in a type error?

VanishMax:
Whoa, I was just trying to understand why it is always empty. Checking if that helps

Valentine1898:
> No type error since we don't type the returns from the services and then use the types from ViewService protobuf – a mistake we have to fix at some point

That's terrible, we definitely need to take care of that

Valentine1898:
I'll hold off on this PR since @TalDerei  is going to provide more clarification

TalDerei (approved):

Valentine1898:
> @Valentine1898 The fixedAltGasPrices struct in the governance proposal was crafted incorrectly with a bracket misalignment, so the prices were interpreted as unknown fields. A new proposal should be submitted with the parameters properly aligned and set correctly.

Ohh 🤯
But it doesn't actually block Max's PR https://github.com/penumbra-zone/web/pull/1614

grod220:
We should return something like:

```ts
return new GasPricesResponse({
    gasPrices,
    altGasPrices,
  });
```

for all services to get typesafety. Will open another issue for this.

PR #1629, opened 2024-08-02
forbid um ibc out to celestia

turbocrime:
~~this PR will conditionally filter um tokens within the ibc out state slice, before they reach the UI asset picker~~

when celestia is selected, filter um tokens from the token picker, when passed into the component prop.

this was easier to confirm behavior than filtering in the state slice.

currently the ibc form will have a pre-filtered set of balances that are already restricted to a chain's native tokens plus um. so, on celestia chain id condition, um is filtered.

### DIFF
--- a/.changeset/polite-otters-cheer.md
+++ b/.changeset/polite-otters-cheer.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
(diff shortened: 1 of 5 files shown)


grod220:
Taking this over (versus leaving feedback) as it is a higher priority fix

PR #1637, opened 2024-08-03
fix: correctly process compact block gas prices

### LINKED ISSUE #1636: Block processor fails to set gas prices from a compact block
https://github.com/penumbra-zone/web/blob/69b083755e83a5d9dc5465345df7981293fb0b1b/packages/query/src/block-processor.ts#L171-L178

When the `assetId` in `compactBlock.gasPrices` is set to `undefined` (not absent, but a present value of undefined), spreading it after attaching an `assetId` to the object will clobber the `assetId` which was just set.

But, the idb table requires an asset id for the keypath. So this `setGasPrices` call will throw an error and prevent sync from progressing.

Screenshot 2024-08-02 at 17 14 06

I believe this is at the root of some of the 'long sync time' issues reported by users.

turbocrime:
fixes #1636

idb interface is updated to require valid input.

block processor saveGasPrices call for native and alt gas prices are both updated to correctly construct the input.

### DIFF
--- a/packages/getters/src/compact-block.ts
+++ b/packages/getters/src/compact-block.ts
@@ -1,0 +1,4 @@
+import { GasPrices } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1/fee_pb.js';
+import { createGetter } from './utils/create-getter.js';
+
+export const getAssetIdFromGasPrices = createGetter((gp?: GasPrices) => gp?.assetId);
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -45,9 +45,10 @@ import { processActionDutchAuctionEnd } from './helpers/process-action-dutch-auc
 import { processActionDutchAuctionSchedule } from './helpers/process-action-dutch-auction-schedule.js';
 import { processActionDutchAuctionWithdraw } from './helpers/process-action-dutch-auction-withdraw.js';
 import { RootQuerier } from './root-querier.js';
-import { GasPrices } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1/fee_pb.js';
 import { IdentityKey } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb.js';
 import { getDelegationTokenMetadata } from '@penumbra-zone/wasm/stake';
+import { toPlainMessage } from '@bufbuild/protobuf';
+import { getAssetIdFromGasPrices } from '@penumbra-zone/getters/compact-block';
 
 declare global {
   // eslint-disable-next-line no-var
@@ -169,16 +170,19 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.indexedDb.saveFmdParams(compactBlock.fmdParameters);
       }
       if (compactBlock.gasPrices) {
-        await this.indexedDb.saveGasPrices(
-          new GasPrices({
-            assetId: this.stakingAssetId,
-            ...compactBlock.gasPrices,
-          }),
-        );
+        console.log('setting native gas prices at', compactBlock.height);
+        await this.indexedDb.saveGasPrices({
+          ...toPlainMessage(compactBlock.gasPrices),
+          assetId: toPlainMessage(this.stakingAssetId),
+        });
       }
       if (compactBlock.altGasPrices.length) {
-        for (const gasPrice of compactBlock.altGasPrices) {
-          await this.indexedDb.saveGasPrices(gasPrice);
+        console.log('setting alt gas prices at', compactBlock.height);
+        for (const altGas of compactBlock.altGasPrices) {
+          await this.indexedDb.saveGasPrices({
+            ...toPlainMessage(altGas),
+            assetId: getAssetIdFromGasPrices(altGas),
+          });
         }
       }
 
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -170,14 +170,12 @@ export class BlockProcessor implements BlockProcessorInterface {
         await this.indexedDb.saveFmdParams(compactBlock.fmdParameters);
       }
       if (compactBlock.gasPrices) {
-        console.log('setting native gas prices at', compactBlock.height);
         await this.indexedDb.saveGasPrices({
           ...toPlainMessage(compactBlock.gasPrices),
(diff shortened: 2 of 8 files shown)


TalDerei (commented):
replicated and confirmed this bug locally. However, this doesn't seem to resolve the problem? https://github.com/penumbra-zone/web/commit/60d0c2e0ea9ec1d6fed0e092ef159922b7cde462 still breaks syncing for me

packages/getters/src/compact-block.ts
```diff
@@ -0,0 +1,4 @@
+import { GasPrices } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/fee/v1/fee_pb.js';
+import { createGetter } from './utils/create-getter.js';
+
+export const getAssetIdFromGasPrices = createGetter((gp?: GasPrices) => gp?.assetId);
```
TalDerei, line 4: If `getAssetIdFromGasPrices` is called with a `GasPrices` object where assetId is undefined, it will still return `undefined`.

packages/query/src/block-processor.ts
```diff
@@ -169,16 +170,17 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+        });
       }
       if (compactBlock.altGasPrices.length) {
-        for (const gasPrice of compactBlock.altGasPrices) {
-          await this.indexedDb.saveGasPrices(gasPrice);
+        for (const altGas of compactBlock.altGasPrices) {
+          await this.indexedDb.saveGasPrices({
+            ...toPlainMessage(altGas),
+            assetId: getAssetIdFromGasPrices(altGas),
+          });
```
grod220, line 183: issue: can we just put the guard inside storage so callers don't have to worry about this?

```ts
  async saveGasPrices(value: PartialMessage<GasPrices>): Promise<void> {
    if (!value.assetId) {
      value.assetId = this.stakingTokenAssetId;
    }
    
    await this.u.update({
      table: 'GAS_PRICES',
      value: new GasPrices(value).toJson() as Jsonified<GasPrices>,
    });
  }
```

This also doesn't require `toPlainMessage` which isn't entirely obvious from a glance how it serializes all the fields.

grod220 (commented):
What environments are passing gas prices where the assetId is set explicitly as `undefined`? Is this on a local chain instance?

packages/query/src/block-processor.ts
TalDerei, line 183: we need an additional conditional check before spreading `compactBlock.gasPrices `, something like

```
const altGasPrices = {
  ...toPlainMessage(altGas),
  assetId: getAssetIdFromGasPrices(altGas),
};

if (altGasPrices.assetId) {
  await this.indexedDb.saveGasPrices(altGasPrices);
} else {
  continue syncing without error
}
```

TalDerei:
> What environments are passing gas prices where the assetId is set explicitly as `undefined`?

can you elaborate on the use of "environments" here?

packages/query/src/block-processor.ts
grod220: Actually, scratch that. We have to handle Um/alts differently. Something like this would work:

```ts
// used in main block processor loop
await this.saveGasPrices(compactBlock);

=====
// additional method

  private async saveGasPrices(block: CompactBlock) {
    const pricesToSave = [...block.altGasPrices];

    // Undefined field implies UM. Manually overriding as this is commonly not passed.
    if (block.gasPrices) {
      pricesToSave.push(
        new GasPrices({
          ...block.gasPrices,
          assetId: this.stakingAssetId,
        }),
      );
    }

    for (const price of pricesToSave) {
      if (!price.assetId) {
        console.warn(
          `BlockHeight: ${block.height}. No asset id for gas prices: ${JSON.stringify(price.toJson())}`,
        );
        continue;
      }
      await this.indexedDb.saveGasPrices(price);
    }
  }
```

grod220: Yep, you're right. Updated my example above.

grod220:
> can you elaborate on what you mean by "environments" here?

Currently on mainnet, the assetId field is either present with data or absent entirely. There is not a block that passes an assetId set to `undefined`.

TalDerei:
from 

> > can you elaborate on what you mean by "environments" here?
> 
> Currently on mainnet, the assetId field is either present with data or absent entirely. There is not a block that passes an assetId set to `undefined`.

currently no environments, which is why this bug hasn't surfaced in the wild. I replicated this on a local chain instance

packages/query/src/block-processor.ts
turbocrime: callers should be explicit in intent. a well-typed parameter means callers don't have to worry - the type system will protect them.

unanticipated fallback behavior upon absent input (not default input, but literally absent input) is a greater source of concern. 

plainmessage does not serialize fields. it represents the message type as a plain object. it is the same as the class instance, sans class instance methods.

it is the simplest way to express that a parameter requires data present for every field of the message type. using `Required` on the class would require non-message input, using the class led us to the present situation, and the third option is just a literal redefinition of what the plainmessage type provides.

any class instance or object literal meeting the message type satisfies it.

this information is available from the docstring on the type and on the function. its existence is explained and its use in specific cases such as this are recommended in bufbuild/connectrpc documentation.

turbocrime:
> What environments are passing gas prices where the assetId is set explicitly as `undefined`? Is this on a local chain instance?  

the compact chain on every um price update, apparently. the issue seems conditional for unknown reasons, but persistent when it occurs.

packages/getters/src/compact-block.ts
turbocrime: hmm. this solved the case on the testnet i was using but it's true the inner could be empty.

true correctness presumably requires much idb refactor and some helper methods.

i will add something here but larger interface work is out of scope

turbocrime: okay - if an AssetId class instance is present, it always contains a uint8array inner. so this

1. guards against the condition of an absent assetid (no inner, invalid keypath)
2. does not guard against the condition of a present assetid with default zero-length inner (bad inner, valid keypath)
3. does not guard against the impossible condition of a present assetid with an absent inner (can't happen by getter input type)

in the case of 3, the parameter of the getter requires a message class instance, and a class instance of gas prices will never have a present assetid with an absent inner. this is because children of the class instance are always other class instances, and so `assetId` may be absent and thus inner is undefined (case 1), or present and thus inner is defined of some length (case 2)

so the getter performs its purpose, but does not perform length verification. thus 1639

PR #1639, opened 2024-08-03
assert lengths of binary field inputs to idb methods

turbocrime:
this is intended to encourage callers to ensure proper input when querying idb, until a more comprehensive and well-typed interface is implemented.

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -63,32 +63,39 @@ import {
   DutchAuctionDescription,
 } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/component/auction/v1/auction_pb.js';
 import { ChainRegistryClient } from '@penumbra-labs/registry';
-import { PartialMessage } from '@bufbuild/protobuf';
+import { PartialMessage, PlainMessage } from '@bufbuild/protobuf';
 import { getAmountFromRecord } from '@penumbra-zone/getters/spendable-note-record';
 import { isZero } from '@penumbra-zone/types/amount';
 import { IDB_VERSION } from './config.js';
 
-const assertBytes = (v?: Uint8Array, expect?: number, name = 'value') => {
+const assertBytes = (v?: Uint8Array, expect?: number, name = 'value'): v is Uint8Array => {
   if (expect !== undefined && v?.length !== expect) {
     throw new Error(`Expected ${name} of ${expect} bytes, but got ${v?.length} bytes`);
   } else if (expect === undefined && !v?.length) {
     throw new Error(`Expected ${name} to be non-empty, but got ${v?.length} bytes`);
   }
+  return true;
 };
 
 // yes these are all 32 bytes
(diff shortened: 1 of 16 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -63,11 +63,33 @@ import {
...
 
+const assertBytes = (v?: Uint8Array, expect?: number, name = 'value') => {
+  if (expect !== undefined && v?.length !== expect) {
+    throw new Error(`Expected ${name} of ${expect} bytes, but got ${v?.length} bytes`);
+  } else if (expect === undefined && !v?.length) {
+    throw new Error(`Expected ${name} to be non-empty, but got ${v?.length} bytes`);
+  }
+};
+
+// yes these are all 32 bytes
```
grod220, line 79: Will they _always_ be 32 bytes in these cases forever?

grod220 (commented):
I'm not really sure what this practically defends us from. Outside of tests, we aren't assembling bytes manually but passing the ones we get from the node. Is this supposed to ensure the quality of data we receive externally?

packages/storage/src/indexed-db/index.ts
TalDerei: currently asset id's are set to 32 bytes and are internally represented as cryptographic [field elements](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/asset/src/asset/id.rs#L31). Theoretically, this could change if we decide to switch cryptographic curves, though that would only be a long-term consideration.

TalDerei (approved):

packages/storage/src/indexed-db/index.ts
grod220: In general, I think we can simply rely upon the bufbuild types and not have to assert byte lengths to ensure the integrity of the data we are passed.

TalDerei: that's a good point

turbocrime:
> I'm not really sure what this practically defends us from.

this defends us from corrupting local state.

underspecified idb interaction has been a persistent cause of critical bugs across minifront, wasm, and consumers like prax

packages/storage/src/indexed-db/index.ts
turbocrime: bufbuild classes do not validate length, and binary fields default init to a zero-length array. some validation is absolutely necessary

grod220:
This PR changes a ibd signature to use `PlainMessage` without reasoning or code comments why. It feels like a bad API when devs don't really know why it's necessary. Also, feels unrelated to the core of this PR.

The serialization/deserialization library should be responsible for asserting that types are valid. In the core repo's case, it's serde. In our case, it's the ConnectRpc library. Validating the exact number of bytes that are in our types feels very strange.

> underspecified idb interaction has been a persistent cause of critical and subtle bugs across minifront, services, wasm, and consumers like prax

Have we ever encountered a case where this code would have defended us? It feels unrelated to any priorities right now and unclear how we'd ever trigger this without manually manipulating the inner bytes of a type, which would be highly unusual. These guarding assertions are not necessary.

PR #1640, opened 2024-08-03
Display jailed validators

grod220:
At the moment, the `delegationsByAddressIndex` rpc method is filtering out jailed validators. This prevents users from being able to see the balance on the staking page.

This PR addresses this by performing two steps:

Step 1: query the current list of validators, send those as normal
Step 2: for the delegation token balances that have not been queried, send those in a separate for loop

A PR should follow this to display status more prominently in the frontend.

### DIFF
--- a/packages/services/src/view-service/delegations-by-address-index.test.ts
+++ b/packages/services/src/view-service/delegations-by-address-index.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { delegationsByAddressIndex } from './delegations-by-address-index.js';
-import { ViewService, StakeService } from '@penumbra-zone/protobuf';
+import { StakeService, ViewService } from '@penumbra-zone/protobuf';
 import {
   createContextValues,
   createHandlerContext,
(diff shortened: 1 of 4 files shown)


packages/services/src/view-service/delegations-by-address-index.test.ts
```diff
@@ -300,35 +305,121 @@ describe('DelegationsByAddressIndex request handler', () => {
...
+          }),
+          mockCtx,
+        )) {
+          results.push(result);
+        }
+
+        expect(results.length).toBe(4);
+      });
     });
   });
```
grod220, line 391: No changes to the tests above, just nested them in a describe block so the mocked functions do not conflict with the new test

Valentine1898 (approved):
LGTM! 

I think it is wrong that the RPC method that is called as `delegationsByAddressIndex` does not return all user delegations by default

grod220:
> I think it's wrong that the RPC method delegationsByAddressIndex doesn't return all user delegations by default

We are following the spec (https://buf.build/penumbra-zone/penumbra/file/main:penumbra/view/v1/view.proto#L714), but I see what you mean.

Minifront next will have to pass the ALL variant as we are relying there on the default case.

PR #1641, opened 2024-08-05
feat: loading spinner for assets table

### LINKED ISSUE #1381: Assets table status + streaming

vacekj:
re-fixes #1381

### DIFF
--- a/apps/minifront/package.json
+++ b/apps/minifront/package.json
@@ -61,6 +61,7 @@
   },
   "devDependencies": {
     "@chain-registry/types": "^0.45.38",
+    "@eslint/compat": "^1.1.0",
     "@types/lodash": "^4.17.4",
     "@types/react": "^18.3.2",
     "@types/react-dom": "^18.3.0",
--- a/apps/minifront/src/components/dashboard/assets-table/index.tsx
+++ b/apps/minifront/src/components/dashboard/assets-table/index.tsx
@@ -20,6 +20,7 @@ import { BalancesByAccount, groupByAccount, useBalancesResponses } from '../../.
 import { AbridgedZQueryState } from '@penumbra-zone/zquery/src/types';
 import { shouldDisplay } from '../../../fetchers/balances/should-display';
 import { sortByPriorityScore } from '../../../fetchers/balances/by-priority-score';
+import { Oval } from 'react-loader-spinner';
 
 const getTradeLink = (balance: BalancesResponse): string => {
   const metadata = getMetadataFromBalancesResponseOptional(balance);
@@ -47,7 +48,18 @@ export default function AssetsTable() {
     shouldReselect: (before, after) => before?.data !== after.data,
   });
 
-  if (balancesByAccount?.length === 0) {
+  /** Are assets still loading */
+  const isLoading = balancesByAccount === undefined;
+
+  if (isLoading) {
+    return (
+      <div className='mt-5 flex w-full flex-col items-center justify-center'>
+        <Oval width={32} height={32} color='white' secondaryColor='white' />
+      </div>
+    );
+  }
+
+  if (balancesByAccount.length === 0) {
     return (
       <div className='mt-5 flex flex-col items-center gap-6'>
         <p>
@@ -64,7 +76,7 @@ export default function AssetsTable() {
   return (
     <div className='w-full overflow-x-auto'>
       <Table>
-        {balancesByAccount?.map(account => (
+        {balancesByAccount.map(account => (
           <Fragment key={account.account}>
             <TableHeader className='group'>
               <TableRow>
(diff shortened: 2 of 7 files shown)


TalDerei (approved):

TalDerei:
> More manual testing is needed, it seems to be loading my balances many times slower than before

seemed relatively performant to me during [testing](https://github.com/penumbra-zone/web/pull/1625#issuecomment-2265456799)?

Valentine1898:
> > More manual testing is needed, it seems to be loading my balances many times slower than before
> 
> seemed relatively performant to me during [testing](https://github.com/penumbra-zone/web/pull/1625#issuecomment-2265456799)?

In the `penumbra-1` chain, my balances take ~21 sec to load 
don't merge this PR

TalDerei:
can you add a changeset?

TalDerei:
> > > More manual testing is needed, it seems to be loading my balances many times slower than before
> > 
> > 
> > seemed relatively performant to me during [testing](https://github.com/penumbra-zone/web/pull/1625#issuecomment-2265456799)?
> 
> In the `penumbra-1` chain, my balances take ~21 sec to load don't merge this PR

> > > More manual testing is needed, it seems to be loading my balances many times slower than before
> > 
> > 
> > seemed relatively performant to me during [testing](https://github.com/penumbra-zone/web/pull/1625#issuecomment-2265456799)?
> 
> In the `penumbra-1` chain, my balances take ~21 sec to load don't merge this PR

interesting, I didn't test `penumbra-1`

Valentine1898:
https://github.com/user-attachments/assets/cfc80062-7894-425a-86a8-2a56ee3576a4

vacekj:
@Valentine1898 ack, debugging..

TalDerei:
following up on this, was the performance regression fixed?

vacekj:
@VanishMax can you try to repro this? I can't seem to repro locally with the test account on mainnet.

TalDerei:
> @VanishMax can you try to repro this? I can't seem to repro locally with the test account on mainnet.

I'll try to reproduce

grod220:
I can't reproduce any slowness, but @JasonMHasperhoven can you test this out on testnet with the testnet whale seedphrase given you have an older macbook?

TalDerei:
I can't reproduce the slowness either using the whale account on mainnet

vacekj:
also can't see why the CI is failing but my local is passing - debugging now.

JasonMHasperhoven:
Using testnet on my 2019 macbook:
https://github.com/user-attachments/assets/1f09cdaf-067e-42a8-8263-32d044f8bf1a

cc @grod220

grod220:
> Using testnet on my 2019 macbook: https://github.com/user-attachments/assets/1f09cdaf-067e-42a8-8263-32d044f8bf1a

This is a compelling test. Why would streaming balances not be basically instant for a slower machine? The balances steps are this:

- Query notes from database
- Combine notes that are the same asset id + account
- Stream results back to the client
- React renders result (via ZQuery)

This leads me to believe there is a react rendering kind of performance drag.

@JasonMHasperhoven, another helpful test would be to make the same query on `main` and see how fast that is compared to this. If it's faster, that means it's a react rendering issue. If not, that means the rpc method (or db calls) in the extension is slower for lower performance machines.

Valentine1898:
> Screen.Recording.2024-08-05.at.17.39.07.mov

Additional context 
I tested on my macbook m3 pro 
The mainnet wallet I tested on has about 3000 notes on it

grod220:
Also @vacekj, just realized we want the spinner to be showing (in a place that doesn't block the content) during the entire stream lifetime. At the moment, it looks like it goes away after the first item is loaded. Can we also use [Linewave](https://github.com/penumbra-zone/web/blob/b732bc580e16d827e55d25a64b1d5d05d85b93a0/apps/node-status/src/components/header.tsx#L2) for the spinner (we use it in other places).

vacekj:
@grod220 good point, will address this now.

vacekj:
@grod220 addressed in the latest commit. the correct loading indicator will now show until the stream has finished.

TalDerei:
still unclear if there's a performance regression?

vacekj:
@TalDerei I can't replicate the perf regression on my machine, and don't see any obvious issues with code. Even with the perf. regression, this feels like better UX than before, so would carefully merge and watch for potential use reports.

Alternatively, I can try debugging with @JasonMHasperhoven on his machine to see what's wrong.

@grod220 would love your call here.

PR #1643, opened 2024-08-05
Jailed validators pt 2: block processor + getValidatorInfo endpoint

grod220:
Follow up from: https://github.com/penumbra-zone/web/pull/1640

Does three things:
- Clears out stale validator info entries in the database anytime we need to refetch
- `getValidatorInfo` view service endpoint has a fallback to the node in the case (like jailed validators) it is not in the db
- Adds a component to prominently display bad statuses of validators

Screenshot 2024-08-05 at 3 08 03 PM


Valentine1898 (approved):
LGTM!

PR #1647, opened 2024-08-05
adr 006 changes

turbocrime:
static and instance methods on `PenumbraClient`

- client static methods allow consumers to inspect available providers
- client instance is created to interact with a specific provider

implementation following soon

### DIFF
--- a/docs/adrs/006-web-apis.md
+++ b/docs/adrs/006-web-apis.md
@@ -41,9 +41,19 @@ Creating the `client` should be the starting point for any application working w
 ```ts
 import { PenumbraClient } from '@penumbra-zone/client';
 
-export const providerManifests: Record<string, PenumbraManifest> = await PenumbraClient.manifests();
-const providerOrigin: keyof providerManifests = '....';
-export const client = new PenumbraClient(providerOrigin);
+const providerManifests: Record<
+  string,
+  Promise<PenumbraManifest>
+> = PenumbraClient.providerManifests();
+const someProviderOrigin: keyof providerManifests = '....';
+
+// connect will fetch and verify manifest before initiating connection, then return an active client
+const someProviderClient = await PenumbraClient.connect(someProviderOrigin);
+
+// or, a caller could do it independently
+const someProviderManifest = await providerManifests[someProviderOrigin];
+const createdProviderClient = new PenumbraClient(someProviderOrigin);
+void createdProviderClient.connect();
 ```
 
 The flow of work with the `client` would be as follows:
--- a/docs/adrs/006-web-apis.md
+++ b/docs/adrs/006-web-apis.md
@@ -100,15 +100,15 @@ import type { PromiseClient } from '@connectrpc/connect';
 interface PenumbraClient {
 
   // static methods
-  public static manifests: () => Record<string, Promise<PenumbraManifest>>;
+  public static providerManifests: () => Record<string, Promise<PenumbraManifest>>;
 
   // provider-specific static methods
-  public static manifest: (providerOrigin?: string) =>  Promise<PenumbraManifest>;
-  public static isConnected: (providerOrigin?: string) => boolean;
-  public static state: (providerOrigin?: string) => PenumbraState;
+  public static providerManifest: (providerOrigin?: string) =>  Promise<PenumbraManifest>;
+  public static providerIsConnected: (providerOrigin?: string) => boolean;
+  public static providerState: (providerOrigin?: string) => PenumbraState;
 
   /** Initiates initiates connection, returns a client instance. */
-  public static connect: (providerOrigin?: string) => Promise<PenumbraClient>;
+  public static providerConnect: (providerOrigin?: string) => Promise<PenumbraClient>;
 
   // instance features
 
--- a/docs/adrs/006-web-apis.md
+++ b/docs/adrs/006-web-apis.md
@@ -41,19 +41,11 @@ Creating the `client` should be the starting point for any application working w
 ```ts
 import { PenumbraClient } from '@penumbra-zone/client';
 
-const providerManifests: Record<
-  string,
-  Promise<PenumbraManifest>
-> = PenumbraClient.providerManifests();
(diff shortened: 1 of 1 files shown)


docs/adrs/006-web-apis.md
```diff
@@ -111,80 +108,61 @@ import type { PenumbraService } from '@penumbra-zone/protobuf';
...
+  public readonly transport?: Transport;
+  public readonly port?: MessagePort;
 
-The client library should separately export the creation of service clients. It is not going to be integrated into the `client` instance to save the initial bundle size. Instead, it should be exported from `@penumbra-zone/client/service`:
+  /** Direct re-exports the from appropriate provider */
+  public readonly disconnect: () => Promise<void>;
+  public readonly isConnected: () => boolean | undefined;
+  public readonly state: () => PenumbraState;
+  public readonly addEventListener: EventTarget['addEventListener'];
+  public readonly removeEventListener: EventTarget['removeEvenListener'];
```
VanishMax, line 140: thought (possible complication): i'm still hesitant of this typing. It requires importing a helper function:

```ts
client.addEventListener('penumbrastate', (event) => {
  if (isPenumbraStateEvent(evt)) {
    // Do something
  }
});
```

This import makes the use way more complicated. Especially that most of the methods of the library come from the `PenumbraClient`, it makes it ambiguous.

suggestion: simplify the typings

VanishMax: question: what happened to the `onConnectionChange`? It worked like a charm here: https://github.com/penumbra-zone/nextjs-penumbra-client-example/pull/5/files#diff-6e5a42ea42819066f6aa0ff0e0d91ef85f5b435355e6af424cc2e25e9d03a8f0R38

docs/adrs/006-web-apis.md
```diff
@@ -39,9 +39,21 @@ that manages the injected connections and provides methods for interacting with
...
 
-export const client = createPenumbraClient();
+const providerManifests: Record<
+  string,
+  Promise<PenumbraManifest>
+> = PenumbraClient.providerManifests();
+const someProviderOrigin: keyof providerManifests = '....';
+
+// connect will fetch and verify manifest before initiating connection, then return an active client
+const someProviderClient = await PenumbraClient.connect(someProviderOrigin);
```
VanishMax, line 51: issue: the `providerClient` has to be reused across the users' apps. If we require them to know the origin beforehand, idk how they are going to reuse it...

suggestion: allow the creation of a client without the need of the origin. Seriously, why changing this?

docs/adrs/006-web-apis.md
turbocrime: in the minifront implementation, there were issues with race conditions in the callback. the event handler would actually fire, and then activate the callbacks in the map, before the new state was available in the client object.

i can re-add this

docs/adrs/006-web-apis.md
turbocrime: specifying an origin is selecting a provider.

there is no way to get around selecting a provider deliberately.

selection of a provider is creation of state, and the object instance is returned to track that state and represent the client connection. the consumer is expected to handle the client instance and convey it to where it should be available.

if consumers want to recreate clients to the same provider, that should be supported, but then they have to be responsible for their own state tracking.

in react apps this can be provided by hooks in a react package, but all of that pends merge of a settled client interface.

docs/adrs/006-web-apis.md
```diff
@@ -111,80 +108,61 @@ import type { PenumbraService } from '@penumbra-zone/protobuf';
...
-import { createPenumbraClient } from '@penumbra-zone/client';
-import { createServiceClient } from '@penumbra-zone/client/servce';
+import { PenumbraClient } from '@penumbra-zone/client';
 import { ViewService } from '@penumbra-zone/protobuf';
 
-export const client = createPenumbraClient();
+const penumbraClient = new PenumbraClient.connect('some origin');
 
-const viewService = createServiceClient(client, ViewService);
+const viewClient = penumbraClient.service(ViewService);
```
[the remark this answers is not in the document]
turbocrime: this wasn't in the adr before, and i'd like to stop adding features

using a servicetype definition allows the provider to technically support services we haven't defined. in #1648 now there's an optional 'transport options' parameter to connect, which would enable this.

docs/adrs/006-web-apis.md
turbocrime: a callback function is restored now

docs/adrs/006-web-apis.md
```diff
@@ -39,9 +39,21 @@ that manages the injected connections and provides methods for interacting with
...
+> = PenumbraClient.providerManifests();
+const someProviderOrigin: keyof providerManifests = '....';
+
+// connect will fetch and verify manifest before initiating connection, then return an active client
+const someProviderClient = await PenumbraClient.connect(someProviderOrigin);
+
+// or, a caller could do it independently
+const someProviderManifest = await providerManifests[someProviderOrigin];
+const createdProviderClient = new PenumbraClient(someProviderOrigin);
+void createdProviderClient.connect();
```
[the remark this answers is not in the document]
turbocrime: some of this was typo

the static PenumbraClient.connect() is available, as well as manual client construction for more control. we can encourage PenumbraClient.connect() by encouraging that in documentation, but the style used in minifront prefers manual management of a manually constructed client, and changing that would require refactor.

i think it's good to support extra control, but we can also prioritize PenumbraClient.connect() in docs. it's updated to do so.

docs/adrs/006-web-apis.md
turbocrime: notably if this is not changed, the API previously defined in this doc provides no way for a page to inspect a provider's connection state before attempting to connect, and given the requirement that client creation not automatically attempt to connect, then that results in a user experience of initiating every connection manually.

VanishMax (approved):
praise for `PenumbraClient.providers()` – I love the naming of it, better than `getWallets`

PR #1648, opened 2024-08-06
adr 006 changes and implementation

turbocrime:
~~meeting suggestions in #1647~~

suggestions available in adr doc in this pr

still removes react package, it would be broken by implementation

### currently

implements PenumbraClient as a class with static and instance methods.

- static methods support inspection of available providers to support making a choice of provider
- a class instance is created to encapsulate client configuration and state
    - provider selection
    - transport config
    - connection and connection state
    - callbacks

a class instance may be created without configuration, but must be configured with provider to connect.

an unconfigured client instance
- can select a provider by parameter of its `.connect` method
- supports adding client connection state change callbacks
- can't forward the provider's EventTarget methods (no provider to forward)
- returns 'undefined' from most other methods depending on a configured state
- will async reject connection to an undefined provider
- sync throw use of services on an undefined provider

### ~~questions remaining~~

~~should an unconfigured client instance which possesses callbacks activate the callbacks immediately when a provider is configured with an event representing the new provider's present state, or wait until the provider emits a state change event?~~

~~should a callback added to a client be immediately activated with an event representing the present state, or wait until the provider emits a state change event?~~

~~should an existing unconfigured client instance provide a feature to select a provider without attempting connection?~~

~~is it appropriate to throw creation of a service client when no provider is configured?~~

~~is it appropriate to implicitly request a connection if a service client is created when a provider is configured but not connected?~~

### i'd like to add

when parsing the manifest, the client could either reformat the icon URLs to a fully-qualified path, or simply fetch them so they are available synchronously.


packages/client/src/client.ts
```diff
@@ -0,0 +1,303 @@
...
+    if (this.config?.origin && isPenumbraStateEvent(evt, this.config.origin)) {
+      this.stateListeners.forEach(listener => listener(evt.detail));
+    }
+  };
+
+  /** Construct an instance representing connection to a specific provider,
+   * identified by the origin parameter, but take to specific action. */
+  private constructor(
+    /** A provider origin may be provided during client construction. */
+    providerOrigin?: string,
```
VanishMax, line 114: suggestion: instead of listing arguments, make it one extendable arguments interface like:

```ts
interface PenumbraClientConstructorOptions {
  providerOrigin?: string,
  transportOptions?: Omit<ChannelTransportOptions, 'getPort'>,
}

private constructor(options: PenumbraClientConstructorOptions) { ... }
```

We don't know how in the future we are going to extend the constructor but we can prepare for it. I think providing object options should apply here and in all functions for consistency and extendability

packages/client/src/client.ts
```diff
@@ -0,0 +1,303 @@
...
+  public service<T extends ServiceType>(service: T) {
+    return (
+      (this.serviceClients.get(service.typeName) as PromiseClient<T> | undefined) ??
+      (this.serviceClients
+        .set(service.typeName, this.connectService(service))
+        .get(service.typeName) as PromiseClient<T>)
+    );
+  }
+
+  get addEventListener(): PenumbraProvider['addEventListener'] | undefined {
```
VanishMax, line 283: todo: add JSDoc for these and the rest of the functions.

JSDocs should be detailed for each of the function, even if it just re-exports the injected provider's methods. This way, when users create the `client`, they will hover over the method (or choose it in the IDE's autocomplete) with the comments.

packages/client/src/client.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /** Return a promised `PenumbraManifest` associated with this provider, fetched
+   * at time of client configuration, or `undefined` if this client is not
+   * configured with a provider.
+   *
+   * May reject if the manifest is not available. If you hold an active
+   * connection to this provider, this promise should successfully resolve.
+   */
+  fetchManifest(): Promise<PenumbraManifest> | undefined {
```
VanishMax, line 212: question: is is `public`, `static`, or `private`?

packages/client/src/index.ts
```diff
@@ -1,48 +1,9 @@
...
-export const getAllPenumbraManifests = (
-  signal?: AbortSignal,
-): Record<string, Promise<PenumbraManifest>> =>
-  Object.fromEntries(
-    Object.keys(assertGlobalPresent()).map(providerOrigin => [
-      providerOrigin,
-      getPenumbraManifest(providerOrigin, signal),
-    ]),
-  );
+export { PenumbraClient } from './client.js';
```
VanishMax, line 5: suggestion: export `createPenumbraClient` too and change `static` methods to `public`. The `createPenumbraClient` makes the API consistent and finished. 

Consider two usages:

```ts
// penumbra.ts
import { PenumbraClient } from '@penumbra-zone/client';
export const client = PenumbraClient.create();

// hooks.ts
import { PenumbraClient } from '@penumbra-zone/client';
import { client } from './penumbra';

const reconnect = async () => {
    const providers = PenumbraClient.getAllProviderOrigins();
    const connected = providers.find((origin) => PenumbraClient.isConnected(origin));
    if (!connected) return;
    await client.connect(connected);
};
```

and

```ts
// penumbra.ts
import { createPenumbraClient } from '@penumbra-zone/client';
export const client = createPenumbraClient();

// hooks.ts
import { client } from './penumbra';

const reconnect = async () => {
    const providers = client.getAllProviderOrigins();
    const connected = providers.find((origin) => client.isConnected(origin));
    if (!connected) return;
    await client.connect(connected);
};
```

the first example is taken straight from https://github.com/penumbra-zone/nextjs-penumbra-client-example/pull/6. The difference seems minor, and still it complicates the DX a little. Why using `PenumbraClient`'s static methods when `client` is already created? And it will always be created.

and a nitpick: maybe it's just me but when I see libraries that make me use classes, I want to find another library instead. Classes always look old-fashioned, idk.

docs/adrs/006-web-apis.md
```diff
@@ -0,0 +1,226 @@
...
+
+```ts
+import { PenumbraClient } from '@penumbra-zone/client';
+
+const providers: Record<string, PenumbraProvider> = PenumbraClient.providers();
+const someProviderOrigin: keyof providers = '....'; // dapp chooses a provider
+
+// connect will fetch and verify manifest before initiating connection, then return an active client
+const someProviderClient = await PenumbraClient.providerConnect(someProviderOrigin);
+```
```
VanishMax, line 47: issue: outdated info

VanishMax (changes requested):
We are extremely close! I think that the requested changes are final

VanishMax:
@turbocrime regarding the questions in the PR description, sharing my opinion:

> should an unconfigured client instance which possesses callbacks activate the callbacks immediately when a provider is configured with an event representing the new provider's present state, or wait until the provider emits a state change event?

no need for activating them immediately – users can always check the state an any time. 

> should a callback added to a client be immediately activated with an event representing the present state, or wait until the provider emits a state change event?

is it different from the previous question? don't really see the difference, so the answer remains

> should an existing unconfigured client instance provide a feature to select a provider without attempting connection?

this could be useful – will save users at least one function call

> is it appropriate to throw creation of a service client when no provider is configured?

probably no

> is it appropriate to implicitly request a connection if a service client is created when a provider is configured but not connected?

this might be an unexpected UX. Maybe in the future we'll add some more wrappers or handlers to the service calls with some logic but I'd leave it as-is for now

> i'd like to add: when parsing the manifest, the client could either reformat the icon URLs to a fully-qualified path, or simply fetch them so they are available synchronously.

+1 on this, could be here or we can create an issue for this and leave for later

turbocrime:
> @turbocrime regarding the questions in the PR description, sharing my opinion:
> 
> > should an unconfigured client instance which possesses callbacks activate the callbacks immediately when a provider is configured with an event representing the new provider's present state, or wait until the provider emits a state change event?
> 
> no need for activating them immediately – users can always check the state an any time.
 👍 
  
> > should a callback added to a client be immediately activated with an event representing the present state, or wait until the provider emits a state change event?
> 
> is it different from the previous question? don't really see the difference, so the answer remains

it is different, because this is a real client state change, despite no provider state change.
 
> > should an existing unconfigured client instance provide a feature to select a provider without attempting connection?
> 
> this could be useful – will save users at least one function call

ok, will expose existing private method that performs this function

> > is it appropriate to throw creation of a service client when no provider is configured?
> 
> probably no

the alternative is the return type of the `service` function must include `undefined`, which will be indistinguishable from the next case. is this acceptable? should calling `connect` on an unconfigured provider match this behavior?

> > is it appropriate to implicitly request a connection if a service client is created when a provider is configured but not connected?
> 
> this might be an unexpected UX. Maybe in the future we'll add some more wrappers or handlers to the service calls with some logic but I'd leave it as-is for now

ok, this will necessarily return `undefined`

> > i'd like to add: when parsing the manifest, the client could either reformat the icon URLs to a fully-qualified path, or simply fetch them so they are available synchronously.
> 
> +1 on this, could be here or we can create an issue for this and leave for later

ok

VanishMax:
@turbocrime  

> - is it appropriate to throw creation of a service client when no provider is configured?
> - probably no
> - the alternative is the return type of the service function must include undefined, which will be indistinguishable from the next case. is this acceptable? should calling connect on an unconfigured provider match this behavior?

but we can create a PromiseClient that will return empty port when the client is not configured, aren't we? This way there's no need to return `undefined` or throw a on service creation – instead the service method call with throw.

as for the other questions (about callbacks, for example) i don't have strong opinions, so i'd be ok with any implementation

turbocrime:
> but we can create a PromiseClient that will return empty port when the client is not configured, aren't we? This way there's no need to return `undefined` or throw a on service creation – instead the service method call with throw.

this is likely confusing to the consumer
- why deliberately allow creation of a broken client?
- will consumers understand they need to recreate the client?
- consumers are now likely to develop use techniques containing race conditions

turbocrime:
at some point, if the caller is doing something wrong, we have to throw.

it is better to throw at the location of the problem, when the problem is noticed. creating broken outputs like that is essentially a null propagation error. we should enable developers to use this library correctly by rejecting conditions that are incorrect.

i dislike that 'void' or 'undefined' are acceptable returns from many of these functions. it seemed very desired to we avoid things like 'undefined' return, but they've been added back now in response to discussion.

simply returning 'undefined', when there is an explainable problem, obscures what's happening and is ultimately confusing rather than robust.

VanishMax:
@turbocrime alright, let's throw an error from the service creation. If one day we find a more neat solution to change the behavior, it won't be a breaking change

packages/client/src/index.ts
VanishMax: ok, I'd say that making static methods also public could be a non-breaking addition for the future. Let's way for what our users have to say and, if they would have this feedback, we'll implement it.

can't see the `createPenumbraClient` for now

packages/client/src/client.ts
turbocrime: unannotated is public, will annotate

packages/client/src/client.ts
turbocrime: just removing the forwarded provider method here as
1. the callback provides the functionality
2. the provider is accessible from the client if desired

packages/client/src/client.ts
turbocrime: i kept the origin identifier as a top-level parameter, as i'm fairly confident it should always be necessary. the second parameter now contains an options object.

turbocrime:
with the addition of this check,

https://github.com/penumbra-zone/web/blob/52e7c84f119a0278f4eb6666885b23477194dbcc/packages/client/src/client.ts#L328-L333

and the removal of the event type change,

https://github.com/penumbra-zone/web/commit/c39c5022f63321c3c250e5e1612dd1dc565c8751

prax 11.13.1 is functional with all features of this ADR.

prax must still be updated to support more than one connection within a document lifecycle, but that work is in https://github.com/prax-wallet/web/pull/134

packages/client/src/index.ts
turbocrime: removed the static effective methods `PenumbraClient.create` and `PenumbraClient.connect` in favor of your suggested `createPenumbraClient`

this feels much cleaner. now all static methods are 'read-only', and only instance methods will take action.

turbocrime:
### questions answered 

>should an unconfigured client instance which possesses callbacks activate the callbacks immediately when a provider is configured with an event representing the new provider's present state, or wait until the provider emits a state change event?

yes. an unconfigured client with callbacks will notify listeners when it attaches to a provider. this is how you expected it to function in your nextjs demo.

>should a callback added to a client be immediately activated with an event representing the present state, or wait until the provider emits a state change event?

no. a caller who wishes to identify the present state must do so by checking the present state.

>should an existing unconfigured client instance provide a feature to select a provider without attempting connection?

this is provided and termed `attach`, but it seems fairly useless. theoretically, one may maximally exercise this feature to

1. create an unconfigured client
2. add a callback
3. manually attach the client to a provider without initiating connection
4. receive the provider's non-connected state in the callback

>is it appropriate to throw creation of a service client when no provider is configured?

creating a service client before attaching to a provider will throw.

this will prevent nonsensical use patterns, or use patterns creating subtle timeout issues or race conditions with configuration.

notably: consumers should not use an unconfigured client to create a service client at a top-level scope and then export it.

>is it appropriate to implicitly request a connection if a service client is created when a provider is configured but not connected?

- if the client is attached and reports connected === `true`, then it will construct a transport if necessary and return a client expected to work
- if the client is attached and reports connected === `false`, then it will throw a connection error specifying its attached origin if creation of a service client is attempted
- if a client is not attached, it reports connected === `undefined` and it will throw a connection error specifying an `undefined` origin if creation of a service client is attempted

turbocrime:
notably, this implementation involves a field indicating a three-state true/false/undefined connection status, with undefined representing the connection state of a client which is not attached to a provider. i just want to point that out.

VanishMax (approved):
🎉🎉🎉

PR #1649, opened 2024-08-06
Tweaks to v2 components based on Sam's notes

jessepinho:
Based on a recent convo with Sam, we decided to make some a few tweaks to existing v2 components:

- Button's no longer have to be of `primary` priority by default. There are a number of contexts in which they can be `secondary`. Thus, `priority` is now a prop, rather than a context, to make it easier to set priority. 
- The first button in a button group is no longer required to be `primary`, because there are many cases where all buttons in a group will be of equal priority (e.g., the Staking page, where someone can delegate or undelegate).
  - Instead, button groups now have a `hasPrimaryButton` boolean prop. When set to `true`, the first button in a button group will be marked as `primary`.
- When a button has `iconOnly='adornment'`, there will no longer be a focus outline on the button when it is clicked, since it didn't really look good.
- Use a real `<Button />` for closing a dialog.


packages/ui/src/ButtonGroup/index.tsx
```diff
@@ -88,24 +96,31 @@ export const ButtonGroup = ({
       */}
       {isIconOnly(props) &&
         props.buttons.map((button, index) => (
-          <ButtonPriorityContext.Provider key={index} value={index === 0 ? 'primary' : 'secondary'}>
```
jessepinho, line 91: We got rid of the context, so now we just pass the priority directly to the button.

packages/ui/src/Dialog/index.tsx
```diff
@@ -226,11 +228,13 @@ const Content = <IconOnlyButtonGroupProps extends boolean | undefined>({
...
-                <Icon IconComponent={X} size='md' />
-              </CloseButton>
-            </RadixDialog.Close>
+            <Density compact>
+              <RadixDialog.Close asChild>
+                <Button icon={X} iconOnly priority='secondary'>
+                  Close
+                </Button>
+              </RadixDialog.Close>
+            </Density>
```
jessepinho, line 237: image

VanishMax (approved):
Dialog is 🔥🔥🔥

PR #1652, opened 2024-08-06
feat: ADR-006 implementation

VanishMax:
Implements ADR-006.

This PR has its own, newer version of the ADR, that completely matches the implementation. Based on previously-closed #1567 

To check the implementation, do this:
1. Check out the branch of this PR
2. Use https://github.com/prax-wallet/web/pull/134 PR in Prax while using this PR's packages
3. Use https://github.com/penumbra-zone/nextjs-penumbra-client-example/pull/5 PR in NextJS example to see how this ADR affects real apps


docs/adrs/006-web-apis.md
```diff
@@ -0,0 +1,171 @@
...
+   * If `providerUrl` argument is not provided, tries to connect to the first injected provider.
+   * Returns the manifest URL of the connected provider or throws an error otherwise.
+   */
+  readonly connect: (providerUrl?: string) => Promise<string>;
+
+  /**
+   * Tries to reconnect to the injected provider (the first one with `connected` state)
+   * without asking for user approval
+   */
+  readonly reconnect: (providerUrl?: string) => Promise<string>;
```
turbocrime, line 112: i don't think this is appropriate.

1. any provider may identify itself as connected
2. multiple providers may identify themself as connected
3. a user likely wishes to use a specific provider

implementing any automatic selection behavior in this client api is likely to guide developers to create a negative user experience for a user that wishes to select a provider. so the api should avoid any automatic selection behavior.

requiring the developer to deliberately select a frontend will force developers to anticipate this ambiguity. at that point they can either

- prompt users to select a specific provider
- implement some technique for their dapp to remember
- encode their own priority

you've removed the `PenumbraClient.providers` method, but with that method a developer could easily implement a similar behavior by filtering the returned record by testing `isConnected` on each provider.

docs/adrs/006-web-apis.md
```diff
@@ -0,0 +1,171 @@
...
+
+export type getPenumbraManifest = (
+  penumbraOrigin: string,
+  signal?: AbortSignal,
+) => Promise<PenumbraManifest>;
+
+export type getAllPenumbraManifests = () => Record<
+  keyof (typeof window)[typeof PenumbraSymbol],
+  Promise<PenumbraManifest>
+>;
```
turbocrime, line 151: It seemed to be required by previous discussion that a `PenumbraClient` object should collect methods a developer would be interested in for selecting and then initiating a provider connection

is there a reason these are broken out again?

docs/adrs/006-web-apis.md
```diff
@@ -0,0 +1,171 @@
...
+
+Moreover, the `client` package might export but not limited to the following useful functions and types:
+
+```ts
+export type PenumbraManifest = Partial<chrome.runtime.ManifestV3> &
+  Required<Pick<chrome.runtime.ManifestV3, 'name' | 'version' | 'description' | 'icons'>>;
+
+export type getInjectedProvider = (penumbraOrigin: string) => Promise<PenumbraProvider>;
+
+export type getAllInjectedProviders = () => string[];
```
turbocrime, line 141: i'd like to avoid the term 'injected' because in the future the api may not actually require injection, given possible `externally_connectable` features

at that point to avoid misleading terminology, either the api would require a breaking change to rename, or another set of methods identifying a distinct set of providers that don't actually require a practically distinct api

by avoiding the term 'injected' and referring solely to 'providers' then there's no potential terminology problem in the future.

packages/client/src/client.ts
```diff
@@ -0,0 +1,165 @@
...
+  readonly isConnected: () => boolean | undefined;
+  /** Reexports the `state` function from injected provider */
+  readonly getState: () => PenumbraState;
+  /** Provides a simplified callback interface to `PenumbraStateEvent`s. */
+  readonly onConnectionChange: (cb: (connection: PenumbraStateEventDetail) => void) => void;
+
+  /** Returns a new or re-used `PromiseClient<T>` for a specific `PenumbraService` */
+  readonly service: <T extends PenumbraService>(
+    service: T,
+    options: Omit<ChannelTransportOptions, 'getPort'>,
```
turbocrime, line 40: transport options parameter should be optional - it's only necessary to configure a custom timeout or custom ser/de behavior. a good set of defaults is available in #1648

packages/client/src/client.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return this.port;
+  }
+
+  private assertProvider() {
+    if (!this.provider) {
+      throw new PenumbraProviderNotConnectedError(this.origin);
+    }
+    return this.provider;
+  }
```
turbocrime, line 68: given discussion requesting that the client avoid throwing, these methods were removed in #1648

notably, a `PenumbraClient` instance is useless without a selected provider. #1648 moved the origin parameter into the constructor - is there a good reason to revert that?

docs/adrs/006-web-apis.md
```diff
@@ -0,0 +1,171 @@
...
+import type { PenumbraService } from '@penumbra-zone/protobuf';
+import type { PromiseClient } from '@connectrpc/connect';
+
+interface PenumbraClient {
+  /**
+   * Connects to the injected provider, asks for user approval if wasn't connected before.
+   * If `providerUrl` argument is not provided, tries to connect to the first injected provider.
+   * Returns the manifest URL of the connected provider or throws an error otherwise.
+   */
+  readonly connect: (providerUrl?: string) => Promise<string>;
```
turbocrime, line 106: returning manifest url from the `connect` method seems like the least useful choice.

it seems more better to return the `MessagePort` or a created `Transport`, allowing the caller to handle the `Transport` directly if they wish, as designed in #1648

if the manifest is really desired, just go ahead and provide a parsed manifest - the connection init process must fetch it anyway to confirm it exists.

packages/client/src/client.ts
```diff
@@ -0,0 +1,165 @@
...
+    return this.provider?.isConnected() ?? false;
+  }
+
+  public getState() {
+    return this.assertProvider().state();
+  }
+
+  public onConnectionChange(callback: (detail: PenumbraStateEventDetail) => void) {
+    this.callbacks.add(callback);
+  }
```
turbocrime, line 139: it seems like there's now no way to remove a callback

turbocrime (changes requested):
i think there were some steps backwards here. review is based on reading the changeset

packages/client/src/client.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    await assertProviderManifest(origin);
+
+    this.origin = origin;
+    this.provider = provider;
+    this.port = await provider.connect();
+    this.callbacks.forEach(cb => cb(new PenumbraStateEvent(origin, provider.state(), true).detail));
+
+    return origin;
+  }
```
turbocrime, line 120: given multiple connected providers, which provider is preferred by this method?

docs/adrs/006-web-apis.md
VanishMax: Fair enough. Somehow I thought checking for manifest validity would be enough but I see know it isn't. Especially at the 'multiple providers may identify themself as connected' point.

Probably will delete it for now

turbocrime:
if some of these changes were intended to address state management concerns discussed in DM, i don't think they achieve that objective

docs/adrs/006-web-apis.md
VanishMax: That's true – the provider origin URL is not that useful. The same can be said about `MessagePort` or `Transport` – they are of no value for the library users. It might be better to return `Promise<void>` for now until we find out the  really useful return, to not produce breaking changes in the future

packages/client/src/client.ts
VanishMax: removed the `reconnect`

packages/client/src/client.ts
VanishMax: copied the AbortController usage here

docs/adrs/006-web-apis.md
VanishMax: changed to return `Promise<void>`

packages/client/src/client.ts
VanishMax: good point, removed these assertion methods in favor of non-throwing.

as for the question – i will continue pursuing the client without the necessary `origin` in the constructor. It is not useless – it allows the creation of client in the setup of the users' app and the future reuse, a good practice popularized by many libraries (listed in the ADR as the source of inspiration). This approach is proven to be effective in terms of DX

docs/adrs/006-web-apis.md
VanishMax: removed these functions in favor of the `get.ts` from #1648

docs/adrs/006-web-apis.md
VanishMax: moved useful utility methods into the `client` with `get` prefixes

VanishMax:
closed in favor of #1648

PR #1662, opened 2024-08-08
feat: separate rust unit and wasm-pack tests

### LINKED ISSUE #1661: Re-separate unit and wasm-pack tests and make unit tests run in CI
addresses additional feedback from #1616

vacekj:
fixes #1661


turbo.json
```diff
@@ -41,6 +41,10 @@
     "test:rust": {
       "dependsOn": ["compile"],
       "inputs": ["crate/src/**", "crate/Cargo.toml", "crate/Cargo.lock", "crate/tests/**"]
+    },
+    "test:rust-unit": {
+      "dependsOn": ["compile"],
```
grod220, line 46: todo: these tests should not require compile to run. Compile step is only for those tests that require the wasm-pack wasm binaries to be built.

packages/wasm/package.json
```diff
@@ -20,7 +20,8 @@
     "lint:strict": "tsc --noEmit && eslint src --max-warnings 0",
     "postcompile": "touch ./wasm/.npmignore",
     "test": "vitest run",
-    "test:rust": "cd crate && wasm-pack test --headless --firefox -- --target wasm32-unknown-unknown --release --features 'mock-database'"
+    "test:rust": "cd crate && wasm-pack test --headless --firefox -- --target wasm32-unknown-unknown --release --features 'mock-database'",
+    "test:rust-unit": "cd crate && cargo test --release"
```
grod220, line 24: issue: think rust vs rust-unit isn't the clearest differentiation. What if we did `test:wasm` vs `test:rust` (or `test:cargo`)?

grod220 (approved):
A few comments, feel free to merge after addressing

turbo.json
vacekj: Removed the depends on compile.

packages/wasm/package.json
vacekj: thanks, good idea - went with test:wasm and test:cargo here.

PR #1682, opened 2024-08-10
Create Toggle and AccountSelector components

jessepinho:
Creating more components for the v2 send/receive pages

## In this PR
- Create the v2 `<AccountSelector />` component
    image
- Create a v2 `<Toggle />` component
  - Note that this component is incomplete, as it doesn't yet visually support disabled and focused states. But those are coming soon in mockups.
- Add `disabled` prop to `<CopyToClipboardButton />`.
- Extract a `truncate` CSS mixin.
- Add support for `startAdornment` and `endAdornment` props in `<TextInput />`, as these are needed for `<AccountSelector />`. Also, add `max`/`min` prop support.

### DIFF
--- a/packages/ui/src/AccountSelector/AccountSelectorAddress.tsx
+++ b/packages/ui/src/AccountSelector/AccountSelectorAddress.tsx
@@ -1,0 +1,61 @@
+import { Address } from '@buf/penumbra-zone_penumbra.bufbuild_es/penumbra/core/keys/v1/keys_pb';
+import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
+import styled from 'styled-components';
+import { useDensity } from '../hooks/useDensity';
+import { Density } from '../types/Density';
+import { CopyToClipboardButton } from '../CopyToClipboardButton';
+import { Shrink0 } from '../utils/Shrink0';
+import { technical, truncate } from '../utils/typography';
+
+const Root = styled.div<{ $ephemeral: boolean; $loading: boolean; $density: Density }>`
+  border: 1px solid ${props => props.theme.color.other.tonalStroke};
+  padding: ${props => props.theme.spacing(2)} ${props => props.theme.spacing(3)};
+
+  display: flex;
+  gap: ${props => props.theme.spacing(2)};
+
+  color: ${props =>
+    props.$loading
+      ? props.theme.color.text.muted
+      : props.$ephemeral
+        ? props.theme.color.text.special
+        : props.theme.color.text.primary};
+
+  ${props => props.$density === 'sparse' && 'word-break: break-all;'}
+`;
+
+const TextWrapper = styled.div<{ $density: Density }>`
+  flex-grow: 1;
+
+  ${props => props.$density === 'compact' && truncate}
+  ${technical}
+`;
+
+export interface AccountSelectorAddressProps {
+  address?: Address;
+  ephemeral: boolean;
+  loading: boolean;
+}
+
+export const AccountSelectorAddress = ({
+  address,
+  ephemeral,
+  loading,
+}: AccountSelectorAddressProps) => {
+  const density = useDensity();
+
+  return (
+    <Root $ephemeral={ephemeral} $loading={loading} $density={density}>
+      <TextWrapper $density={density}>
+        {address ? bech32mAddress(address) : `penumbra1...`}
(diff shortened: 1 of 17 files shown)


packages/ui/src/AccountSelector/AccountSelectorAddress.tsx
jessepinho, line 1: image

packages/ui/src/AccountSelector/IbcDepositToggle.tsx
jessepinho, line 1: image

Note that this isn't complete per the designs yet, as it's missing the (i) icon that appears in the Figma mockup (since we don't yet have designs for the dialog that should open when clicking that icon).

packages/ui/src/AccountSelector/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1682-jessepinho-more-v2-c-1lgp5gcw.web.app/?path=/docs/ui-library-accountselector--docs

packages/ui/src/CopyToClipboardButton/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1682-jessepinho-more-v2-c-1lgp5gcw.web.app/?path=/docs/ui-library-copytoclipboardbutton--docs

packages/ui/src/utils/button.ts
```diff
@@ -8,6 +8,7 @@ export const buttonBase = css`
   background: transparent;
   border: none;
   cursor: pointer;
+  padding: 0;
```
jessepinho, line 11: Hadn't realized that buttons have built-in padding by default in Chrome, so I removed that in `buttonBase`.

packages/ui/src/Toggle/index.stories.tsx
jessepinho, line 1: https://penumbra-ui-preview--pr1682-jessepinho-more-v2-c-1lgp5gcw.web.app/?path=/docs/ui-library-toggle--docs

packages/ui/src/AccountSelector/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+          const valueAsNumber = Number(value);
+          const valueLength = value.replace(/^0+/, '').length;
+
+          if (valueAsNumber > MAX_INDEX || valueLength > MAX_INDEX.toString().length) {
+            return;
+          }
+
+          handleChange(valueAsNumber);
+        }}
+        startAdornment={<StartAdornment>Account #</StartAdornment>}
```
VanishMax, line 121: question: maybe it's worth making not `div` but `label` with `cursor: pointer`, so that users will have a larger hover area to select the input? So far it is still not obvious the AddressSelector is the input

packages/ui/src/Toggle/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+export interface ToggleProps {
+  /** An accessibility label. */
+  label: string;
+  value: boolean;
+  onChange: (value: boolean) => void;
+  /** @todo: Implement disabled state visually. */
+  disabled?: boolean;
+}
+
+export const Toggle = ({ label, value, onChange, disabled }: ToggleProps) => {
```
VanishMax, line 50: nitpick (non-blocking, i'd leave it as-is like an easter egg): check how fancy the outline is.

in the `false` state – outline is ok:

image

in the `true` state – outline is funny:

image

VanishMax (approved):

packages/ui/src/Toggle/index.tsx
jessepinho: Crap. I'll look into this

packages/ui/src/AccountSelector/index.tsx
jessepinho: Great idea. I didn't go with your exact implementation, though: `AccountSelector` is a field that will often be nested inside `<FormField />`, which itself renders a `<label />`. Nesting `<label />`s is invalid HTML. So, to make this element usable inside or outside of a `<FormField />`, I've changed `<StartAdornment />` to be an (accessible) clickable div with a pointer cursor: https://github.com/penumbra-zone/web/pull/1682/commits/97f12855d5de48f324feeb87c5fb8c8439f56c0b

PR #1683, opened 2024-08-12
Compress fonts

grod220:
Using woff2 format over ttf saves lots of kbs


packages/ui/styles/globals.css
grod220, line 1: Diffs here due to formatting

packages/ui/styles/globals.css
```diff
@@ -3,149 +3,154 @@
...
+    @font-face {
+        font-family: 'Faktum';
+        font-weight: 600;
+        src: url('../fonts/Faktum-SemiBold.woff2') format('woff2');
+    }
+
+    @font-face {
+        font-family: 'Devanagari Sangam';
+        font-weight: 400;
+        src: url('../fonts/Devanagari-Sangam-Regular.woff2') format('woff2');
```
grod220, line 124: The key here is that ttf fonts are now using the woff2 formats

packages/ui/styles/globals.css
grod220, line 1: For some reason formatter did not like those kinds of comments to the side

VanishMax (approved):
wow, huge reduction of size!

PR #1687, opened 2024-08-13
react hooks

turbocrime:
this contains some sketches for a new react package.

context-based, which there was some interest in moving away from. but
- this is early wip, chill
- context model facilitates scoping components to a specific provider, so a developer could use multiple providers

the `usePenumbraQuery` hook is the interesting part, as it wraps a `PromiseClient` with `useQuery`, so that callers may automatically consume penumbra requests as react queries. streams are consumed with `Array.fromAsync` so it is not suitable for long-streaming requests.

### DIFF
--- a/packages/react/src/components/penumbra-context-provider.tsx
+++ b/packages/react/src/components/penumbra-context-provider.tsx
@@ -1,0 +1,45 @@
+import { penumbraContext } from '../context/penumbra-context.js';
+import { ReactNode, useCallback, useEffect, useMemo, useState } from 'react';
+import {
+  PenumbraContextInput,
+  resolvePenumbraContextInput,
+} from '../context/penumbra-context-input.js';
+import { PenumbraEventDetail } from '@penumbra-zone/client';
+
+export const PenumbraContextProvider = ({
+  client,
+  provider,
+  children,
+}: PenumbraContextInput & { children?: ReactNode }) => {
+  const penumbra = useMemo(
+    () => resolvePenumbraContextInput({ client, provider }),
+    [client, provider],
+  );
+
+  const [providerState, setProviderState] = useState(penumbra.state);
+  const [providerConnected, setProviderConnected] = useState(penumbra.connected);
+
+  useEffect(() => {
+    console.log('setting providerConnected', providerState, penumbra.connected);
+    setProviderConnected(penumbra.connected);
+  }, [penumbra, providerState]);
+
+  const listener = useCallback((update: PenumbraEventDetail<'penumbrastate'>) => {
+    console.log('listener', update);
+    setProviderState(update.state);
+  }, []);
+
+  useEffect(() => {
+    const ac = new AbortController();
+    penumbra.onConnectionStateChange(listener, ac.signal);
+    return () => ac.abort();
+  }, [penumbra, listener]);
+
+  console.log('providerConnected', providerConnected);
+
+  return (
+    <penumbraContext.Provider value={providerConnected ? penumbra : penumbra}>
+      {children}
+    </penumbraContext.Provider>
+  );
+};
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -36,6 +36,13 @@
   ],
   "versionGroups": [
     {
+      "label": "Ignore @tanstack/react-query",
+      "dependencies": [
+        "@tanstack/react-query"
+      ],
+      "isIgnored": true
+    },
+    {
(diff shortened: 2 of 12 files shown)


packages/react/src/components/penumbra-context-provider.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    const ac = new AbortController();
+    penumbra.onConnectionStateChange(listener, ac.signal);
+    return () => ac.abort();
+  }, [penumbra, listener]);
+
+  console.log('providerConnected', providerConnected);
+
+  return (
+    <penumbraContext.Provider value={providerConnected ? penumbra : penumbra}>
```
VanishMax, line 41: should it be like this?

```suggestion
    <penumbraContext.Provider value={penumbra}>
```

VanishMax: probably the value of the context should be a reactive object similar to `PenumbraClient` but without static methods, `service`, `onConnectionStateChange` and others non-needed functions replaced by hooks

VanishMax (commented):
## Right now

If I were to write the examples for this package, I'd write smt like this:

### Wrap the app in the Penumbra context provider

```tsx
import { PenumbraContextProvider } from '@penumbra-zone/react';
import App from './app.tsx';

export const Main = () => {
	return (
		<React.StrictMode>
			<PenumbraContextProvider>
				<App />
			</PenumbraContextProvider>
		</React.StrictMode>
	);
};

const rootElement = document.getElementById('root') as HTMLDivElement;
createRoot(rootElement).render(<Main />);
```

### Get the Penumbra connection state 

```tsx
import { usePenumbra } from '@penumbra-zone/react';

const Component = () => {
	const { connected, connect } = usePenumbra();

	if (!connected) {
		return (
			<button type="button" onClick={connect} />
		);
	}

	return (
		<button type="button" onClick={disconnect} />
	);
};
```

### Request blockchain data by calling the queries

```tsx
import { usePenumbraQuery } from '@penumbra-zone/react';
import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';

const Component = () => {
	const viewService = usePenumbraQuery(ViewService);
	const { data, error, isPending } = viewService.getAddressByIndex({ account: 0 });

	if (isPending) {
		return <p>loading...</p>;
	}

	return <p>You account address is {bech32mAddress(data)}</p>;
}
```

### Or request blockchain data without wrappers 

```ts
import { usePenumbraService } from '@penumbra-zone/react';

const viewService = usePenumbraService(ViewService);

const response = await viewService.getAddressByIndex({ account: 0 });
```

## Proposal

Having the idea of the current implementation state, I think of more ways of using the React package:

1. Firstly, we need hooks to represent static methods of `PenumbraClient`: `getProviders`, `getProviderManifest`, and the rest. It could be, for instance, the `useProviders` hook. It might wrap the provider list into the react query.
2. Queries with simplified namings: `useBalances`, `useAccount`, etc. It will use the `usePenumbraQuery` under the hook but at least the names will be clear.
3. Something to simplify the creation of a transaction. It might also be a naming issue but for me personally it is extremely hard to compose a transaction. This can also go into the `client` package, not only for react.

@turbocrime @grod220 wdyt?

turbocrime:
parts of minifront slices that might be helpful in developing hooks

- assembleTransaction https://github.com/penumbra-zone/web/blob/main/apps/minifront/src/state/send/index.ts#L128-L162

- assembleSwapRequest https://github.com/penumbra-zone/web/blob/74527dfbaf57cde3ecf3141998142def640efec3/apps/minifront/src/state/swap/instant-swap.ts#L173-L206

- issueSwapClaim https://github.com/penumbra-zone/web/blob/74527dfbaf57cde3ecf3141998142def640efec3/apps/minifront/src/state/swap/instant-swap.ts#L211-L249

- assembleUndelegateClaimRequest https://github.com/penumbra-zone/web/blob/74527dfbaf57cde3ecf3141998142def640efec3/apps/minifront/src/state/staking/assemble-undelegate-claim-request.ts#L38-L58

- ibc getPlanRequest https://github.com/penumbra-zone/web/blob/74527dfbaf57cde3ecf3141998142def640efec3/apps/minifront/src/state/ibc-out.ts#L192-L236

there's plenty more, these are just some suggestions. anything in the minifront slices that assembles a plan request or a transaction request could become a provided querier.

TalDerei:
@grod220 should we close these QOL improvements as not planned?

Issue #1666, opened 2024-08-09
Flag incompatible environments


grod220:
Users are frequently using minifront + prax in environments we do not support. This is causing support questions for an usual class of bugs. 

**Todo:** Add a banner (similar to the [testnet banner](https://github.com/penumbra-zone/web/blob/8fc6c85a184c5d7d95c2372c703ece780ce8f440/packages/ui/components/ui/testnet-banner/index.tsx#L4)) that alerts users when they are using an incompatible environment (and that we cannot guarantee it works in this env).

Should detect:
- Browser type (should be chrome)
- Chrome version (should be at minimum what's in the [manifest](https://github.com/prax-wallet/web/blob/main/apps/extension/public/manifest.json#L7))
- Device type (should be desktop)

Note: A number of users are using [Mises](https://www.mises.site/), we should test this out to ensure the banner triggers in this environment as well.

PR #1693, opened 2024-08-14
Add Banner and IncompatableBrowserBanner

JasonMHasperhoven:
Addresses #1666 

**Changes**
- Added generic Banner component
- Added IncompatableBrowserBanner component to splash screen and layout (header component)
- Fixed splash screen UI on mobile
- Fixed buttons not to shrink

Before:
image

After:
image

### DIFF
--- a/packages/ui/components/ui/incompatable-browser-banner/get-compatability.ts
+++ b/packages/ui/components/ui/incompatable-browser-banner/get-compatability.ts
@@ -1,0 +1,62 @@
+import { UAParser, UAParserInstance, IDevice } from 'ua-parser-js';
+
+const parser: UAParserInstance = new UAParser();
+
+const CompatableBrowsers: Record<string, number> = {
+  // browser name: min version
+  Chrome: 119,
+};
+
+interface CompatabilityResponse {
+  isIncompatable: boolean;
+  title: string;
+  content: string;
+}
+
+// from: https://github.com/faisalman/ua-parser-js/issues/182
+function isDesktop(device: IDevice): boolean {
+  return device.type === undefined || !['wearable', 'mobile'].includes(device.type);
+}
+
+function getCompatability(): CompatabilityResponse {
+  const browser = parser.getBrowser();
+  const device = parser.getDevice();
+
+  if (!CompatableBrowsers[browser.name!]) {
+    return {
+      isIncompatable: true,
+      title: 'Incompatable Browser Detected',
+      content:
+        'You are currently using an incompatible browser. For the best experience, we recommend using Chrome as your browser.',
+    };
+  }
+
+  if (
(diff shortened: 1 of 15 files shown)


TalDerei:
can you add a [changeset](https://github.com/penumbra-zone/web/blob/main/docs/documenting-changes.md)?

packages/ui/components/ui/testnet-banner/index.tsx
```diff
@@ -1,12 +1,13 @@
+import { Banner } from '../banner';
+
 export const TestnetBanner = ({ chainId }: { chainId?: string }) =>
```
TalDerei, line 3: think we can remove `TestnetBanner`

packages/ui/components/ui/incompatable-browser-banner/get-compatability.ts
```diff
@@ -0,0 +1,62 @@
+import { UAParser, UAParserInstance, IDevice } from 'ua-parser-js';
+
+const parser: UAParserInstance = new UAParser();
+
+const CompatableBrowsers: Record<string, number> = {
+  // browser name: min version
+  Chrome: 119,
```
TalDerei, line 7: Is there a specific reason why Chrome 119 was chosen as the minimum version?

TalDerei (approved):

packages/ui/components/ui/testnet-banner/index.tsx
JasonMHasperhoven: @grod220 confirm please

packages/ui/components/ui/incompatable-browser-banner/get-compatability.ts
grod220: Can we add a code comment pointing to that json?

packages/ui/components/ui/testnet-banner/index.tsx
grod220: Nah, let's keep this for now. Think it remains important to somehow distinguish visually mainnet vs testnet.

packages/ui/components/ui/incompatable-browser-banner/get-compatability.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+interface CompatabilityResponse {
+  isIncompatable: boolean;
+  title: string;
+  content: string;
+}
```
grod220, line 14: nonblocking: wonder if the type should be something like a rust enum:

```ts
type CompatabilityResponse = { isIncompatable: true } | { isIncompatable: false, title: string, content: string }
```

chairulakmal:
Is IncompatableBrowserBanner is also intended for Brave browser on desktop? I have been using Prax on Brave desktop for a while without any problem.

Issue #1660, opened 2024-08-08
Re-enable spend max feature 


TalDerei:
### **Context**

https://github.com/penumbra-zone/web/pull/1612 and https://github.com/penumbra-zone/web/pull/1611 resulted in a feature regression that should be re-enabled and properly implemented. Below is a high-level product plan for achieving this.

###  **Updates required to enable send max**
- check user flows
   - test a set of user flows associated with the feature to ensure all use cases are covered
- enable the feature in the transaction planner request within minifront
- fix the feature in the wasm planner
   - ensure that the remaining balance from the fee note is returned to the originating address of the transaction
   
cc @grod220

TalDerei:
Currently, we leverage our multi-fee system to support maximum sends in certain scenarios. Consider the following cases:

----------------------------------------------------------------------------------------------------------------------------
## **Active Functionality**:

These are constructed using an `output` transaction planner request.

**1. Sending an alternative asset (GM) while paying fees with the native asset (UM).**

<img width="531" alt="Screenshot 2024-08-14 at 8 47 55 AM" src="https://github.com/user-attachments/assets/1dcd2729-2a7b-47ff-9b81-c5a81009f272">
<img width="1728" alt="Screenshot 2024-08-14 at 8 48 38 AM" src="https://github.com/user-attachments/assets/3531e058-8626-4842-ab2a-baacee8da8d1">

**2. Sending an alternative asset (Pizza) while paying fees with another alternative asset (GN).**

<img width="1728" alt="Screenshot 2024-08-14 at 8 48 19 AM" src="https://github.com/user-attachments/assets/34c80882-525f-479c-b520-35b54f6a3594">

----------------------------------------------------------------------------------------------------------------------------
## **Inactive Functionality**:

**3. Sending the maximum amount of the native asset (UM).**
**4. Sending the maximum amount of an alternative asset (GM) without any native asset (UM) balance, and paying the fee with the same alternative asset (GM).**

----------------------------------------------------------------------------------------------------------------------------
## **Proposals**:

Invariants (3) and (4) fundamentally both rely on the presence of UM. In the transaction planner service implementation, we need to handle these cases by checking the following conditions:

- If the user sends the maximum amount of the native asset (UM):
  - craft a spend action.
- If the user sends max alternative asset (GM), and UM is _absent_:
  - craft a spend action.
- If the user sends the maximum amount of an alternative asset (GM) and the native asset (UM) is _not_ absent:
[...]

PR #1698, opened 2024-08-15
 re-enable spend max feature

TalDerei:
references https://github.com/penumbra-zone/web/issues/1660 and depends on https://github.com/penumbra-zone/web/pull/1801.

to address the feature regression, we implement additional safety checks in both the frontend and services package. 

1. The frontend uses [checkSendMaxInvariants](https://github.com/penumbra-zone/web/pull/1698/files#diff-f91054ffb202d0abdebea8130d45bc1e9b6af6dc75f0a893b6b0ebbab8670332R73-R101) to determine if certain invariants are satisfied and decide whether to construct the transaction as a `Spend` or `Output` transaction planner request. This determines the appropriate planner request that the wasm planner will consume to craft the fully-formed transaction.

2. In the transaction planner service implementation, we've added additional constraint checks in [assertSpendMax](https://github.com/penumbra-zone/web/pull/1698/files#diff-934196153c1f43ccb98be7f6051a2ff95f7ae80f9ed4e66f3f7cd775c634d05cR6) on the fully-formed transaction. However, these checks occur only after the planning step because, with the current code, there's no effective way to enforce these constraints directly in the wasm planner. This provides a greater degree of validation and safety.

In thinking about this more, while these approaches increase safety, they don't completely cover the _entire_ domain space. There is still an external risk to third-party consumers of our wasm package since we don't enforce these checks directly in the wasm planner. To address this, the final step is to implement the same checks in `assertSpendMax` directly within the wasm planner. This requires public getters to access the private fields inside the [ActionList](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/action_list.rs#L21) struct. This is currently implemented in this [PR](https://github.com/penumbra-zone/penumbra/pull/4856) but is blocked until a minor point release is cut on the protocol side. 

**Update –** separated the wasm code into https://github.com/penumbra-zone/web/pull/1801 and added additional safety checks directly within the wasm planner, along with wasm-bindgen tests.


TalDerei:
enumerating various test cases following https://github.com/penumbra-zone/web/issues/1660#issuecomment-2289254108. 

If both `invariantOne` and `invariantTwo` are false, `isSendingMax` will be false, crafting an `output` TPR. However, if either of the invariants is satisfied, a `spend` TPR will be crafted.

**1. User sends the maximum amount of an alternative fee asset (GM), and pays fees with the native asset (UM).**

Screenshot 2024-08-14 at 8 47 55 AM

Screenshot 2024-08-14 at 10 03 41 PM

```
invariantOne:  false
invariantTwo:  false
isSendingMax:  false
```

**2. User sends the maximum amount of an another asset (Pizza), and pays fees with an alternative fee asset (GN).**

Screenshot 2024-08-14 at 10 09 56 PM

```
invariantOne:  false
invariantTwo:  false
isSendingMax:  false
```

**3. User sends the maximum amount of the native asset (UM).**

Screenshot 2024-08-14 at 10 11 26 PM

```
invariantOne:  true
invariantTwo:  false
isSendingMax:  true
```

**4. User sends the maximum amount of alternative asset (GM), and pays fees with the same alternative asset (GM) since the native token (UM) is absent.**

Screenshot 2024-08-14 at 10 15 36 PM

```
invariantOne:  false
invariantTwo:  true
isSendingMax:  true
```

TalDerei:
@Valentine1898 optimistically requested review. I'll implement more comprehensive unit testing and incorporate the additional [constraint](https://github.com/penumbra-zone/web/issues/1660#issuecomment-2289395481) checks.

apps/minifront/src/components/shared/non-native-fee-warning.tsx
```diff
@@ -38,6 +40,9 @@ const hasTokenBalance = ({
       getAddressIndex.optional()(asset)?.account === account,
   );
 
+  // Set the staking token status in the state
+  setStakingToken(hasStakingToken);
```
VanishMax, line 44: issue: the `hasTokenBalance` function is not meant to modify the state. Apart from that, it is hard for me to follow the flow of data from `gasPrices` and `stakingToken` that you added to the state.

suggestion: instead of adding new values to the send and swap Zustand states, move it to the shared state `src/state/shared.ts`. It exists for this very case, when the data from some query is shared between pages. 

You can copy the `assets` zQuery slice and modify it a little to request the `gasPrices`. This way you will eliminate the need to request and set the state in a purely informative component (non-native-fee-warning). As for the `hasStakingToken`, it could just be a hook based on the shared `gasPrices` and `balancesResponses` states

VanishMax (changes requested):

apps/minifront/src/components/shared/non-native-fee-warning.tsx
TalDerei: great suggestion, I overlooked the shared state

TalDerei:
**update**

this has taken a backseat to more pressing sync work, but reprioritizing this by including  unit tests that have been sitting in a local branch for some time.

TalDerei:
@penumbra-zone/web-team please test with sample transactions for extra assurance folks!

packages/wasm/crate/src/planner.rs
grod220, line 1: issue: Can we split the wasm code off into its own PR? Also, definitely think we need tests for this. See `test_planner_delegator_vote.rs` as a reference for passing in a mock db.

TalDerei: good idea, will open separate PR

TalDerei: https://github.com/penumbra-zone/web/pull/1801

TalDerei:
@grod220 @VanishMax please review, but hold off on merging since I’m doing some final testing.

apps/minifront/src/components/send/send-form/index.tsx
```diff
@@ -31,11 +35,42 @@ export const SendForm = () => {
...
+    updateStakingTokenAndGasPrices();
+  }, [
+    transferableBalancesResponses,
+    stakingTokenMetadata,
+    selection,
+    gasPrices,
+    isStakingToken,
+    setGasPrices,
+    setStakingToken,
+  ]);
```
grod220, line 72: issue: It appears this logic is not critical for the display of this component. This looks to be only used in `assembleRequest()` right? Think you should move it entirely out of react-land and put it there so you don't need to worry about the hooks stuff.

apps/minifront/src/components/send/send-form/index.tsx
```diff
@@ -31,11 +35,42 @@ export const SendForm = () => {
...
   } = useStore(sendSelector);
 
   useRefreshFee();
 
+  // Determine if the selected token is the staking token based on the current balances and metadata
+  const isStakingToken = hasStakingToken(
+    transferableBalancesResponses?.data,
+    stakingTokenMetadata.data,
+    selection,
+  );
```
grod220, line 50: issue: you should convert this to a zustand selector (maybe living in the slice?). That way, you'll get memoization.

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
```diff
@@ -74,6 +75,15 @@ export const TokenSwapInput = () => {
...
   };
 
+  const stakingTokenMetadata = useStakingTokenMetadata();
+
+  // Determine if the selected token is the staking token based on the current balances and metadata
+  const isStakingToken = hasStakingToken(
+    balancesResponses?.data,
+    stakingTokenMetadata.data,
+    assetIn,
+  );
```
grod220, line 85: Looks like the selector can be re-used here too

apps/minifront/src/state/send/helpers.test.ts
grod220, line 1: Nice tests!

apps/minifront/src/state/send/index.ts
```diff
@@ -122,10 +132,28 @@ export const createSendSlice = (): SliceCreator<SendSlice> => (set, get) => {
...
+  feeTier,
+  recipient,
+  selection,
+  memo,
+  gasPrices,
+  stakingToken,
+}: SendSlice) => {
   const spendOrOutput:
     | PartialMessage<TransactionPlannerRequest_Spend>
     | PartialMessage<TransactionPlannerRequest_Output> = {
```
grod220, line 159: suggestion: These lines seem to be the source of some funky typing downstream. Would recommend simplifying this to:

```ts
interface SpendOrOutput { 
  address: Address; 
  value: Value 
}

  const spendOrOutput: SpendOrOutput = {
    address: new Address({ altBech32m: recipient }),
    value: new Value({
      amount: toBaseUnit(
        BigNumber(amount),
        getDisplayDenomExponentFromValueView.optional(selection?.balanceView),
      ),
    }),
  };
```

and propagating to all things downstream

apps/minifront/src/state/shared.ts
```diff
@@ -88,16 +90,31 @@ export const { assets, useAssets } = createZQuery({
...
+  fetch: getGasPrices,
+  getUseStore: () => useStore,
+  get: state => state.shared.gasPrices,
+  set: setter => {
+    const newState = setter(useStore.getState().shared.gasPrices);
+    useStore.setState(state => {
+      state.shared.gasPrices = newState;
+    });
+  },
+});
```
grod220, line 104: thought: If this fetch is moved to `assembleRequest()` can delete

packages/services/src/view-service/transaction-planner/index.ts
```diff
@@ -7,8 +7,14 @@ import { TransactionPlannerRequest } from '@penumbra-zone/protobuf/penumbra/view
 import { fvkCtx } from '../../ctx/full-viewing-key.js';
 import { extractAltFee } from '../fees.js';
 import { assertTransactionSource } from './assert-transaction-source.js';
+import { TransactionPlan } from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
+import { assertSpendMax } from './assert-max-spends.js';
+// import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
```
grod220, line 12: nit: can delete

grod220:
After Prax v14 comes out, we'll likely need to hold off on merging this for like 2 weeks

VanishMax (approved):
Approving to remove the "request changes" label from this PR

apps/minifront/src/components/send/send-form/index.tsx
TalDerei: I specifically wanted to _avoid_ memoization here because **(1)** it wasn’t updating correctly due to edge cases where the `transferableBalancesResponses` and `selection` states remained unchanged across pages, causing the `isStakingToken` boolean to not properly update. Additionally, `hasStakingToken` is a lightweight operation, and I think it’s safer for each page to recalculate this boolean independently.

apps/minifront/src/components/send/send-form/index.tsx
TalDerei: modified the useEffect in the react component to 

```
  useEffect(() => {
    const updateStakingToken = () => {
      // Update the zustand store and local state
      setStakingToken(isStakingToken);
    };

    updateStakingToken();
  }, [
    transferableBalancesResponses,
    stakingTokenMetadata,
    selection,
    isStakingToken,
    setStakingToken,
  ]);
```

since `isStakingToken` is relevant for the `NonNativeFeeWarning`, it makes sense to set the state if we’re already computing it. Moving this logic into `assembleRequest` would require access to `hasStakingToken()`, which in turn would necessitate creating a dedicated Zustand slice. As noted below, that approach wasn’t working effectively.

apps/minifront/src/components/swap/swap-form/token-swap-input.tsx
TalDerei: the above comment also applies here

apps/minifront/src/state/send/index.ts
TalDerei: modified type signature

apps/minifront/src/state/shared.ts
TalDerei: kept this and moved `await getGasPrices()` fetch into async function in [createSendSlice](https://github.com/penumbra-zone/web/pull/1698/files#diff-b771e10f5243fc1cb84e351a8420cc7585eef71a057e4cf373871a8e59ec435aR52).

packages/services/src/view-service/transaction-planner/index.ts
TalDerei: deleted

packages/storage/src/indexed-db/index.ts
```diff
@@ -956,4 +958,45 @@ export class IndexedDb implements IndexedDbInterface {
...
     });
   }
+
+  async accumulateNoteBalance(accountIndex: number, assetId: AssetId): Promise<Amount> {
+    assertAssetId(assetId);
+    const spendableNotes = await this.db.getAllFromIndex(
+      'SPENDABLE_NOTES',
+      'assetId',
+      uint8ArrayToBase64(assetId.inner),
+    );
```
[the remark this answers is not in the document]
TalDerei: these seems to be equivalent, but the documentation suggests you need to additionally iterate over the cursor and collect the records into an array.

packages/storage/src/indexed-db/index.ts
```diff
@@ -956,4 +958,45 @@ export class IndexedDb implements IndexedDbInterface {
       );
     });
   }
+
+  async accumulateNoteBalance(accountIndex: number, assetId: AssetId): Promise<Amount> {
```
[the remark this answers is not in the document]
TalDerei: changed to `totalNoteBalance`

packages/storage/src/indexed-db/index.ts
```diff
@@ -956,4 +958,45 @@ export class IndexedDb implements IndexedDbInterface {
...
+        const newAmount = addLoHi(
+          { lo: accumulator.lo, hi: accumulator.hi },
+          { lo: BigInt(noteAmount.lo), hi: BigInt(noteAmount.hi) },
+        );
+
+        accumulator = new Amount(newAmount);
+      }
+    });
+
+    return accumulator;
```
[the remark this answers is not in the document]
TalDerei: nice simplification, changed

TalDerei:
@grod220 testing the latest round of changes based on your suggestions, please hold off on merging for now.

TalDerei:
@grod220 the screenshot is from one of henry's sample transactions. Could you replicate this and test send max for an "IBC Deposit Address"? We should also perform sample transactions with OSMO and ATOM, as I currently don't have a balance in either. 

Once that's done, we can release this PR to the beta testers. You previously suggested setting up a permissioned beta link with a password and directing your fork of the minifront deployer to this branch?

Actions

apps/minifront/src/components/shared/non-native-fee-warning.tsx
```diff
@@ -119,19 +32,10 @@ export const NonNativeFeeWarning = ({
...
-  const shouldRender =
-    !!amount &&
-    hasTokenBalance({
-      source,
-      balancesResponses,
-      gasPrices,
-      stakingAssetMetadata: stakingTokenMetadata.data,
-    });
-
+  const shouldRender = !!amount && !hasStakingToken;
```
grod220, line 38: This means the notice will show also when using an account with neither UM or an alt gas fee token. That feels like it would add confusion right? 

Screenshot 2024-10-22 at 12 08 34 PM

grod220 (commented):
My suggestions: https://github.com/penumbra-zone/web/pull/1867

happy to pair 🍐 on this!

apps/minifront/src/components/shared/non-native-fee-warning.tsx
TalDerei: I see your point, are you suggesting that we display the privacy warning only for accounts without UM, but with an active alternative gas token balance?

PR #1704, opened 2024-08-16
feat: DropdownMenu

VanishMax:
Closes https://github.com/prax-wallet/web/issues/160

Implement not exactly `MenuItem` but `DropdownMenu` because the `MenuItem` from Figma designs is meant to be used within the dropdown. Radix-ui has great support for these component with a11y enhancement.

The `MenuItem` from Figma is used for `DropdownMenu.Item`, `DropdownMenu.RadioItem`, and `DropdownMenu.CheckboxItem` components.

Design – https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=778-49265&m=dev

Storybook – https://penumbra-ui-preview--pr1704-feat-prax-160-menu-mvj0lz31.web.app/?path=/docs/ui-library-dropdownmenu--docs


packages/ui/src/Button/index.tsx
```diff
@@ -200,6 +201,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
             ? theme.borderRadius.sm
             : theme.borderRadius.full
         }
+        {...props}
```
VanishMax, line 204: In Radix, `asChild` prop will only work if the nested component passes the rest props. 

It doesn't mean we allow users to pass whatever they want - at least TypeScript disagrees with it

packages/ui/src/Popover/index.tsx
```diff
@@ -122,19 +94,10 @@ export const Popover = ({ children, onClose, isOpen }: PopoverProps) => {
...
   children: ReactNode;
-  /**
-   * Change the default rendered element for the one passed as a child, merging
-   * their props and behavior.
-   *
-   * Uses Radix UI's `asChild` prop under the hood.
-   *
-   * @see https://www.radix-ui.com/primitives/docs/guides/composition
-   */
-  asChild?: boolean;
```
VanishMax, line 133: Removed `asChild` prop from the `Popover.Trigger` and `DropdownMenu.Trigger`. Since we do not users to change the component styles, it makes to sense to allow `asChild={false}` – it will always be an unstyled HTML-element

packages/ui/src/DropdownMenu/root.tsx
```diff
@@ -0,0 +1,90 @@
...
+    </RadixDropdownMenuRoot>
+  );
+};
+
+DropdownMenu.Trigger = Trigger;
+DropdownMenu.Content = Content;
+DropdownMenu.RadioGroup = RadioGroup;
+DropdownMenu.RadioItem = RadioItem;
+DropdownMenu.CheckboxItem = CheckboxItem;
+DropdownMenu.Item = Item;
```
VanishMax, line 90: These components represent the `MenuItem` from the design

packages/ui/src/DropdownMenu/shared.ts
```diff
@@ -0,0 +1,63 @@
...
+  color: ${props => getColorByActionType(props.theme, props.$actionType)};
+  cursor: pointer;
+
+  background-color: transparent;
+  transition: background-color 0.15s;
+
+  &:focus:not(:disabled) {
+    background-color: ${props => props.theme.color.action.hoverOverlay};
+    outline: 2px solid ${props => getOutlineColorByActionType(props.theme, props.$actionType)};
+  }
```
VanishMax, line 53: the styles do not have `:hover` and we can't change it – Radix does not distinguish between hover and focus for the menu items. It means that hovering an item also makes it focused to not confuse the states: https://github.com/radix-ui/primitives/issues/956#issuecomment-961819460

packages/ui/src/Button/index.tsx
jessepinho: Ah! Great solution. I was kinda bothered by the fact that I wasn't previously spreading remaining props when Radix wants us to, but I didn't want to change it for TypeScript reasons.

request: could you add a note documenting this? so that future maintainers don't think that there should be add'l props in TypeScript?

packages/ui/src/utils/popover.ts
```diff
@@ -0,0 +1,29 @@
...
+  width: 240px;
+  max-width: 320px;
+  padding: ${props => props.theme.spacing(3)} ${props => props.theme.spacing(2)};
+
+  background: ${props => props.theme.color.other.dialogBackground};
+  border: 1px solid ${props => props.theme.color.other.tonalStroke};
+  border-radius: ${props => props.theme.borderRadius.sm};
+  backdrop-filter: blur(${props => props.theme.blur.lg});
+
+  transform-origin: var(--radix-tooltip-content-transform-origin);
```
jessepinho, line 27: Ah, just noticed this variable name. It should be `--radix-popover-content-transform-origin` ([docs](https://www.radix-ui.com/primitives/docs/components/popover#content)). I must have missed this when reviewing the popover PR.

Since Radix's dropdown menu also has a CSS variable (`--radix-dropdown-menu-content-transform-origin` — [docs](https://www.radix-ui.com/primitives/docs/components/dropdown-menu#content)), the CSS variable name should probably be passed in as a parameter.

packages/ui/src/Popover/index.tsx
jessepinho: OK. I could see there being a use case where users want to just have plain text as the trigger. But maybe that's not even a good idea. So, no strong opinion here.

packages/ui/src/DropdownMenu/checkbox-item.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Text } from '../Text';
+import { DropdownMenuItemBase, MenuItem } from './shared.ts';
+
+export interface DropdownMenuCheckboxItemProps extends DropdownMenuItemBase {
+  children?: ReactNode;
+  checked?: boolean;
+  onChange?: (value: boolean) => void;
+}
+
+export const CheckboxItem = ({
```
jessepinho, line 17: suggestion: keep the dropdown menu open when selecting checkbox items

I couldn't figure out a way to do this in the docs, nor do I see in the code where you're closing the dropdown menu once a user clicks a menu item (maybe Radix does it automatically?). But it'd be nice to keep the dropdown menu open when clicking checkbox items specifically, since the user may wish to click more than one.

Not a blocker for this PR, though !

jessepinho (approved):
Really nice work 👏🏻 I feel confident that Penumbra UI is in good hands :D 

A few comments here and there, but nothing major. Feel free to merge after those are addressed.

TalDerei:
> Really nice work 👏🏻 I feel confident that Penumbra UI is in good hands :D

one last PR for the books 📚

packages/ui/src/DropdownMenu/checkbox-item.tsx
VanishMax: We should probably investigate the use cases with time. Leaving as-is for now

PR #1706, opened 2024-08-18
sync: add bundled genesis compact block to block processor 

TalDerei:
references https://github.com/prax-wallet/web/issues/161 and https://github.com/prax-wallet/web/pull/163

most of the changes in the diff are formatting, related to moving the core logic for processing a compact block into a dedicated [processBlock](https://github.com/penumbra-zone/web/pull/1706/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daR186) method.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -41,10 +41,11 @@ import { processActionDutchAuctionWithdraw } from './helpers/process-action-dutc
 import { RootQuerier } from './root-querier.js';
 import { IdentityKey } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import { getDelegationTokenMetadata } from '@penumbra-zone/wasm/stake';
-import { toPlainMessage } from '@bufbuild/protobuf';
+import { JsonObject, toPlainMessage } from '@bufbuild/protobuf';
 import { getAssetIdFromGasPrices } from '@penumbra-zone/getters/compact-block';
 import { getSpendableNoteRecordCommitment } from '@penumbra-zone/getters/spendable-note-record';
 import { getSwapRecordCommitment } from '@penumbra-zone/getters/swap-record';
+import { CompactBlock } from '@penumbra-zone/protobuf/penumbra/core/component/compact_block/v1/compact_block_pb';
 
 declare global {
   // eslint-disable-next-line no-var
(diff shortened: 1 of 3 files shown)


packages/query/src/block-processor.ts
```diff
@@ -100,8 +101,8 @@ export class BlockProcessor implements BlockProcessorInterface {
 
   // If sync() is called multiple times concurrently, they'll all wait for
   // the same promise rather than each starting their own sync process.
-  public sync = (): Promise<void> =>
-    (this.syncPromise ??= backOff(() => this.syncAndStore(), {
+  public sync = (genesisBlock?: JsonObject): Promise<void> =>
```
grod220, line 104: Think it would be easier for the consumer to pass this into the class instance: https://github.com/prax-wallet/web/pull/165/files#diff-3483cabb881c2f2296418077d1f80daeadfb36525092b964211b678724314192R110-R111 and have it as a private property

TalDerei: this is a great suggestion; it's much cleaner to pass this into the class instance

packages/query/src/block-processor.ts
```diff
@@ -125,7 +136,7 @@ export class BlockProcessor implements BlockProcessorInterface {
     this.numeraires = numeraires;
   }
 
-  private async syncAndStore() {
+  private async syncAndStore(genesisBlock?: CompactBlock) {
```
TalDerei, line 139: should we make type signature here more explicit, `genesisBlock: CompactBlock | undefined`?

grod220: Given it is a class field we don't need to pass it here. We can just reference it below:

```diff
- if (currentHeight === -1n && genesisBlock) {
+ if (currentHeight === -1n && this.genesisBlock) {
```

packages/query/src/block-processor.ts
```diff
@@ -169,171 +187,178 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      );
+    }
+
+    // We only query Tendermint for the latest known block height once, when
+    // the block processor starts running. Once we're caught up, though, the
+    // chain will of course continue adding blocks, and we'll keep processing
+    // them. So, we need to update `latestKnownBlockHeight` once we've passed
+    // it.
+    if (compactBlock.height > latestKnownBlockHeight) {
+      latestKnownBlockHeight = compactBlock.height;
```
grod220, line 349: This should move to the loop within `syncAndStore()` so we don't need to override the input

TalDerei: we'll have to move 

```
      if (isLastBlockOfEpoch) {
        await this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
      }
```

since handling the epoch transition also relies on `latestKnownBlockHeight`

packages/query/src/block-processor.ts
```diff
@@ -329,10 +202,160 @@ export class BlockProcessor implements BlockProcessorInterface {
       if (isLastBlockOfEpoch) {
         await this.handleEpochTransition(compactBlock.height, latestKnownBlockHeight);
       }
```
grod220, line 204: This can remain in `processBlock()`

grod220 (approved):
We should do a lot of assurance testing for this. Added one more comment, but other than that, everything looks good to me.

Issue #1278, opened 2024-06-11
Quick sync for freshly generated wallets


turbocrime:
If a user onboards prax with a passphrase, they may have existing activity on chain.

But if a user generates a new wallet in onboarding, there will be no activity on the chain, so the block processor can skip trial decryption until the present height.

This would greatly accelerate initial sync for many new users.

hdevalence:
this is a great idea and would be a big ux boost. we'd need to figure out what data is necessary to store -- one thing that comes to mind is the SCT state, which would ideally be initialized with a bare root (as if the user had synced and then forgotten all internal nodes). maybe @plaidfinch would know if there's a way to initialize an SCT with just the root like that.

Valentine1898:
Theoretically, it is even possible not to scan with skip trial decryption, but to use a snapshot of view service data for the last epoch (if there will be a service that will provide these snapshots)

TalDerei:
some relevant helper functions: 

querying the latest block height
`const latestHeight = await this.querier.tendermint.latestBlockHeight();`

fetching merkle root associated with the latest height
`let root = await this.querier.cnidarium.fetchRemoteRoot(latestHeight!);`

fetching epoch index from compact block associated with latest height
`let epochIndex = this.querier.compactBlock.compactBlockRange(latestHeight).epochIndex`

TalDerei:
> The TCT cannot be initialized just as a bare root, or else no updates could be performed to it. It always needs to have a frontier. The smallest TCT is the one when an epoch has just ended, which contains only a path down from the root to an epoch root.

would we need to implement https://github.com/penumbra-zone/penumbra/issues/1082 and https://github.com/penumbra-zone/penumbra/issues/1131?

turbocrime:
more specifically

```patch
diff --git a/packages/wasm/crate/src/view_server.rs b/packages/wasm/crate/src/view_server.rs
index af5240690..ac1b83be3 100644
--- a/packages/wasm/crate/src/view_server.rs
+++ b/packages/wasm/crate/src/view_server.rs
@@ -111,7 +111,11 @@ impl ViewServer {
     /// Use `flush_updates()` to get the scan results
     /// Returns: `bool`
     #[wasm_bindgen]
-    pub async fn scan_block(&mut self, compact_block: &[u8]) -> WasmResult<bool> {
+    pub async fn scan_block(
+        &mut self,
+        compact_block: &[u8],
+        skip_trial_decrypt: bool,
[...]

PR #1707, opened 2024-08-19
sync: fresh and existing wallets skip trial decryption

TalDerei:
references https://github.com/penumbra-zone/web/issues/1278 and related to https://github.com/prax-wallet/web/pull/164


packages/query/src/block-processor.ts
```diff
@@ -189,7 +193,13 @@ export class BlockProcessor implements BlockProcessorInterface {
         throw new Error(`Unexpected block height: ${compactBlock.height} at ${currentHeight}`);
       }
 
-      await this.processBlock(compactBlock, latestKnownBlockHeight);
+      // Set the skip_trial_decrypt flag
+      const skipTrialDecrypt = Boolean(
+        this.walletCreationBlockHeight && currentHeight < BigInt(this.walletCreationBlockHeight),
```
TalDerei, line 198: skips trial decryption for fresh wallets until wallet creation height

TalDerei: Screenshot 2024-08-28 at 8 03 02 PM 2

packages/wasm/crate/src/view_server.rs
grod220, line 1: todo: think we should try to get a few tests to ensure this works. Though, it would be the first of its kind for the view server wasm code. But I think it's a good thing.

packages/query/src/block-processor.ts
```diff
@@ -203,7 +212,11 @@ export class BlockProcessor implements BlockProcessorInterface {
   }
 
   // logic for processing a compact block
-  private async processBlock(compactBlock: CompactBlock, latestKnownBlockHeight: bigint) {
+  private async processBlock(
+    compactBlock: CompactBlock,
+    latestKnownBlockHeight: bigint,
+    skipTrialDecrypt: boolean,
+  ) {
```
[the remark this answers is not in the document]
TalDerei: great idea, added `ProcessBlockParams` interface

packages/query/src/block-processor.ts
TalDerei: If `walletCreationBlockHeight` is set to zero, the wallet will need to sync the entire chain and trial decrypt every compact block. This occurs when the user does not specify a wallet birthday height during the import of an existing seed phrase. From manual testing, this behavior works correctly when the field is set to zero, but I'll still write some unit tests for completeness.

TalDerei: Screenshot 2024-08-30 at 10 30 59 PM

packages/query/src/block-processor.ts
```diff
@@ -92,6 +93,7 @@ export class BlockProcessor implements BlockProcessorInterface {
   private readonly stakingAssetId: AssetId;
   private syncPromise: Promise<void> | undefined;
   private genesisBlock: CompactBlock | undefined;
+  private walletCreationBlockHeight: number;
```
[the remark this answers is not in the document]
TalDerei: updated type signatures to use `bigint` and converted `genesisBlock` and `walletCreationBlockHeight` to readonly properties, ensuring they are immutability.

packages/query/src/block-processor.ts
```diff
@@ -225,11 +238,12 @@ export class BlockProcessor implements BlockProcessorInterface {
       }
     }
 
+    // TODO: add logic for determining when to skip trial decryption
```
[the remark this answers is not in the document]
TalDerei: stale comment, removed

packages/types/src/block-processor.ts
```diff
@@ -1,7 +1,7 @@
 import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
 
 export interface BlockProcessorInterface {
-  sync(): Promise<void>;
+  sync(isFreshWallet?: boolean, walletCreationBlockHeight?: number): Promise<void>;
```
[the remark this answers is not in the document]
TalDerei: removed, instead passing as class instances

packages/wasm/crate/src/view_server.rs
```diff
@@ -113,7 +113,11 @@ impl ViewServer {
     /// Use `flush_updates()` to get the scan results
     /// Returns: `bool`
     #[wasm_bindgen]
-    pub async fn scan_block(&mut self, compact_block: &[u8]) -> WasmResult<bool> {
+    pub async fn scan_block(
+        &mut self,
+        compact_block: &[u8],
+        skip_trial_decrypt: bool,
+    ) -> WasmResult<bool> {
```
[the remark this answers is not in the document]
TalDerei: currently, this does _not_ skip poseidon hashing. You'll notice that we still call `self.sct.insert(Forget, payload.note_commitment)` in the [None](https://github.com/penumbra-zone/web/pull/1707/files#diff-12730df368e71c68b53c35221f7dfc3452bdf61024705a1c6315afb479178df3R164) case, meaning we're inserting _every_ note commitment in _every_ block to construct our filtered TCT instance, which involves performing many poseidon hashes per note commitment insertion under the hood. Instead, if a block isn't relevant to us, we should only be inserting the top-level block root commitment. This way, we still need to perform some poseidon hashing, but not nearly as much.

packages/wasm/crate/src/view_server.rs
```diff
@@ -125,7 +129,9 @@ impl ViewServer {
 
             match state_payload {
                 StatePayload::Note { note: payload, .. } => {
-                    match payload.trial_decrypt(&self.fvk) {
+                    match bool::then_some(!skip_trial_decrypt, payload.trial_decrypt(&self.fvk))
```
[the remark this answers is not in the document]
TalDerei: modified to use closure

```
let note_opt = (!skip_trial_decrypt)
   .then(|| payload.trial_decrypt(&self.fvk))
   .flatten();
match note_opt {    
```

packages/wasm/crate/src/view_server.rs
TalDerei: will play around with implementing some view server wasm tests

.changeset/rude-camels-attack.md
```diff
@@ -0,0 +1,7 @@
+---
+'@penumbra-zone/query': major
+'@penumbra-zone/types': major
+'@penumbra-zone/wasm': minor
```
grod220, line 4: question: wasm is a major bump too given there's a breaking function signature change right?

packages/query/src/helpers/skip-trial-decrypt.test.ts
grod220, line 1: issue(blocking): while these test cases are valid, there isn't any functionality it imports from `block-processor.ts`. That means another developer later can make a change to the logic in the block processor and these tests would pass all the same. 

You should create a helper function that does this:

```ts
// Some code comment for context
export const skipTrialDecrypt = (walletCreationBlockHeight: number, currentHeight: number) => Boolean(walletCreationBlockHeight && currentHeight < BigInt(walletCreationBlockHeight));
```

and it should both be imported into the block processor and this test suite.

packages/query/src/helpers/skip-trial-decrypt.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+  it('should not skip trial decryption when wallet creation block height equals current height', () => {
+    const currentHeight = 100n;
+    const walletCreationBlockHeight = 100n;
+
+    const skipTrialDecrypt = Boolean(
+      walletCreationBlockHeight && currentHeight < BigInt(walletCreationBlockHeight),
+    );
+
+    expect(skipTrialDecrypt).toBe(false);
```
grod220, line 44: You can use this as well:

```ts
expect(skipTrialDecrypt).toBeFalsy();
expect(skipTrialDecrypt).toBeTruthy();
```

packages/wasm/crate/src/view_server.rs
grod220: I suppose we can skip these for now, but we should really prioritize coverage for the wasm code in the future. After these sync improvements, we should treat additions to wasm as blocking without tests.

grod220 (approved):
One blocking comment. After addressing, feel free to merge.

Issue #1413, opened 2024-07-04
display better error message for multi-asset fees


TalDerei:
https://github.com/penumbra-zone/web/pull/1412 properly switched to using the alt_gas_fees field from the compact block to determine the price of an asset. During testing, however, we discovered that the compact blocks were mostly returning zero alternative gas prices.

In the context of multi-asset fees, this results in either (1) zero-valued fees for the alternative asset in the approval dialog (since that's what was in the compact block), or (2) a "GasPrices not available" error. If we receive real prices in the compact blocks, these error messages would be resolved.

<img width="770" alt="Screenshot 2024-07-03 at 6 34 50 PM" src="https://github.com/penumbra-zone/web/assets/70081547/88844c65-d90f-49a4-910a-c9cc84c6c0c7">

-----------------------------------------------------------------------------------------------------------------------------------

<img width="1728" alt="Screenshot 2024-07-03 at 2 06 41 PM" src="https://github.com/penumbra-zone/web/assets/70081547/b74b4ceb-457c-4ebf-8dc0-9006dd327aa8">

-----------------------------------------------------------------------------------------------------------------------------------

for something like this, we should display a friendly error message indicating that the user has an insufficient balance to pay for fees. This way, users are better informed and understand why the transaction cannot proceed.

<img width="1171" alt="19988 118228" src="https://github.com/penumbra-zone/web/assets/70081547/b0ecc629-6141-4e70-9020-665f7c529b8b">

TalDerei:
The first two error messages have already been improved and will now display the [error](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L99) in extractAltFees. The final error message in the wasm planner should be changed to a more user-friendly message, for instance "`Transaction failed due to insufficient funds`," if the user runs out of notes.

PR #1708, opened 2024-08-19
fees: clearer error message

TalDerei:
references https://github.com/penumbra-zone/web/issues/1413 and https://github.com/prax-wallet/web/issues/148

Screenshot 2024-08-18 at 11 37 45 PM

### DIFF
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -601,13 +601,7 @@ pub async fn plan_transaction_inner<Db: Database>(
         let note = notes_by_asset_id
             .get_mut(&required.asset_id)
             .and_then(|notes| notes.pop())
-            .ok_or_else(|| {
-                anyhow!(
-                    "Failed to retrieve or ran out of notes for asset {}, required amount {}",
-                    required.asset_id,
-                    required.amount
-                )
-            })?;
+            .ok_or_else(|| anyhow!("Transaction failed due to insufficient funds"))?;
 
         // Add a spend for that note to the action list.
         actions_list.push(SpendPlan::new(&mut OsRng, note.note, note.position));
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -137,6 +137,28 @@ async fn save_metadata_if_needed<Db: Database>(
     }
 }
 
+// Used to augment error message with denom metadata for a more readable error
+pub async fn insufficient_funds_err<Db: Database>(
+    storage: &Storage<Db>,
+    required: &Value,
+) -> WasmError {
+    let metadata_result = storage.get_asset(&required.asset_id).await;
+    let error_message = match metadata_result {
+        Ok(Some(asset_metadata)) => {
+            let display_unit = asset_metadata.default_unit();
+            format!(
+                "Transaction failed due to insufficient funds. 
+                Required amount: {} {}",
+                display_unit.format_value(required.amount),
+                display_unit
+            )
+        }
+        // Gracefully handle cases when database errors or does not have metadata for asset
+        _ => "Transaction failed due to insufficient funds".to_string(),
+    };
+    anyhow!(error_message).into()
+}
+
 /// Process a `TransactionPlannerRequest`, returning a `TransactionPlan`
 #[wasm_bindgen]
 pub async fn plan_transaction(
@@ -598,10 +620,13 @@ pub async fn plan_transaction_inner<Db: Database>(
     // Now iterate over the action list's imbalances to balance the transaction.
     while let Some(required) = actions_list.balance_with_fee().required().next() {
         // Find a single note to spend towards the required balance.
-        let note = notes_by_asset_id
+        let maybe_note = notes_by_asset_id
             .get_mut(&required.asset_id)
-            .and_then(|notes| notes.pop())
-            .ok_or_else(|| anyhow!("Transaction failed due to insufficient funds"))?;
+            .and_then(|notes| notes.pop());
+        let note = match maybe_note {
+            Some(note) => Ok(note),
+            None => Err(insufficient_funds_err(&storage, &required).await),
+        }?;
 
         // Add a spend for that note to the action list.
         actions_list.push(SpendPlan::new(&mut OsRng, note.note, note.position));
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -147,8 +147,7 @@ pub async fn insufficient_funds_err<Db: Database>(
         Ok(Some(asset_metadata)) => {
             let display_unit = asset_metadata.default_unit();
             format!(
-                "Transaction failed due to insufficient funds. 
-                Required amount: {} {}",
+                "Transaction failed due to insufficient funds. Required amount: {} {}.",
                 display_unit.format_value(required.amount),
                 display_unit
             )
--- a/packages/wasm/crate/tests/test_insufficient_notes.rs
+++ b/packages/wasm/crate/tests/test_insufficient_notes.rs
@@ -1,8 +1,8 @@
 mod utils;
 use crate::utils::planner_setup::seed_params_in_db;
 use decaf377::Fq;
-use penumbra_asset::asset::{Id, Metadata};
-use penumbra_asset::{Value, STAKING_TOKEN_ASSET_ID};
+use penumbra_asset::asset::Id;
+use penumbra_asset::{Value, STAKING_TOKEN_ASSET_ID, STAKING_TOKEN_DENOM};
 use penumbra_num::Amount;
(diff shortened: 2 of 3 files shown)


grod220:
This also seems to address: https://github.com/prax-wallet/web/issues/148

packages/wasm/crate/src/planner.rs
```diff
@@ -601,13 +601,7 @@ pub async fn plan_transaction_inner<Db: Database>(
...
             .get_mut(&required.asset_id)
             .and_then(|notes| notes.pop())
-            .ok_or_else(|| {
-                anyhow!(
-                    "Failed to retrieve or ran out of notes for asset {}, required amount {}",
-                    required.asset_id,
-                    required.amount
-                )
-            })?;
+            .ok_or_else(|| anyhow!("Transaction failed due to insufficient funds"))?;
```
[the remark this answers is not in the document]
TalDerei: I think it’s a good idea, but how do we display UM? For instance, logging these variables returns

```
let asset: Metadata = storage.get_asset(&required.asset_id).await?.unwrap();
let asset_id = asset.id();
let asset_denom = asset.base_denom();
let asset_unit = asset.base_unit();
```

Screenshot 2024-08-19 at 8 08 57 PM

I don't see a getter for retrieving the [name](https://github.dev/penumbra-zone/penumbra/blob/56cb0025e0cece24a6301d72d46bfe7094266daf/crates/core/asset/src/asset/denom_metadata.rs#L49-L50) field?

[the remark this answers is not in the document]
grod220: What do you think about this?

```rust
    while let Some(required) = actions_list.balance_with_fee().required().next() {
        // Find a single note to spend towards the required balance.
        let maybe_note = notes_by_asset_id
            .get_mut(&required.asset_id)
            .and_then(|notes| notes.pop());
        let note = match maybe_note {
            Some(note) => Ok(note),
            None => Err(insufficient_funds_err(&storage, &required).await),
        }?;
       // .....
    }
```
And then a helper function:
```rust
// Used to augment error message with denom metadata for a more readable error
async fn insufficient_funds_err<Db: Database>(
    storage: &Storage<Db>,
    required: &Value,
) -> WasmError {
    let metadata_result = storage.get_asset(&required.asset_id).await;
    let error_message = match metadata_result {
        Ok(Some(asset_metadata)) => {
            let display_unit = asset_metadata.default_unit();
            format!(
                "Transaction failed due to insufficient funds. Required: {}{}",
                display_unit.format_value(required.amount),
                display_unit
            )
        }
        // Gracefully handle cases when database errors or does not have metadata for asset
        _ => "Transaction failed due to insufficient funds".to_string(),
    };
    anyhow!(error_message).into()
}

```

grod220:
Also, we should have tests for this please 🙏

packages/wasm/crate/src/planner.rs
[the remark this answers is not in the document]
TalDerei: this is great! I was looking for `display_unit`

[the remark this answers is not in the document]
TalDerei: Screenshot 2024-08-19 at 8 08 41 PM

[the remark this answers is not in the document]
TalDerei: Screenshot 2024-08-20 at 12 13 47 PM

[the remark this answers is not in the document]
TalDerei: need to remove transfer channels for IBC assets 

Screenshot 2024-08-20 at 12 39 56 PM

packages/wasm/crate/src/planner.rs
```diff
@@ -137,6 +137,28 @@ async fn save_metadata_if_needed<Db: Database>(
...
+    storage: &Storage<Db>,
+    required: &Value,
+) -> WasmError {
+    let metadata_result = storage.get_asset(&required.asset_id).await;
+    let error_message = match metadata_result {
+        Ok(Some(asset_metadata)) => {
+            let display_unit = asset_metadata.default_unit();
+            format!(
+                "Transaction failed due to insufficient funds. 
+                Required amount: {} {}",
```
grod220, line 151: nit: can you add a period?

```diff
- Required amount: {} {}
+ Required amount: {} {}.
```

packages/wasm/crate/tests/test_insufficient_notes.rs
grod220, line 1: issue: we need at least two tests (maybe three):

- Asset id metadata is not in storage (show backup generic error)
- Asset id metadata is in storage (show the display denom + amount) 
- Db throws an error while retrieving (show backup generic error). This last one may be tricky to pull off with the mock database. Feel free to exclude it too difficult.

packages/wasm/crate/src/planner.rs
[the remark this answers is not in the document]
grod220: We can leave the ibc denom's as is. If it proves confusing to users, can revisit. But regardless, far more clear than the current error.

packages/wasm/crate/tests/test_insufficient_notes.rs
TalDerei: expanded testing coverage

packages/wasm/crate/tests/test_insufficient_notes.rs
```diff
@@ -0,0 +1,124 @@
...
+    let key = byte_array_to_base64(&asset_id_staking_token.to_proto().inner);
+    let metadata_in_db_proto = pb::Metadata {
+        base: asset_id_staking_token.to_string(),
+        ..Default::default()
+    };
+
+    mock_db
+        .put_with_key(&tables.assets, key.clone(), &metadata_in_db_proto)
+        .await
+        .unwrap();
```
grod220, line 109: question: this is not necessary for this test right?

packages/wasm/crate/tests/test_insufficient_notes.rs
```diff
@@ -0,0 +1,124 @@
...
+    let tables = get_mock_tables();
+    seed_params_in_db(&mock_db, &tables).await;
+
+    let asset_id_staking_token = *STAKING_TOKEN_ASSET_ID;
+
+    let key = byte_array_to_base64(&asset_id_staking_token.to_proto().inner);
+    let metadata_in_db_proto = pb::Metadata {
+        base: asset_id_staking_token.to_string(),
+        ..Default::default()
+    };
```
grod220, line 65: addon: interestingly, you found another case I didn't think about. That is: what if the database has metadata but no display info? Think it's worth having an additional test for this.

Revised test case:
```rust
async fn metadata_in_storage_but_no_display_info() {
...
// The current code you are using
    let metadata_in_db_proto = pb::Metadata {
        base: asset_id_staking_token.to_string(),
        ..Default::default()
    };
...
 assert_eq!(
        format!("{}", error),
       // I'd consider putting the whole id there so it's clear to folks
         "Transaction failed due to insufficient funds. Required amount: 100 passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm."
    );

```

Additional one:
```rust
async fn metadata_in_storage() {
...
// Store whole metadata struct
let metadata_in_db_proto: pb::Metadata = STAKING_TOKEN_DENOM.clone().into();
..
    assert_eq!(
        format!("{}", error),
       // note the decimal change
        "Transaction failed due to insufficient funds. Required amount: 0.0001 penumbra."
    );
```

packages/wasm/crate/tests/test_insufficient_notes.rs
```diff
@@ -0,0 +1,124 @@
...
+use penumbra_wasm::database::mock::{get_mock_tables, MockDb};
+use penumbra_wasm::planner::insufficient_funds_err;
+use penumbra_wasm::storage::{byte_array_to_base64, Storage};
+use rand_core::OsRng;
+use wasm_bindgen_test::wasm_bindgen_test;
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+use penumbra_proto::DomainType;
+
+#[wasm_bindgen_test]
+// Database throws an error while retrieving
```
grod220, line 18: issue: this test hits the same case as `test_insufficient_notes_not_in_storage`. Triggering a database error is actually not supported in our mock db right now. To not confuse others, I'd leave a comment like:

```
// TODO: add a test for database error when mock storage can support errors
```

grod220 (approved):
Approving, but left some comments for another iteration on some things in the tests

packages/wasm/crate/tests/test_insufficient_notes.rs
TalDerei: good catch, this was a copy of the other test.

packages/wasm/crate/tests/test_insufficient_notes.rs
TalDerei: added the additional test case

Issue #1709, opened 2024-08-19
UI alignment regressions


grod220:
Looks like a recent UI change (perhaps related to UI v2 changes?) was merged that caused a few UI misalignments:

1. Buttons are not vertically centered correctly:

<img width="254" alt="Screenshot 2024-08-19 at 1 48 07 PM" src="https://github.com/user-attachments/assets/11cfc1d2-b60f-4836-a2a4-254b7a1beaef">
<img width="304" alt="Screenshot 2024-08-19 at 1 48 34 PM" src="https://github.com/user-attachments/assets/f9789083-891e-4176-ac66-d3b81f7d55bb">
<img width="402" alt="Screenshot 2024-08-19 at 1 49 26 PM" src="https://github.com/user-attachments/assets/33b3002d-2e42-4c2e-8b9c-f42890454445">

2. Pill denoms are not aligning with amounts (not denom label is lower):

<img width="175" alt="Screenshot 2024-08-19 at 1 49 58 PM" src="https://github.com/user-attachments/assets/8da72507-c65b-4932-ba78-7c6dd6614cb3">

3. Sync modal width does not match the content inside:

<img width="1111" alt="Screenshot 2024-08-19 at 1 50 25 PM" src="https://github.com/user-attachments/assets/82e5ac1f-4332-4f63-bfad-5bcada7dd33e">

VanishMax:
@JasonMHasperhoven i tried this out of curiosity and found that the misalignment is probably in the font itself. On the screenshot, I used the flex lines to draw the center of the box. Both the text from the description and the button do not match the center.

<img width="1116" alt="image" src="https://github.com/user-attachments/assets/afe2c18f-6691-433c-a4ff-44811faedb93">

PR #1711, opened 2024-08-19
Fix Alignments

JasonMHasperhoven:
Fixes https://github.com/penumbra-zone/web/issues/1709

image


packages/ui/fonts/Devanagari-Sangam-Regular.woff2
grod220, line 1: question: these fonts seem to be back to the old larger size? Is there a way we can compress them? The former ones were sub 20kbs.

JasonMHasperhoven: Actually the tff file were much bigger.
image

grod220 (approved):

Issue #1576, opened 2024-07-28
block scanning should use block acceleration features of the TCT


TalDerei:
clients stream compact blocks and construct a filtered instance of the SCT. Currently, we're trial-decrypting every encrypted payload for every block retrieved from the full node. However, our wasm view server implementation doesn't yet take advantage of the TCT acceleration features detailed in [Penumbra issue #842](https://github.com/penumbra-zone/penumbra/issues/842). In that context, we can use the fact that penumbra commits the intermediate root hash of every block, and client synchronization can stream the root hashes of each block _without_ needing to know the individual commitments. This approach would reduce the number of Merkle hashing operations needed to insert a block into the SCT.

our [scan_block](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/view_server.rs#L116) implementation was built using `mock_client` as a template. Instead, we should employ the TCT's API, specifically the [insert_block](https://github.com/penumbra-zone/penumbra/blob/main/crates/crypto/tct/src/epoch.rs#L202) and [insert_epoch](https://github.com/penumbra-zone/penumbra/blob/main/crates/crypto/tct/src/tree.rs#L556) methods, to accelerate block syncing.

we've essentially built this highly optimized merkle tree (**Tierd Commitment Tree**) designed for efficient client sync, but we're not taking advantage of it's fundamental properties.

hdevalence:
the relevant code in the rust view server is here: https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/sync.rs#L111-L117

the optimization is: if the result of scanning the block is that there were no decryptions, use `insert_block`.

there's no need to use the epoch-level insertion features at this time as that is more complicated to employ (scanning goes block by block).

hdevalence:
this optimization isn't completely trivial to port to the wasm view server, because it inlines together the trial decryption and the SCT insertions into one sequence of operations:

https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/view_server.rs#L123-L130

instead, these operations can be cleanly separated by using in-memory advice, as in the native code: https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/sync.rs#L86-L109 : first, build a lookup table of successful decryptions (the "advice"), then use it when processing each note payload. that way, it's easy to write code that skips over an empty block, as all the trial decryption is done up-front.

hdevalence:
[...]

Issue #1142, opened 2024-05-22
profile block processor to improve sync performance


TalDerei:
profile the block processor and sync mechanism in the web, tracked by https://docs.google.com/document/d/1OCVsyi4qn5DlJmdoY9x2mk0dMgcLF577hlC0TD_fpjo/edit

TalDerei:
expanding scope of benchmarking to check cost of poesidon hashing for TCT related operations.

TalDerei:
benchmarking profile for syncing **800k** blocks:

- baseline: **280s**
- skip trial decryption: **200s** (_~25%_ speedup)
- skip trial decryption and poseidon hashing: **100s** (_~75%_ speedup)
- streaming 800k blocks in async block querier: ~**95s**
- binary serialization and wasm decoding: ~**5s**

TalDerei:
![screenshot_2024-08-27_at_11 43 46___am](https://github.com/user-attachments/assets/d3d2b10d-4dcc-4781-a4c4-0d0ada029546)

TalDerei:
minimizing storage accesses and disk i/o will be profiled and investigated as other potential low-hanging fruit.

PR #1716, opened 2024-08-20
sync: use TCT block acceleration 

hdevalence:
Relevant to #1576; I believe this code does the correct behavior but I've only checked that it compiles. Posting so it can get pulled into other work streams


hdevalence:
Looks like the repo could disable the `clippy::clone_on_copy` lint

TalDerei (approved):
this implementation appears to achieve parity with the core. Preliminary benchmarks indicate a performance improvement of approximately **55%**, which aligns with our analysis [here](https://github.com/penumbra-zone/web/issues/1142#issuecomment-2323898812).

benchmarks:
- mainnet
- testnet whale seedphrase
- ~800k blocks
- crouton rpc

_without_ TCT block acceleration: **285 seconds**
_with_ TCT block acceleration: **183 seconds**

TalDerei:
the only open question for me is if the compact block contains an undefined root block, what are we actually [inserting](https://github.com/penumbra-zone/web/pull/1716/files#diff-12730df368e71c68b53c35221f7dfc3452bdf61024705a1c6315afb479178df3R157-R159)?

screenshot_2024-09-01_at_9 39 04___pm

PR #1718, opened 2024-08-20
local genesis PR suggestions

turbocrime:
review suggestions. use work up to 8fe9234c988abf4b284b3996da98963ec430b631

commits after 8fe9234c988abf4b284b3996da98963ec430b631 should become a new issue

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -136,6 +136,15 @@ export class BlockProcessor implements BlockProcessorInterface {
     this.numeraires = numeraires;
   }
 
+  /**
+   * Sync local state to present. This method will
+   * - identify current synced height (or `-1n` to represent a 'pre-genesis' state)
+   * - query remote rpc for the chain's latest block height
+   * - pre-genesis, initialize validator info
+   * - pre-genesis, process a local genesis block if provided
+   * - query remote rpc to begin streaming at the next block
+   * - iterate
+   */
   private async syncAndStore() {
     // start at next block, or genesis if height is undefined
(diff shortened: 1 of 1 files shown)


packages/query/src/block-processor.ts
```diff
@@ -153,24 +162,16 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-
-    // if this is the first block being synced and the bundled genesis block data is available,
-    // process the genesis block first
-    if (currentHeight === -1n && this.genesisBlock) {
-      await this.processBlock(this.genesisBlock, latestKnownBlockHeight);
-      currentHeight = 0n;
+      if (this.genesisBlock?.height === currentHeight + 1n) {
+        // begin the chain with bundled genesis block if provided
+        await this.processBlock(this.genesisBlock, latestKnownBlockHeight);
+        currentHeight = this.genesisBlock.height;
```
turbocrime, line 170: i believe these should be transposed (height first, then process) as that's the technique used in the loop

https://github.com/penumbra-zone/web/commit/8fe9234c988abf4b284b3996da98963ec430b631#r145572205

TalDerei (approved):

packages/query/src/block-processor.ts
TalDerei: note from offline conversation; if processing the genesis fails, and we update the current height before that, what happens in the the exponential backoff? [this.viewServer.resetTreeToStored()](https://github.com/penumbra-zone/web/blob/8fe9234c988abf4b284b3996da98963ec430b631/packages/query/src/block-processor.ts#L124) re-queries idb for local sync height, effectively preventing this class of issues from occurring.

Issue #1719, opened 2024-08-20
IDB contains block processing logic


turbocrime:
**Describe the bug**
IDB contains block processing logic such as epoch genesis.

**To Reproduce**

1. Initialize indexed-db. Do not initialize a block processor.
2. See that idb already contains a 0th epoch

**Expected behavior**
IDB should not arbitrarily create chain state without external input

TalDerei:
not sure how broad you're intending the scope to be, but we should also check for other instances of divergence where IDB may be performing tasks that should be handled by the block processor.

grod220:
It's not super clear to me why the database can't be responsible for its own initialization+default values logic.

https://github.com/penumbra-zone/web/blob/75ef4f5bb7bb07c4ea0478bee360beba12220f28/packages/storage/src/indexed-db/index.ts#L174-L180

This is basically saying, on initialization see if default value has been set. If not, set it.

grod220:
Ah, I see there's a draft PR, will take a look at that to see if there's something I'm missing

PR #1720, opened 2024-08-20
move epoch genesis to block processor

turbocrime:
idb is presently performing epoch genesis, instead of the block processor directing idb through the process.

block processor should contain all genesis special cases.

https://github.com/penumbra-zone/web/issues/1719


TalDerei:
taking this over

TalDerei:
I wanted to finalize the draft PR and took the opportunity to audit some of the database and block processor code for other instances of divergence where IDB may be performing tasks that should be handled by the block processor.

packages/query/src/block-processor.ts
```diff
@@ -162,12 +162,12 @@ export class BlockProcessor implements BlockProcessorInterface {
 
     // special case genesis sync
     if (currentHeight === -1n) {
-      // initialize validator info at genesis
-      // TODO: use batch endpoint https://github.com/penumbra-zone/penumbra/issues/4688
```
grod220, line 166: suggestion: think it would be helpful to keep this comment

packages/query/src/block-processor.ts
```diff
@@ -162,12 +162,12 @@ export class BlockProcessor implements BlockProcessorInterface {
...
+      // create first epoch and initialize validator info at genesis
+      await this.indexedDb.addEpoch(currentHeight + 1n);
       void this.updateValidatorInfos(currentHeight + 1n);
 
-      // begin the chain with local genesis block if provided
```
grod220, line 169: suggestion: feel like the comment placement before was better

packages/query/src/block-processor.ts
```diff
@@ -373,8 +373,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       );
     }
 
-    const isLastBlockOfEpoch = !!compactBlock.epochRoot;
-    if (isLastBlockOfEpoch) {
+    if (compactBlock.epochRoot) {
```
grod220, line 376: issue: we lost a bit of documentation here that informed the dev that the presence of this field means it's the last block of the epoch. Perhaps a code comment can bring that info back?

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -565,67 +562,6 @@ describe('IndexedDb', () => {
     });
   });
 
-  describe('epochs', () => {
```
grod220, line 568: thought: makes me uneasy removing tests without a replacement. However, maybe it's fine in this case.

packages/query/src/block-processor.ts
```diff
@@ -162,12 +162,12 @@ export class BlockProcessor implements BlockProcessorInterface {
 
     // special case genesis sync
     if (currentHeight === -1n) {
```
grod220, line 164: question: what is the difference between this conditional and the one nested inside?

```
if (currentHeight === -1n) {
vs
if (this.genesisBlock?.height === currentHeight + 1n) {
```

packages/query/src/block-processor.ts
```diff
@@ -162,12 +162,12 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-      // initialize validator info at genesis
-      // TODO: use batch endpoint https://github.com/penumbra-zone/penumbra/issues/4688
+      // create first epoch and initialize validator info at genesis
+      await this.indexedDb.addEpoch(currentHeight + 1n);
```
grod220, line 166: suggestion: feels more clear if we just do:

```
await this.indexedDb.addEpoch(0n);
void this.updateValidatorInfos(0n);
```

packages/query/src/block-processor.ts
grod220: issue: Ah just realized the second conditional only runs if there is a bundled genesis block for the chain. Can you add a code comment?

packages/query/src/block-processor.ts
```diff
@@ -138,7 +138,7 @@ export class BlockProcessor implements BlockProcessorInterface {
    * Sync local state to present. This method will
    * - identify current synced height (or `-1n` to represent a 'pre-genesis' state)
```
grod220, line 139: issue: I have been having trouble wrapping my mind around what currentHeight of -1 meant. But finally understood when I saw this comment. Can we do something like this:

```
const PRE_GENESIS_SYNC_HEIGHT = -1n;
...
let currentHeight = (await this.indexedDb.getFullSyncHeight()) ?? PRE_GENESIS_SYNC_HEIGHT;
...
if (currentHeight === PRE_GENESIS_SYNC_HEIGHT) {
```

packages/query/src/block-processor.ts
```diff
@@ -162,12 +162,12 @@ export class BlockProcessor implements BlockProcessorInterface {
 
     // special case genesis sync
```
grod220, line 163: issue: Can we add more to this code comment? Something like:

```
// Special case if no syncing has yet been done. Will prep for syncing 
// and see if there's a bundled genesis block passed (saves on network request time)
```

grod220 (approved):
Mostly documentation review suggestions. After you address, feel free to merge.

packages/query/src/block-processor.ts
TalDerei: added

packages/query/src/block-processor.ts
TalDerei: added comment

packages/storage/src/indexed-db/indexed-db.test.ts
TalDerei: I think it's alright since I opted for a more minimal diff.

packages/query/src/block-processor.ts
TalDerei: added more information on the conditionals

packages/query/src/block-processor.ts
TalDerei: good observation, changed

packages/query/src/block-processor.ts
TalDerei: that's more clear, changed

packages/query/src/block-processor.ts
TalDerei: changed

PR #1733, opened 2024-08-26
feat(ui): #1703: publish UI library

### LINKED ISSUE #1703: Publish v2 UI components to npm
Currently the package is not being published: https://github.com/penumbra-zone/web/blob/main/packages/ui/package.json#L4

Given the good progress with the v2 UI components, it's a good time to expose this externally. There are already a few consumers that want to make use of them.

Objective:
- Delete private indication on package.json (will trigger changeset publishing)
- Expose v2 UI components (not v1 ones)
- Demonstrate external example (probably in the [nextjs example repo](https://github.com/penumbra-zone/nextjs-penumbra-client-example) or a new one?)

VanishMax:
Closes #1703

Prepares the UI package for publishing to NPM. Added the build step with Vite that collects all `index.tsx` components from the `src` in the UI package. Then, it bundles them as ES modules to the `dist` directory in the corresponding paths. The shared code is reused, so the components do not get bloated in size. Moreover, it puts type declarations next to the bundled code and configures peer dependencies to reduce the resulting bundle.

**The new `@penumbra-zone/ui` does not bundle the old components from the `components/ui` directory and does not encourage their use. Yet, it is still available for internal use.**

The PR in the NextJS example demonstrates the use of the updated package: https://github.com/penumbra-zone/nextjs-penumbra-client-example/pull/8

image


vacekj (commented):
preliminary approval from me, will finish review tomorrow

vacekj (approved):
lgtm

Issue #1715, opened 2024-08-19
Implement `<Slider/>` UI component


grod220:
Designs: https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=400-31027&m=dev

VanishMax:
Implemented in #1738

PR #1738, opened 2024-08-28
Add Slider component

JasonMHasperhoven:
Implements https://github.com/penumbra-zone/web/issues/1715

image


packages/ui/src/Slider/index.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { Slider } from '.';
+import { PenumbraUIProvider } from '../PenumbraUIProvider';
+
+describe('<Slider />', () => {
+  it('renders correctly', () => {
```
VanishMax, line 7: thought: that's not a meaningful test. UI interactions could be properly tested.

suggestion: for this test, you can render the component, simulate a click to some point of the slider, mock the `onChange` function and expect the correct value to be passed to it.

packages/ui/src/Slider/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+  valueDetails?: string;
+  showTrackGaps?: boolean;
+  trackGapBackground?: string;
+  showFill?: boolean;
+  fontSize?: string;
+}
+
+const THUMB_SIZE = '16px';
+const TRACK_HEIGHT = '4px';
+const TRACK_GAP_WIDTH = '4px';
```
VanishMax, line 24: todo (blocking): use `${props => props.theme.spacing(4)}` to get the correct value for any spacing property in css. Sam tries to use all theme values everywhere in Figma (see the screenshot), so should we

image

packages/ui/src/Slider/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+const ValueDisplay = styled.div`
+  color: ${props => props.theme.color.text.primary};
+`;
+
+const ValueDetails = styled.div`
+  margin-left: ${props => props.theme.spacing(1)};
+  color: ${props => props.theme.color.text.secondary};
+`;
+
+export const Slider: React.FC<SliderProps> = ({
```
VanishMax, line 112: issue (blocking): this Slider is not a11y-compliant. You can check the [W3C standards for a Slider](https://www.w3.org/WAI/ARIA/apg/patterns/slider/) – it describes the patterns we should implement to provide the keyboard accessibility and help the users with low vision. For example, the standards say that the value can be controlled via pressing left or arrow keys when the Slider is focused.

So far, Jesse, Sam and me tried to comply with a11y standards in the design and code. If you open v2 pages, tabbing around shows correct focus state for any component, and it responds not only to the "enter press" but also to any other keyboard key. I believe we could do the same with Slider 🙏

suggestion: use [Radix's Slider component](https://www.radix-ui.com/primitives/docs/components/slider) as a base to handle the a11y and even simplify the code.

VanishMax (changes requested):

VanishMax (approved):
Loving the result!

TalDerei (commented):
LGTM

Issue #1702, opened 2024-08-15
Wasm out of memory


Valentine1898:
**Describe the bug**
Building a “withdraw all” transaction for a large number of auctions (about 40) leads to wasm memory overflow
```
ConnectError: [unknown] Worker ErrorEvent chrome-extension://lkpmkhpnhknhmibgnmmhdhgdilepfghe/764.js:1:80331 Uncaught RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance
    at c (service-worker.js:1:311355)
    at service-worker.js:1:441479
    at async service-worker.js:1:441396
``` 

**Screenshots**
<img width="1508" alt="image" src="https://github.com/user-attachments/assets/9e1caa5e-2745-4d54-8bad-3d9872ffc934">

TalDerei:
auxiliary; we should remove the _serial_ build method from wasm to further reduce the binary size. 

<img width="1728" alt="Screenshot 2024-08-22 at 10 13 15 AM" src="https://github.com/user-attachments/assets/7bb85916-7e04-4fb9-b547-7d37d2b17cc6">

TalDerei:
my initial thought is that this issue could be related to a memory leak or limited memory allocation that might need to be increased.

TalDerei:
> auxiliary consideration; we should remove the `serial` build method from wasm to further reduce the binary size.

removing this doesn't reduce the wasm binary size (still **5 MB**), likely because dead code elimination has already taken effect during the compiler pass?

PR #1741, opened 2024-08-29
Lower batch limit (auctions)

grod220:
Temporary mitigation for: https://github.com/penumbra-zone/web/issues/1702

A full 24 action auction batch takes ~2mins on my computer (m1 max). I wonder if this is a good setting just generally.

### DIFF
--- a/apps/minifront/src/components/swap/auction-list/end-or-withdraw-all-button.tsx
+++ b/apps/minifront/src/components/swap/auction-list/end-or-withdraw-all-button.tsx
@@ -44,6 +44,14 @@ export const assembleAuctionBatch = (
   return { auctions: filteredBySeqAndAddressIndexAuctions, source: firstFoundAddressIndex! };
 };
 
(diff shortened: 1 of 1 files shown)


apps/minifront/src/components/swap/auction-list/end-or-withdraw-all-button.tsx
```diff
@@ -89,26 +93,22 @@ export const EndOrWithdrawAllButton = () => {
...
                 auctionBatch.auctions,
                 // TODO Should use the index of the selected account after the account selector for the auction is implemented
                 auctionBatch.source,
               );
             }}
-            aria-label='Withdraw all ended auctions, with a limit of 48 auctions per transaction'
+            aria-label={withdrawTooltipContent}
           >
-            <div className='w-[95px] shrink-0'>
+            <div className='w-[120px] shrink-0'>
```
grod220, line 105: Making button wider to accommodate larger text length 

Screenshot 2024-08-29 at 3 56 19 PM

JasonMHasperhoven (approved):

PR #1749, opened 2024-08-30
Fix reselect conditional

grod220:
On other `reselect` usages, the conditional is opposite. It makes sense for the reselector to run when the data object has changed (not when remaining the same).

On testnet noticed this fixes a bug with navigating the staking page (arrows illuminated only when balances in another account). Not seen on mainnet as I believe the longer validator list causes re-renders (?). Not sure why.

https://github.com/penumbra-zone/web/blob/5616a7a6d54ebdc37127f398ec773edbc0e8ccf7/apps/minifront/src/components/dashboard/assets-table/index.tsx#L55

https://github.com/penumbra-zone/web/blob/5616a7a6d54ebdc37127f398ec773edbc0e8ccf7/apps/minifront/src/components/v2/dashboard-layout/assets-page/index.tsx#L39

### DIFF
--- a/apps/minifront/src/components/staking/account/use-staking-tokens-and-filter.ts
+++ b/apps/minifront/src/components/staking/account/use-staking-tokens-and-filter.ts
@@ -60,7 +60,7 @@ export const useStakingTokensAndFilter = (
   const { data: stakingTokenMetadata } = useStakingTokenMetadata();
   const balancesByAccount = useBalancesResponses({
     select: balancesByAccountSelector,
-    shouldReselect: (before, after) => before?.data === after.data,
+    shouldReselect: (before, after) => before?.data !== after.data,
   });
 
   const stakingTokensByAccount = useMemo(() => {


turbocrime (approved):
this seems like the correct comparison to make, by the docstrings

https://github.com/penumbra-zone/web/blob/5616a7a6d54ebdc37127f398ec773edbc0e8ccf7/packages/zquery/src/types.ts#L25-L45

it also seems to be the default behavior

https://github.com/penumbra-zone/web/blob/5616a7a6d54ebdc37127f398ec773edbc0e8ccf7/packages/zquery/src/get-use-hook.ts#L119-L120

so maybe this can actually just be deleted here, and the other places it appears?

apps/minifront/src/components/staking/account/use-staking-tokens-and-filter.ts
```diff
@@ -60,7 +60,7 @@ export const useStakingTokensAndFilter = (
   const { data: stakingTokenMetadata } = useStakingTokenMetadata();
   const balancesByAccount = useBalancesResponses({
     select: balancesByAccountSelector,
-    shouldReselect: (before, after) => before?.data === after.data,
+    shouldReselect: (before, after) => before?.data !== after.data,
```
turbocrime, line 63:
```suggestion
```

PR #1721, opened 2024-08-20
wip: new eslint rules and fix/ignore errors

vacekj:
fixes #1668 

I adjusted some rules as they felt unnecessarily restrictive for our usecase and added some rules to prevent misbehaviour and ease reviews.

Most of the checks are passing, but I rebased some commits recently so those might still need work.

PR #1750, opened 2024-08-30
remove react-refresh lint

turbocrime:
react fast refresh is an acceleration to HMR, and it doesn't affect production code.

it is supported by vite and enabled by `@vitejs/plugin-react-swc`, so technically it is used in this repository. https://github.com/vitejs/vite-plugin-react-swc#consistent-components-exports

the lint rule was included in eslint config at some point, but its use was unclear. the eslint rules from the eslint plugin aren't actually enabled, so the eslint plugin is unused.

work in #1721 did not re-enable this rule that was provisionally disabled.

the rule should be enabled or deleted.

many components do not respect the requirements of react-fast-refresh, so enabling it would require many changes, including breaking changes to exports.

this PR deletes the eslint plugin and the disabled rule.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -9,7 +9,6 @@ import * as import_ from 'eslint-plugin-import';
 import prettier from 'eslint-config-prettier';
 import react from 'eslint-plugin-react';
 import react_hooks from 'eslint-plugin-react-hooks';
-import react_refresh from 'eslint-plugin-react-refresh';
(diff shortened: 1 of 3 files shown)


VanishMax (approved):
agree that this should just be deleted

vacekj:
Since L is gone, taking the liberty to merge this while taking responsiblity.

PR #1751, opened 2024-08-30
improve manifest type

turbocrime:
manifest access was slightly awkward in #1734 so this change improves the manifest type to provide more consistent access (fields no longer use index access, allow dot notation, don't optionalize existing required fields)


VanishMax (approved):

VanishMax:
merging to apply this to #1734

PR #1755, opened 2024-09-02
fix(ui): #1744: `ValueViewComponent` update

### LINKED ISSUE #1743: Popover 2.1 update
Update: https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=778-47220&node-type=FRAME&cc-version-id=6463618495&m=dev

Added technical variant

### LINKED ISSUE #1744: ValueView 2.1 update
Updates: https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=239-19457&node-type=FRAME&cc-version-id=6488188598&m=dev

- Styles changed
- Variants added

VanishMax:
Closes #1744
Closes #1743 

Style updates to align with the latest design updates for `ValueViewComponent`, `AssetIcon`, and `Popover` components.

The main change is adding the `size` prop to `AssetIcon`.

Results: https://penumbra-ui-preview--pr1755-fix-1744-value-view-ilyok8sx.web.app/?path=/docs/ui-library-valueviewcomponent--docs


packages/ui/src/AssetIcon/index.tsx
```diff
@@ -1,52 +1,62 @@
...
+    /**
+     * @todo: Render a custom unbonding token for validators that have a
+     * logo -- e.g., with the validator ID superimposed over the validator logo.
+     */
+    assetIcon = <UnbondingTokenIcon displayDenom={display} />;
+  } else {
+    assetIcon = <Identicon uniqueIdentifier={metadata?.symbol ?? '?'} type='solid' />;
+  }
+
+  return <BorderWrapper $size={size}>{assetIcon}</BorderWrapper>;
```
grod220, line 61: observation: this was done as we have a eslint rule against nested ternaries now? I wonder if this is in fact a better pattern, idk.

grod220 (approved):
LGTM

packages/ui/src/AssetIcon/index.tsx
VanishMax: i personally prefer using `useMemo` for computed values but it wasn't needed in this component since it would subscribe to all props. Probably this could also be moved to a separate hook. Anyway better than nested ternaries

PR #1761, opened 2024-09-03
fix(minifront): #1120: sort the balances by priority score, fix styles

### LINKED ISSUE #1120: Better rendering display of assets
Currently, it's a bit confusing the manner in which we display assets of the same type...we should group by account or asset type. 

Screenshot 2024-05-15 at 4 29 10 PM

VanishMax:
Closes #1120 

- Added the sorting "by priority score" for balances in the swap asset selector
- Fixed the paddings in the asset selectors

Might look a little unexpected first but it is designed for better UX – users will see their top-valued assets first

image

### DIFF
--- a/.changeset/young-mice-drive.md
+++ b/.changeset/young-mice-drive.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
+Sort the balances by priority in the swap and send pages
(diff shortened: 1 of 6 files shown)


vacekj (approved):
lgtm, lint only

grod220:
Can you check with Sam on this ordering first? Does this come out of the UIv2 designs where account groups are not together?

VanishMax:
discussed better alternatives and finally came up with account index sorting + priority sorting within the account groups

image

PR #1769, opened 2024-09-05
remove offscreen and parallelize transaction building

turbocrime:
references https://github.com/penumbra-zone/web/issues/1787

this PR removes the 'offscreen' logic and thus all chrome runtime calls from services. parallel builds now call `new Worker` directly. the package now uses DOM lib in tsconfig.

keys package now more correctly supports consumers attempting to identify asset paths.

'internal message' types are now removed from the types package.


packages/services/src/view-service/util/parallel.ts
```diff
@@ -0,0 +1,77 @@
...
+  WitnessData,
+} from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
+import { WorkerBuildAction } from './build-action-worker.js';
+import { ConnectError } from '@connectrpc/connect';
+
+/**
+ * Build actions in parallel by launching a worker for each action in the plan.
+ * @returns An individually-promised list of build results.
+ */
+export const launchActionWorkers = (
```
TalDerei, line 15: **comment:** for external reviewers, this delineates some of the design considerations based on offline conversations. 

The service worker is a single-threaded runtime, and objective is to lift the block processor to operate in an offscreen context, preventing it from blocking the service worker.  The possible approaches to achieve this are:

1. **target standard dom**
2. **require offscreen client to be provided**

-------------------------------------------------------------------------------------------------------------------
**target standard dom**

previously, the services package was responsible for managing the offscreen state by relying on chrome's [offscreen API](https://developer.chrome.com/docs/extensions/reference/api/offscreen). The services package was _refactored_ to abandon the offscreen API entirely, instead using standard DOM (built-in [Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Worker)) to manage workers. 

This approach is more general and works outside of chrome extension limitations. If standard DOM is available, the Worker API is used directly. If not, a [polyfill](https://github.com/prax-wallet/web/pull/190) simulates the same environment (e.g., using offscreen workers). Services and block processors can directly create and manage workers using a standardized API, and there’s no dependency on chrome’s runtime or messaging. 

-------------------------------------------------------------------------------------------------------------------

**require offscreen client to be provided (via context, or constructor)**

this approach abstracts the worker-launching mechanism so that services don’t manage workers directly. Instead, workers are launched through a generalized interface, which could be either a standard DOM worker to a chrome offscreen worker. The services packages would be _agnostic_ to the specific worker implementation and lifecycle.

This would require moving the offscreen client into the handler context for services and pass as a parameter for the block processor on initialization, so they can use the same interface to launch and manage offscreen. 

In comparison to the first approach, this serves as a generalized worker launching interface provided externally to launch workers, where you're calling a  generic interface method rather than a worker constructor (standard DOM). 

-------------------------------------------------------------------------------------------------------------------

We opted for the first approach, as the difference between the two is minimal in terms of the actual implementation effort.

packages/services/src/view-service/util/parallel.ts
```diff
@@ -0,0 +1,77 @@
...
+    fullViewingKey: fullViewingKey.toJson() as JsonObject,
+  };
+
+  const workerOutputs = transactionPlan.actions.map(async (_, actionPlanIndex) => {
+    const buildReq: WorkerBuildAction = {
+      ...partialRequest,
+      actionPlanIndex,
+    };
+
+    const actionWorker = new Worker(new URL('./build-action-worker.js', import.meta.url));
```
TalDerei, line 33: **comment:** previously we were spawning the worker in the extension's [offscreen handler](https://github.com/prax-wallet/web/pull/190/files#diff-8c2640a27ba27f5d2d0715fe7abd9f49980168a00aae7d28528e5daf1c4c316bL48), and now it's happening directly in the services package. 

Based on offline discussion, a generic offscreen utility called [OffscreenWorker](https://github.com/prax-wallet/web/pull/190/files#diff-e510b8ebc1f7156cc731b9eeea856aaf3081a4d5e82c0f38a2a6afbb751dc408R5) now manages the lifecycle of the offscreen document. A polyfill was created where `new Worker` references the global `Worker` variable. So when the service worker sets [globalThis.Worker = OffscreenWorker](https://github.com/prax-wallet/web/pull/190/files#diff-b9be68cac5d37d21289e20d8d7fcb88d993de52b16e69a6500255f3fde092ae0R48), all `Worker` references in the script now point to `OffscreenWorker`. Each new `Worker` call creates an instance of `OffscreenWorker`, which communicates via a dedicated port to an actual `Worker` running in the offscreen document.

TalDerei (commented):
@turbocrime the block processor component still needs to be migrated within the offscreen client – what would that generally look like and how significant of a lift would it be?

TalDerei:
We should benchmark the web worker performance to identify any potential regressions, though there shouldn't be any.

packages/services/src/view-service/util/parallel.ts
TalDerei: we should also benchmark the web worker performance to identify any potential regressions in the proving time, though there shouldn't be any.

TalDerei:
has there been extensive manual testing that the optimistic proving mechanism doesn't somehow break the [OffscreenWorker](https://github.com/prax-wallet/web/pull/190/files#diff-e510b8ebc1f7156cc731b9eeea856aaf3081a4d5e82c0f38a2a6afbb751dc408R5) utility which manages the lifecycle of the offscreen document?

TalDerei:
closing this intermediate step in favor of proceeding directly with the migration.

TalDerei:
closing in favor of revisiting when this work stream becomes prioritized when we have free cycles

PR #1770, opened 2024-09-05
fix(minifront): #1675: decimals validation

### LINKED ISSUE #1675: Typing prevented w/ some decimal cases
In the send form, paste in an amount like this:

Screenshot 2024-08-09 at 1 29 41 PM

Then highlight the total amount and try to type another number. The UI prevents this. It is likely due to maximum precision reasons.

**Task**: Remove restrictions in the send/swap form on typing in _any_ amounts. Instead, add a new validation check that looks for max precision. If the input is not valid, just communicate it in red (like the way we do for insufficient funds). Violating the new rule should prevent sending as well.

VanishMax:
Closes #1675 

Renders the errors in case of insufficient funds or incorrect decimals in the balance selector in the swap page.

image

image

### DIFF
--- a/.changeset/three-suns-complain.md
+++ b/.changeset/three-suns-complain.md
@@ -1,0 +1,5 @@
+---
+'minifront': patch
+---
+
+Render the error in swap page in case of incorrect decimals or insufficient funds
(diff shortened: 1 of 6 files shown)


JasonMHasperhoven (approved):
LGTM

Issue #1792, opened 2024-09-17
LP position not detected as closed


grod220:
On mainnet, an LP position for the testnet whale is stored in Prax as opened when it's actually in the withdrawn state. This means our block processing is not detecting the state transitions correctly.

**To Reproduce**:

1. Install Prax using [testnet whale seed phrase](https://rustdoc.penumbra.zone/main/penumbra_keys/test_keys/constant.SEED_PHRASE.html)
2. Connect to mainnet rpc
3. Sync chain
4. Right click Prax, click inspect, click application tab, click IndexedDB on the left, and visit `POSITIONS` table. Note the state of position `84HVlCF1pWPdKQ2VIWtFxHlLi2dWjyl9OpHOCXPNVE4=` as opened.
5. See [position history in dex explorer](https://dex.penumbra.sevenseas.capital/lp/plpid17wqat9ppwkjk8hffpk2jz669c3u5hzm8268jjlf6j88qju7d238qak905k) as withdrawn.

PR #1793, opened 2024-09-17
Only process relevant transactions

grod220:
Closes https://github.com/penumbra-zone/web/issues/1792

We should only be processing transactions that have nullifiers, spends, or swaps we recognize. At the moment, if there is a relevant tx in a block, we later process the entire block. This caused the original issue of an LP position misattributed to the owner when it was external.

Note, we need to version the DB given users may have in their db's LP NFTs and Auction NFTs that happened to be in the same block as another person's.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -333,7 +333,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       // TODO: this is the second time we save these records, after "saveScanResult"
       await this.saveRecoveredCommitmentSources(recordsWithSources);
 
-      await this.processTransactions(blockTx);
+      await this.processTransactions(relevantTx);
 
       // at this point txinfo can be generated and saved. this will resolve
       // pending broadcasts, and populate the transaction list.
@@ -591,8 +591,8 @@ export class BlockProcessor implements BlockProcessorInterface {
    * Identify various pieces of data from the transaction that we need to save,
    * such as metadata, liquidity positions, etc.
    */
-  private async processTransactions(txs: Transaction[]) {
-    for (const tx of txs) {
+  private async processTransactions(txs: Map<TransactionId, Transaction>) {
+    for (const [, tx] of txs) {
       for (const { action } of tx.body?.actions ?? []) {
         await Promise.all([this.identifyAuctionNfts(action), this.identifyLpNftPositions(action)]);
       }
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -305,6 +305,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       // this is a network query
       const blockTx = await this.querier.app.txsByHeight(compactBlock.height);
 
+      // Filter down to transactions & note records in block relevant to user
       const { relevantTxs, recoveredSourceRecords } = await identifyTransactions(
         spentNullifiers,
         recordsByCommitment,
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -4,21 +4,13 @@ import {
   PositionState,
   PositionState_PositionStateEnum,
 } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
-import {
-  CommitmentSource,
-  Nullifier,
-} from '@penumbra-zone/protobuf/penumbra/core/component/sct/v1/sct_pb';
+import { Nullifier } from '@penumbra-zone/protobuf/penumbra/core/component/sct/v1/sct_pb';
 import { ValidatorInfoResponse } from '@penumbra-zone/protobuf/penumbra/core/component/stake/v1/stake_pb';
-import {
-  Action,
-  Transaction,
-} from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
-import { TransactionId } from '@penumbra-zone/protobuf/penumbra/core/txhash/v1/txhash_pb';
+import { Action } from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
 import { StateCommitment } from '@penumbra-zone/protobuf/penumbra/crypto/tct/v1/tct_pb';
 import { SpendableNoteRecord, SwapRecord } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
 import { auctionIdFromBech32 } from '@penumbra-zone/bech32m/pauctid';
 import { bech32mIdentityKey } from '@penumbra-zone/bech32m/penumbravalid';
-import { sha256Hash } from '@penumbra-zone/crypto-web/sha256';
 import { getAssetId } from '@penumbra-zone/getters/metadata';
 import {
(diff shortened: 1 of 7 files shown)


TalDerei (approved):
nice catch. Is bumping the idb version necessary, since we haven't modified the database schema?

grod220:
=== Part 2 ===

The original simple bug fix actually exposed another bug [here](https://github.com/penumbra-zone/web/pull/1793/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872daL488). Adding to that map duplicate entries caused multiple attempts to save the same auction state (certain invalid combos raise exceptions).

Extracted `identifyTransactions()` to a helper file, refactored it, patched the bug, and added tests.

packages/query/src/block-processor.ts
```diff
@@ -129,9 +118,7 @@ export class BlockProcessor implements BlockProcessorInterface {
       numOfAttempts: Infinity,
       maxDelay: 20_000, // 20 seconds
       retry: async (e, attemptNumber) => {
-        if (globalThis.__DEV__) {
-          console.debug('Sync failure', attemptNumber, e);
-        }
+        console.error(`Sync failure #${attemptNumber}: `, e);
```
grod220, line 121: Think we shouldn't require a special flag to see sync failures. Will help with debugging if users get stuck.

packages/query/src/block-processor.ts
```diff
@@ -450,64 +432,6 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-          case 'swapClaim':
-            return action.value.body?.nullifier;
-          default:
-            return;
-        }
-      });
-
-      for (const spentNullifier of spentNullifiers) {
-        if (txNullifiers.some(txNullifier => spentNullifier.equals(txNullifier))) {
-          txId = new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
```
grod220, line 488: This line was the source of the bug. Because the key of the map is an object, when you create a new object (even though it has the same values), it will be added as a new entry in the map. Later on, the redundant entries in the map can cause problems. So this should have been like the loop below:

```diff
- txId = new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
+ txId ??= new TransactionId({ inner: await sha256Hash(tx.toBinary()) });
```

packages/query/src/block-processor.ts
```diff
@@ -450,64 +432,6 @@ export class BlockProcessor implements BlockProcessorInterface {
...
-          relevantTx.set(txId, tx);
-          spentNullifiers.delete(spentNullifier);
```
grod220, line 490: We seem to be mutating the inputs `spentNullifiers` and `commitmentRecordsByStateCommitment`. This seems like a brittle pattern, especially from within a loop that iterates over that object. I could not see a reason why we needed to do this.

packages/query/src/helpers/identify-txs.ts
```diff
@@ -0,0 +1,137 @@
...
+  blockTx: Transaction[],
+): Promise<{
+  relevantTxs: RelevantTx[];
+  recoveredSourceRecords: RecoveredSourceRecords;
+}> => {
+  const relevantTxs: RelevantTx[] = [];
+  const recoveredSourceRecords: RecoveredSourceRecords = [];
+
+  const searchPromises = blockTx.map(tx => searchRelevant(tx, spentNullifiers, commitmentRecords));
+  const results = await Promise.all(searchPromises);
```
grod220, line 126: New feature: we now process transactions in parallel versus the previous synchronous loop

packages/storage/src/indexed-db/config.ts
```diff
@@ -2,4 +2,4 @@
  * The version number for the IndexedDB schema. This version number is used to manage
  * database upgrades and ensure that the correct schema version is applied.
  */
-export const IDB_VERSION = 46;
+export const IDB_VERSION = 47;
```
grod220, line 5: Necessary to bump storage version as users may have NFTs (auction/LP positions) that are misattributed to them as being the owner.

packages/query/src/block-processor.ts
```diff
@@ -318,28 +305,23 @@ export class BlockProcessor implements BlockProcessorInterface {
       // this is a network query
       const blockTx = await this.querier.app.txsByHeight(compactBlock.height);
 
-      // identify tx that involve a new record
-      // - compare nullifiers
-      // - compare state commitments
-      // - collect relevant tx for info generation later
-      // - if matched by commitment, collect record with recovered source
```
TalDerei, line 325: reason for removing code comment?

packages/query/src/helpers/identify-txs.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return undefined;
+};
+
+// identify transactions that involve a new record by comparing nullifiers and state commitments
+// also returns records with recovered sources
+export const identifyTransactions = async (
+  spentNullifiers: Set<Nullifier>,
+  commitmentRecords: Map<StateCommitment, SpendableNoteRecord | SwapRecord>,
```
TalDerei, line 116: think we might have another potential bug. I don’t think we've ever addressed the case where the map holding `StateCommitment` objects contains [duplicate commitments](https://github.com/penumbra-zone/penumbra/issues/3945)?

packages/query/src/helpers/identify-txs.ts
TalDerei: Is there a sense of how the concurrency performs when handling large blocks with many transactions? Is the reduction in unnecessary computations and storage operations pronounced when focusing on processing only relevant transactions?

TalDerei (approved):

packages/query/src/block-processor.ts
grod220: Carried over to external function. But will add a code comment above where its used again.

packages/query/src/helpers/identify-txs.ts
grod220: > handling large blocks with many transactions

This should perform better as it's no longer sequentially peforming `await sha256Hash(tx.toBinary())`, but it's happening in parallel. However, the hashing cost may be trivial and this optimization may only be detectable with a very large amount of txs. 

> impact of avoiding unnecessary computations and storage operations

This code impacts this by making the storage of irrelevant transactions unnecessary. In a situation with quite large blocks with a ton of transactions, this would likely be noticeable at some point had we not fixed it.

packages/query/src/helpers/identify-txs.ts
grod220: In that case, the worst that could happen is in this line: https://github.com/penumbra-zone/web/pull/1793/files#diff-df9a3308a5ae26e8f5e912c4bdd659a76d12e6f751ec7c3eb8de3c0b81147bf0R91-R98 where there are two duplicate recovered records. And in that case, we just [save to the database](https://github.com/penumbra-zone/web/blob/75ef4f5bb7bb07c4ea0478bee360beba12220f28/packages/query/src/block-processor.ts#L424-L440) twice. Don't see a failure case we need to worry about.

Issue #1798, opened 2024-09-19
Cannot scroll assets in shield asset selector


zmanian:
If you have a large number of assets in your wallet on source chain, you can't scroll the asset selector widget and and thus can't select the desired asset.

<img width="294" alt="Screenshot 2024-09-20 at 2 54 30 AM" src="https://github.com/user-attachments/assets/aa0d3336-0d22-48f2-862e-559ed8fb5790">

zmanian:
You can work around this with IBC.fun

PR #1799, opened 2024-09-20
Update Select component

grod220:
Fixes: https://github.com/penumbra-zone/web/issues/1798

Updates shadcnui component to latest version which comes with scrolling. Only in use in shield page.

https://github.com/user-attachments/assets/7869c1c7-0f80-41a0-a358-5e5f74423e9d

### DIFF
--- a/apps/minifront/src/components/ibc/ibc-out/chain-selector.tsx
+++ b/apps/minifront/src/components/ibc/ibc-out/chain-selector.tsx
@@ -6,7 +6,6 @@ import {
   SelectValue,
 } from '@penumbra-zone/ui/components/ui/select';
 import { cn } from '@penumbra-zone/ui/lib/utils';
-import { useState } from 'react';
 import { AllSlices } from '../../../state';
 import { Chain } from '@penumbra-labs/registry';
 import { useStoreShallow } from '../../../utils/use-store-shallow';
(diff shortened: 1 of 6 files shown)


apps/minifront/src/components/ibc/ibc-out/chain-selector.tsx
```diff
@@ -20,18 +19,15 @@ const chainSelectorSelector = (state: AllSlices) => ({
...
-  const [openSelect, setOpenSelect] = useState(false);
 
   return (
     <div className='flex flex-col gap-3 rounded-lg border bg-background px-4 pb-5 pt-3'>
       <p className='text-base font-bold'>Chain</p>
       <Select
         value={chain?.displayName ?? ''}
         onValueChange={v => setChain(chains.data?.find(i => i.displayName === v))}
-        open={openSelect}
-        onOpenChange={open => setOpenSelect(open)}
```
grod220, line 32: Used to flip chevron, feels a bit unnecessary

apps/minifront/src/components/ibc/ibc-out/chain-selector.tsx
```diff
@@ -41,7 +37,7 @@ export const ChainSelector = () => {
             )}
           </SelectValue>
         </SelectTrigger>
-        <SelectContent className='left-[-17px]'>
```
grod220, line 44: Manually offsetting not necessary any longer

packages/ui/components/ui/select/select.tsx
grod220, line 1: Most updates here came from the new shadcnui update. However, the light/dark mode does not feel like the best API & usage here. Given it will be thrown away with v2, think we can live with it for now

VanishMax (approved):
@grod220 don't we have the Storybook deployment to a preview URL now? I wanted to check if the v2 select is ok after the version bump but couldn't

grod220:
V2 selector's do not use the `<Select>` component from Shadcnui. Only the two drop downs on the shield page. Checked locally regardless, all good!

PR #1800, opened 2024-09-20
fix(ui, minifront): v2 nav style fixes

VanishMax:
- Adds a temporary `zIndex` prop to the Dialog component. It is only needed for minifront v1 and must be removed when we stop supporting the v1.
- Improves the styles of the v2 sync bar and popover like this (notice the beautiful background):

image

image


grod220 (approved):
LGTM

PR #1805, opened 2024-09-23
Detect relevant IbcRelay actions

grod220:
At the moment, after a user makes a Ics20Transfer, Prax is not indexing the transaction for the IbcRelay action. This can cause confusion for users wondering the status of transaction.

This PR detects the relay action and, if the user is the recipient, adds it to the tx db.

### DIFF
--- a/apps/minifront/src/components/tx-details/tx-viewer.tsx
+++ b/apps/minifront/src/components/tx-details/tx-viewer.tsx
@@ -48,7 +48,7 @@ export const TxViewer = ({ txInfo }: { txInfo?: TransactionInfo }) => {
 
   // use React-Query to invoke custom hooks that call async translators.
   const { data: receiverView } = useQuery(
-    ['receiverView', txInfo, option],
+    ['receiverView', txInfo?.toJson({ typeRegistry }), option],
     () =>
       fetchReceiverView(
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- TODO: justify
--- a/packages/protobuf/src/registry.ts
+++ b/packages/protobuf/src/registry.ts
@@ -1,4 +1,4 @@
-import { IMessageTypeRegistry, createRegistry } from '@bufbuild/protobuf';
+import { createRegistry, IMessageTypeRegistry } from '@bufbuild/protobuf';
 
 import * as ibcCore from './services/cosmos-ibc-core.js';
 import * as penumbraCnidarium from './services/penumbra-cnidarium.js';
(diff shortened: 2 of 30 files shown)


packages/protobuf/src/registry.ts
```diff
@@ -41,6 +42,9 @@ export const typeRegistry: IMessageTypeRegistry = createRegistry(
   // gen/ibc/core/channel/v1/tx_pb
   MsgRecvPacket,
 
+  // gen/ibc/core/client/v1/tx_pb
+  MsgUpdateClient,
+
```
grod220, line 47: Required to add the the type registry as an IbcRelay transaction sent from the extension to the frontend needs to be able to serialize the `Any` inside of it & a `MsgUpdateClient` is accompanied with a `MsgRecvPacket`

grod220:
Some type registry updates are needed for this PR. Will re-request review when ready.

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
grod220, line 1: Screenshot 2024-09-24 at 11 19 12 AM

apps/minifront/src/components/tx-details/tx-viewer.tsx
```diff
@@ -48,7 +48,7 @@ export const TxViewer = ({ txInfo }: { txInfo?: TransactionInfo }) => {
 
   // use React-Query to invoke custom hooks that call async translators.
   const { data: receiverView } = useQuery(
-    ['receiverView', txInfo, option],
+    ['receiverView', txInfo?.toJson({ typeRegistry }), option],
```
grod220, line 51: Now that we are saving relay transactions, there is a change of an `Any` field needing to be serialized. For that reason, it needs access to the registry to know how to serialize/deserialize. Hence, why you see `typeRegistry` being added to a number of from/toJson() methods.

packages/protobuf/src/registry.ts
```diff
@@ -38,9 +36,6 @@ export const typeRegistry: IMessageTypeRegistry = createRegistry(
   ClientState,
   Header,
 
-  // gen/ibc/core/channel/v1/tx_pb
-  MsgRecvPacket,
-
```
grod220, line 43: Not necessary to list specifically in the registry as it exists in `IbcChannelMsgService` already

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
```diff
@@ -0,0 +1,141 @@
...
+const parsePacket = ({ packet }: MsgRecvPacket): FungibleTokenPacketData | undefined => {
+  if (!packet?.data) {
+    return undefined;
+  }
+
+  try {
+    const dataString = new TextDecoder().decode(packet.data);
+    return FungibleTokenPacketData.fromJsonString(dataString);
+  } catch (e) {
+    return undefined;
```
vacekj, line 24: Suggestion: should we error-log here, or display something to the user even?

vacekj (approved):
lgtm

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
grod220: Hmm, perhaps we do. I'll handle the case where the packet cannot be parsed and display `Unknown packet data` in the row.

PR #1810, opened 2024-09-25
Ack & timeout relay msgs

grod220:
A follow up from: https://github.com/penumbra-zone/web/pull/1805

Able to display IBC relay acknowledgements & timeouts in the transaction details page:

Screenshot 2024-09-25 at 10 37 04 AM
Screenshot 2024-09-25 at 10 38 29 AM

### DIFF
--- a/packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
+++ b/packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
@@ -16,6 +16,7 @@ import { uint8ArrayToBase64 } from '@penumbra-zone/types/base64';
 import { ReactElement } from 'react';
 import { Packet } from '@penumbra-zone/protobuf/ibc/core/channel/v1/channel_pb';
 import { getUtcTime } from './isc20-withdrawal.tsx';
+import { CircleX } from 'lucide-react';
 
 // Attempt to parse data w/ fallbacks if unknown or failures
 const ParsedPacketData = ({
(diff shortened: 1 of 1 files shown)


packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
```diff
@@ -4,93 +4,119 @@ import {
...
 
-const MsgResvComponent = ({ packet }: { packet: MsgRecvPacket }) => {
-  const packetData = useMemo(() => parsePacket(packet), [packet]);
+const PacketRows = ({
+  packet,
+  dataParser,
+}: {
+  packet: Packet;
+  dataParser?: (arg: Uint8Array) => ReactElement;
+}) => {
```
grod220, line 49: Because a number of messages need to display `Packet` data, abstracted it here

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
```diff
@@ -140,5 +232,23 @@ export const IbcRelayComponent = ({ value }: { value: IbcRelay }) => {
     return <UpdateClientComponent update={update} />;
   }
 
+  if (value.rawAction?.is(MsgTimeout.typeName)) {
+    const timeout = new MsgTimeout();
+    value.rawAction.unpackTo(timeout);
```
TalDerei, line 237: suggestion: the proto definition of `rawAction` is `Any`. could there potentially be errors when unpacking? It might be a good idea to wrap the [unpackTo](https://github.com/penumbra-zone/web/pull/1810/files#diff-b9fb2f34ee8a60d23b71b17c7677f6248bdf75115b3f05d8699b7b7828adae6eR237) method in a try-catch block when deserializing the rawAction into a message type.

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
```diff
@@ -4,93 +4,119 @@ import {
...
+  packet,
+  dataParser,
+}: {
+  packet: Packet;
+  dataParser?: (arg: Uint8Array) => ReactElement;
+}) => {
+  return (
+    <>
+      <ParsedPacketData data={packet.data} dataParser={dataParser} />
+      {!!packet.sequence && (
```
TalDerei, line 53: comment: what does a sequencer number of 0 represent?

TalDerei (approved):

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
grod220: Good catch, will change any field that returns a primitive (required field) to show without a conditional check.

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
grod220: Given the conditional block it's in, it _should_ unbpack fine. However, we want this to be fault-tolerant for unexpected values or deserialization errors. Will wrap in try/catch.

packages/ui/components/ui/tx/actions-views/ibc-relay.tsx
```diff
@@ -4,93 +4,119 @@ import {
...
+        <ActionDetails.Row label='Timeout timestamp'>
+          {getUtcTime(packet.timeoutTimestamp)}
+        </ActionDetails.Row>
+      )}
+    </>
+  );
+};
+
+// Packet data stored as json string encoded into bytes
+const parseRecvPacket = (packetData: Uint8Array) => {
```
grod220, line 90: Was informed that all of these messages expect to have a `FungibleTokenPacketData` inside. Will refactor a bit for that case.

PR #1817, opened 2024-09-30
feat(ui): #1807: add Toast v2 UI component

### LINKED ISSUE #1807: Uiv2 Toasts
We currently do not have v2 designs for the Toast, however, we should re-implement our Toast component within the spirit of v2 so it can be immediately used for external projects (like Dex explorer). Task:

- Use styled components
- Aria compliant
- Re-evaluate API for low-hanging fruit improvements
- Expose via NPM

VanishMax:
Closes #1807 

Demo: https://penumbra-ui-preview--pr1817-feat-toasts-v2-sr0jtwq9.web.app/?path=/docs/ui-library-toast--docs

Implements Toasts v2. It uses the `sonner` library just like the v1 but in a different API that better aligns with the ideas of the v2 library. Example:

```tsx
import { ToastProvider, openToast } from '@penumbra-zone/ui/Toast';
import { ToastProvider, openToast } from '@penumbra-zone/ui/Button';

const Component = () => {
  const open = () => {
    const toast = openToast({
      type: 'loading',
      message: 'Loading...',
    });

    setTimeout(() => {
      toast.update({
        type: 'error',
        message: 'Failed!',
        description: 'Unknown error'
      });
    }, 2000);
  };

  return (
    <>
      <ToastProvider />
      <Button onClick={open}>Open</Button>
    </>
  );
};
```

As you can see, this API supports not only opening toasts but also mutating their props dynamically.


vacekj (approved):
lgtm, good work

PR #1826, opened 2024-10-03
Add className prop to Card for overrides

JasonMHasperhoven:
For the RouteBook we need to be able to remove the padding here:

image


grod220:
As discussed in the team meeting, think we want to avoid accepting className as prop

JasonMHasperhoven:
Yes, closing this one.

PR #1827, opened 2024-10-03
Add React.HTMLAttributes to Text prop type

JasonMHasperhoven:
This is to fix:
image


grod220:
This is only if we were to want to support className as a prop right?

JasonMHasperhoven:
I will make a new PR for this one. I believe the Text component should support a few additional props, another other styling changes/adjustments should be done in child components.

PR #1828, opened 2024-10-04
Remove validator info throttler

grod220:
Previously, the loop was using a version of `updatePriceForValidatorDelegationToken()` that made a request for metadata from the node. This caused rate-limiting issues that a throttle was put into place to slow down the requests: https://github.com/penumbra-zone/web/pull/1392.

Now that delegation metadata is generated locally (versus a request) with https://github.com/penumbra-zone/web/issues/1189, we can get rid of the throttle. 

This will also be helpful to have the speedup so the staking page is immediately populated on a fresh sync and doesn't have to wait for +1 minute to get all validator infos.

### DIFF
--- a/.changeset/tough-mayflies-collect.md
+++ b/.changeset/tough-mayflies-collect.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/query': patch
+---
+
+Remove validator info throttle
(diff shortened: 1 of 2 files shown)


VanishMax (approved):
good improvement 👍

PR #1847, opened 2024-10-04
Add Slim Density

JasonMHasperhoven:
For the dex-explorer we need to support another variation of density that's smaller than compact: slim.

This PR changes the API slightly for the Density component. Instead of passing compact or sparse as prop directly, it now needs to be passed as value for the density prop.

It also adds the slim variation to the table cell css.

### DIFF
--- a/.changeset/rotten-gorillas-brush.md
+++ b/.changeset/rotten-gorillas-brush.md
@@ -1,0 +1,6 @@
+---
+'minifront': minor
+'@penumbra-zone/ui': minor
+---
+
+Add Slim Density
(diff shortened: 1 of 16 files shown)


grod220:
Just checking, have you synced with @smmhrdmn on the proposed density update?

smmhrdmn:
@JasonMHasperhoven is this a response to just the tables, or a broader issue? if it's just the tables, i've been refining them and have some updates that should address this. in general, i think it makes more sense to refine our existing props instead of adding new ones, since many of the components won't be affected by the new prop. but obviously im open to whatever addresses the needs more directly

JasonMHasperhoven:
Closing this as it's outdated.

PR #1848, opened 2024-10-04
Add Text props

JasonMHasperhoven:
Adding additional props that are common and widely used to the Text component.

If you can think of more props that would be good to add, please mention it here.

### DIFF
--- a/packages/ui/src/Text/index.tsx
+++ b/packages/ui/src/Text/index.tsx
@@ -19,75 +19,84 @@ import { ReactNode } from 'react';
 interface StyledProps {
   $truncate?: boolean;
   $color?: (color: DefaultTheme['color']) => string;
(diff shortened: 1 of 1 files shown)


packages/ui/src/Text/index.tsx
```diff
@@ -293,45 +318,60 @@ const omit = <ObjectType extends Record<string, unknown>>(
...
  * ```
  */
-export const Text = ({ truncate, color, ...props }: TextProps) => {
+export const Text = ({
+  truncate,
+  color,
+  align,
+  decoration,
+  transform,
+  break: breakProp,
```
JasonMHasperhoven, line 327: break is a reserved js word

VanishMax (approved):
good to me!

vacekj (approved):

PR #1854, opened 2024-10-14
fix(ui): #1803: fix `copyable` prop on `AddressViewComponent`

### LINKED ISSUE #1803: isCopyable attribute on AddressView doesn't work
**Describe the bug**
Passing the `isCopyable` attribute as `false` on `<AddressView />` doesn't seem to disable and hide the copy button.

**To Reproduce**

1. Use AddressView in a React component
2. Pass `isCopyable={false}`
3. Observe the copy button still being present and clickable

**Expected behavior**
`isCopyable={false}` should hide and disable the copy button.

VanishMax:
Fixes #1803

### DIFF
--- a/.changeset/perfect-rocks-type.md
+++ b/.changeset/perfect-rocks-type.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/ui': patch
+---
+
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
nullish coalescing operator works well here

PR #1855, opened 2024-10-15
Noble forwarding wasm helpers

grod220:
Related to: https://github.com/prax-wallet/prax/issues/206 & https://github.com/penumbra-zone/penumbra/pull/4878

Exposes helpers needed to generate noble forwarding addresess.

### DIFF
--- a/packages/wasm/crate/tests/test_keys.rs
+++ b/packages/wasm/crate/tests/test_keys.rs
@@ -6,8 +6,8 @@ use penumbra_keys::keys::{
 use penumbra_keys::{Address, FullViewingKey};
 use penumbra_proto::{DomainType, Message};
 use penumbra_wasm::keys::{
-    generate_spend_key, get_address_by_index, get_full_viewing_key, get_wallet_id,
-    is_controlled_address,
+    forwarding_addr_inner, generate_spend_key, get_address_by_index, get_full_viewing_key,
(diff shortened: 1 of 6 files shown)


JasonMHasperhoven (approved):

JasonMHasperhoven:
Recommend that someone else looks at this before merging since I'm not too familiar with rust.

vacekj (approved):
LGTM.

packages/wasm/crate/tests/test_keys.rs
```diff
@@ -87,3 +87,39 @@ fn returns_false_on_unknown_addr() {
             .0;
     assert!(!is_controlled_address(&fvk.encode_to_vec(), &other_address.encode_to_vec()).unwrap());
 }
+
+#[test]
```
TalDerei, line 91: what about additional test coverage for:

1. edge cases near the sequence range boundaries (eg. check addresses are different for sequence numbers 0 and 2^16 - 1)

2. same sequence and account generate the same address

3. different full viewing keys yield different addresses

TalDerei (approved):
wasm achieves parity with core implementation

packages/wasm/crate/tests/test_keys.rs
grod220: Adding those additional cases 👍

Issue #1814, opened 2024-09-26
minimize storage accesses 


TalDerei:
`"During the development of the native view server, our biggest cost was actually disk i/o, and we did some optimizations to avoid writes when possible. Why do we expect that parallelizing trial decryption will give a higher performance ROI than minimizing storage accesses?"`

Minimizing storage accesses and disk I/O should be profiled and investigated as potential low-hanging fruit for improving performance.

Issue #1726, opened 2024-08-22
introduce atomicity to database saves in block processor


TalDerei:
processing blocks in the block processor isn't actually atomic, which introduces the possibility of local state corruption.

references this [comment](https://github.com/penumbra-zone/web/issues/34#issuecomment-2301783429)

grod220:
Spent a bit of time thinking about this today. In order to make the remaining saves atomic, we need:
1. A staging area for saves
2. Updates in-block db getters to look in staging area prior to db

Idb has support for atomic updates. We make use of it for [tct updates](https://github.com/penumbra-zone/web/blob/75ef4f5bb7bb07c4ea0478bee360beba12220f28/packages/storage/src/indexed-db/index.ts#L213-L222). However, the staging area is important. We need a way to queue up all of the save actions that can later (within that same block) be accessed as if they were already committed.

For example, the infamous [auction upsert loop](https://github.com/penumbra-zone/web/blob/75ef4f5bb7bb07c4ea0478bee360beba12220f28/packages/query/src/block-processor.ts#L345-L373). The comment + code placement is there because it depends on the auction NFT metadata already saved in the database before it can associate a note with the NFT.

Another example is [handling epoch transitions](https://github.com/penumbra-zone/web/blob/75ef4f5bb7bb07c4ea0478bee360beba12220f28/packages/query/src/block-processor.ts#L701-L703). It relies upon requesting the appParams from the database. However, the block _could_ have updated these. In an atomic world, it would not know for sure without checking the staging area first.

Queuing up saves (and at the end committing them all at once) is definitely possible. But I believe we need to add a staging area and update the getters to reference these prior to database lookups.

TalDerei:
precursor to this should be addressing https://github.com/penumbra-zone/web/pull/1720, which distinguishes between tasks that should be managed by idb versus those that belong to the block processor. 

staging areas are idiomatic for implementing atomicity, but we should consider the granularity of these atomic operations. for instance, should we perform atomic updates every block, or buffer updates and then apply them atomically every 1000 blocks during [flushing](https://github.com/penumbra-zone/web/blob/main/packages/query/src/block-processor.ts#L283)?

TalDerei:
maybe we should prioritize this next sprint following https://github.com/penumbra-zone/web/pull/1858?

PR #1858, opened 2024-10-16
database batching 

TalDerei:
references https://github.com/penumbra-zone/web/issues/1814

collectively yields **~8%** syncing improvement. 

Screenshot 2024-10-15 at 10 01 53 PM

Profiling indicates that the block processor performs substantially more database reads than writes. The most resource-intensive DB operations are "saveScanResult," "identifyNewAssets," and "resolveNullifiers," which account for the majority of disk I/O operations.

Currently, storage operations are executed sequentially (ie. saving transactions, metadata, etc.) in indexDB. Storage R/W blocks sync since each storage operation has an associated `await`, causing each to wait for the previous one to finish before continuing.

Those operations were refactored to use `Promise.all()`, ensuring the block processor isn't _artificially_ blocked on each `await` call. This means a concurrent execution of the promises, at the javascript level, meaning the calls will be _initiated_ at the same time.  But, the actual execution at the IndexedDB level won't be truly parallel because IndexedDB itself is single-threaded and performs DB operations one at a time. IndexedDB will handle them in a queue and execute them sequentially in an atomic transaction. 

The benefit of using `Promise.all()` here is the concurrent execution of **non-blocking** operations, reducing wait times between operations like network fetches and other async tasks. Therefore, other operations are not blocked while waiting for each individual write to complete. 

tldr; disk I/O doesn't seem to be a primary bottleneck. Still, disk I/O especially when filtered through the browser API / network stack is always going to be relatively slow.

### DIFF
--- a/packages/query/src/block-processor.ts
+++ b/packages/query/src/block-processor.ts
@@ -263,11 +263,11 @@ export class BlockProcessor implements BlockProcessorInterface {
 
     // flushing is slow, avoid it until
     // - wasm says
-    // - every 1000th block
+    // - every 5000th block
     // - every block at tip
     const flushReasons = {
       scannerWantsFlush,
-      interval: compactBlock.height % 1000n === 0n,
+      interval: compactBlock.height % 5000n === 0n,
       new: compactBlock.height > latestKnownBlockHeight,
     };
 
@@ -431,14 +431,18 @@ export class BlockProcessor implements BlockProcessorInterface {
   }
 
   private async identifyNewAssets(notes: SpendableNoteRecord[]) {
+    const saveOperations = [];
+
     for (const note of notes) {
       const assetId = note.note?.value?.assetId;
       if (!assetId) {
         continue;
       }
 
-      await this.saveAndReturnMetadata(assetId);
+      saveOperations.push(this.saveAndReturnMetadata(assetId));
     }
+
+    await Promise.all(saveOperations);
   }
 
   // TODO: refactor. there is definitely a better way to do this.  batch
@@ -491,11 +495,25 @@ export class BlockProcessor implements BlockProcessorInterface {
   // Nullifier is published in network when a note is spent or swap is claimed.
   private async resolveNullifiers(nullifiers: Nullifier[], height: bigint) {
     const spentNullifiers = new Set<Nullifier>();
+    const readOperations = [];
+    const writeOperations = [];
 
     for (const nullifier of nullifiers) {
-      const record =
-        (await this.indexedDb.getSpendableNoteByNullifier(nullifier)) ??
-        (await this.indexedDb.getSwapByNullifier(nullifier));
+      const readPromise = (async () => {
+        const record =
+          (await this.indexedDb.getSpendableNoteByNullifier(nullifier)) ??
+          (await this.indexedDb.getSwapByNullifier(nullifier));
+        return { nullifier, record };
+      })();
+
+      readOperations.push(readPromise);
+    }
+
+    // Await all reads in parallel
+    const readResults = await Promise.all(readOperations);
+
+    // Process the read results and queue up write operations
+    for (const { nullifier, record } of readResults) {
       if (!record) {
         continue;
       }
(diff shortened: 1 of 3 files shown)


packages/query/src/block-processor.ts
```diff
@@ -263,11 +263,11 @@ export class BlockProcessor implements BlockProcessorInterface {
 
     // flushing is slow, avoid it until
     // - wasm says
-    // - every 1000th block
+    // - every 5000th block
     // - every block at tip
     const flushReasons = {
       scannerWantsFlush,
-      interval: compactBlock.height % 1000n === 0n,
+      interval: compactBlock.height % 5000n === 0n,
```
TalDerei, line 270: Increasing the chunk size for flush operations improves performance.

packages/storage/src/indexed-db/updater.ts
```diff
@@ -41,10 +41,19 @@ export class IbdUpdater {
...
+        .objectStore(update.table)
+        .put(update.value, update.key)
+        .then(() => {
+          this.notifySubscribers(update);
+        });
+      batchOperations.push(putOperation);
     }
+    await Promise.all(batchOperations);
 
     await tx.done;
```
grod220, line 58: thought: while I think the Promise.all() change is a good one, this doesn't change the atomicity of this operation. The transaction is building up the operations, but not actually committing anything until the even loop is released. The `await tx.done` tells us when that is. Perhaps you have a different understanding of this though?

packages/storage/src/indexed-db/updater.ts
```diff
@@ -41,10 +41,19 @@ export class IbdUpdater {
...
+
+    // Batch all the updates into promises
     for (const update of updates.all) {
-      await tx.objectStore(update.table).put(update.value, update.key);
-      this.notifySubscribers(update);
+      const putOperation = tx
+        .objectStore(update.table)
+        .put(update.value, update.key)
+        .then(() => {
+          this.notifySubscribers(update);
```
grod220, line 52: issue: we should put this below `await tx.done` for all the updates given they are not committed yet

packages/query/src/block-processor.ts
grod220: Of all the changes in the PR, is it not this change that represents the nearlly all of the sync speed improvement?

grod220 (commented):
I wonder if this PR kinda skirts around the core issue which is https://github.com/penumbra-zone/web/issues/1726. Aka, should we not convert our IDB implementation to a transaction+commit model instead?

packages/query/src/block-processor.ts
TalDerei: > This is something worth thinking about. Why not make the interval say 10k blocks or even bigger? I wonder what is the limiting principle for this.

the trade-off here is between memory pressure and disk I/O pressure. The longer the block processor waits to flush, the more updates accumulate in memory, but less frequent writes leading to lower I/O costs. 

In my testing, I there was a noticeable improvement when increasing the interval from 1k to 5k, but only a marginal change from 5k to 10k.

TalDerei: > Also, we sync on `scannerWantsFlush`. This will be true when the fvk successfully decrypts something from a block. What if we don't do that though. Would it be bad if the user just waits until the next interval? Wonder if that would save time as well.

I had the same thought, and it was noticeably more performant, but it resulted in SCT state corruption.

TalDerei: > Of all the changes in the PR, is it not this change that represents the nearlly all of the sync speed improvement?

Increasing the interval improved performance on its own, but combining it with these other changes enhanced it even further.

packages/storage/src/indexed-db/updater.ts
TalDerei: shouldn't we just keep the status quo here and just handle block processor atomicity separately in https://github.com/penumbra-zone/web/issues/1726?

TalDerei: good call 👍

packages/query/src/block-processor.ts
grod220: > I had the same thought, and it was noticeably more performant, but resulted in SCT state corruption. I can investigate further.

That feels like it should not be the case 🤔 . Can you create an issue to follow up investigating why we can't flush on intervals but are required to flush on finding new notes/swaps?

grod220 (approved):

packages/storage/src/indexed-db/updater.ts
```diff
@@ -41,15 +41,12 @@ export class IbdUpdater {
     const tables = updates.all.map(u => u.table);
     const tx = this.db.transaction(tables, 'readwrite');
 
-    const batchOperations = [];
-
     // Batch all the updates into promises
     for (const update of updates.all) {
-      const putOperation = tx.objectStore(update.table).put(update.value, update.key);
-      batchOperations.push(putOperation);
+      tx.objectStore(update.table).put(update.value, update.key);
```
TalDerei, line 46: Gabe rightly identified that performing an await inside IndexedDB transaction can risk closing the transaction prematurely in the event loop, which we shouldn't do.

Issue #1829, opened 2024-10-04
Get and order by priority from registry


grod220:
At the moment, minifront displays assets in arbitrary order:

<img width="687" alt="Screenshot 2024-10-04 at 12 34 11 PM" src="https://github.com/user-attachments/assets/88a9b439-e27a-442d-b17c-db44b1ce5b82">

The issue is that not all of these assets are equally important. Ideally, the most relevant ibc assets get shown on top.

For that, minifront needs to rely upon [Metadata priority score](https://buf.build/penumbra-zone/penumbra/docs/430d52207ecc47dbacae84b596bc99db:penumbra.core.asset.v1#penumbra.core.asset.v1.Metadata) which it gets when it requests from the rpc (in `BalancesRequest` and `AssetsRequest`).

Task:
- Add priority scores to registry for assets: https://github.com/prax-wallet/registry
- Update rpc methods that return denom metadata to reference registry priority scores
- Update minifront to utilize priority score for ordering in assets selector

TalDerei:
closed by https://github.com/prax-wallet/registry/pull/96?

grod220:
That's only part 1, we need to implement it into minifront & use in rpc. In progress: https://github.com/penumbra-zone/web/pull/1860

PR #1860, opened 2024-10-17
Sort assets by priority score

JasonMHasperhoven:
As part of https://github.com/penumbra-zone/web/issues/1829

image

### DIFF
--- a/apps/minifront/src/fetchers/assets.ts
+++ b/apps/minifront/src/fetchers/assets.ts
@@ -6,7 +6,9 @@ import { penumbra } from '../penumbra';
 
 export const getAllAssets = async (): Promise<Metadata[]> => {
   const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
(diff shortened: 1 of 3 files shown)


grod220 (approved):
Ah so much better

TalDerei:
still think shitmos should be hardcoded with the highest priority

apps/minifront/src/fetchers/assets.ts
```diff
@@ -6,7 +6,9 @@ import { penumbra } from '../penumbra';
 
 export const getAllAssets = async (): Promise<Metadata[]> => {
   const responses = await Array.fromAsync(penumbra.service(ViewService).assets({}));
-  return responses.map(getDenomMetadata);
+  return responses
+    .map(getDenomMetadata)
+    .toSorted((a, b) => Number(b.priorityScore) - Number(a.priorityScore));
```
VanishMax, line 11: question: can we also apply this to the `groupAndSort.ts` file in the UI package?

PR #1867, opened 2024-10-22
SpendMax PR Suggestions

grod220:


TalDerei (approved):
This looks good to me, merge at your discretion!

here's what I'm generally concerned about – are there potential edge cases where components render before the relevant data (`useGasPrices` or `useBalancesResponses`) is fully loaded? Could this lead to stale data during the rendering phase that waterfalls into inconsistent behavior?

one example that comes to mind might be the `NonNativeFeeWarning` component showing a warning even when fees could be covered, and this inconsistency down streams into the `hasStakingToken` function call?

apps/minifront/src/components/shared/non-native-fee-warning.tsx
```diff
@@ -32,10 +94,18 @@ export const NonNativeFeeWarning = ({
    * />
    * ```
    */
-  hasStakingToken: boolean;
   wrap?: (children: ReactNode) => ReactNode;
 }) => {
-  const shouldRender = !!amount && !hasStakingToken;
+  const { data } = useGasPrices();
+  const stakingTokenMetadata = useStakingTokenMetadata();
```
TalDerei, line 100: nit; can pass the `stakingTokenMetadata` field directly into `NonNativeFeeWarning` instead of fetching it again within the react component.

grod220:
> are there potential edge cases where components render before the relevant data (specifically useGasPrices or useBalancesResponses) is fully loaded?

Whenever we do async in hooks, we should always propagate data, error, and loading states to the consumers of them. If those haven't fully loaded, I'm pretty sure the empty array makes it so it doesn't display. But then after loading, it comes into view if relevant.

Going to merge, but feel free to make any final adjustments to the main PR and I'll check with the beta testers to have at it 👍

apps/minifront/src/components/shared/non-native-fee-warning.tsx
```diff
@@ -32,10 +94,18 @@ export const NonNativeFeeWarning = ({
    * />
    * ```
    */
-  hasStakingToken: boolean;
   wrap?: (children: ReactNode) => ReactNode;
 }) => {
-  const shouldRender = !!amount && !hasStakingToken;
+  const { data } = useGasPrices();
```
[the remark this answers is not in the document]
grod220: I think we should always be handling error/loading states

apps/minifront/src/components/shared/non-native-fee-warning.tsx
grod220: Given they pull from the same cache, it isn't like it fetches more than once. But up to you!

TalDerei: right, good point

PR #1868, opened 2024-10-22
Fix asset selector onchange event

grod220:
Before:

https://github.com/user-attachments/assets/27af6cf7-87de-4a12-b56c-683ea6ced49b

After:

https://github.com/user-attachments/assets/5248f3db-83f1-48ce-b7ac-9c8d06c5b980

When a user clicks on the item, the `onMouseDown` event fires, the unmounting of the component prevents the `onClick` from firing.

### DIFF
--- a/.changeset/proud-ducks-marry.md
+++ b/.changeset/proud-ducks-marry.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


JasonMHasperhoven (approved):

grod220:
Fyi, made a fix that restored keyboard controls for this element

Issue #1830, opened 2024-10-04
Convert build_action() tests


grod220:
Current [build tests](https://github.com/penumbra-zone/web/blob/8bf66eafdc7d9654d6a9df65921f2b532745d9b3/packages/wasm/crate/tests/test_build.rs#L49) uses deserializable input types. This makes it more difficult to maintain as one has to manually write json to run the test.

Task: Migrate this test to use `MockDb`. Will need to make a `build_action_inner()` function to test that does not use any JsValue types as arguments. 

Should also delete runtime `mock_test_database`.

===

Additional: `build()` and `build_parallel()` should be relocated to build.rs file

PR #1875, opened 2024-10-26
wasm: build action tests

TalDerei:
references https://github.com/penumbra-zone/web/issues/1830


packages/wasm/crate/tests/test_build.rs
```diff
@@ -1,577 +1,398 @@
...
+    let spendable_note_json = r#"
+    {
+        "note_commitment": {
+            "inner": "MY7PmcrH4fhjFOoMIKEdF+x9EUhZ9CS/CIfVco7Y5wU="
+        },
+        "note": {
+            "value": {
+                "amount": {
+                    "lo": "1000000",
+                    "hi": "0"
```
grod220, line 109: Maintaining these strings manually feels like a chore. Can we just test the inner function of the build and not feel like we need to test the deserialization logic?

packages/wasm/crate/tests/utils/sct.rs
grod220, line 1: Can't we import these from core?

TalDerei: Some of the internal fields are private, which is why I emulated the types to match their structure.

packages/wasm/crate/tests/test_build.rs
TalDerei: replaced the hardcoded json with their corresponding typed structures.

packages/wasm/crate/tests/test_build.rs
```diff
@@ -1,577 +1,374 @@
...
+
+    // Execute parallel spend transaction and generate proof.
+    let parallel_transaction = build_parallel_inner(
+        actions,
+        transaction_plan.clone(),
+        witness_data.clone(),
+        authorization_data.clone(),
+    )
+    .unwrap();
+    console_log!("Parallel transaction is: {:?}", parallel_transaction);
```
grod220, line 355: nit: we probably don't want these console logs right?

packages/wasm/crate/tests/utils/planner_setup.rs
```diff
@@ -9,6 +9,7 @@ use penumbra_wasm::database::interface::Database;
 use penumbra_wasm::database::mock::MockDb;
 use penumbra_wasm::storage::{byte_array_to_base64, Tables};
 
+#[allow(dead_code)]
```
grod220, line 12: question: should we just delete this?

packages/wasm/crate/src/database/indexed_db.rs
```diff
@@ -16,59 +15,10 @@ pub async fn open_idb_database(constants: &DbConstants) -> WasmResult<IdbDatabas
     #[allow(unused_mut)]
     let mut db_req: OpenDbRequest = IdbDatabase::open_u32(&constants.name, constants.version)?;
 
-    // Conditionally mock sample `IdbDatabase` database for testing purposes
-    #[cfg(feature = "mock-database")]
-    let db_req = mock_test_database(db_req).into_future().await;
-
```
grod220, line 22: It feels very nice deleting this

grod220 (approved):
Great job!

packages/wasm/crate/tests/test_build.rs
TalDerei: removed

packages/wasm/crate/tests/utils/planner_setup.rs
TalDerei: removed

PR #1887, opened 2024-10-31
Add theme & lib/toast UI exports

JasonMHasperhoven:
Currently we can't access theme and lib/toast within the dex-explorer. This should fix that.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -57,8 +57,8 @@
         "default": "./dist/lib/toast/*.js"
       },
       "./utils/*": {
-        "types": "./src/utils/*.d.ts",
-        "default": "./src/utils/*.js"
+        "types": "./dist/src/utils/*.d.ts",
+        "default": "./dist/src/utils/*.js"
       },
       "./*": {
         "types": "./dist/src/*/index.d.ts",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -33,6 +33,7 @@
     "./tailwind": "./src/tailwindConfig.ts",
     "./postcss.config.js": "./postcss.config.js",
     "./styles/*": "./styles/*",
+    "./theme": "./src/PenumbraUIProvider/theme.ts",
     "./hooks/*": "./src/hooks/*/index.ts",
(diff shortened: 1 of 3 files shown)


TalDerei (approved):

packages/ui/package.json
```diff
@@ -46,6 +48,18 @@
         "types": "./dist/components/ui/*.d.ts",
         "default": "./dist/components/ui/*.js"
       },
+      "./theme": {
+        "types": "./dist/src/PenumbraUIProvider/theme.d.ts",
+        "default": "./dist/src/PenumbraUIProvider/theme.js"
+      },
+      "./lib/toast/*": {
```
VanishMax, line 55: issue: this is the old toast library, it should not be exported. The v2 one is already working in the dex: https://github.com/penumbra-zone/dex-explorer/pull/118/files#diff-e1b5cf1eb29b361c28415989517c8312f1f349ab3e0cac44425ffc02290259c2R62

packages/ui/package.json
```diff
@@ -46,6 +48,18 @@
         "types": "./dist/components/ui/*.d.ts",
         "default": "./dist/components/ui/*.js"
       },
+      "./theme": {
```
VanishMax, line 51: question: why is the theme needed to be exported, especially this way? We use `PenumbraUIProvider` in the `app/app.tsx` in the DEX.

suggestion: let's just export the `theme` object from `PenumbraUIProvider/index.tsx` file

packages/ui/package.json
```diff
@@ -33,6 +33,8 @@
     "./tailwind": "./src/tailwindConfig.ts",
     "./postcss.config.js": "./postcss.config.js",
     "./styles/*": "./styles/*",
+    "./theme": "./src/PenumbraUIProvider/theme.ts",
+    "./utils/*": "./src/utils/*.ts",
```
VanishMax, line 37: question: is there something in the v1 utils we need in the DEX?

VanishMax (commented):
If possible, I believe this PR should be re-done

PR #1899, opened 2024-11-04
Add ReactNode to Toast description prop type

JasonMHasperhoven:
Add ReactNode to Toast description prop type.

Needed to provide a progress bar to the transaction toast.


VanishMax (approved):
nitpick: in `string | ReactNode`, the `string |` part could be removed

JasonMHasperhoven:
> nitpick: in `string | ReactNode`, the `string |` part could be removed

Just curious, why is it preferred?

PR #1903, opened 2024-11-08
Update connection copy

grod220:


TalDerei (approved):
missing changeset, LGTM

grod220:
Not convinced we really need one for minifront as we aren't really versioning it for anything meaningful at the moment. But perhaps we should think about that more.

PR #1907, opened 2024-11-12
concurrency settings

TalDerei:


hdevalence (approved):
Why does this fix it?

TalDerei:
> Why does this fix it?

we added another package, `ui-deprecated`, which includes a package.json with a `dev:pack` step. 
It uses [turbo](https://turbo.build/) to execute the dev:pack script across packages, and turbo failed to allocate enough threads for the task. We can remove the dev:pack config from that package once we fully transition to the v2 UI package.

PR #1908, opened 2024-11-15
Reintroduce Slider component with tailwind

JasonMHasperhoven:
Will be part of the order form:

image


vacekj (approved):

JasonMHasperhoven:
> Checked the Storybook, the component works nicely.
> 
> question: how difficult was it for you personally to switch it to Tailwind?

I'm a newb to tailwind, I learned that it doesn't support template strings for classnames so I bumped into this as an issue. Otherwise it's quite easy and straightforward.

PR #1909, opened 2024-11-15
Script to enable `pnpm link`

JasonMHasperhoven:
Currently pnpm linking does not work because the exports inside the package.json files of the packages refer to the src files:

```
"exports": {
    ".": "./src/index.ts",
    "./*": "./src/*.ts"
  },
 ```
 
 This script can temporarily set the exports to the value of the `publishConfig.exports`, so that the exports will point to the dist folder.

It would work like this:
```
// inside penumbra-zone/web
node scripts/link-externally.js enable
cd packages/ui && dev:pack

// inside dex-explorer
pnpm link ../penumbra-zone/web/packages/ui

// to revert:
node scripts/link-externally.js disable
```
 
> PS: Possible improvement would be to directly build a monorepo's package within the dex-explorer as part of the build pipeline.

### DIFF
--- a/scripts/link-externally.js
+++ b/scripts/link-externally.js
@@ -2,6 +2,17 @@ import fs from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 
+/**
+ * This script is used to link dependencies from this monorepo to other repos.
+ *
+ * @usage:
+ *   node scripts/link-externally.js enable
+ *   node scripts/link-externally.js disable
+ *
+ *   // when enabled, you can link the package to another repo like this:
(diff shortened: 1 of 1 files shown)


VanishMax (approved):
Good solution for this problem!

vacekj:
Nice, lgtm.

vacekj (approved):

grod220:
Would be helpful to get a demo of this local workflow in an upcoming sync

JasonMHasperhoven:
> Would be helpful to get a demo of this local workflow in an upcoming sync

Sure thing!

PR #1913, opened 2024-11-18
feat(ui): toggle component and more

VanishMax:
Changeset:
- Add Toggle UI component – adapt it from `ui-deprecated`
- Add `medium` density – new density type introduced in the DEX Figma, implemented in at least Tabs component
- Add `xxs` Text style – new typography type with 11px font-size
- Improve the styles of `Tabs` component – add support for `medium` density

Check here: https://penumbra-ui-preview--pr1913-fix-tab-styles-tf62cbwe.web.app/

### DIFF
--- a/packages/ui/src/Density/index.tsx
+++ b/packages/ui/src/Density/index.tsx
@@ -1,25 +1,10 @@
 import { ReactNode } from 'react';
 import { Density as TDensity, DensityContext } from '../utils/density';
 
-/**
- * Utility interface to be used below to ensure that only one density type is used at a time.
- */
-interface NeverDensityTypes {
-  sparse?: never;
-  medium?: never;
-  compact?: never;
-}
-
 export type DensityPropType =
-  | (Omit<NeverDensityTypes, 'sparse'> & {
-      sparse: true;
-    })
-  | (Omit<NeverDensityTypes, 'medium'> & {
-      medium: true;
-    })
-  | (Omit<NeverDensityTypes, 'compact'> & {
-      compact: true;
-    });
+  | { sparse: true; medium?: never; compact?: never }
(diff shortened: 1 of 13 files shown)


packages/ui/src/Density/index.tsx
```diff
@@ -1,11 +1,29 @@
...
+  | (Omit<NeverDensityTypes, 'compact'> & {
+      compact: true;
+    });
+
+export type DensityProps = DensityPropType & {
   children?: ReactNode;
-} & (SelectedDensity extends 'sparse'
-  ? { sparse: true; compact?: never }
-  : { compact: true; sparse?: never });
+};
```
grod220, line 26: question: can't we do this?

```ts
export type Density = 'sparse' | 'medium' | 'compact';

export type DensityProps = {
  density: Density;
  children?: ReactNode;
};
```

grod220: though if we specifically need those names as a field, we could also do this:

```ts
export type DensityPropType =
  | { sparse: true; medium?: never; compact?: never }
  | { medium: true; sparse?: never; compact?: never }
  | { compact: true; sparse?: never; medium?: never };

export type DensityProps = DensityPropType & {
  children?: ReactNode;
};
```

grod220 (approved):

packages/ui/src/Density/index.tsx
VanishMax: Density and Text components initially had this style of props that is more neat: `<Density compact/>`, `<Text body/>`. I personally don't mind the union type but it is already a bit too late to change – so many components depend on a shorter syntax.

Changed the types to the version from your second comment

PR #1915, opened 2024-11-19
Shortened amount props for ValueViewComponent

grod220:
Updating `ValueViewComponent` to support displaying for this view on the dex explorer:

Screenshot 2024-11-19 at 5 48 29 PM

At the moment those display raw numbers, but given the metadata of the assets, we need them to be `ValueViewComponent` to handle the display exponents.

### DIFF
--- a/packages/types/src/round.ts
+++ b/packages/types/src/round.ts
@@ -1,0 +1,39 @@
+import { ceil as lodashCeil, floor as lodashFloor, round as lodashRound } from 'lodash';
+
+export type RoundingMode = 'round' | 'ceil' | 'floor';
+
+export interface RoundOptions {
+  value: number;
+  decimals: number;
+  roundingMode?: RoundingMode;
+}
+
+const roundingStrategies = {
+  ceil: lodashCeil,
+  floor: lodashFloor,
+  round: lodashRound,
+} as const;
+
+/**
+ * Rounds a number based on the specified options.
+ *
+ * @param options - An object containing the properties:
+ *   - value: The number to round.
+ *   - decimals: The number of decimal places to round to.
+ *   - roundingMode: The mode of rounding ('round', 'ceil', 'floor'). Defaults to 'round'.
+ *
+ * @returns A string representation of the rounded number.
+ *
+ * @example
+ *
+ * ```typescript
+ * round({ value: 1.2345, decimals: 2, roundingMode: 'ceil' }); // "1.24"
+ * round({ value: 1.2345, decimals: 2, roundingMode: 'floor' }); // "1.23"
+ * round({ value: 1.2345, decimals: 2 }); // "1.23" (default rounding)
+ * ```
+ */
+export function round({ value, decimals, roundingMode = 'round' }: RoundOptions): string {
+  const roundingFn = roundingStrategies[roundingMode];
+  const roundedNumber = roundingFn(value, decimals);
+  return roundedNumber.toFixed(decimals);
+}
--- a/packages/ui/src/ValueView/index.tsx
+++ b/packages/ui/src/ValueView/index.tsx
@@ -8,6 +8,8 @@ import { Text } from '../Text';
 import { AssetIcon } from '../AssetIcon';
 import { Density, useDensity } from '../utils/density';
 import cn from 'clsx';
+import { shortify } from '@penumbra-zone/types/shortify';
+import { detailTechnical, technical } from '../utils/typography.ts';
 
 type Context = 'default' | 'table';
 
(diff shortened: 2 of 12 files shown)


packages/types/src/round.ts
```diff
@@ -0,0 +1,39 @@
...
+ * round({ value: 1.2345, decimals: 2, roundingMode: 'ceil' }); // "1.24"
+ * round({ value: 1.2345, decimals: 2, roundingMode: 'floor' }); // "1.23"
+ * round({ value: 1.2345, decimals: 2 }); // "1.23" (default rounding)
+ * ```
+ */
+export function round({ value, decimals, roundingMode = 'round' }: RoundOptions): string {
+  const roundingFn = roundingStrategies[roundingMode];
+  const roundedNumber = roundingFn(value, decimals);
+  return roundedNumber.toFixed(decimals);
+}
```
grod220, line 39: Both `round()` and `shortify()` come from the dex explorer. However, expanded these quite a bit w/ testing as well.

packages/ui/src/ValueView/index.tsx
```diff
@@ -33,6 +35,18 @@ export interface ValueViewComponentProps<SelectedContext extends Context> {
...
+   */
+  hideIcon?: boolean;
+  /**
+   * If true, the asset symbol will be hidden.
+   */
+  hideSymbol?: boolean;
+  /**
+   * If true, the displayed amount will be shortened.
+   */
+  abbreviate?: boolean;
```
grod220, line 49: I'm not entirely sure if these kinds of props are in the spirit of the v2 UI component prop style

packages/ui/src/ValueView/index.tsx
```diff
@@ -82,12 +120,14 @@ export const ValueViewComponent = <SelectedContext extends Context = 'default'>(
               'border-b-2 border-dashed border-other-tonalStroke',
           )}
         >
-          <div className='min-w-[50px] shrink grow truncate' title={formattedAmount}>
-            <ValueText density={density}>{formattedAmount} </ValueText>
-          </div>
-          <div className='min-w-[50px] shrink grow truncate' title={symbol}>
-            <ValueText density={density}>{symbol}</ValueText>
+          <div className='shrink grow' title={formattedAmount}>
+            <ValueText density={density}>{formattedAmount}</ValueText>
```
grod220, line 124: Bug fix, the min width made it too wide

packages/ui/src/ValueView/index.tsx
grod220: Also these design changes should be validated on the storybook preview in the PR

packages/ui/src/ValueView/index.tsx
```diff
@@ -63,16 +87,30 @@ export const ValueViewComponent = <SelectedContext extends Context = 'default'>(
...
       )}
+      else={children => (
+        <span
+          className={cn(
+            density === 'sparse' ? technical : detailTechnical,
+            priority === 'primary' ? 'text-text-primary' : 'text-secondary-dark',
+          )}
+        >
+          {children}
+        </span>
```
grod220, line 105: Needed to have this else branch as the Pill was giving the font/color in the other branch and this one wasn't getting any.

packages/types/src/round.ts
grod220: If this works, after publishing, will delete the local one in the dex explorer

packages/ui/src/ValueView/index.tsx
TalDerei: comment: validated storybook props in preview

packages/ui/src/ValueView/index.tsx
```diff
@@ -46,14 +60,24 @@ export const ValueViewComponent = <SelectedContext extends Context = 'default'>(
...
   const density = useDensity();
 
   if (!valueView) {
     return null;
   }
 
-  const formattedAmount = getFormattedAmtFromValueView(valueView, true);
+  let formattedAmount: string;
+  if (abbreviate) {
+    const amount = getFormattedAmtFromValueView(valueView, false);
```
TalDerei, line 75: comment: what if `getFormattedAmtFromValueView` returns something unparseable (eg. null, undefined)? should we add validation guards to shortify to prevent breakage?

packages/ui/src/ValueView/index.tsx
```diff
@@ -82,12 +120,14 @@ export const ValueViewComponent = <SelectedContext extends Context = 'default'>(
               'border-b-2 border-dashed border-other-tonalStroke',
           )}
         >
-          <div className='min-w-[50px] shrink grow truncate' title={formattedAmount}>
-            <ValueText density={density}>{formattedAmount} </ValueText>
-          </div>
-          <div className='min-w-[50px] shrink grow truncate' title={symbol}>
-            <ValueText density={density}>{symbol}</ValueText>
+          <div className='shrink grow' title={formattedAmount}>
```
TalDerei, line 123: nit; using both `shrink-0` and `shrink` in the same contexts.

TalDerei (approved):

packages/ui/src/ValueView/index.tsx
grod220: The function returns a `string` type, so all good.

packages/ui/src/ValueView/index.tsx
grod220: The `shrink-0` prevents the icon from shrinking, The uses of `shrink` on the other two say that is' ok to shrink.

grod220 (commented):
Going to add a few more tweaks to better support the compact version

PR #1917, opened 2024-11-20
Round func updates

grod220:
Removing trailing zeros + exponential number support

Updates after https://github.com/penumbra-zone/web/pull/1915


TalDerei (approved):
can we add test cases for extremely small negative numbers with varying rounds modes?

PR #1929, opened 2024-11-27
Truncate Toast description

JasonMHasperhoven:
- Fixes the case where it may take up the whole screen incase we directly log an error message as description

### DIFF
--- a/packages/ui/src/Toast/open.ts
+++ b/packages/ui/src/Toast/open.ts
@@ -73,7 +73,6 @@ export const openToast = (props: ToastProps): Toast => {
 
   const open = () => {
     const fn = toastFnMap[options.type];
-    console.log('TCL: open -> options', options);
 
     id = fn(options.message, {
       id,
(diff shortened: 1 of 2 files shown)


packages/ui/src/Toast/open.ts
```diff
@@ -72,7 +76,10 @@ export const openToast = (props: ToastProps): Toast => {
 
     id = fn(options.message, {
       id,
-      description: options.description,
+      description:
+        typeof options.description === 'string'
+          ? truncateString(options.description, 200)
+          : options.description,
```
VanishMax, line 82: issue: i don't think this should be a part of the UI component. If users need this, they can truncate the string in their code when passing `description` to the Toast. Moreover, UI components should not trim the string programmatically – this worsens the UX. Think of it as a long but important error message – what might happen if users see the start but not the important end of the message? 

From the UI components part, we could add "show more"/"show less" buttons if this is really necessary. But still, the description is defined as `ReactNode`, so this can be done on the users' side.

VanishMax (changes requested):

packages/ui/src/Toast/open.ts
JasonMHasperhoven: I suppose you're right, but I have no idea where this long error message throwIfAborted comes from.

PR #1935, opened 2024-12-04
Add pnum

JasonMHasperhoven:
Currently we have a bunch of conversion functions scattered around in:
- packages/types/src/amount.ts
- packages/types/src/lo-hi.ts
- packages/types/src/round.ts
- packages/types/src/value-view.ts

This aims to unify and simplify number conversions, see the function description at the top for context.

### DIFF
--- a/.changeset/modern-taxis-repair.md
+++ b/.changeset/modern-taxis-repair.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/types': minor
+---
+
+Add pnum
--- a/packages/types/src/pnum.ts
+++ b/packages/types/src/pnum.ts
@@ -21,7 +21,7 @@ import { removeTrailingZeros } from '@penumbra-zone/types/shortify';
  * - in display form with decimals for string and number
  *
  * @param input
- * @param exponent
+ * @param optionsOrExponent
  */
 function pnum(
   input?: string | number | LoHi | bigint | Amount | ValueView | undefined,
--- a/packages/types/src/pnum.ts
+++ b/packages/types/src/pnum.ts
@@ -25,10 +25,11 @@ import { removeTrailingZeros } from '@penumbra-zone/types/shortify';
  */
 function pnum(
   input?: string | number | LoHi | bigint | Amount | ValueView | undefined,
-  options: { exponent?: number } | number = { exponent: 0 },
+  optionsOrExponent: { exponent?: number } | number = { exponent: 0 },
 ) {
   let value: BigNumber;
-  let exponent = typeof options === 'number' ? options : (options.exponent ?? 0);
+  let exponent =
+    typeof optionsOrExponent === 'number' ? optionsOrExponent : (optionsOrExponent.exponent ?? 0);
 
   if (typeof input === 'string' || typeof input === 'number') {
     value = new BigNumber(input).shiftedBy(exponent);
--- a/packages/types/src/pnum.ts
+++ b/packages/types/src/pnum.ts
@@ -24,11 +24,11 @@ import { removeTrailingZeros } from '@penumbra-zone/types/shortify';
  * @param exponent
  */
 function pnum(
-  input: string | number | LoHi | bigint | Amount | ValueView | undefined,
-  exponentInput = 0,
+  input?: string | number | LoHi | bigint | Amount | ValueView | undefined,
+  options: { exponent?: number } = { exponent: 0 },
 ) {
   let value: BigNumber;
-  let exponent = exponentInput;
+  let exponent = options.exponent ?? 0;
 
   if (typeof input === 'string' || typeof input === 'number') {
     value = new BigNumber(input).shiftedBy(exponent);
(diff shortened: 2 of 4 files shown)


packages/types/src/pnum.ts
```diff
@@ -0,0 +1,115 @@
...
+ * - in base units for bigint, LoHi, Amount and ValueView
+ * - in display form with decimals for string and number
+ *
+ * @param input
+ * @param exponent
+ */
+function pnum(
+  input: string | number | LoHi | bigint | Amount | ValueView | undefined,
+  exponentInput = 0,
+) {
```
grod220, line 29: issue: this feels very class-like (private vars, exposed methods, constructor like initialization). Think we should convert it.

packages/types/src/pnum.ts
```diff
@@ -0,0 +1,115 @@
...
+  } else if (typeof input === 'bigint') {
+    value = new BigNumber(input.toString());
+  } else if (input instanceof ValueView) {
+    const amount = getAmount(input);
+    value = new BigNumber(joinLoHi(amount.lo, amount.hi).toString());
+    exponent =
+      input.valueView.case === 'knownAssetId' ? getDisplayDenomExponentFromValueView(input) : 0;
+  } else if (
+    input instanceof Amount ||
+    (typeof input === 'object' && 'lo' in input && 'hi' in input && input.lo && input.hi)
```
grod220, line 44: Suggestion: Wouldn't this work the same?

```ts
else if (input instanceof Amount) {
    value = new BigNumber(joinLoHi(input.lo, input.hi).toString());
  }
```

packages/types/src/pnum.ts
```diff
@@ -0,0 +1,115 @@
...
+    toRoundedNumber(decimals = exponent): number {
+      return Number(round({ value: value.shiftedBy(-exponent).toNumber(), decimals }));
+    },
+
+    toLoHi(): LoHi {
+      return splitLoHi(BigInt(value.toString()));
+    },
+
+    toAmount(): Amount {
+      return new Amount(splitLoHi(BigInt(value.toString())));
```
grod220, line 110: issue: missing tests for this method

.changeset/modern-taxis-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@penumbra-zone/types': minor
+---
```
grod220, line 3: question: should we take those packages you've listed, create a new package, make those specific files private, and only expose pnum?

packages/types/src/pnum.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Likewise for all methods, the outputs are
+ * - in base units for bigint, LoHi, Amount and ValueView
+ * - in display form with decimals for string and number
+ *
+ * @param input
+ * @param exponent
+ */
+function pnum(
+  input: string | number | LoHi | bigint | Amount | ValueView | undefined,
+  exponentInput = 0,
```
grod220, line 28: issue: think we should consider an options obj as a second parameter so it's easier to expand in the future:

```diff
- exponentInput = 0,
+ options: { exponentInput?: number } = { exponentInput: 0 },
```

packages/types/src/pnum.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Likewise for all methods, the outputs are
+ * - in base units for bigint, LoHi, Amount and ValueView
+ * - in display form with decimals for string and number
+ *
+ * @param input
+ * @param exponent
+ */
+function pnum(
+  input: string | number | LoHi | bigint | Amount | ValueView | undefined,
```
grod220, line 27: question: what are the usecases for an `undefined` input?

.changeset/modern-taxis-repair.md
grod220: related: if we make the other stuff private, we'll need to find the previous usages and convert them to pnum

grod220 (commented):
Cool stuff!

packages/types/src/pnum.ts
JasonMHasperhoven: I did this on purpose as it's shorter: `pnum(val)` vs `new PNum(val)`. My guess is that we'll use it quite often and this way it looks cleaner imo, especially if we need to do it inline inside a react component.

packages/types/src/pnum.ts
JasonMHasperhoven: This case catches an instance of Amount **and LoHi**, in your check a LoHi will not be true.

packages/types/src/pnum.ts
JasonMHasperhoven: Perhaps wise indeed

packages/types/src/pnum.ts
JasonMHasperhoven: This: 
image

I guess you can make an argument against it, what do you think?

packages/types/src/pnum.ts
grod220: No test appears to fail when removing that line. Can you add a test for coverage for that?

grod220 (approved):
Feel free to merge after addressing the final comments 👍 

Think it's mostly more tests we need

TalDerei:
nice helper function, I've been wanting something like this for a while 👍

packages/types/src/pnum.ts
JasonMHasperhoven: Sure

packages/types/src/round.ts
```diff
@@ -4,9 +4,10 @@ import { removeTrailingZeros } from './shortify.js';
 export type RoundingMode = 'half-up' | 'up' | 'down';
 
 export interface RoundOptions {
-  value: number;
+  value: number | string;
```
JasonMHasperhoven, line 7: For @grod220 (because I know you want to know why) this is needed for a big number (caught by one of the tests)

PR #1938, opened 2024-12-10
client: fix connection persistence bug

TalDerei:
references https://github.com/penumbra-zone/dex-explorer/issues/154

this is a subtle wallet connection bug. Initially, I thought the culprit was related to mobx's observer HOC, where the connection state wasn't properly being observed. Upon digging deeper, I narrowed in on the client connection implementation, specifically this [no-op](https://github.com/penumbra-zone/web/blob/main/packages/client/src/client.ts#L151-L153) comment. If the connection is set, indicating an active connection, the nullish coalescing operator prevents initializing a new connection. Debugging the connection state revealed that this no-op behavior becomes problematic in a specific (yet common) case: the connection is _set_, but the underlying connection is broken. This could be from either the connection timing out, provider detaching, etc. and the no-op behavior prevents the frontend from re-establishing connection, _even_ when the extension is logged in. 

Screenshot 2024-12-09 at 10 04 23 PM

Problematically, the same error is encountered when the user manually _denies_ the connection. If this error isn't explicitly caught and propagated back to the caller, then a a denied connection would immediately trigger an attempt to re-establish the connection. For other types of errors, the `connect` method was modified to clean up the existing connection resources and attempts to establish a new connection to the attached provider.

We should sanity check the security implications of this remediation given that the last audit caught some connection related vulnerabilities.

--------------------------------------

https://github.com/user-attachments/assets/6da56808-52fb-4e89-a9b6-976202d3bce0

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -17,6 +17,7 @@ import {
 import { PenumbraManifest } from './manifest.js';
 import { PenumbraProvider } from './provider.js';
 import { PenumbraState } from './state.js';
+import { PenumbraRequestFailure } from './error.js';
 
 const isLegacyProvider = (
   provider: PenumbraProvider,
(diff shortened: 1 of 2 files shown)


packages/client/src/client.ts
```diff
@@ -188,7 +189,24 @@ export class PenumbraClient {
...
+      await this.connection.port;
+    } catch (error) {
+      if (error instanceof Error && error.cause) {
+        if (error.cause === PenumbraRequestFailure.Denied) {
+          throw error;
+        }
+      }
+
+      // Clean up dangling connection resources and attempt to establish reconnection
+      this.destroyConnection();
```
TalDerei, line 206: Is this sufficient for ensuring no dangling resources persist?

VanishMax (approved):
This looks good to me. Should we create a new issue for the disconnect from wallet persisting the connection after page reload, which you showed on the video?

PR #1941, opened 2024-12-10
feat(ui): updates for address view 

TalDerei:
aligns `Address View` V2 UI component with the latest figma designs – review the storybook preview link. Some pointers would be helpful cc @VanishMax

### DIFF
--- a/packages/ui/src/AddressView/index.tsx
+++ b/packages/ui/src/AddressView/index.tsx
@@ -59,7 +59,8 @@ export const AddressViewComponent = ({
         </div>
       )}
 
-      <div className={truncate ? 'truncate max-w-[150px]' : ''}>
+      <div className={truncate ? 'max-w-[150px] truncate' : ''}>
+        {/* eslint-disable-next-line no-nested-ternary -- can alternatively use dynamic prop object like {...fontProps} */}
         {addressIndex ? (
           density === 'sparse' ? (
             <Text strong-bold truncate={truncate}>
--- a/packages/ui/src/AddressView/index.tsx
+++ b/packages/ui/src/AddressView/index.tsx
@@ -4,12 +4,35 @@ import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
 import { CopyToClipboardButton } from '../CopyToClipboardButton';
 import { AddressIcon } from './AddressIcon';
 import { Text } from '../Text';
+import { Density, useDensity } from '../utils/density';
+import cn from 'clsx';
 
 export interface AddressViewProps {
   addressView: AddressView | undefined;
   copyable?: boolean;
   hideIcon?: boolean;
+  truncate?: boolean;
 }
+
+export const getIconSize = (density: Density): number => {
+  if (density === 'compact') {
+    return 16;
+  }
+  if (density === 'slim') {
+    return 12;
+  }
+  return 24;
+};
+
+export const getFont = (density: Density): string => {
+  if (density === 'compact') {
+    return cn('flex items-center gap-1.5 text-sm text-text-primary');
+  }
+  if (density === 'slim') {
+    return cn('flex items-center gap-1 text-xs text-text-primary');
+  }
+  return cn('flex items-center gap-2 text-base text-text-primary');
+};
 
 // Renders an address or an address view.
 // If the view is given and is "visible", the account information will be displayed instead.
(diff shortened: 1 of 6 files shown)


packages/ui/src/Text/index.tsx
```diff
@@ -102,7 +102,9 @@ const getTextOptionClasses = ({
   break: breakProp,
   whitespace,
 }: TextProps): string => {
-  const truncateClass = truncate ? cn('truncate') : '';
+  const truncateClass = truncate
+    ? cn('truncate max-w-[150px] overflow-hidden whitespace-nowrap')
+    : ''; // Adds max-width for aggressive truncation
```
TalDerei, line 107: what defines the truncations max width?

Screenshot 2024-12-10 at 10 47 16 AM

packages/ui/src/AddressView/index.tsx
```diff
@@ -4,46 +4,79 @@ import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
...
           {isRandomized && 'IBC Deposit Address for '}
-          {`Sub-Account #${addressIndex.account}`}
+          {getAccountLabel(addressIndex.account)}
         </Text>
       ) : (
-        <Text technical truncate>
+        <Text technical truncate={truncate}>
           {encodedAddress}
         </Text>
       )}
```
TalDerei, line 82: the copy button below has some stylistic inconsistencies. It displays different styles for sparse and compact densities, and it is also getting cut off. Should this be addressed at the button component level, or by adjusting the wrapping div tag to reduce its size?

---------------------

Screenshot 2024-12-10 at 10 47 01 AM

Screenshot 2024-12-10 at 10 47 08 AM

packages/ui/src/Text/index.tsx
```diff
@@ -102,7 +102,9 @@ const getTextOptionClasses = ({
   break: breakProp,
   whitespace,
 }: TextProps): string => {
-  const truncateClass = truncate ? cn('truncate') : '';
+  const truncateClass = truncate
+    ? cn('truncate max-w-[150px] overflow-hidden whitespace-nowrap')
```
VanishMax, line 106: issue: the `max-w-[150px]` class would break many of existing components. Truncation width is generally defined in the parent component.

also, `truncate` class applies both `overflow-hidden` and `whitespace-nowrap` under the hood

VanishMax (changes requested):

packages/ui/src/AddressView/index.tsx
VanishMax: it should be fine in the real app, it the Storybook that cuts the block. Could be solved by adding padding in the Storybook story around AddressView

packages/ui/src/AddressView/index.tsx
```diff
@@ -4,46 +4,79 @@ import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
...
 
   const encodedAddress = bech32mAddress(addressView.addressView.value.address);
 
+  // Sub-account selector logic
+  const getAccountLabel = (index: number) =>
+    index === 0 ? 'Main Account' : `Sub-Account ${index}`;
+
   return (
-    <div className='flex items-center gap-2 text-text-primary'>
+    <div className={cn(getFont(density))}>
```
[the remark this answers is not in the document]
TalDerei: makes sense, was questioning whether it only applied to wrapping div – will modify accordingly

packages/ui/src/Text/index.tsx
TalDerei: I’ll move the truncation width to the parent component. Moving forward, I’ll focus on applying styling at the parent level where possible to avoid introducing breaking changes to existing components.

packages/ui/src/AddressView/index.tsx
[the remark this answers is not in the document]
TalDerei: applied the logic directly within the `Text` component rather than passing in a dynamic prop like `{...fontProps}`, which isn't supported.

packages/ui/src/AddressView/index.tsx
TalDerei: padding fixes it 👍

packages/ui/src/AddressView/index.tsx
```diff
@@ -4,49 +4,84 @@ import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
...
+          density === 'sparse' ? (
+            <Text strong-bold truncate={truncate}>
+              {isRandomized && 'IBC Deposit Address for '}
+              {getAccountLabel(addressIndex.account)}
+            </Text>
+          ) : (
+            <Text small truncate={truncate}>
+              {isRandomized && 'IBC Deposit Address for '}
+              {getAccountLabel(addressIndex.account)}
+            </Text>
```
VanishMax, line 73: thought: this makes me think of creating an alternative prop that would simply accept a string

VanishMax (approved):
Looks good! Need to somehow fix the linting issues, and it's good to go

PR #1948, opened 2024-12-13
Adjust Max button for -0.1 asset

tacolopo:
Max swap would previously fail because it doesn't automatically adjust for gas fees. This change should alter the max to be -0.1, then check to make sure the balance stays positive, before attempting to swap, so it works right away off the bat without user interference. Just an idea as a potential UI improvement. 

image

### DIFF
--- a/apps/minifront/src/state/swap/instant-swap.ts
+++ b/apps/minifront/src/state/swap/instant-swap.ts
@@ -186,16 +186,22 @@ const assembleSwapRequest = async ({
   }
 
   const addressIndex = getAddressIndex(assetIn.accountAddress);
+  const exponent = getDisplayDenomExponentFromValueView(assetIn.balanceView);
+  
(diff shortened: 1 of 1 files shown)


TalDerei:
max functionality for swaps, as you pointed out, is _not_ currently supported and requires more extensive changes, rather than simply hardcoding a subtraction factor in the frontend! However, we do currently support maximum sends between accounts: https://x.com/penumbrazone/status/1855018748014362964.

tacolopo:
@TalDerei Just for personal research, is there a reference for what makes it more complex? Is the idea to do analysis on each coin to fetch the gas fee for a transaction and then use that number to get closer to a full max swap? Regardless, the 0.1 would be sufficient for most (?) chains right now and would enable the "max" button swap to work for new users off the bat. Anyways, cool product, thanks for working on it.

PR #1949, opened 2024-12-17
Add trailingZeros option to ValueView

JasonMHasperhoven:


grod220 (commented):
Perhaps next time the description would help, but could you give more detail on the use-case for this? Why are trailing zeros desirable in some cases?

VanishMax:
I guess the reason for this is that `toFormattedString`'s `trailingZeros` param is `true` by default. This options is not for adding trailing zeros but for removing them. 

Agree that it needs some more description

JasonMHasperhoven:
Oh sorry, here you can see in the design that we should 2 decimals for UM for example:
image

As Max mentioned, toFormattedString shows trailingZero's by default whereas toNumber does not.

grod220 (approved):

PR #1950, opened 2024-12-18
ibc: transparent address support 

TalDerei:
remediates https://forum.penumbra.zone/t/usdc-transfers-temporarily-affected-by-noble-chain-upgrade/129 by detecting whether inbound / outbound IBC transfers involve Noble USDC and uses a transparent address (t-addr) to bypass the `bech32` encoding compatibility issues that have previously blocked transfers. For all other assets sent to Noble, we use the `penumbra1` encoding format.

references https://github.com/penumbra-zone/penumbra/pull/4950

pairs with https://github.com/prax-wallet/prax/pull/254


apps/minifront/src/state/ibc-out.ts
```diff
@@ -215,20 +216,44 @@ const getPlanRequest = async ({
...
         ),
-        denom: { denom: getMetadata(selection.balanceView).base },
+        denom: { denom },
         destinationChainAddress,
         returnAddress,
         timeoutHeight,
         timeoutTime,
         sourceChannel: chain.channelId,
         useCompatAddress: bech32ChainIds.includes(chain.chainId),
+        useTransparentAddress,
```
TalDerei, line 256: on this branch, performing a regular IBC withdrawal with `useTransparentAddress` set to false is throwing a binding signature failure. cc @erwanor  

Screenshot 2024-12-18 at 11 53 30 AM

apps/minifront/src/state/ibc-out.ts
```diff
@@ -215,20 +216,44 @@ const getPlanRequest = async ({
...
           getDisplayDenomExponentFromValueView(selection.balanceView),
         ),
-        denom: { denom: getMetadata(selection.balanceView).base },
+        denom: { denom },
         destinationChainAddress,
         returnAddress,
         timeoutHeight,
         timeoutTime,
         sourceChannel: chain.channelId,
         useCompatAddress: bech32ChainIds.includes(chain.chainId),
```
erwanor, line 255: This field is deprecated, so it can be omitted entirely and set to false

TalDerei: if we want to be explicit about maintaining backwards compatibility with older frontends, wouldn't we want to keep this field?

packages/wasm/crate/src/keys.rs
```diff
@@ -181,3 +181,24 @@ pub fn forwarding_addr_inner(sequence: u16, account: Option<u32>, fvk: &FullView
...
+    /// The t-address encoding of the transparent address
+    pub encoding: String,
+}
+
+/// Returns the "truncated" address (t-addr) associated with the account.
+#[wasm_bindgen]
+pub fn get_transparent_address(full_viewing_key: &[u8]) -> WasmResult<TransparentAddrResponse> {
+    let fvk: FullViewingKey = FullViewingKey::decode(full_viewing_key)?;
+    let encoding = fvk.incoming().transparent_address();
+    let address: Address = encoding.parse().expect("encoding transparent address");
```
TalDerei, line 198: maybe `Address::from_str(&encoding).expect("encoding transparent address")` is better here?

TalDerei:
I’d like to add unit tests for compact and transparent addresses to complement the manual testing and provide greater assurance.

apps/minifront/src/state/ibc-out.ts
TalDerei: removed the deprecated field

apps/minifront/src/state/ibc-in/index.tsx
```diff
@@ -207,6 +208,24 @@ async function execute(
...
+
+  // Temporary: detect USDC Noble inbound transfers, and use transparent (t-addr) encoding
+  // to ensure bech32m encoding compatibility.
+  if (
+    transferToken.denom.includes('uusdc') &&
+    (selectedChain.chainId === 'noble-1' || selectedChain.chainId === 'grand-1')
+  ) {
+    // Set the reciever address to the t-addr encoding.
+    penumbraAddress = encoding;
+  }
```
hdevalence, line 227: Shouldn't this also be setting the flag in the withdrawal message that tells the Penumbra chain to encode the address as a tpenumbra1... address? Otherwise I think it will be the `penumbra1...` encoding of the user's t-addr.

hdevalence: Oh, no, because this is ibc-IN

apps/minifront/src/state/ibc-out.ts
hdevalence: What happened here?

TalDerei: this was pre-upgrade, resolved now

packages/bech32m/src/format/index.ts
```diff
@@ -31,12 +31,6 @@ export default {
     byteLength: ByteLength.penumbra,
     innerName: Inner.penumbra,
   },
-  penumbracompat1: {
```
erwanor, line 34: It might be worth keeping it around to render historical txs etc.

erwanor: However, it would be good to add the transparent address prefix so that the bech32 library has the ability to encode an `Address` etc.

Here's the code in core: https://github.com/penumbra-zone/penumbra/blob/020986747245270a00d039306458be9e7f1f33f5/crates/core/keys/src/address.rs#L232

apps/minifront/src/state/ibc-out.ts
erwanor: I think this happened using a new client on an old network, with changes to the proto actions, the auth hash looks invalid to the old full node (because the extra proto fields are ignored when creating the domain type).

packages/bech32m/src/format/index.ts
TalDerei: that's a good suggestion, would enable rendering a t-addr in prax 👍

erwanor (approved):

packages/bech32m/src/format/index.ts
TalDerei: added back the `compat` bech32 encoding and added the [transparent](https://github.com/penumbra-zone/web/pull/1950/commits/486e9a8d98ec87964c72695fbe9233e6b6fa07a7) address prefix to the bech32 library.

packages/services/src/view-service/util/transaction-view.ts
```diff
@@ -0,0 +1,28 @@
+import { TransactionView } from '@penumbra-zone/protobuf/penumbra/core/transaction/v1/transaction_pb';
+import { getTransmissionKeyByAddress } from '@penumbra-zone/wasm/keys';
+
+// Some transaction views (TXVs) require additional preprocessing before being rendered
+// in the UI component library. For example, when handling IBC withdrawals with transparent
+// addresses, this component transforms ephemeral addresses into their bech32-encoded
+// transparent form to ensure the proper data is being displayed.
+export const txvTranslator = async (view: TransactionView): Promise<TransactionView> => {
```
TalDerei, line 8: renders the proper address format in the transaction view

Screenshot 2024-12-22 at 12 29 48 AM


Screenshot 2024-12-22 at 12 29 39 AM

PR #1953, opened 2024-12-19
Fix pnum parsing ValueView

JasonMHasperhoven:
- `input instanceof ValueView` didn't work apparently.


JasonMHasperhoven:
Also updated linking so that you can do:
```
node scripts/link
node scripts/unlink
```

PR #1960, opened 2024-12-23
Updates for Positions Table

JasonMHasperhoven:
There's an issue reported with Density [here](https://penumbralabs.slack.com/archives/C074P1TPBH9/p1734715004799049).
TLDR: Different density components use the same context variable, so when you have multiple densities within the page there's a case where the children of one density get the value of another density.

This PR includes a temporary solution and is not meant to be a permanent solution, but merely a way to not block our progress with the position table in the DEX.

Other changes:
- Change button display type from flex to inline-flex
- Improved Table styling to match dex designs for "slim" density
- Added table-headings support to the Text component

### DIFF
--- a/packages/ui/src/Button/index.tsx
+++ b/packages/ui/src/Button/index.tsx
@@ -102,13 +102,15 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       actionType = 'default',
       type = 'button',
       priority = 'primary',
+      density: densityProp,
       ...attrs
       // needed for the Radix's `asChild` prop to work correctly
       // https://www.radix-ui.com/primitives/docs/guides/composition#composing-with-your-own-react-components
     },
     ref,
   ) => {
-    const density = useDensity();
+    const densityContext = useDensity();
+    const density = densityProp ?? densityContext;
     const styleAttrs = { actionType, iconOnly, density, priority };
 
     return (
@@ -133,7 +135,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
 
           'relative',
           'text-neutral-contrast',
-          'flex items-center justify-center',
+          'inline-flex items-center justify-center',
           density === 'sparse' ? 'gap-2' : 'gap-1',
         )}
       >
--- a/packages/ui/src/Table/index.tsx
+++ b/packages/ui/src/Table/index.tsx
@@ -3,6 +3,7 @@ import cn from 'clsx';
 import { tableHeading, tableItem } from '../utils/typography';
 import { Density, useDensity } from '../utils/density';
 import { ConditionalWrap } from '../ConditionalWrap';
+import { getThemeColorClass, ThemeColor } from '../utils/color';
 
 export interface TableProps {
   /** Content that will appear above the table. */
@@ -10,6 +11,8 @@ export interface TableProps {
   children: ReactNode;
   /** Which CSS `table-layout` property to use. */
   tableLayout?: 'fixed' | 'auto';
+  /** The background color of the table. */
+  bgColor?: ThemeColor;
 }
 
 /**
@@ -53,7 +56,12 @@ export interface TableProps {
  * </Table>
  * ```
  */
-export const Table = ({ title, children, tableLayout }: TableProps) => (
+export const Table = ({
+  title,
+  children,
+  tableLayout,
+  bgColor = 'other.tonalFill5',
+}: TableProps) => (
   <ConditionalWrap
     if={!!title}
     then={children => (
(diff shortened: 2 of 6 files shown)


packages/ui/src/Button/index.tsx
```diff
@@ -133,7 +136,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
 
           'relative',
           'text-neutral-contrast',
-          'flex items-center justify-center',
+          'inline-flex items-center justify-center',
```
VanishMax, line 139: issue: this is a breaking change we shouldn't introduce for such purpose. 

suggestion: from button's parent, apply `[&>button]:inline-flex` or `[&_button]:inline-flex` class. Might feel hacky, but it's Tailwind – would look nicer with plain css.

packages/ui/src/utils/typography.ts
```diff
@@ -61,4 +61,4 @@ export const button = cn('font-default text-textBase font-medium leading-textBas
 
 export const buttonMedium = cn('font-default text-textBase font-medium leading-textBase');
 
-export const buttonSmall = cn('font-default text-textBase font-medium leading-textBase');
+export const buttonSmall = cn('font-default text-textXs font-medium leading-textBase');
```
VanishMax, line 64: issue: it's `textBase` in Figma

image

packages/ui/src/ValueView/index.tsx
```diff
@@ -64,12 +77,15 @@ export const ValueViewComponent = <SelectedContext extends Context = 'default'>(
   valueView,
   context,
   priority = 'primary',
+  showIcon = true,
```
VanishMax, line 80: praise: waited for this prop for so long! 🙏

packages/ui/src/Text/index.tsx
```diff
@@ -135,6 +138,9 @@ const VARIANT_MAP: Record<TextVariant, { element: ElementType; classes: string }
   detailTechnical: { element: 'span', classes: detailTechnical },
   technical: { element: 'span', classes: technical },
   body: { element: 'span', classes: body },
+  tableHeading: { element: 'span', classes: tableHeading },
+  tableHeadingMedium: { element: 'span', classes: tableHeadingMedium },
+  tableHeadingSmall: { element: 'span', classes: tableHeadingSmall },
```
VanishMax, line 143: suggestion: add `tableItem`, `tableItemMedium`, and `tableItemSmall` too. You would need it for several cells in the positions table

image

packages/ui/src/Table/index.tsx
```diff
@@ -82,23 +91,30 @@ Table.Thead = Thead;
...
+const densityPaddingMapping = {
+  slim: 'px-3 py-2',
+  compact: 'px-3 py-3',
+  sparse: 'px-3 py-4',
+};
+
+const getCell = (density: Density) => {
+  const padding = densityPaddingMapping[density];
+  return cn('box-border', padding);
+};
```
VanishMax, line 109: question: i know it might be too much but maybe it's worth to implement the `TableCell` unified component that is designed in Figma? We would have to do it one day anyway, so why not now? I believe it would suit all our needs for position table

https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?node-id=333-17181&m=dev

VanishMax (changes requested):

packages/ui/src/AssetIcon/index.tsx
```diff
@@ -12,7 +12,7 @@ type Size = 'lg' | 'md' | 'sm';
 const sizeMap: Record<Size, string> = {
   lg: cn('w-8 h-8'),
   md: cn('w-6 h-6'),
-  sm: cn('w-4 h-4'),
+  sm: cn('w-6 h-6'),
```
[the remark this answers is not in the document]
JasonMHasperhoven: Is it? It's 24px here:
image

packages/ui/src/Button/index.tsx
JasonMHasperhoven: Ok will do that

packages/ui/src/utils/typography.ts
JasonMHasperhoven: I think that's out of date, see here:
image

packages/ui/src/AssetIcon/index.tsx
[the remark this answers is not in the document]
VanishMax: well that's md size, isn't it? think AssetIcon's own page should be the source of truth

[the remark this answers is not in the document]
VanishMax: Yeah that's truly md 
image

packages/ui/src/utils/typography.ts
VanishMax: There's also the mobile size, which is not yet supported in the frontend

image

packages/ui/src/Table/index.tsx
VanishMax: That we probably don't need `Table`, `TableRow`, `TableHeading` anymore with the new `TableCell` component from Figma

VanishMax: Not pushing for this to be in this PR. We can do it later and merge this pr

VanishMax (approved):

packages/ui/src/Table/index.tsx
JasonMHasperhoven: Got it

PR #1975, opened 2025-01-07
feat: chunk genesis sync 

TalDerei:
## Description of Changes

chunking genesis sync for lower powered devices. For beta testing, we’d specifically like to request and test against airdrop recipients. 

pairs with https://github.com/prax-wallet/prax/pull/257

## Related Issue

https://github.com/prax-wallet/prax/issues/230


TalDerei:
note to reviewers: putting on hold for now in favor of dex-explorer work

packages/wasm/crate/src/view_server.rs
```diff
@@ -97,10 +98,128 @@ impl ViewServer {
...
+                        // it doesn't matter what kind of payload it was either.
+                        // Just insert and forget
+                        self.sct
+                            .insert(tct::Witness::Forget, *payload.commitment())
+                            .expect("inserting a commitment must succeed");
+                    }
+                }
+
+                // End the block in the commitment tree
+                self.sct.end_block().expect("ending the block must succed");
```
erwanor, line 214: We're sealing a new block for every commitment, I think this is a mistake right?

packages/wasm/crate/src/view_server.rs
```diff
@@ -97,10 +98,128 @@ impl ViewServer {
...
+            self.sct
+                .insert_block(full_block.block_root)
+                .expect("inserting a block root must succeed");
+        } else {
+            // If we found at least one note for us in this block, we have to explicitly construct the
+            // whole block in the SCT by inserting each commitment one at a time
+            for payload in full_block.state_payloads.into_iter() {
+                // We proceed commitment by commitment, querying our in-memory advice
+                // to see if we have any data for the commitment and act accordingly
+                match genesis_advice.get(payload.commitment()) {
```
erwanor, line 172: I think we can use a different method to drain the advices as we process them.

packages/wasm/crate/src/view_server.rs
TalDerei: > We're sealing a new block for every commitment, I think this is a mistake right?

good catch, think what needs to happen here is after processing all genesis chunks in `genesis_advice()`, finally seal the block.

packages/wasm/crate/src/view_server.rs
```diff
@@ -97,10 +98,128 @@ impl ViewServer {
...
             last_position: None,
             last_forgotten: None,
+            genesis_advice: None,
         };
         Ok(view_server)
     }
 
+    /// Scans a chunk of the genesis block for notes that can be trial decrypted with the viewing key.
+    #[wasm_bindgen]
+    pub async fn trial_decrypt_genesis_chunk(
```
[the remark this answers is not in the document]
TalDerei: modified the method name

packages/wasm/crate/src/view_server.rs
TalDerei: moved block sealing outside the for loop

packages/wasm/crate/src/view_server.rs
TalDerei: since the underlying data structure chosen was a `BTreeMap`, the complexity is already O(log n). what did you have in mind?

erwanor: it's not going to change the asymptotic of a lookup, but it will make the constant factor go down. come to think of it, the number of notes is so small this almost certainly doesn't matter.

erwanor (approved):

PR #1980, opened 2025-01-15
chore: extend alternative fees to LPs

TalDerei:
## Description of Changes

extends alternative fee capability to liquidity positions in the [view service](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/transaction-planner/index.ts#L33). 

pairs with https://github.com/penumbra-zone/dex-explorer/pull/272

this will be staged in the next prax minor point release https://github.com/prax-wallet/prax/pull/263

## Related Issue

fixes https://github.com/penumbra-zone/dex-explorer/issues/246 


packages/services/src/view-service/fees.ts
```diff
@@ -89,6 +89,40 @@ export const extractAltFee = async (
...
+    if (assetId) {
+      return assetId;
+    }
+  }
+
+  const positionClose = request.positionCloses.map(a => a.positionId).find(Boolean);
+  if (positionClose) {
+    const closePosition = await indexedDb.getPosition(positionClose);
+    if (closePosition?.phi?.pair?.asset1) {
+      const inputAssetId = closePosition.phi.pair.asset1;
```
erwanor, line 106: I don't understand what's the basis for selecting the gas fee asset as the `asset_1` of the position?

TalDerei: It was a placeholder. I had originally added a comment (not sure why I removed it) and was planning to clarify this when I requested the review. Is there a canonical way for selecting the proper LP asset here? The tricky case here is when both assets in the pair are valid alternative fee assets (eg. osmo <> atom)

packages/services/src/view-service/fees.ts
```diff
@@ -89,6 +89,40 @@ export const extractAltFee = async (
     }
   }
 
+  const positionOpen = request.positionOpens
```
TalDerei, line 92: https://github.com/user-attachments/assets/61fa088f-d5f8-464b-a5cf-dacba0ecd6d9

TalDerei:
CI rust linter seems to be failing on all open PRs

cronokirby:
One thing I don't quite understand about how the planner does alternate fees: why do we need to do particular work for any given action? Can the planner not just try and use whatever fee tokens it has, in order of "priority"?

TalDerei:
> One thing I don't quite understand about how the planner does alternate fees: why do we need to do particular work for any given action? Can the planner not just try and use whatever fee tokens it has, in order of "priority"?

we have a more dynamic and sophisticated fee selection mechanism (reference https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L117-L137). The motivation was to avoid situations where the user is spending one alternative fee token and paying with another (there are nuanced edge cases here). Instead, we ensure that the same fee token is used, which is extracted from the transaction parameters in the wasm planner and supplied as input to `getAssetFromGasPriceTable`, and that is what the first if statement block handles. Otherwise, we query the assets in the user's `GasPrices` table and select the first available gas token, and those are ordered by priority in IndexDB.

packages/services/src/view-service/fees.ts
cronokirby: Could we just use the fallback default for fee selection for closing positions? I don't see a reason to pick one asset over the other.

TalDerei: sure will change

TalDerei: https://github.com/penumbra-zone/web/pull/1980/commits/15f8399f7957248fc3391cc4c4c12448ea76572e

https://github.com/user-attachments/assets/5d689b39-fddc-4a72-ac09-03e9d2850fb0

packages/services/src/view-service/fees.ts
TalDerei: modified ordering of the gas prices to prioritize usdc https://github.com/penumbra-zone/web/pull/1980/commits/8e2f22b5827b254ae8bab42b730cd393e5866b5a (compare with fee selection using atom above)


https://github.com/user-attachments/assets/2925f8b4-4b87-4446-af98-0c2306d5674e

conorsch:
Now that #1983 has landed, consider rebasing this PR, to get CI passing again—or at least the `rust:lint` portion of it!

packages/storage/src/indexed-db/index.ts
```diff
@@ -472,9 +472,19 @@ export class IndexedDb implements IndexedDbInterface {
...
+    // Retrieve gas prices from the database and prioritize USDC as the preferred asset.
     return allGasPrices
       .map(gp => GasPrices.fromJson(gp))
-      .filter(gp => !gp.assetId?.equals(this.stakingTokenAssetId));
+      .filter(gp => gp.assetId?.inner && !gp.assetId.equals(this.stakingTokenAssetId))
+      .sort((a, b) => {
+        const assetA = a.assetId?.inner ? uint8ArrayToBase64(a.assetId.inner) : '';
+        const assetB = b.assetId?.inner ? uint8ArrayToBase64(b.assetId.inner) : '';
+        return (
+          (assetA === usdcPriorityAssetId ? -1 : 0) - (assetB === usdcPriorityAssetId ? -1 : 0)
```
erwanor, line 485: I found it a bit gnarly to read, not knowing how TS does `Ordering::`
```suggestion
    if (assetA === usdcPriorityAssetId) return -1;
    if (assetB === usdcPriorityAssetId) return 1;
    return 0;
```

TalDerei: thanks for the suggestion

packages/services/src/view-service/fees.ts
```diff
@@ -116,12 +150,14 @@ export const getAssetFromGasPriceTable = async (
 
   // If a specific asset ID is provided, extracted from the transaction request, check its balance is
   // positive and GasPrices for that asset exist.
-  if (assetId) {
-    const balance = await indexedDb.hasTokenBalance(request.source.account, assetId);
-    // This check ensures that the alternative fee token is a valid fee token, for example, TestUSD is not.
-    const isInGasTable = altGasPrices.find(gp => gp.assetId?.equals(assetId));
-    if (balance && isInGasTable) {
-      return assetId;
+  if (!prioritySelection(request)) {
```
erwanor, line 153: I don't totally understand the extra control flow here, we gate all of this method's logic on whether `prioritySelection` evaluates to true. But zooming in:

```
const prioritySelection = (request: TransactionPlannerRequest): boolean => {
  return (
    request.positionOpens.length > 0 ||
    request.positionCloses.length > 0 ||
    request.positionWithdraws.length > 0
  );
};
```

erwanor: Why is it correct? what's the intended feature set before this PR and after this PR?

TalDerei: also relevant here: https://github.com/penumbra-zone/web/pull/1980#issuecomment-2594393421

it's worth exploring the logic embedded in [extractAltFee](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L6-L9) that comprises our fee selection mechanism. 

1. Firstly, [identify](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L13-L97) the relevant TPR, and attempt to extract an asset ID from the TPR (eg. [output](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L13)). 

2. call [getAssetFromGasPriceTable](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L102-L105) to check a few invariants: 

- if the asset ID from the TPR that was (1) optionally passed to `getAssetFromGasPriceTable` and (2) is a valid gas token, [use that for the fee](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/fees.ts#L117-L126). 

- otherwise, if the (1) asset ID (eg. [position TPRs](https://github.com/penumbra-zone/web/pull/1980/files#diff-caad9b17fa66f6badd2a44a36749fc51c4452f35e34f7fe38905467885f60f0bR102-R124)) wasn't supplied or (2) the asset ID isn't a valid gas fee token, or (3) the TPR is in [prioritySelection()](https://github.com/penumbra-zone/web/pull/1980/files#diff-caad9b17fa66f6badd2a44a36749fc51c4452f35e34f7fe38905467885f60f0bR8),  then use the fallback default for fee selection (prioritizes USDC first, and then the next fee asset in the gas prices table for which the user has a positive balance). 

-----------

**Why is it correct?** 

It doesn't affect the fee logic for existing TPRs (spends, outputs, swaps, etc.), just extends the alternative fee functionality to liquidity positions. There are nuanced edge cases here, so we only want to scope this fee logic for LPs to simply fix https://github.com/penumbra-zone/dex-explorer/issues/246. 

**Q. what's the intended feature set before this PR and after this PR?**

[prioritySelection()](https://github.com/penumbra-zone/web/pull/1980/files#diff-caad9b17fa66f6badd2a44a36749fc51c4452f35e34f7fe38905467885f60f0bR8) basically says: "for liquidity positions, we just want to fallback to a default priority fee selection", and we [skip trying to extract](https://github.com/penumbra-zone/web/pull/1980/files#diff-caad9b17fa66f6badd2a44a36749fc51c4452f35e34f7fe38905467885f60f0bR153) the asset ID for the fee from the TPR (which we were [previously trying to do](https://github.com/penumbra-zone/web/pull/1980#discussion_r1917301704)).

TalDerei:
note to reviewers: to summarize, LP-actions (limit orders and range liquidity) will default to using USDC as the [priority asset](https://github.com/penumbra-zone/web/pull/1980/files#diff-8298d111bc94531ec943d5f024a9ab57657ad28670caea51e3aa9eda88bf3954R475) for the _alternative_ fee. If the user's USDC balance is zero, it will select another alternative asset for which the user has an positive balance. This PR was _not_ intended to modify the existing fee mechanism we've already built for existing actions, but rather extend it to fix a bug related to liquidity provisioning on fresh accounts. 

I'd like to stage this PR as part of the upcoming prax release, `v15.1` https://github.com/prax-wallet/prax/pull/263

erwanor (approved):

TalDerei:
note to reviewers: will merge this later tonight after one last holistic review, but it's good.

Issue #1363, opened 2024-06-28
Transactions table should populate recent transactions first


turbocrime:
On the minifront transactions table, transaction info is streamed to the page.

They populate in essentially random order: they stream in lexical order of their key, which is a base64 encoding of their transaction hash.

Ideally, the most recent transactions known would populate first, and earlier transactions would load as the user scrolls down the page.

turbocrime:
I experimented with this issue today. Unfortunately it looks like this is impossible given current idb schema constraints

1. we should add an index of transactions by height. this is doable.
2. we should query by this index. this is doable.

But, the index is also a string which sorts in lexical order, because the height in database is a jsonified bigint.

The situation is improved but still clearly incorrect.

https://github.com/penumbra-zone/web/assets/134443988/e9960065-07b5-4b07-9981-65d6f5b37770

turbocrime:
if idb schema used real types instead of stringifying everything, many operations like this would be more sane and just work as expected.

turbocrime:
if the string key bounds are interpreted lexically instead of numerically, it may not even be possible to achieve numeric ranges with idb key bounds. this could create problems long term.

TalDerei:
closed by https://github.com/penumbra-zone/web/pull/1996 and improvements in https://github.com/penumbra-zone/web/issues/2009

PR #1996, opened 2025-01-25
transaction table: index store by height 

TalDerei:
## Description of Changes

Previously, transactions were populating in a random order. To address this, we load transactions in order of block height and the [storage](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts) package has been updated to:
- extend the idb schema to _Index_ the `TRANSACTIONS` object store by height,
- introduce a new `TRANSACTION_INFO` object store to save the transaction perspectives (TxP) and transaction views (TxV)
- introduce new `saveTransactionInfo` and `getTransactionInfo` storage methods

We lazily compute and store TxP + TxV from expensive calls like [generateTransactionInfo](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/transaction-info.ts#L23) rather than redundantly recalculating them every page reload. 

These changes introduce a breaking change to backwards compatibility due to database schema changes and necessary DB version bump, requiring the client to resync.

## Related Issue

references https://github.com/penumbra-zone/web/issues/1363

------------------------

loading ~400 transactions is much faster!

https://github.com/user-attachments/assets/00bca6cd-99d8-44cd-9d4b-aa88292ce4f4

### DIFF
--- a/.changeset/sixty-fishes-travel.md
+++ b/.changeset/sixty-fishes-travel.md
@@ -1,0 +1,7 @@
+---
+'@penumbra-zone/storage': major
+'@penumbra-zone/services': minor
+'@penumbra-zone/types': minor
+---
+
+transaction table indexes by height and save txp and txv in indexdb
--- a/packages/services/src/test-utils.ts
+++ b/packages/services/src/test-utils.ts
@@ -34,6 +34,8 @@ export interface IndexedDbMock {
   upsertAuction?: Mock;
   hasTokenBalance?: Mock;
   saveGasPrices?: Mock;
+  saveTransactionInfo?: Mock;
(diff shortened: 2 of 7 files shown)


.changeset/sixty-fishes-travel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@penumbra-zone/storage': major
```
TalDerei, line 2: idb storage bump

VanishMax (approved):
Oh wow, love the result!

Probably we should wait a little with applying this to the newer version of Prax since I also plan to introduce new changes to IDB that would best work after resync, so we could combine the changes together

vacekj:
Great work Tal.

Issue #1394, opened 2024-07-02
make transaction approval dialogue actually understandable


TalDerei:
One long-term action item we discussed is improving the readability of the extension's approval dialogue. Users perform actions (user-defined state transitions), and the transaction internally balances by spending UTXO notes and receiving change notes. In the current approval dialogue, we display the raw actions but fail to provide sufficient context—specifically, the concept of a UTXO note and the balancing logic necessary to understand these actions. We need to transform the approval dialogue to be more human-readable, minimally displaying the input and output amounts. This may require an additional layer in our translator logic to perform this transformation.

grod220:
I really like Rabby wallet's pre-sign simulation:

<img width="496" alt="Screenshot 2024-07-02 at 9 37 54 AM" src="https://github.com/penumbra-zone/web/assets/16624263/6514557a-accc-410f-9075-91509e9f51a8">

It makes it super clear what's going out and what's coming in. I feel like we could do something similar if we aggregate in/out note amounts.

PR #2000, opened 2025-01-27
feat: transaction summary

TalDerei:
## Description of Changes

transaction summary scaffolding that will be hooked into https://github.com/penumbra-zone/web/issues/2051 (implemented by https://github.com/penumbra-zone/web/pull/2083). It remained a draft until the `TransactionSummary` UI component was ready. 

## Related Issue

references https://github.com/penumbra-zone/penumbra/pull/4943 and https://github.com/penumbra-zone/web/issues/1394


packages/services/src/view-service/transaction-info.ts
```diff
@@ -38,6 +41,9 @@ export const transactionInfo: Impl['transactionInfo'] = async function* (_req, c
         indexedDb.constants(),
       );
 
+      let tx_summary = await generateTransactionSummary(txv);
```
VanishMax, line 44: question: do we need to add a summary field to the TransactionInfoResponse protobuf?

VanishMax: tried doing it myself: https://github.com/penumbra-zone/penumbra/pull/5102

TalDerei: that's what I was thinking, or alternatively a view service rpc for transaction summaries that's decoupled from `transactionInfo`

packages/storage/src/indexed-db/config.ts
```diff
@@ -2,4 +2,4 @@
  * The version number for the IndexedDB schema. This version number is used to manage
  * database upgrades and ensure that the correct schema version is applied.
  */
-export const IDB_VERSION = 49;
+export const IDB_VERSION = 50;
```
VanishMax, line 5: question: what happens if we don't bump this version? The changes seem to be backwards compatible and would work without resync... or not?

VanishMax (approved):
generally approve but confused about the version bump

VanishMax:
@TalDerei committed an update to `transactionInfoByHash` method too, the DEX needs it even more

TalDerei:
> @TalDerei committed an update to `transactionInfoByHash` method too, the DEX needs it even more

ah yes, thanks! will remove the idb version bump and make the summary field optional, which can later be updated. It shouldn't break backwards compatibility

PR #2006, opened 2025-01-28
feat: #1989: subaccount filter in `ownedPositionIds`

### LINKED ISSUE #1989: view: implement subaccount filtering for `ownedPositionIds`
**Is your feature request related to a problem? Please describe.**
In https://github.com/penumbra-zone/penumbra/pull/4992 we amended the Penumbra view protocol protos to incorporate an optional account filter in the `OwnedPositionIds` RPC of the `View` service.

**Describe the solution you'd like**
Implement this RPC in the web view service so that the DEX explorer can do proper segmentation by sub-account.

VanishMax:
Closes #1989

Relates to https://github.com/prax-wallet/prax/pull/279 (Prax) and https://github.com/penumbra-zone/dex-explorer/pull/329 (DEX)

Changes:
- protobuf: sync latest changes in penumbra protobufs. `subaccount` field appears in OwnedPositionIdsResponse buf
- wasm: change `isControlledAddress` method to `getIndexByAddress`. Needed for the block processor to return not just a boolean but the found subaccount
- types: sync types with updated wasm package
- storage: add subaccount filter to `getOwnedPositionIds` method. Database should be populated with subaccounts by the latest changes to the block processor
- services: add subaccount filter to `ownedPositionIds` method in ViewService

Note to reviewers. Before running, call `pnpm gen:penumbra` in `packages/protobuf`

### DIFF
--- a/packages/storage/src/indexed-db/indexed-db.test.ts
+++ b/packages/storage/src/indexed-db/indexed-db.test.ts
@@ -28,6 +28,8 @@ import {
   tradingPairGmGn,
   transaction,
   transactionId,
+  mainAccount,
+  firstSubaccount,
 } from './indexed-db.test-data.js';
 import { AddressIndex, WalletId } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import {
@@ -496,13 +498,13 @@ describe('IndexedDb', () => {
     it('position should be added and their state should change', async () => {
       const db = await IndexedDb.initialize({ ...generateInitialProps() });
 
-      await db.addPosition(positionIdGmPenumbraBuy, positionGmPenumbraBuy);
+      await db.addPosition(positionIdGmPenumbraBuy, positionGmPenumbraBuy, mainAccount);
       await db.updatePosition(
         positionIdGmPenumbraBuy,
         new PositionState({ state: PositionState_PositionStateEnum.CLOSED }),
       );
       const ownedPositions: PositionId[] = [];
-      for await (const positionId of db.getOwnedPositionIds(undefined, undefined)) {
+      for await (const positionId of db.getOwnedPositionIds(undefined, undefined, undefined)) {
         ownedPositions.push(positionId as PositionId);
       }
       expect(ownedPositions.length).toBe(1);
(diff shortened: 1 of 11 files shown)


vacekj (commented):
lgtm

vacekj (approved):

packages/wasm/crate/src/view_server.rs
```diff
@@ -307,13 +308,17 @@ impl ViewServer {
         Ok(root.encode_to_vec())
     }
 
-    /// Checks if address is controlled by view server full viewing key
+    /// Get the address index by provided address
+    /// Returns: `Uint8Array representing inner Address`
     #[wasm_bindgen]
-    pub fn is_controlled_address(&self, address: &[u8]) -> WasmResult<bool> {
+    pub fn get_index_by_address(full_viewing_key: &[u8], address: &[u8]) -> WasmResult<JsValue> {
```
TalDerei, line 314: question: what's the functional difference between this and the other [get_index_by_address](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/keys.rs#L115) method in the keys dir. are these not duplicates?

packages/storage/src/indexed-db/indexed-db.test.ts
```diff
@@ -496,13 +498,13 @@ describe('IndexedDb', () => {
     it('position should be added and their state should change', async () => {
```
TalDerei, line 498: suggestion: can we add some other tests 

1. call `getOwnedPositionIds` with no positions in the database
2. check that combining all filters (PositionState, tradingPair, subaccount) works
3. check that it filters by subaccount works when multiple subaccounts exist

TalDerei (approved):
nice work! added some minor feedback

packages/wasm/crate/src/view_server.rs
TalDerei: inspecting the `wasm/index.d.ts`, it seems like the caller `getAddressIndexByAddress` is using the `get_index_by_address` method in keys

VanishMax: yes, that's right. The block processor uses view server implementation from wasm, so i tried to stick to it without importing methods from other parts of the code

TalDerei: block processor calls `getIndexByAddress`, which internally calls `getAddressIndexByAddress` view service method, which calls `get_index_by_address`, which is _agnostic_ to which file the wasm method lives in because it's calling the method from `../wasm/index.js` which is a global view of all `wasm_bindgen` methods.

TalDerei:
@VanishMax let's also hold on merging the [version packages](https://github.com/penumbra-zone/web/pull/1999) PR in favor of also including https://github.com/penumbra-zone/web/pull/1996 (I have to push an update to that changeset) to bundle all the web changes to be consumed by prax.

packages/types/src/indexed-db.ts
```diff
@@ -97,9 +97,14 @@ export interface IndexedDbInterface {
   getOwnedPositionIds(
     positionState: PositionState | undefined,
     tradingPair: TradingPair | undefined,
+    subaccount: AddressIndex | undefined,
   ): AsyncGenerator<PositionId, void>;
-  addPosition(positionId: PositionId, position: Position): Promise<void>;
-  updatePosition(positionId: PositionId, newState: PositionState): Promise<void>;
+  addPosition(positionId: PositionId, position: Position, subaccount?: AddressIndex): Promise<void>;
```
TalDerei, line 102: ACK subaccount in `POSITIONS` table 

Screenshot 2025-01-28 at 7 55 29 PM

TalDerei:
note: modifying the db schema requires bumping the `IDB_VERSION`, but extending the storage methods doesn't. Since we're already bumping it in a separate PR, it's fine.

packages/wasm/crate/src/view_server.rs
VanishMax: agree, this change wasn't really needed. Removed all WASM code updates in favor of the old version

packages/storage/src/indexed-db/indexed-db.test.ts
VanishMax: added 1 and 2, number 3 is sufficiently covered by 'should get all position with given subaccount index' test

PR #2018, opened 2025-02-04
don't retain port in `CRSessionClient`, remove singleton pattern

### LINKED ISSUE #1736: `CRSessionClient` attempts to retain reference to a transferred `MessagePort` object
initial design of `@penumbra-zone/transport-chrome` and `CRSessionClient` made the assumption that one `chrome.runtime.Port`/`MessageChannel` pair should exist per document lifecycle. this simplified reasoning and implementation, and did not affect actual `Transport` lifecycle, which is independent of the connection.

this assumption was extended to the initial design of the `PenumbraProvider` page api of `@penumbra-zone/client`. But the development of ADR-006 changed this, as api consumers desired multiple lifecycles per page.

in 88b16a7f055ce79fb5ec594f8b39ba86d464843e, `CRSessionClient` was updated minimally to support this, but it seems this minimal change was ultimately incorrect: desiring to re-use the same channel, it now attempts to retain reference to the page's `MessagePort` which is ultimately transferred out of its isolated content-script execution context and into the mainworld execution context. transferred objects are no longer available in their original context, and any reference will throw on access.

this didn't affect prax's impelementation of ADR-006 directly, but did affect prax's attempt at resolving a backwards compatibility issue https://github.com/prax-wallet/web/pull/173#issuecomment-2309810495

### don't retain reference to any transferred object

the current implemetation attempts to re-provide the same channel when called again, but this is incorrect and any implementation of that is impossible. consuming content-scripts likewise cannot be responsible for retaining and re-sending the `MessagePort` when necessary, as this violates the same restriction of transferred objects.

### don't support re-providing the same `MessagePort`

result of the above objective. this will affect implementation of consumer's isolated content scripts, but should be simplifying.

### don't require any `PenumbraProvider` behavioral change

interfaces and behavior defined by ADR-006 should not be affected. unlikely to be a problem.

## implementation

there are a few possible implementations. A and B contain potential handler leaks, but solving that as in C requires more extensive lifecycle management which is unclear.

### option A: create multiple `MessagePort`s internally

instead of trying to retain the transferred `MessagePort`, simply create a new `MessageChannel`, attach the existing listeners, and transfer a new port.

benefits: remain the same externally.

downsides: intentionally, no routing exists at this level, but now some basic routing must be implemented. potential leak of handlers to unused channels. potential leak of streams buffering to irrelevant channels.

it seems initially simple to just broadcast each response to every channel, as the transport channel clients discard responses labelled with an unrecognized request id. but `ReadableStream`s are created here, and ultimately transferred out of the execution context - so either the streams must be teed with some kind of cleanup to prevent memory leak, or some basic routing is required.

### option B: remove the singleton restriction

instead of trying to retain the transferred `MessagePort`, only provide it once. require callers to create a new session client to obtain a port again.

beneftis: minimal changes, but not the same externally. instead of conditional init on first request, an isolated content script would init a session client for every connection request.

downsides: it's not currently possible to identify the appropriate end of session client lifetime, so this has the same potential listener leak as A.

### option C: option B plus more involved transport-level lifecycle management

this would require changes in `@penumbra-zone/transport-dom` and the channel-level messaging, so the session client may be aware of the presently connected transports. current expectations of transport use don't provide an obvious way to identify an unused transport.

## suggestion

i prefer option B. this reduces new behaviors and code diff, at the cost of a small api diff. with option B, changes are restricted to `@penumbra-zone/transport-chrome`. its use is purely internal to consumers, so consumers may upgrade the package at their leisure.

its worst case is that a page may make a connection request and a new transport for every method call, and the providing mainworld content-script posts a new connection request to get a new port for each of those.

but users of `@penumbra-zone/client` are expected to retain a port and not make prolific requests. with the removal of the initial assumptions about channel lifecycle, and the additional state tracking required by ADR-006, providers implementations can correctly understand when to discard/reacquire a port.

turbocrime:
## Description of Changes

`CRSessionClient` of `@penumbra-zone/transport-dom` no longer retains a `MessagePort` and no longer uses a singleton pattern.

## Related Issue

fixes #1736 

rebased onto #2040 

Tested in combination with other work, not yet tested alone.

### DIFF
--- a/packages/transport-chrome/src/session-client.ts
+++ b/packages/transport-chrome/src/session-client.ts
@@ -1,7 +1,6 @@
 /**
  * CRSessionClient is a Chrome runtime session client: it handles channel
- * Transport client sessions in the Chrome runtime.  Intended for use as a
- * document singleton, in a content script.
+ * Transport client sessions in the Chrome runtime.
  *
  * Simple handlers unconditionally forward messages back and forth. Chrome
  * runtime disconnect is detected and surfaced as an error.
@@ -56,27 +55,24 @@ const localErrorJson = (err: unknown, relevantMessage?: unknown) =>
       };
 
 export class CRSessionClient {
-  private static singleton?: CRSessionClient;
-  private servicePort: chrome.runtime.Port;
-  private clientPort: MessagePort;
-  public inputPort: MessagePort;
+  private readonly sessionName: string;
+  private readonly servicePort: chrome.runtime.Port;
 
-  private constructor(private prefix: string) {
-    if (CRSessionClient.singleton) {
-      throw new Error('Already constructed');
-    }
+  private constructor(
+    private readonly prefix: string,
+    private readonly clientPort: MessagePort,
+  ) {
+    this.sessionName = nameConnection(prefix, ChannelLabel.TRANSPORT);
 
-    const { port1, port2 } = new MessageChannel();
-    this.clientPort = port1;
-    this.inputPort = port2;
-
+    // listen to service
     this.servicePort = chrome.runtime.connect({
       includeTlsChannelId: true,
-      name: nameConnection(prefix, ChannelLabel.TRANSPORT),
+      name: this.sessionName,
     });
-
-    this.servicePort.onMessage.addListener(this.serviceListener);
     this.servicePort.onDisconnect.addListener(this.disconnect);
+    this.servicePort.onMessage.addListener(this.serviceListener);
+
+    // listen to client
     this.clientPort.addEventListener('message', this.clientListener);
     this.clientPort.start();
   }
@@ -88,16 +84,18 @@ export class CRSessionClient {
    * @returns a `MessagePort` that can be provided to DOM channel transports
    */
   public static init(prefix: string): MessagePort {
-    CRSessionClient.singleton ??= new CRSessionClient(prefix);
-    return CRSessionClient.singleton.inputPort;
+    const { port1, port2 } = new MessageChannel();
+    new CRSessionClient(prefix, port1);
+    return port2;
   }
 
   private disconnect = () => {
     this.clientPort.removeEventListener('message', this.clientListener);
     this.clientPort.postMessage(false);
     this.clientPort.close();
+    this.servicePort.onMessage.removeListener(this.serviceListener);
+    this.servicePort.onDisconnect.removeListener(this.disconnect);
     this.servicePort.disconnect();
-    CRSessionClient.singleton = undefined;
   };
 
   private clientListener = (ev: MessageEvent<unknown>) => {
--- a/.changeset/violet-coats-remember.md
+++ b/.changeset/violet-coats-remember.md
@@ -1,0 +1,9 @@
+---
+'@penumbra-zone/transport-chrome': minor
+---
+
+Transport session reliability improvements.
+
+- No external API change.
+- Remove singleton restriction of `CRSessionClient`.
+- Don't retain port reference to respect object transfer rules.
--- a/packages/transport-chrome/src/session-client.test.ts
+++ b/packages/transport-chrome/src/session-client.test.ts
@@ -61,7 +61,7 @@ describe('CRSessionClient', () => {
   });
 
   describe('repeated calls to init', () => {
-    it('should return the same port for each call', () => {
+    it.fails('should return the same port for each call', () => {
       const ports: MessagePort[] = [];
 
       for (let i = 0; i < 3; i++) {
@@ -72,7 +72,7 @@ describe('CRSessionClient', () => {
       }
     });
 
-    it.fails('should return a new port for each call', () => {
+    it('should return a new port for each call', () => {
       const ports: MessagePort[] = [];
 
       for (let i = 0; i < 3; i++) {


packages/transport-chrome/src/session-client.ts
```diff
@@ -56,27 +55,25 @@ const localErrorJson = (err: unknown, relevantMessage?: unknown) =>
...
-    this.clientPort = port1;
-    this.inputPort = port2;
+  private constructor(
+    private readonly prefix: string,
+    private readonly clientPort: MessagePort,
+  ) {
+    this.sessionName = nameConnection(prefix, ChannelLabel.TRANSPORT);
 
     this.servicePort = chrome.runtime.connect({
+      name: this.sessionName,
```
TalDerei, line 70: ACK – unique `sessionName` from prefix + label

packages/transport-chrome/src/session-client.ts
```diff
@@ -56,27 +55,25 @@ const localErrorJson = (err: unknown, relevantMessage?: unknown) =>
...
       includeTlsChannelId: true,
-      name: nameConnection(prefix, ChannelLabel.TRANSPORT),
     });
-
     this.servicePort.onMessage.addListener(this.serviceListener);
-    this.servicePort.onDisconnect.addListener(this.disconnect);
+    this.servicePort.onDisconnect.addListener(this.serviceReconnecter);
```
TalDerei, line 74: ACK – seems to maintain session identity across reconnections! `onDisconnect()` triggers `serviceReconnecter`, which creates new connection using the _same_ `sessionName` stored during construction. It then seems like `chrome.runtime.connect()` wakes up the service worker and reattaches the listeners.

packages/transport-chrome/src/session-client.ts
```diff
@@ -88,16 +85,29 @@ export class CRSessionClient {
    * @returns a `MessagePort` that can be provided to DOM channel transports
    */
   public static init(prefix: string): MessagePort {
-    CRSessionClient.singleton ??= new CRSessionClient(prefix);
-    return CRSessionClient.singleton.inputPort;
+    console.debug('session-client init', prefix);
+    const { port1, port2 } = new MessageChannel();
```
TalDerei, line 89: ACK – ephemeral input and client ports

packages/transport-chrome/src/session-client.ts
```diff
@@ -88,16 +85,29 @@ export class CRSessionClient {
...
+    const session = new CRSessionClient(prefix, port1);
+    CRSessionClient.sessions.add(session);
+    console.debug('session-client init', session, CRSessionClient.sessions);
+    return port2;
```
TalDerei, line 93: ACK – input port to be forwarded and exposed in webpage for sending messages

packages/transport-chrome/src/session-client.ts
```diff
@@ -88,16 +85,29 @@ export class CRSessionClient {
...
   }
 
   private disconnect = () => {
     this.clientPort.removeEventListener('message', this.clientListener);
     this.clientPort.postMessage(false);
     this.clientPort.close();
     this.servicePort.disconnect();
-    CRSessionClient.singleton = undefined;
+    CRSessionClient.sessions.delete(this);
```
TalDerei, line 101: ACK – session cleanup; removes session from set on disconnect

packages/transport-chrome/src/session-client.ts
```diff
@@ -88,16 +85,29 @@ export class CRSessionClient {
...
+  };
+
+  private serviceReconnecter = () => {
+    this.servicePort = chrome.runtime.connect({
+      name: this.sessionName,
+      includeTlsChannelId: true,
+    });
+    this.servicePort.onMessage.addListener(this.serviceListener);
+    this.servicePort.onDisconnect.addListener(this.serviceReconnecter);
```
TalDerei, line 110: comment: potential listener leaks aren't handled (stale / duplicate listeners) – wondering where else we may be leaking memory?

TalDerei (commented):
this is great! went through the `CRSessionClient` connection flow and left a comment about addressing listener leaks.

packages/transport-chrome/src/session-client.ts
```diff
@@ -88,16 +85,29 @@ export class CRSessionClient {
...
   };
```
turbocrime, line 111: todo: remove listeners from serviceport before connecting a new serviceport

packages/transport-chrome/src/session-client.ts
```diff
@@ -133,7 +143,9 @@ export class CRSessionClient {
...
 
   private acceptChannelStreamResponse = ({ requestId, channel: name }: TransportInitChannel) => {
-    const stream = new ReadableStream(new PortStreamSource(chrome.runtime.connect({ name })));
+    const stream = new ReadableStream(
+      new PortStreamSource(chrome.runtime.connect({ name, includeTlsChannelId: true })),
+    );
```
TalDerei, line 148: comment: remove and handle in separate PR

TalDerei (approved):
changed base branch to `connection-lifecycle-branch`

we can think more about handling [memory leaks](https://github.com/penumbra-zone/web/pull/2018#discussion_r1941751713) separately in the reconnection PR.

packages/transport-chrome/src/session-client.ts
turbocrime: reconnector is now scoped to a different pr. cleaned up the disconnect callback a bit

packages/transport-chrome/src/session-client.ts
turbocrime: again, reconnector is now scoped to a different pr

packages/transport-chrome/src/session-client.ts
turbocrime: maintained session name attribute even though reconnector is now omitted

packages/transport-chrome/src/session-client.ts
turbocrime: after research, it is apparent that chrome runtime will handle listener destruction automatically. there is no listener leak concern on disconnected ports.

PR #2019, opened 2025-02-04
fix: latest swaps

VanishMax:
After discussing Prax <–> Pindexer connection with protocol devs, it turned out that Pindexer stores swap execution data based on the time when the swap happened, but `latestSwaps` method of the ViewService was returning the height when user claimed the swap. This PR fixes the issue, so that `latestSwaps` can be easily merged with Pindexer data


TalDerei (approved):
👍

PR #2034, opened 2025-02-07
session client reconnect

turbocrime:
session-client will now automatically request to reconnect when disconnected.

to support this, session-manager now uses a callback parameter to validate connection requests, instead of depending on external management of port injection.

considerations:

 - reconnect won't automatically retry individual failed requests
 - reconnect can't prevent in-flight requests from being interrupted by a disconnect
 - reconnect can't prevent a page from persistently using an actually-dead transport

discovered issues:

- permissions check must be async, so listener attachment became async. this causes a race condition under which early requests may arrive before a listener is attached.

potential issues:

- some clients may be using structural typing or string matching to detect errors. continuing to investigate possible error detection problems.

### DIFF
--- a/packages/transport-chrome/src/session-client.ts
+++ b/packages/transport-chrome/src/session-client.ts
@@ -48,7 +48,7 @@ export class CRSessionClient {
       includeTlsChannelId: true,
       name: this.sessionName,
     });
-    this.servicePort.onDisconnect.addListener(this.disconnect);
+    this.servicePort.onDisconnect.addListener(this.reconnect);
     this.servicePort.onMessage.addListener(this.serviceListener);
 
     // listen to client
@@ -108,8 +108,7 @@ export class CRSessionClient {
   };
 
   /**
-   * Used to tear down this session when the client announces channel closure,
-   * or when the extension channel disconnects.
+   * Used to tear down this session when the client announces channel closure.
    *
    * Announces closure from this side towards the document, and ensures closure
    * of both ports. Listeners are automatically gargbage-collected. This session
@@ -122,6 +121,19 @@ export class CRSessionClient {
 
     this.clientPort.close();
     this.servicePort.disconnect();
+  };
+
+  /**
+   * Used when the service port disconnects. Tears down the service port and
+   * reconnects, creating a new service port.
+   */
+  private reconnect = () => {
+    this.servicePort = chrome.runtime.connect({
+      name: this.sessionName,
+      includeTlsChannelId: true,
+    });
+    this.servicePort.onMessage.addListener(this.serviceListener);
+    this.servicePort.onDisconnect.addListener(this.reconnect);
   };
 
   /**
--- a/packages/transport-chrome/src/session-manager.ts
+++ b/packages/transport-chrome/src/session-manager.ts
@@ -28,25 +28,6 @@ type SenderWithOrigin = chrome.runtime.MessageSender & { origin: string };
 type PortWithOrigin = chrome.runtime.Port & { sender: SenderWithOrigin };
 export type CheckPortSenderFn = (port: chrome.runtime.Port) => Promise<PortWithOrigin>;
 
-// Move port validation logic to a distinct function, anticipating to parameterize it.
-const defaultCheckPortSender: CheckPortSenderFn = port => {
-  // Allow connections from the same extension, from https pages, or from http://localhost
-  if (port.sender?.origin) {
-    const fromThisExtension = port.sender.id === chrome.runtime.id;
-    const fromPageHttps =
-      !port.sender.frameId && !!port.sender.tab?.id && port.sender.origin.startsWith('https://');
-    const isLocalhost =
-      port.sender.origin.startsWith('http://localhost:') ||
-      port.sender.origin === 'http://localhost';
-
-    if (isLocalhost || fromPageHttps || fromThisExtension) {
-      return Promise.resolve(port as unknown as PortWithOrigin);
-    }
-  }
-
-  throw new Error('Invalid sender');
-};
-
 /**
  * Only for use as an extension-level singleton by the extension's main
  * background worker.
@@ -81,7 +62,7 @@ export class CRSessionManager {
    * @param handler your router entry function
    * @param checkPortSender a function used to validate the sender of a connection
    */
-  private constructor(
+  constructor(
     private readonly managerId: string,
     private readonly handler: ChannelHandlerFn,
     private readonly checkPortSender: CheckPortSenderFn,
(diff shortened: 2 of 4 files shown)


packages/transport-chrome/src/session-client.ts
```diff
@@ -15,17 +15,23 @@ import { ChannelLabel, nameConnection } from './channel-names.js';
...
+  let header: HeadersInit | undefined = undefined;
+  if (o != null && typeof o === 'object' && 'header' in o) {
+    try {
+      header = new Headers(o.header as never);
+    } catch (e) {
+      // not headers
+    }
+  }
+  return header;
+};
```
TalDerei, line 28: `getHeader` is only used in unsupported cases in `clientListener`, so I think we can remove it.

packages/transport-chrome/src/session-client.ts
```diff
@@ -15,17 +15,23 @@ import { ChannelLabel, nameConnection } from './channel-names.js';
...
 /**
  * Transparently adapts `Transport`s using `MessageChannel` DOM connections to
  * use Chrome `chrome.runtime.Port` extension connections.
  *
  * Simple handlers unconditionally forward messages back and forth. Channel
- * disconnects are detected, and will terminate the session.
+ * disconnects are detected, and transparently re-established if possible.
```
TalDerei, line 34: 👍

packages/transport-chrome/src/session-client.ts
```diff
@@ -37,7 +43,8 @@ declare global {
  */
 export class CRSessionClient {
   private readonly sessionName: string;
-  private readonly servicePort: chrome.runtime.Port;
+  private servicePort: chrome.runtime.Port;
+  private pendingChannels = new Map<string, AbortController>();
```
TalDerei, line 47: `pendingChannels` is specific to client streaming requests which we don't support, so we can remove it. For completeness, we should also remove any client streaming stubs elsewhere, and reimplement separately if a use case arrises.

packages/transport-chrome/src/session-client.ts
```diff
@@ -52,7 +59,7 @@ export class CRSessionClient {
       includeTlsChannelId: true,
       name: this.sessionName,
     });
-    this.servicePort.onDisconnect.addListener(this.disconnect);
+    this.servicePort.onDisconnect.addListener(this.reconnect);
```
TalDerei, line 62: ACK – references `reconnect` method in https://github.com/penumbra-zone/web/pull/2018#discussion_r1941747552 that was moved into this PR

packages/transport-chrome/src/session-client.ts
```diff
@@ -80,7 +87,7 @@ export class CRSessionClient {
   /** Just some typed wrappers for consistency, in a namespace. */
   private send = {
     /** Forward a service-emitted response or error (response failure) to the client. */
-    response: (msg: TransportMessage | TransportStream | TransportError<string>) => {
+    response: (msg: TransportMessage | TransportStream | TransportError) => {
```
TalDerei, line 90: ACK – `transport-dom` messages have changed by remove the type param

packages/transport-chrome/src/session-client.ts
```diff
@@ -120,11 +126,25 @@ export class CRSessionClient {
   private disconnect = () => {
     // announce closure, ok even if already closed
     this.clientPort.postMessage(false);
+    this.pendingChannels.forEach(ac => ac.abort());
```
TalDerei, line 129: remove based on comments above.

packages/transport-chrome/src/session-client.ts
```diff
@@ -148,6 +168,12 @@ export class CRSessionClient {
         if (isTransportAbort(tev, requestId)) {
           // abort control for a specific request
           this.send.request(tev);
+
+          // only pending client-stream requests are aborted here.
+          // - active client-stream requests abort via stream control.
+          // - server-stream responses abort via stream control.
+          // - non-streaming messages need no abort control.
+          this.pendingChannels.get(requestId)?.abort();
```
TalDerei, line 176: remove based on comments above.

packages/transport-chrome/src/session-client.ts
```diff
@@ -159,7 +185,7 @@ export class CRSessionClient {
           const unsupported = new ConnectError(
             'Unsupported request from client',
             Code.Unimplemented,
-            tev.header,
+            getHeader(tev),
```
TalDerei, line 188: remove based on comments above.

packages/transport-chrome/src/session-client.ts
```diff
@@ -241,26 +267,30 @@ export class CRSessionClient {
...
    */
   private requestChannelStream = ({ requestId, stream }: TransportStream): TransportInitChannel => {
+    if (!globalThis.__DEV__) {
+      throw new ConnectError('Unknown request kind', Code.Unimplemented);
+    }
     const channel = nameConnection(this.managerId, ChannelLabel.STREAM);
 
+    const ac = new AbortController();
+    this.pendingChannels.set(requestId, ac);
+
```
TalDerei, line 277: remove based on comments above.

TalDerei (approved):
Looks good to me, left some suggestions 👍

I don’t think we need to split the diff for the session manager changes — we can just merge it directly

packages/transport-chrome/src/session-manager.ts
```diff
@@ -1,23 +1,33 @@
...
+  signal: AbortSignal;
+  sessionId: string;
   port: chrome.runtime.Port;
   origin: string;
+  requests: Map<string, AbortController>;
 }
 
+type SenderWithOrigin = chrome.runtime.MessageSender & { origin: string };
+type PortWithOrigin = chrome.runtime.Port & { sender: SenderWithOrigin };
+export type CheckPortSenderFn = (port: chrome.runtime.Port) => Promise<PortWithOrigin>;
```
TalDerei, line 29: ACK – origin validation in the session manager

packages/transport-chrome/src/session-manager.ts
```diff
@@ -85,65 +132,107 @@ export class CRSessionManager {
...
+      session.requests.forEach(request => request.abort(session.signal.reason));
+      if (this.sessions.delete(sessionId)) {
+        port.disconnect();
       }
-    });
+    };
+
+    const sessionDisconnectListener = () => {
+      console.debug('sessionDisconnectListener', sessionId);
+      if (this.sessions.delete(sessionId)) {
```
TalDerei, line 187: ACK – session cleanup: removes session from sessions map, which previously happened in the abort controller. should this be done directly in the disconnect listener?

packages/transport-chrome/src/session-client.ts
turbocrime: resolution discussed on call: client streaming support is defined out with `globalThis.__DEV__` at build time

packages/transport-chrome/src/session-client.ts
turbocrime: type changes moved to https://github.com/penumbra-zone/web/pull/2058/

PR #2040, opened 2025-02-11
mock chrome runtime messaging and test transport-chrome

turbocrime:
implement mocks for chrome runtime messaging in a new package. this is an alternative approach to implementing a new testing harness with puppeteer and dedicated dapp/ext fixtures

use mocks to test transport-chrome. ~~session client tests are mostly written, session manager tests are less complete. after unit tests of session-client and session-manager are implemented, the mocks could be used to test the full transport and connectrpc stack.~~

many tests written.

this will be ordered before #2018, then #2034 will break down into two sets of changes: an incremental improvement PR, and then the lifecycle changes.

testing will help ensure the series of changes in connection-lifecycle-branch aren't changing unexpected behavior/functionality. a complete solution testing real interaction with puppeteer is still ideal, but the mock approach has already been useful.

edit: some review commentary below.

### DIFF
--- a/packages/transport-chrome/src/session-client.test.ts
+++ b/packages/transport-chrome/src/session-client.test.ts
@@ -1,250 +1,556 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment -- tests */
-import { vi, describe, it, expect, beforeEach, Mock } from 'vitest';
+import { ConnectError } from '@connectrpc/connect';
+import { errorToJson } from '@connectrpc/connect/protocol-connect';
+import { mockChannel, MockedChannel, MockedPort } from '@penumbra-zone/mock-chrome/runtime/connect';
+import type { TransportMessage, TransportStream } from '@penumbra-zone/transport-dom/messages';
+import { afterEach, beforeEach, describe, expect, it, type Mock, onTestFinished, vi } from 'vitest';
+import { ChannelLabel, nameConnection } from './channel-names.js';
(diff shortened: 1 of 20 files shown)


packages/transport-chrome/src/session-client.test.ts
```diff
@@ -0,0 +1,323 @@
...
+    await vi.waitFor(() =>
+      expect(mockConsole.warn).toHaveBeenCalledWith('Unknown item from client', bustedMessage),
+    );
+
+    expect(extOnMessageListener).not.toHaveBeenCalled();
+    expect(domMessageErrorHandler).not.toHaveBeenCalled();
+    expect(domMessageHandler).not.toHaveBeenCalled();
+  });
+
+  it.fails('should respond with failures if the input is not an expected type', async () => {
```
turbocrime, line 194: this test is for newer functionality and is expected to fail until that merges

turbocrime (commented):
this is huge, sorry, but it's all tests.

mocks are used to test transport features and session manager features.

some complex things like streaming weren't possible to test with the mocks, but even partial coverage is an improvement.

some testing of improvement to come in later PRs is already present here, and skipped with `todo`. where that is the case, tests for present functionality affected is under a comment denoting this.

TalDerei (commented):
@turbocrime holding off final review for this changeset for now

I'd really like to first consolidate all the diffs here so we can create a more streamlined environment for manual testing and observing how components interact. Is that still the intention?

It's great to finally have a testing framework we can point to for validation 👍

Issue #2022, opened 2025-02-05
Tracking issue: LQT mechanism


TalDerei:
References https://github.com/penumbra-zone/penumbra/issues/5010; this is a tracking issue for end-to-end LQT support in the web / prax stack.

---------------

**1. view service component**
   - [x] [protocol support: view service rpcs](https://github.com/penumbra-zone/penumbra/pull/5062)
   - [x] #2027
   - [x] #2028

**2. wasm component**

   - [x] #2029

**3. IndexedDB storage component** 

   - [x] #2024
   - [x] #2025

**4. block processor component**

   - [x] #2026

**5. action views component**

   - [x] #2036

TalDerei:
closed by https://github.com/penumbra-zone/web/pull/2041 and https://github.com/prax-wallet/prax/pull/292. cc @erwanor

Issue #2009, opened 2025-01-31
Optimization: more idb queries could use indicies or key ranges


turbocrime:
A frequent pattern is to query an entire table, or cursor-scan an entire table, then filter the result in-memory to obtain a subset of items. Often this involves parsing of item json, conversion to/from binary, and/or regex operations on every table item.

This is slow. This has been discussed before:

- https://github.com/penumbra-zone/web/issues/312
- https://github.com/penumbra-zone/web/issues/486

Though not as featured as SQL, idb does provide query features beyond 'give me the whole table' that should be used.

### asset types

Commonly, this is for the purpose of identifying asset types - delegation assets, for example. Two approaches:

- By table index. Note tables and asset tables could add a field identifying asset type of the item, and add an index for this field. Then queries conditional on this field could be made directly and handled by the database engine.

- By key range. Tables keyed by bech32 asset id could be queried by key range, since bech32 strings for each type would be lexically ordered together.

### spend height

Commonly, notes are filtered by spend height - either detecting unspent notes, or in comparison to a specific height.

Note tables could add an index by spend height.

### considerations

Be aware that idb has some awkward requirements around compound queries https://stackoverflow.com/questions/23806635/searching-for-compound-indexes-in-indexeddb

Any idb change requires a migration.

TalDerei:
some early design decisions like [stringifying](https://github.com/penumbra-zone/web/issues/1363#issuecomment-2196189474) idb contents were proving to be less than desirable towards implementing https://github.com/penumbra-zone/web/pull/1996, but was still able to yield some speed benefits.

turbocrime:
more considerations

## bigint

idb does not support bigint indicies, but, bigints could be stringified and left-padded with zeroes to achieve lexical ordering equivalent to numeric ordering.

some desirable indicies such as height which are nominally bigint will effectively never exceed float64 values, so could be converted to js number for this purpose.

## byte strings

idb fully supports storage and indexing of byte strings

PR #2041, opened 2025-02-12
web: LQT integration

TalDerei:
## Description of Changes

end-to-end testing across view service, block processor, wasm, and indexedDB is challenging across separate PRs. This serves as a staging area to test and build on the code drafted in those individual draft PRs . Includes wasm, vitest, and testing against pcli for assurance. 

references tracking issue: https://github.com/penumbra-zone/web/issues/2022 and pairs with https://github.com/prax-wallet/prax/pull/292.

----------------------

**View Service:**

- `LqtVotingNotes` view service RPC retrieves the user's eligible voting notes for the current epoch, and performs nullifier queries against the nullifier set tracked by the nullifier state key in the funding service. Leverages existing IndexedDB storage method `getNotesForVoting` to retrieve eligible voting notes, and exposes an equivalent utility storage method in wasm. 

- `TournamentVotes` view service RPC retrieves the user vote for the associated incentivized asset in the current epoch.

----------------------

**Wasm:**

- necessary wasm plumbing to extend planner support for the `ActionLiquidityTournamentVote` action, which plans a transaction by directing all available voting power to a single asset. 

----------------------

**IndexedDB:**

- `LqtHistoricalVotes` table records historical votes, and corresponding storage method for saving historical votes and rewards after voting and during syncing. 

---------------

**Action Views:**

- Implements visible and opaque action view and translators in _deprecated_ UI library for testing purposes.

---------------------------------

note: https://github.com/penumbra-zone/web/pull/2041#issuecomment-2669626657 will be handled separately in https://github.com/penumbra-zone/web/issues/2065 as follow-up work.

-----------------

eg. LQT vote and reward detection during syncing and saving to IndexedDB.

Screenshot 2025-02-13 at 1 52 46 PM


packages/storage/src/indexed-db/index.ts
```diff
@@ -740,6 +793,23 @@ export class IndexedDb implements IndexedDbInterface {
...
+    for await (const cursor of this.db.transaction('EPOCHS', 'readonly').store) {
+      const currentEpoch = Epoch.fromJson(cursor.value);
+      if (currentEpoch.index === epoch_index) {
+        epoch = currentEpoch;
+      }
+    }
+
+    return epoch;
+  }
+
```
cronokirby, line 812: In what order does this iterate? This might be a bit slow if we can't just index into it, and we're iterating in order of earliest to latest epoch.

packages/wasm/crate/src/planner.rs
```diff
@@ -655,6 +656,56 @@ pub async fn plan_transaction_inner<Db: Database>(
...
+                    .to_string();
+            return Err(WasmError::Anyhow(anyhow!(error_message)));
+        }
+
+        // TODO: Consolidate staking notes into a single note and create a single action plan.
+        for staked_note in staked_notes {
+            let incentivized: Denom = incentivized
+                .clone()
+                .into_iter()
+                .next()
```
cronokirby, line 678: We should just make the proto have a single asset being incentivized, it's silly to make the field repeated and then ignore everything but the first value

packages/wasm/crate/src/planner.rs
```diff
@@ -655,6 +656,56 @@ pub async fn plan_transaction_inner<Db: Database>(
...
+                .ok_or_else(|| anyhow!("missing incentivized asset in liquidity tournament"))?
+                .try_into()?;
+            let rewards_recipient: Address = rewards_recipient
+                .clone()
```
cronokirby, line 682:
```suggestion
```

Cloning shouldn't be necessary since we feed this into try_into

cronokirby (commented):
A few small suggested changes, otherwise this looks good ; I think we should modify the TPR to only accept a single denom for incentivization, which is the current behavior of the code anyways. I'll submit a PR to that effect in the main repo.

packages/wasm/crate/src/planner.rs
cronokirby: https://github.com/penumbra-zone/penumbra/pull/5091

packages/perspective/src/plan/view-action-plan.ts
```diff
@@ -456,5 +456,8 @@ export const viewActionPlan =
         return new ActionView({
           actionView: actionPlan.action,
         });
+
+      default:
+        throw new Error(`Unsupported action case: ${actionPlan.action.case}`);
```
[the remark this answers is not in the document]
TalDerei: yeah stubbing for now – we can handle actions views separately in https://github.com/penumbra-zone/web/issues/2053

packages/storage/src/indexed-db/index.ts
```diff
@@ -410,6 +413,56 @@ export class IndexedDb implements IndexedDbInterface {
     });
   }
 
+  /**
+   * Retrieves liquidity tournament votes and rewards for a given epoch.
+   */
+  async getLQTHistoricalVotes(epoch: bigint): Promise<
```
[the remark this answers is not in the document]
TalDerei: artifact of operating under the assumption we'd be supporting variable voting weights, and therefore multiple votes in a single epoch. will change to `getLQTHistoricalVote` accordingly

packages/storage/src/indexed-db/index.ts
TalDerei: iterates by key order, which is the default when no index is specified. we also use [autoIncrement](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L161) for generating sequential keys. efficient lookups by index would require using a `createIndex` (eg. [like this](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L140)) on the object store, which we can certainly add.

packages/wasm/crate/src/planner.rs
TalDerei: yeah this is certainly misleading, will modify once protos are updated

packages/wasm/crate/src/planner.rs
TalDerei: surprised clippy didn't flag, will remove

packages/wasm/crate/src/planner.rs
```diff
@@ -655,6 +656,56 @@ pub async fn plan_transaction_inner<Db: Database>(
...
+    } in request.action_liquidity_tournament_vote
+    {
+        if staked_notes.is_empty() {
+            let error_message =
+                "Invalid transaction: zero delegation notes for voting in the liquidity tournament"
+                    .to_string();
+            return Err(WasmError::Anyhow(anyhow!(error_message)));
+        }
+
+        // TODO: Consolidate staking notes into a single note and create a single action plan.
```
TalDerei, line 673: TODO

TalDerei:
@cronokirby where is the rolling over of notes to prevent nullifier linkability handled?

packages/services/src/view-service/lqt-voting-notes.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (!lqtCheckNullifierResponse.alreadyVoted) {
+        spendableNoteRecords.push(votingNote.noteRecord as SpendableNoteRecord);
+      }
+    }
+  }
+
+  // Yield the SNRs that haven't been used for voting yet.
+  for (const record of spendableNoteRecords) {
+    yield new LqtVotingNotesResponse({ noteRecord: record });
+  }
```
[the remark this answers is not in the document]
TalDerei: we should also filter by subaccount here

cronokirby:
> @cronokirby where is the rolling over of notes to prevent nullifier linkability handled?

https://github.com/penumbra-zone/penumbra/blob/protocol/lqt_branch/crates/bin/pcli/src/command/tx/lqt_vote.rs#L107-L121

We insert a spend of every delegator vote we use, creating a new note.

(In writing this, I noticed that we needlessly create outputs, instead of letting change work its magic, so in implementing this logic I think all that's need is just adding a spend for each note)

TalDerei:
> We insert a spend of every delegator vote we use, creating a new note.

doesn't that go against this https://discord.com/channels/824484045370818580/983778100821262398/998722753085591612?

cronokirby:
> > We insert a spend of every delegator vote we use, creating a new note.
> 
> doesn't that conflict against the messaging in https://discord.com/channels/824484045370818580/983778100821262398/998722753085591612?

The voting action itself doesn't require a spend ; however, it does reveal information about the note being spent, which allows linking the use of the same note for voting across multiple epochs. However, what you can do is, in addition to voting, *actually* spend the notes, consolidating all the voting power you have into a single note, which will now be unlinkable, and allow you to vote in the next round with a single action instead of several. This is an additional feature we can add when actually planning such actions.

This can be done at:
1. the planner level (would advise against, it's useful to be able to not do this)
2. at the planner request level (unsure if we should do here, probably not)
3. when constructing the request (if we ever implement lqt voting in several places having to reimplement the logic might be a bit annoying, but a priori this seems like the best place to put the logic)

packages/perspective/src/plan/view-action-plan.ts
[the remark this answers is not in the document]
TalDerei: ended up implement the opaque and visible action views and translators in https://github.com/penumbra-zone/web/pull/2041/commits/febf46dc5e462905b95719b70c7dd6812293c2b4

--------------

_before_

Screenshot 2025-02-16 at 7 52 10 PM

_after_

Screenshot 2025-02-16 at 10 38 54 PM

-------------

_before_

Screenshot 2025-02-16 at 9 55 13 PM

_after_ 

Screenshot 2025-02-16 at 10 25 07 PM

Screenshot 2025-02-16 at 10 25 30 PM

packages/storage/src/indexed-db/index.ts
TalDerei: performance testing confirmed that iterating through epoch table does _not_ meaningfully impact transaction times. This optimization can be addressed as part of the larger effort in https://github.com/penumbra-zone/web/issues/2009

cronokirby (commented):
I submitted a review in prax which is closely linked to this, but I think we need to shift things here to have multiple votes per epoch and transaction, to reflect what's possible, and that we need the view for the LQT action to show all the information present in the opaque case.

packages/perspective/src/plan/view-action-plan.ts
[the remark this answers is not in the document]
TalDerei: refs https://github.com/prax-wallet/prax/pull/292#discussion_r1962283219

-------------------

https://github.com/user-attachments/assets/80884540-150b-4b72-bf09-b57fa5ab6857

packages/ui-deprecated/components/ui/tx/actions-views/liquidity-tournament-vote.tsx
```diff
@@ -31,7 +45,25 @@ export const LiquidityTournamentVoteComponent = ({
...
+            }
+            {value.liquidityTournamentVote.value.vote?.body?.incentivized && (
+              <ActionDetails.Row label='Voted Asset'>
+                {value.liquidityTournamentVote.value.vote.body.incentivized.denom}
+              </ActionDetails.Row>
+            )}
+          </ActionDetails>
+        }
+      />
+    );
```
cronokirby, line 66: We could try and consolidate these two branches to share the common stuff, but it's not worth messing with the deprecated component much anyways, so let's not do anything.

packages/storage/src/indexed-db/index.ts
```diff
@@ -410,6 +413,67 @@ export class IndexedDb implements IndexedDbInterface {
...
+    assetMetadata: Metadata,
+    voteValue: Value,
+    rewardValue?: Amount,
+    id?: string,
+  ): Promise<void> {
+    assertTransactionId(transactionId);
+
+    // This is a unique identifier to force unique primary keys. If the field isn't provided,
+    // a random one is generated and used to store an object.
+    const uniquePrimaryKey = id ?? Math.random().toString(36).substring(2, 10);
```
cronokirby, line 460:
```suggestion
    const uniquePrimaryKey = id ?? crypto.randomUUID();
```

cronokirby (commented):
Let's try and get CI green, but otherwise looks good!

packages/storage/src/indexed-db/index.ts
TalDerei: didn't go with that package because it generates 128-bit UUID (36 chars) which felt extraneous

packages/ui-deprecated/components/ui/tx/actions-views/liquidity-tournament-vote.tsx
TalDerei: yeah don't want to touch it more beyond this, because max will implemented the updated views in couple days anyway.

cronokirby (approved):

Issue #2014, opened 2025-02-01
data clone errors persist


TalDerei:
user's are still seeing this error when attempting to swap, which persists across switching rpcs. However, they are still able to fully sync.

<img width="761" alt="Image" src="https://github.com/user-attachments/assets/aad8ad49-bf7f-489b-a359-bfd36a8c83d8" />

PR #2043, opened 2025-02-12
incremental transport-chrome session improvements

turbocrime:
## Description of Changes

splitting the difference from #2018  towards #2034

contains all changes except reconnect, mostly focusing on reliability.

- typed senders make it harder to emit a malformed message
- incoming message checking is more thorough
- poisons transport in fewer cases
- matches errors up with relevant requests in more cases
- fewer data clone failures #2014

### DIFF
--- a/packages/transport-chrome/src/session-client.ts
+++ b/packages/transport-chrome/src/session-client.ts
@@ -1,68 +1,47 @@
-/**
- * CRSessionClient is a Chrome runtime session client: it handles channel
- * Transport client sessions in the Chrome runtime.
- *
- * Simple handlers unconditionally forward messages back and forth. Chrome
- * runtime disconnect is detected and surfaced as an error.
- *
- * Only a basic same-window origin check and structural typing are performed.
- * Content scripts are ultimately untrusted, so services are responsible for
- * confirming legitimacy.
- *
- * A `ReadableStream` or `AsyncIterable` cannot cross the runtime, so streaming
- * methods sink/source a dedicated `chrome.runtime.Port` at this boundary.
- */
-
+import { Code, ConnectError } from '@connectrpc/connect';
+import { errorToJson } from '@connectrpc/connect/protocol-connect';
 import {
   isTransportAbort,
   isTransportError,
+  isTransportEvent,
   isTransportMessage,
   isTransportStream,
-  TransportStream,
+  type TransportAbort,
+  type TransportError,
+  type TransportMessage,
+  type TransportStream,
 } from '@penumbra-zone/transport-dom/messages';
 import { ChannelLabel, nameConnection } from './channel-names.js';
-import { isTransportInitChannel, TransportInitChannel } from './message.js';
+import { isTransportInitChannel, type TransportInitChannel } from './message.js';
 import { PortStreamSink, PortStreamSource } from './stream.js';
 
-const localErrorJson = (err: unknown, relevantMessage?: unknown) =>
-  err instanceof Error
-    ? {
-        message: err.message,
-        details: [
-          {
-            type: err.name,
-            value: err.cause,
-          },
-          relevantMessage,
-        ],
-      }
-    : {
-        message: String(err),
-        details: [
-          {
-            type: String(
-              // eslint-disable-next-line no-nested-ternary -- readable ternary nesting
-              typeof err === 'function'
-                ? err.name
-                : typeof err === 'object'
-                  ? ((Object.getPrototypeOf(err) as unknown)?.constructor?.name ?? String(err))
-                  : typeof err,
-            ),
-            value: err,
-          },
-          relevantMessage,
-        ],
-      };
-
+/**
+ * Transparently adapts `Transport`s using `MessageChannel` DOM connections to
+ * use Chrome `chrome.runtime.Port` extension connections.
+ *
+ * Simple handlers unconditionally forward messages back and forth. Channel
+ * disconnects are detected, and transparently re-established if possible.
+ *
+ * Content scripts are ultimately untrusted, so the counterpart session manager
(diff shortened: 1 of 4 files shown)


turbocrime:
todo: specifically tag errors originating in the session client and session manager

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
+  type TransportError,
+  type TransportMessage,
+  type TransportStream,
 } from '@penumbra-zone/transport-dom/messages';
 import { ChannelLabel, nameConnection } from './channel-names.js';
-import { isTransportInitChannel, TransportInitChannel } from './message.js';
+import { isTransportInitChannel, type TransportInitChannel } from './message.js';
 import { PortStreamSink, PortStreamSource } from './stream.js';
 
-const localErrorJson = (err: unknown, relevantMessage?: unknown) =>
```
TalDerei, line 28: ACK – we discussed removing in favor of standardizing with connectError (using `connectRPC`) serialization. 

You mentioned that the error behavior remains the same with this change? Should we also tag errors coming out of this file (maybe dedicated error class?) with stack information emitted by connectRPC, or that may be unnecessary.

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
+  private readonly servicePort: chrome.runtime.Port;
 
-    const { port1, port2 } = new MessageChannel();
-    this.clientPort = port1;
-    this.inputPort = port2;
+  /** Private constructor. Launch a session with `CRSessionClient.init`. */
+  private constructor(
+    private readonly clientPort: MessagePort,
+    private readonly managerId: string,
+  ) {
```
TalDerei, line 46: ACK – internal private constructor signature changed, but init method signature has not.

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
-    CRSessionClient.singleton ??= new CRSessionClient(prefix);
-    return CRSessionClient.singleton.inputPort;
+  public static init(managerId: string): MessagePort {
+    const { port1, port2 } = new MessageChannel();
+    new CRSessionClient(port1, managerId);
+    return port2;
   }
 
+  /** Just some typed wrappers for consistency, in a namespace. */
+  private send = {
```
TalDerei, line 81: ACK – namespaced wrapper for sending transport events back and forth, making `postMessge` typesafe.

can we modify the naming convention used here for clarity?

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
+   * Used when the client announces channel closure, or when the extension
+   * channel disconnects.
+   *
+   * Announces closure from this side towards the document, and ensures closure
+   * of both ports. Listeners are automatically gargbage-collected. This session
+   * is complete.
+   */
   private disconnect = () => {
-    this.clientPort.removeEventListener('message', this.clientListener);
+    // announce closure, ok even if already closed
```
TalDerei, line 121: ACK – listeners (`clientListener`, `serviceListener`, `disconnect`) automatically gargbage collected. references https://issues.chromium.org/issues/40341793 and https://groups.google.com/a/chromium.org/g/chromium-extensions/c/nAvia8r1cvo for visibility

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
+   * If an unidentifiable message arrives, this handler will report the failure
+   * back to the specific request if possible, or, as a top-level transport
+   * error.
+   *
+   * Since a `MessagePort` has no event or callback indicating closure, the
+   * client may announce port closure by posting a `false` value.
+   *
+   * @param tev `TransportEvent` or `false` message payload
+   */
+  private clientListener = ({ data: tev }: MessageEvent<unknown>) => {
```
TalDerei, line 140: ACK – more robust failures and better serialization.

packages/transport-chrome/src/session-client.ts
```diff
@@ -1,152 +1,267 @@
...
       }
-      chrome.runtime.onConnect.removeListener(sinkListener);
-      void stream.pipeTo(new WritableStream(new PortStreamSink(p))).catch(() => null);
     };
+
     chrome.runtime.onConnect.addListener(sinkListener);
-    return { requestId, channel } satisfies TransportInitChannel;
+
+    /** avoid leak. @todo improve more */
+    setTimeout(() => chrome.runtime.onConnect.removeListener(sinkListener), 60_000);
```
TalDerei, line 263: question: what's this?

TalDerei (approved):
Looks good to me overall, left some comments 👍

It seems like this incorporates the diff in https://github.com/penumbra-zone/web/pull/2018, so we can probably just close that PR?

packages/transport-chrome/src/session-client.ts
turbocrime: will work on identifying any necessary error detection changes over weekend

packages/transport-chrome/src/session-client.ts
turbocrime: i un-namespaced these and named them more like the previous names

packages/transport-chrome/src/session-client.ts
turbocrime: this was updated since review. it's a fallback to clean up channel listeners that never activate.

this is only for experimental client-streaming support, in which the session client attaches a listener and instructs the service worker to activate the listener.

client streaming is defined out in production builds, and can't activate unless `globalThis.__DEV__` is truthy at build time.

PR #2049, opened 2025-02-12
fix: bump skip widget to fix penumbra noble transparent addresses

vacekj:
## Description of Changes

bump skip widget version and bump minifront version


erwanor (approved):
Thank you

PR #2058, opened 2025-02-14
disable client-streaming in transport-dom outside of dev mode

turbocrime:
## Description of Changes

Moves client streaming request support to a `globalThis.__DEV__` conditional, which will be codesplit by bundlers or skipped in execution unless defined.

### DIFF
--- a/packages/transport-dom/src/create.ts
+++ b/packages/transport-dom/src/create.ts
@@ -26,6 +26,11 @@ import {
 
 import ReadableStream from './ReadableStream.from.js';
 
+declare global {
+  // eslint-disable-next-line no-var -- global dev mode flag
+  var __DEV__: boolean | undefined;
+}
+
 const forceTransportOptions = {
   httpClient: null as never,
   baseUrl: 'https://in-memory',
@@ -276,6 +281,9 @@ export const createChannelTransport = ({
             case MethodKind.BiDiStreaming:
               // send as an actual stream
               {
+                if (!globalThis.__DEV__) {
+                  throw new ConnectError('MethodKind not supported', Code.Unimplemented);
+                }
                 const stream: ReadableStream<JsonValue> = ReadableStream.from(input).pipeThrough(
                   new TransformStream({
                     transform: (chunk: PartialMessage<I>, cont) =>
--- a/packages/transport-dom/vitest.config.ts
+++ b/packages/transport-dom/vitest.config.ts
@@ -2,7 +2,8 @@
 
 import { defineConfig } from 'vite';
 
-export default defineConfig({
+export default defineConfig(({ mode }) => ({
+  define: { __DEV__: mode !== 'production' },
   test: {
     include: ['src/*.test.ts'],
     browser: {
@@ -12,4 +13,4 @@ export default defineConfig({
       headless: true,
     },
   },
-});
+}));
--- a/.changeset/empty-ads-repeat.md
+++ b/.changeset/empty-ads-repeat.md
@@ -1,0 +1,8 @@
+---
+'@penumbra-zone/transport-chrome': patch
+'@penumbra-zone/transport-dom': patch
+---
+
+disable unsupported client-streaming requests. if you are experimenting with
+development of client-streaming requests, define the `globalThis.__DEV__` flag
+in your bundler to enable them.
--- a/packages/transport-chrome/src/session-client.ts
+++ b/packages/transport-chrome/src/session-client.ts
@@ -112,6 +112,9 @@ export class CRSessionClient {
         console.warn('Unknown item from client', ev.data);
       }
     } catch (e) {
+      if (globalThis.__DEV__) {
+        console.warn('Error in client listener', e);
+      }
       this.clientPort.postMessage({ error: localErrorJson(e, ev.data) });
     }
   };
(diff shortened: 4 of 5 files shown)


packages/transport-dom/src/create.ts
```diff
@@ -168,9 +175,9 @@ export const createChannelTransport = ({
...
         new Promise<TransportMessage>((resolve, reject) => {
           pending.set(requestId, (tev: TransportEvent) => {
-            if (isTransportMessage(tev, requestId)) {
+            if (isTransportMessage(tev) && tev.requestId === requestId) {
               resolve(tev);
-            } else if (isTransportError(tev, requestId)) {
+            } else if (isTransportError(tev)) {
               reject(errorFromJson(tev.error, tev.metadata, new ConnectError('Unary failed')));
             } else {
               reject(ConnectError.from(tev));
```
turbocrime, line 183: behavior of the response callback is slightly changed.

id parameter of the type guard was removed, and the types were made simpler. the id parameter was used to reject events that did not contain a specific `requestId`, so to achieve the same behavior it should be checked deliberately.

but, anything activating this callback has got it from the pending map by its id, so it's not as critical as it might seem. this change is more permissive.

1. the same: a successful response resolving the request is confirmed to match `requestId`
2. changed: a serialized error rejecting the request is not confirmed to match `requestId`

case 2 should never happen, because the message listener shouldn't pick up this callback unless the event it's handling contains the appropriate `requestId`.

if it somehow did happen, this is still an improvement because:

- the request will fail immediately with that error instead of timing out,
- the error content will be properly deserialized instead of passed to the generic fallback rejection

packages/transport-dom/src/create.ts
```diff
@@ -150,13 +155,15 @@ export const createChannelTransport = ({
...
       timeoutMs: number | undefined = defaultTimeoutMs,
-      header: HeadersInit | undefined,
+      headersInit: HeadersInit | undefined,
       input: PartialMessage<I>,
     ): Promise<UnaryResponse<I, O>> {
       transportFailure.signal.throwIfAborted();
       port ??= await connect();
 
       const requestId = crypto.randomUUID();
+      const header = headersInit instanceof Headers ? Array.from(headersInit) : headersInit;
```
turbocrime, line 165: this prevents `DataCloneError` or transfer failure of a `Headers` class instance, by converting it to an equivalent json-compatible format.

the message types have been updated to require the json-compatible format.

packages/transport-dom/src/create.ts
```diff
@@ -273,18 +281,20 @@ export const createChannelTransport = ({
...
                       cont.enqueue(Any.pack(new method.I(chunk)).toJson(jsonOptions)),
                   }),
                 );
                 port.postMessage({ requestId, stream, header } satisfies TransportStream, [stream]);
+                break;
               }
-              break;
+            // eslint-disable-next-line no-fallthrough -- deliberate fallthrough
             default:
               throw new ConnectError('MethodKind not supported', Code.Unimplemented);
```
turbocrime, line 299: unless `globalThis.__DEV__` is defined as truthy, streaming requests will fall through to the default case and throw.

it's present in the released package after simple tsc compilation, but bundlers will tree-shake this as a constant conditional without any configuration.

packages/transport-dom/vitest.config.ts
```diff
@@ -3,6 +3,7 @@
 import { defineConfig } from 'vite';
 
 export default defineConfig({
+  define: { __DEV__: true },
```
turbocrime, line 6: enable dev mode in tests, to pass the existing tests of client-streaming behavior.

turbocrime (commented):
commentary on minor changes to transport-dom

packages/transport-dom/src/messages.ts
turbocrime, line 1: types changes:
- specify only the jsonifiable subset of  `HeadersInit`, for `header`/`trailer`/`metadata` fields. prevents data clone error
- no longer constrain `requestId` by type parameter

previous type guard functionality was used to condition on requestId, so guard behavior is maintained, but the type no longer exposes this information. this makes types, especially `TransportError` which was incorrectly defined, easier to work with.

packages/transport-chrome/src/message.ts
```diff
@@ -1,7 +1,7 @@
 import { isTransportEvent, TransportEvent } from '@penumbra-zone/transport-dom/messages';
 
-export const isTransportInitChannel = (c: unknown): c is TransportInitChannel =>
-  isTransportEvent(c) && 'channel' in c && typeof c.channel === 'string'; // sub-channel stream
+export const isTransportInitChannel = (c: unknown, id?: string): c is TransportInitChannel =>
+  isTransportEvent(c, id) && 'channel' in c && typeof c.channel === 'string'; // sub-channel stream
```
turbocrime, line 4: technically unnecessary, but now this is more consistent with the transport-dom type guards.

TalDerei (approved):
I suggested removing, but I think hiding client streaming functionality behind this flag is a good alternative

turbocrime:
closing in favor of smaller changesets

packages/transport-dom/src/messages.ts
turbocrime: reverted

packages/transport-chrome/src/message.ts
turbocrime: reverted

turbocrime:
scope narrowed to just disable client-streaming

TalDerei:
> scope narrowed to just disable client-streaming

nice, LGTM

Issue #2023, opened 2025-02-05
remove temporary workaround fix for noble USDC transfers


TalDerei:
revert minifront changes in https://github.com/penumbra-zone/web/pull/1950 once Noble's mainnet upgrade is live.

erwanor:
One thing to note is that we shouldn't just revert the commit but totally remove compat/transparent handling and just use ordinary ephemeral addresses everywhere.

PR #2060, opened 2025-02-17
t-addr: revert temporary support in web

TalDerei:
## Description of Changes

removes compat / transparent handling in preparation for noble's upcoming mainnet upgrade. 

## Related Issue

references https://github.com/penumbra-zone/web/issues/2023 and pairs with https://github.com/prax-wallet/prax/pull/295


cronokirby (commented):
What if we just limited the blast radius to the ibc-in folder, changing the behavior to not use transparent addresses, and to the deprecated ui component. That way the functionality is still there, which is something we have to reckon with, it being, permanently, part of the protocol, but just not exercising it.

TalDerei:
> What if we just limited the blast radius to the ibc-in folder

closing in favor of https://github.com/penumbra-zone/web/pull/2101 and https://github.com/prax-wallet/prax/pull/304

Issue #2053, opened 2025-02-14
UI v2: ActionView component


VanishMax:
- [x] `SpendView` (P1)
  - [x] `OutputView` (P1)
  - [x] `SwapView` (P1)
  - [ ] `SwapClaimView` (P1)
  - [ ] `DelegatorVoteView` (P3)
  - [ ] `ValidatorDefinition` (P3)
  - [ ] `IbcRelay` (P1)
  - [ ] `ProposalSubmit` (P3)
  - [ ] `ProposalWithdraw` (P3)
  - [ ] `ValidatorVote` (P3)
  - [ ] `ProposalDepositClaim` (P3)
  - [ ] `PositionOpen` (P1)
  - [ ] `PositionClose` (P1)
  - [ ] `PositionWithdraw` (P1)
  - [ ] `Delegate` (P2)
  - [ ] `Undelegate` (P2)
  - [ ] `CommunityPoolSpend` (P4)
  - [ ] `CommunityPoolOutput` (P4)
  - [ ] `CommunityPoolDeposit` (P4)
  - [ ] `ActionDutchAuctionScheduleView` (P2)
  - [ ] `ActionDutchAuctionEnd` (P2)
  - [ ] `ActionDutchAuctionWithdrawView` (P2)
  - [ ] `UndelegateClaim` (P2)
  - [ ] `Ics20Withdraw` (P1)
  - [ ] `ActionLiquidityTournmentVote`

TalDerei:
@VanishMax added `actionLiquidityTournmentVote` action view

PR #2061, opened 2025-02-18
feat(ui): #2053: implement `SwapClaim` action view

VanishMax:
Relates to #2053 

Implements SwapClaim action view. The old version didn't display the opaque view, this one does.

Live: https://penumbra-ui-preview--pr2061-feat-2053-action-vi-khqajj28.web.app/?path=/docs/ui-library-actionview--docs

Before

Visible             |  Opaque
:-------------------------:|:-------------------------:
image  |  image

Now

Visible             |  Opaque
:-------------------------:|:-------------------------:
image  |  image

### DIFF
--- a/packages/ui/src/utils/bufs/action-view.ts
+++ b/packages/ui/src/utils/bufs/action-view.ts
@@ -143,7 +143,7 @@ export const SwapClaimAction = new ActionView({
           }),
           output2: new NoteView({
             address: ADDRESS_VIEW_DECODED,
-            value: PENUMBRA_VALUE_VIEW_ZERO,
+            value: PENUMBRA_VALUE_VIEW,
           }),
           swapTx: new TransactionId({
             inner: new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7]),
@@ -154,6 +154,35 @@ export const SwapClaimAction = new ActionView({
                 amount: AMOUNT_999,
                 assetId: PENUMBRA_METADATA.penumbraAssetId,
               }),
+            },
+          },
+        },
+      },
+    }),
+  },
+});
+
+export const SwapClaimActionOpaque = new ActionView({
+  actionView: {
+    case: 'swapClaim',
+    value: new SwapClaimView({
+      swapClaimView: {
+        case: 'opaque',
+        value: {
+          swapClaim: {
+            body: {
+              fee: new Fee({
+                amount: AMOUNT_999,
+                assetId: PENUMBRA_METADATA.penumbraAssetId,
+              }),
+              outputData: {
+                tradingPair: {
+                  asset1: USDC_METADATA.penumbraAssetId,
+                  asset2: PENUMBRA_METADATA.penumbraAssetId,
+                },
+                lambda1: AMOUNT_123_456_789,
+                lambda2: AMOUNT_999,
+              },
             },
           },
         },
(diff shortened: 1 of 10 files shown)


packages/ui/src/utils/bufs/action-view.ts
```diff
@@ -161,3 +161,32 @@ export const SwapClaimAction = new ActionView({
...
+                amount: AMOUNT_999,
+                assetId: PENUMBRA_METADATA.penumbraAssetId,
+              }),
+              outputData: {
+                tradingPair: {
+                  asset1: USDC_METADATA.penumbraAssetId,
+                  asset2: PENUMBRA_METADATA.penumbraAssetId,
+                },
+                lambda1: AMOUNT_123_456_789,
+                lambda2: AMOUNT_999,
```
TalDerei, line 184: comment: privacy leak here. referencing the fifth bullet point in https://github.com/penumbra-zone/web/issues/889, we should only display the swap claim fee for `SwapClaim` actions in opaque views (and visible views for consistency, because the swap claim already links to the associated swap in the user's visible view – see attached example)

-----------------

https://github.com/user-attachments/assets/1942faf3-0582-451f-a64a-9b371c252b29

VanishMax: @TalDerei thanks for the reference to another issue – i understand the context now. 

Question: how _not displaying_ the input/output amounts would leak the information having that this data is still _in the protobuf message_?

cc @hdevalence

VanishMax: I guess I don't fully understand the idea after re-reading the comment for a 10th time. When talking about cosmetic changes, do you mean we shouldn't display the input/output amounts?

TalDerei: In this context, cosmetic changes mean that we should display the same data as the current deprecated UI, with only styling, typography, etc. updated according to the figma design. Like the data modeling shouldn't change, for instance we shouldn't be adding new _opaque_ (public view) fields that weren't there before. 

to clarify, the opaque swap claim view should display only the swap claim fee and not the input/output amounts.

VanishMax: Yeah, but what's the reason for this? 

> Like the data modeling shouldn't change, for instance we shouldn't be adding new opaque (public view) fields that weren't there before.

I'm only changing the rendering without touching the fields. If data exists, why can't we display it?

TalDerei: well okay, the BSOD is part of the opaque swap claim view, however I initally thought displaying that information might be confusing as apposed to just the claimed amount since we also already link to the swap. It's _not_ actually a privacy leak because I had forgotten the BSOD field was part of the swap claim body. I think up to your discretion whether to include it

VanishMax:
@TalDerei Updated the PR, seems to be correct now

TalDerei:
> @TalDerei Updated the PR, seems to be correct now

I'd like to double check with @erwanoron this. Why is the

packages/ui/src/utils/bufs/action-view.ts
hdevalence: Yeah there's no privacy issue in the UI here — if the data is there, better to display it (as long as that makes sense in the context of the designs)

PR #2062, opened 2025-02-18
Prevent listener race condition

turbocrime:
Prevent race conditions discovered in implementation of session-client-reconnect.

Thoroughly manage port state at all locations.


TalDerei (approved):
wasn't able to repro the race condition on slower public wifi, and don't fully understand all the justifications here, but discussed out of band about the assurance testing you performed on this.

turbocrime:
closing in favor of #2066

PR #2067, opened 2025-02-20
Mock chrome messaging, implement tests for transport-chrome, improve tests for transport-dom

turbocrime:
PR including tests implemented for transport-chrome and transport-dom


TalDerei (approved):
discussed out of band, looks good to me after green CI!

can we add a changeset here for npm publishing?

PR #2069, opened 2025-02-21
feat(ui): #2053: ActionView component part 3

VanishMax:
Continuation of #2053

In this PR:
- `AssetIcon` component v2 design – added support for `badge` field representing asset network
- `AssetGroup` UIv2 component
- `PositionOpen`
- `PositionClose`
- `PositionWithdraw`
- `PositionRewardClaim`

PositionOpen now displays asset Metadata and shows correct amount exponents for reserves in comparison with old ActionViews. The rest of the views only display positionId due to lack of data in protobufs. Should we modify them to return full Position?


TalDerei (commented):
this looks like really great work! why does the storybook preview seem to be only rendering the reserves for open positions rather than the full position?

---------------

Screenshot 2025-02-21 at 8 26 33 PM

VanishMax:
@TalDerei as per new requirements, position actions should render the reserves, fees, and the position id. Forgot to show fees in Storybook. As for positionId, it's an open question about how to compute it in TypeScript without calling wasm functions

TalDerei (approved):
thanks for the clarification – per the new requirements, this looks good to me! for now, let's display the fees in the storybook and consider using the `ownedPositionIds` view service method to render the relevant positionId? We can iterate on it as needed.

Issue #2070, opened 2025-02-21
`@penumbra-zone/transport-dom` type `TransportEvent` permits a non-transferable field


turbocrime:
transport-dom's `TransportEvent` and `TransportError` types permit field values that are not transferable. failure to serialize nontransferable values results in `DataCloneError` when the event is posted to a message port.

```ts
// typescript lib.dom.d.ts
type HeadersInit = [string, string][] | Record<string, string> | Headers
```

Headers is a constructable.

https://github.com/penumbra-zone/web/blob/d0cc2ee9bb5cb0e77238238d2546751fc70286a1/packages/transport-dom/src/messages.ts#L5-L9

https://github.com/penumbra-zone/web/blob/d0cc2ee9bb5cb0e77238238d2546751fc70286a1/packages/transport-dom/src/messages.ts#L15-L20

turbocrime:
closed by #2071

PR #2071, opened 2025-02-21
correctly serialize headers and timeout

turbocrime:
## Description of Changes

Restrict the use of `HeadersInit` in a `TransportEvent` or `TransportError` to only permit jsonifiable members.

Serialize timeout into the request header as in typical connectrpc.

Technically, an exported type becomes narrower, but the now-omitted member was invalid, and no external API has changed its accepted input/output formats. 

## Related Issue

- [x] https://github.com/penumbra-zone/web/issues/2070
- [x] https://github.com/penumbra-zone/web/issues/2072

### DIFF
--- a/.changeset/curly-icons-sniff.md
+++ b/.changeset/curly-icons-sniff.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/transport-dom': patch
+---
+
+convey timeout in request headers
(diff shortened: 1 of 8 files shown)


TalDerei (approved):
this looks good, but seems to conflict with https://github.com/penumbra-zone/web/pull/2034#discussion_r1956663960 that was moved in #2058? approving optimistically, and should we narrow the scope of that PR to only disable client streaming?

turbocrime:
yes, breaking up PR #2058, it was kind of a random collection

Issue #2073, opened 2025-02-21
`@penumbra-zone/transport-dom` never times out a stalled stream


turbocrime:
at the request level, a request that does not receive a response by the configured timeout will be rejected, including any unary or streaming request.

for streams, initialization of a stream satisfies this deadline. but then, streams are not held to any time requirements.

so a stream established before its request-level deadline may stall indefinitely, even if it never actually delivers a value.

turbocrime:
closed by #2074

PR #2074, opened 2025-02-22
time out stalled streams, emit stream abort events

turbocrime:
## Description of Changes

Streaming responses now use the per-request or transport default timeout configuration to abort a stalled stream.

## Related Issue

- [x] https://github.com/penumbra-zone/web/issues/2073

### DIFF
--- a/packages/transport-dom/src/create.ts
+++ b/packages/transport-dom/src/create.ts
@@ -159,7 +159,9 @@ export const createChannelTransport = ({
       header: HeadersInit | undefined,
       input: PartialMessage<I>,
     ): Promise<UnaryResponse<I, O>> {
-      transportFailure.signal.throwIfAborted();
+      if (transportFailure.signal.aborted) {
+        throw transportFailure.signal.reason;
+      }
       port ??= await connect();
 
       const requestId = crypto.randomUUID();
@@ -229,7 +231,9 @@ export const createChannelTransport = ({
       header: HeadersInit | undefined,
       input: AsyncIterable<PartialMessage<I>>,
     ): Promise<StreamResponse<I, O>> {
-      transportFailure.signal.throwIfAborted();
+      if (transportFailure.signal.aborted) {
+        throw transportFailure.signal.reason;
+      }
       port ??= await connect();
 
       const requestId = crypto.randomUUID();
--- a/packages/transport-dom/src/create.ts
+++ b/packages/transport-dom/src/create.ts
@@ -154,7 +154,7 @@ export const createChannelTransport = ({
     async unary<I extends Message<I> = AnyMessage, O extends Message<O> = AnyMessage>(
       service: ServiceType,
       method: MethodInfo<I, O>,
-      signal: AbortSignal | undefined,
+      signal: AbortSignal | undefined = new AbortController().signal,
       timeoutMs: number | undefined = defaultTimeoutMs,
       header: HeadersInit | undefined,
       input: PartialMessage<I>,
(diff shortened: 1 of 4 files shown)


packages/transport-dom/src/create.ts
```diff
@@ -303,24 +303,36 @@ export const createChannelTransport = ({
...
+                chunkTimeout = setTimeout(chunkDeadlineExceeded, timeoutMs);
                 const o = new method.O();
                 Any.fromJson(chunk, jsonOptions).unpackTo(o);
                 cont.enqueue(o);
               },
             }),
-            { signal },
-          ),
-        ),
+            { signal: AbortSignal.any([signal, chunkDeadline]) },
```
turbocrime, line 333: this maintains the present behavior of aborting a streaming response with the original request's abort controller, but also adds the chunk deadline.

packages/transport-dom/src/create.ts
```diff
@@ -303,24 +303,36 @@ export const createChannelTransport = ({
...
       }
 
+      const chunkAc = new AbortController();
+      const chunkDeadline = chunkAc.signal;
+      const chunkDeadlineExceeded = () => {
+        console.debug('chunkDeadlineExceeded');
+        if (timeoutMs) {
+          chunkAc.abort(ConnectError.from('Stream stalled', Code.DeadlineExceeded));
+        }
+      };
```
turbocrime, line 313: once passed to `pipeThrough` an abort signal cannot be replaced. so i'm using a signal controlled by timeout instead of `AbortSignal.timeout`, so the timer can be reset at each chunk arrival.

alternatively, an abort signal/controller could be avoided, and `setTimeout` could be used with `cont.error`

packages/transport-dom/src/create.ts
```diff
@@ -149,7 +149,7 @@ export const createChannelTransport = ({
     async unary<I extends Message<I> = AnyMessage, O extends Message<O> = AnyMessage>(
       service: ServiceType,
       method: MethodInfo<I, O>,
-      signal: AbortSignal | undefined,
+      signal: AbortSignal | undefined = new AbortController().signal,
```
turbocrime, line 152: this just creates a signal that will never abort, if an abort signal is not provided by the caller.

packages/transport-dom/src/create.ts
TalDerei: to clarify, the `AbortSignal.any([signal, chunkDeadline])` will combine the main abort signal with the chunk-level abort signal, and if either is triggered, it cancels the stream and posts an abort message to the port?

packages/transport-dom/src/create.ts
```diff
@@ -160,10 +160,10 @@ export const createChannelTransport = ({
       const requestId = crypto.randomUUID();
 
       const requestFailure = new AbortController();
-      const deadline = timeoutMs ? AbortSignal.timeout(timeoutMs) : undefined;
+      const requestDeadline = timeoutMs ? AbortSignal.timeout(timeoutMs) : undefined;
```
TalDerei, line 163: this `timeoutMs` is the same as the default transport timeout in `createChannelTransport` right?

TalDerei (approved):
handling stalled streams is a good improvement 👍

packages/transport-dom/src/create.ts
turbocrime: by default, yes, but it may be zero (indefinite) or a custom value for any individual request

TalDerei:
@turbocrime general question, how do these stream timeouts impact the reliability of background syncing?

turbocrime:
> @turbocrime general question, how do these stream timeouts impact the reliability of background syncing?

replied in DM, but repeating here for clarity of discussion:

background syncing itself is not affected. the block processor has its own retry logic, and doesn't rely on this transport.
the exposure of background syncing state in the frontend becomes more fragile, but also the displayed state is more accurate and truthful.

the minifront PR enabling status stream retry is to address the minifront experience https://github.com/penumbra-zone/web/pull/2080

packages/transport-dom/src/create.ts
```diff
@@ -159,7 +159,9 @@ export const createChannelTransport = ({
       header: HeadersInit | undefined,
       input: PartialMessage<I>,
     ): Promise<UnaryResponse<I, O>> {
-      transportFailure.signal.throwIfAborted();
+      if (transportFailure.signal.aborted) {
+        throw transportFailure.signal.reason;
+      }
```
turbocrime, line 164: retried streams on a dead transport would create an ever-expanding error message, due to `throwIfAborted` creating a new error message for each failure. this change will allow the same error to be rethrown.

PR #2080, opened 2025-02-25
minifront: retry status stream

turbocrime:
## Description of Changes

Retry the status stream in minifront, when it fails due to https://github.com/penumbra-zone/web/pull/2074

zquery doesn't really have a stream-retry this feature, so state management in the slice and in the zquery stream callbacks will collect the stream status for interested components, and schedule re-fetch attempts.

### DIFF
--- a/apps/minifront/src/components/header/header.tsx
+++ b/apps/minifront/src/components/header/header.tsx
@@ -1,16 +1,24 @@
 import { CondensedBlockSyncStatus } from '@penumbra-zone/ui-deprecated/components/ui/block-sync-status';
 import { IncompatibleBrowserBanner } from '@penumbra-zone/ui-deprecated/components/ui/incompatible-browser-banner';
 import { TestnetBanner } from '@penumbra-zone/ui-deprecated/components/ui/testnet-banner';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useMemo } from 'react';
 import { Link } from 'react-router-dom';
 import { getChainId } from '../../fetchers/chain-id';
-import { useStatus } from '../../state/status';
+import { statusStreamStateSelector, useInitialStatus, useStatus } from '../../state/status';
 import { PagePath } from '../metadata/paths';
 import { MenuBar } from './menu/menu';
+import { useStore } from '../../state';
 
 export const Header = () => {
-  const { data, error } = useStatus();
   const [chainId, setChainId] = useState<string | undefined>();
+  const initialStatus = useInitialStatus();
+  const status = useStatus();
+  const { error: streamError } = useStore(statusStreamStateSelector);
+
+  const syncData = useMemo(
+    () => ({ ...initialStatus.data, ...status.data }),
+    [initialStatus.data, status.data],
+  );
 
   useEffect(() => {
     void getChainId().then(id => setChainId(id));
@@ -21,9 +29,9 @@ export const Header = () => {
       <IncompatibleBrowserBanner />
       <TestnetBanner chainId={chainId} />
       <CondensedBlockSyncStatus
-        fullSyncHeight={data?.fullSyncHeight}
-        latestKnownBlockHeight={data?.latestKnownBlockHeight}
-        error={error}
+        fullSyncHeight={syncData.fullSyncHeight}
+        latestKnownBlockHeight={syncData.latestKnownBlockHeight}
+        error={streamError}
       />
       <div className='flex w-full flex-col items-center justify-between px-6 md:h-[82px] md:flex-row md:gap-12 md:px-12'>
         <HeaderLogo />
--- a/apps/minifront/src/components/syncing-dialog/index.tsx
+++ b/apps/minifront/src/components/syncing-dialog/index.tsx
@@ -1,56 +1,95 @@
 import { Dialog } from '@penumbra-zone/ui-deprecated/Dialog';
-import { statusSelector, useStatus } from '../../state/status';
+import { statusStreamStateSelector, useInitialStatus, useStatus } from '../../state/status';
 import { SyncAnimation } from './sync-animation';
 import { Text } from '@penumbra-zone/ui-deprecated/Text';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useMemo } from 'react';
 import { useSyncProgress } from '@penumbra-zone/ui-deprecated/components/ui/block-sync-status';
+import { useStore } from '../../state';
 
 export const SyncingDialog = () => {
-  const status = useStatus({
-    select: statusSelector,
-  });
+  const initialStatus = useInitialStatus();
+  const status = useStatus();
+  const { error: streamError } = useStore(statusStreamStateSelector);
 
   const [isOpen, setIsOpen] = useState(false);
+  const [didClose, setDidClose] = useState(false);
+  const [dialogText, setDialogText] = useState<string | undefined>();
+
+  const syncData = useMemo(
+    () => ({ ...initialStatus.data, ...status.data }),
+    [initialStatus.data, status.data],
+  );
+
+  const isSynced = useMemo(
+    () =>
+      syncData.fullSyncHeight &&
+      syncData.latestKnownBlockHeight &&
+      syncData.fullSyncHeight >= syncData.latestKnownBlockHeight,
+    [syncData],
+  );
 
   useEffect(() => {
-    if (status?.isCatchingUp) {
-      setIsOpen(true);
-    } else {
-      setIsOpen(false);
+    if (streamError) {
+      setDialogText('Retrying...');
+    } else if (!initialStatus.data) {
+      setDialogText('Getting local block height...');
+    } else if (!status.data) {
+      setDialogText('Getting remote block height...');
+    } else if (!isSynced) {
+      setDialogText('Syncing...');
     }
-  }, [status?.isCatchingUp]);
+
+    const shouldShow = !isSynced || !!streamError;
+    setIsOpen(shouldShow && !didClose);
+  }, [didClose, initialStatus.data, status.data, isSynced, streamError]);
 
   return (
-    <Dialog isOpen={isOpen} onClose={() => setIsOpen(false)}>
+    <Dialog isOpen={isOpen} onClose={() => setDidClose(true)}>
       <Dialog.Content title='Your client is syncing...' zIndex={9999}>
         <SyncAnimation />
 
         <div className='text-center'>
-          <Text body as='p'>
-            Decrypting blocks to update your local state
+          <Text body>
+            {dialogText}
+            {streamError ? (
+              <Text technical color={theme => theme.caution.main} as='div'>
+                {streamError instanceof Error ? streamError.message : String(streamError)}
+              </Text>
+            ) : undefined}
+            {syncData.fullSyncHeight && syncData.latestKnownBlockHeight ? (
+              <>
+                <BlockProgress
+                  fullSyncHeight={syncData.fullSyncHeight}
+                  latestKnownBlockHeight={syncData.latestKnownBlockHeight}
+                />
+                <RemainingTime
+                  fullSyncHeight={syncData.fullSyncHeight}
+                  latestKnownBlockHeight={syncData.latestKnownBlockHeight}
+                />
+              </>
+            ) : undefined}
           </Text>
           <Text small as='p'>
             You can click away, but your data <i>may</i> not be current
           </Text>
-          <div className='mt-6'>
-            {!!status?.isCatchingUp && (
-              <Text technical>
-                {!!status.percentSynced && `${status.percentSynced} Synced – `} Block{' '}
-                {status.fullSyncHeight.toString()}{' '}
-                {!!status.latestKnownBlockHeight &&
-                  `of ${status.latestKnownBlockHeight.toString()}`}
-              </Text>
-            )}
-          </div>
-          {!!status?.isCatchingUp && status.latestKnownBlockHeight ? (
-            <RemainingTime
-              fullSyncHeight={status.fullSyncHeight}
-              latestKnownBlockHeight={status.latestKnownBlockHeight}
-            />
-          ) : null}
         </div>
       </Dialog.Content>
     </Dialog>
+  );
+};
+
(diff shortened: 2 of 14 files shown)


apps/minifront/src/components/syncing-dialog/index.tsx
turbocrime, line 1: dialog is significantly modified

apps/minifront/src/state/status.ts
```diff
@@ -1,21 +1,66 @@
...
+      };
+    },
+    onError: (prevData, error) => {
+      useStore.getState().status.streamState.setStreamError(error);
+      return prevData;
+    },
+    onEnd: prevData => {
+      useStore.getState().status.streamState.scheduleRefetch();
+      return prevData;
+    },
```
TalDerei, line 63: ACK – reconnection logic has `onEnd` callback which triggers `scheduleRefetch()`

-----------

https://github.com/user-attachments/assets/a10c706e-9807-4dcc-8dd4-c6851daf69d0

apps/minifront/src/state/status.ts
```diff
@@ -27,69 +72,69 @@ export const { status, useStatus } = createZQuery({
...
+      },
+
+      scheduleRefetch: () => {
+        clearTimeout(get().status.streamState.timer);
+
+        const timer = setTimeout(() => {
+          const { streamState, status } = get().status;
+          if (!streamState.running) {
+            status.revalidate();
+          }
```
TalDerei, line 107: ACK – calls `status.revalidate()` to restart the stream

apps/minifront/src/state/status.ts
```diff
@@ -27,69 +72,69 @@ export const { status, useStatus } = createZQuery({
...
+        clearTimeout(get().status.streamState.timer);
+
+        set(state => {
+          state.status.streamState.error = undefined;
+          state.status.streamState.running = true;
+          state.status.streamState.timer = undefined;
+        });
+      },
+
+      scheduleRefetch: () => {
```
TalDerei, line 100: comment: potential race condition here? `scheduleRefetch` checks if `streamState.running` is false before attempting reconnection, however what if the stream status changes between when the timeout is set and when it executes?

apps/minifront/src/state/status.ts
```diff
@@ -27,69 +72,69 @@ export const { status, useStatus } = createZQuery({
...
-    };
-  }
+export const syncPercentSelector = (
+  zQueryState: AbridgedZQueryState<PlainMessage<StatusStreamResponse>>,
+) => {
+  const { fullSyncHeight, latestKnownBlockHeight } = { ...zQueryState.data };
+
+  const synced =
+    Number(latestKnownBlockHeight ?? 0) &&
+    Math.min(1, Number(fullSyncHeight) / Number(latestKnownBlockHeight));
```
TalDerei, line 131: comment: sync percentage calculations may run into a division by zero error if `latestKnownBlockHeight` is zero?

apps/minifront/src/state/status.ts
```diff
@@ -1,21 +1,66 @@
...
+      prevData: PlainMessage<StatusStreamResponse> | undefined,
+      item: PlainMessage<StatusStreamResponse>,
+    ): PlainMessage<StatusStreamResponse> => {
+      useStore.getState().status.streamState.setStreamRunning();
+      return {
+        ...prevData,
+        ...item,
+      };
+    },
+    onError: (prevData, error) => {
```
TalDerei, line 56: comment: what's the difference here between `onError` versus `onEnd` in terms of reconnections?

apps/minifront/src/state/status.ts
```diff
@@ -1,21 +1,66 @@
...
+}
+
+export interface StatusSlice {
+  initialStatus: ZQueryState<PlainMessage<StatusResponse>>;
+  status: ZQueryState<PlainMessage<StatusStreamResponse>>;
+  streamState: StatusStreamState;
 }
 
-export const { status, useStatus } = createZQuery({
+export const { initialStatus, useInitialStatus, useRevalidateInitialStatus } = createZQuery({
```
TalDerei, line 29: comment: `useRevalidateInitialStatus` not used

apps/minifront/src/components/syncing-dialog/index.tsx
TalDerei: comment: the dialogue feels too cramped – can we revert to the previous dialogue (with the sync percentage) while keeping the message "Streaming new blocks to update private state"?

-------------

_before_

Screenshot 2025-02-27 at 12 33 09 PM

-----------------

_after_

Screenshot 2025-02-27 at 12 33 15 PM

TalDerei: thoughts? cc @smmhrdmn

turbocrime: sure, no aesthetic preference here, i was primarily seeking to have the displayed status reflect the new state information

apps/minifront/src/components/header/header.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { PagePath } from '../metadata/paths';
 import { MenuBar } from './menu/menu';
+import { useStore } from '../../state';
 
 export const Header = () => {
-  const { data, error } = useStatus();
   const [chainId, setChainId] = useState<string | undefined>();
+  const initialStatus = useInitialStatus();
+  const status = useStatus();
+  const { error: streamError } = useStore(statusStreamStateSelector);
```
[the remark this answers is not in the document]
turbocrime: it's important to display some kind of failure to the user, when the request has failed.

if you want to conceal the failure message, can you suggest a replacement?

should different failures use different replacements? this quickly gets complex.

apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -1,66 +1,80 @@
...
-    } else {
-      setIsOpen(false);
+    if (streamError) {
+      setDialogText('Retrying...');
+    } else if (!initialStatus.data) {
+      setDialogText('Querying local block height...');
+    } else if (!status.data) {
+      setDialogText('Fetching remote block height...');
+    } else if (!isSynced) {
+      setDialogText('Streaming new blocks to update private state...');
```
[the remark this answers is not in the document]
turbocrime: these are distinct states.

the cases you've tested involve a quick transition, and it is often a quick transition, but the lack of distinction between these states is at the root of the issue in #2081.

specifically, the 'Querying local block height' state is the 'initial status' unary request that will pend and possibly time out if the extension is connected, but taking a long time to respond to RPC queries.

i don't think making these states ambiguous is a good idea. there is no downside to making these states distinct. the upside is that users are more informed of information that is highly useful for issue reports and investigation.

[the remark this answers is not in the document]
turbocrime: i'm disinclined to make this change. please suggest a different dialog message if you think it could be better

apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -1,66 +1,80 @@
...
         <SyncAnimation />
 
         <div className='text-center'>
-          <Text body as='p'>
-            Decrypting blocks to update your local state
+          <Text body>
+            {dialogText}
+            {!!streamError && (
+              <Text technical color={theme => theme.caution.main} as='div'>
+                {streamError instanceof Error ? streamError.message : String(streamError)}
```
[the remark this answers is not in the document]
turbocrime: clearing the cache through this menu will restart the extension service worker. this can't be fixed until connection lifecycle work lands.

apps/minifront/src/state/status.ts
```diff
@@ -1,21 +1,66 @@
...
+  StatusResponse,
+  StatusStreamResponse,
+} from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+import { PlainMessage, toPlainMessage } from '@bufbuild/protobuf';
 
-export interface Status {
-  fullSyncHeight?: bigint;
-  latestKnownBlockHeight?: bigint;
+// Time in milliseconds to wait before attempting to reconnect the status stream
+const RECONNECT_TIMEOUT = 5_000;
```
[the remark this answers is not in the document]
turbocrime: attempting an exponential backoff implementation here would become significantly more complex, due to the fact that the query is not a single awaited promise over which we have direct control, but a zquery stream.

i agree backoff would be better, but this PR is essentially a spike to land a single case of stream re-request without getting deep into zquery. any more than this becomes significant zquery work.

i would like to consider new zquery features to be out of scope for this PR.

these are low-cost queries, so i don't think it's a big deal.

apps/minifront/src/state/status.ts
turbocrime: if the stream is running when the timeout executes, a reconnect is not necessary.

apps/minifront/src/state/status.ts
turbocrime: no - if `latestKnownBlockHeight` is zero, the conditional is short-circuited and `synced` is simply zero.

apps/minifront/src/state/status.ts
turbocrime: according to zquery documentation, `onEnd` executes on every stream end. `onError` executes before `onEnd`, if there is an error from the fetch.

turbocrime: there is another method, `onAbort`, which i investigated but it does not seem to be involved in the cases we care about here.

apps/minifront/src/state/status.ts
turbocrime: export of the `useRevalidate` generated method is consistent with other zquery use

apps/minifront/src/components/syncing-dialog/index.tsx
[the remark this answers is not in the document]
turbocrime: eventually this should be addressed with a more comprehensive attempt at addressing context loss.

i would like to consider this case out of scope for this PR, because no part of minifront handles this case well, and it's a significant effort to address, with many considerations.

apps/minifront/src/components/header/header.tsx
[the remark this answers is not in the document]
TalDerei: maybe a more general "connection issue detected" message, rather than anything specific to streams?

apps/minifront/src/components/syncing-dialog/index.tsx
smmhrdmn: I would agree with Tal that the top layout feels less cramped, although it still doesn't provide a great solution for displaying status info. Imo this dialog could use a proper redesign at some point. One small addition is we should probably include comma separators in the block numbers to make them more legible – the percentage is also helpful for this

turbocrime: and the required reload case,

Screenshot 2025-03-03 at 11 23 53

TalDerei (approved):

apps/minifront/src/components/syncing-dialog/index.tsx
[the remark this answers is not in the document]
TalDerei: discussed offline, we should open a separate issue for this case.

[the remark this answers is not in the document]
TalDerei: the scope there should also include locking the wallet

TalDerei (commented):
fell into an infinite loop of "querying local block height" after locking the wallet mid-sync, then reloading the page following the "connection closed" error. The connection then times out, and reloading the page doesn’t resolve the issue, leaving syncing stuck in a state where it can't self-correct or make progress. Using current prax main.

----------------

Screenshot 2025-03-03 at 4 11 24 PM

Screenshot 2025-03-03 at 4 14 51 PM

apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -73,11 +73,11 @@ export const SyncingDialog = () => {
   return (
     <Dialog isOpen={(!isSynced || !!streamError) && !didClose} onClose={() => setDidClose(true)}>
       <Dialog.Content
-        title={dialogErrorMessage ? 'Syncing interrupted' : 'Syncing...'}
+        title={streamError != null ? 'Syncing interrupted' : 'Syncing...'}
         zIndex={9999}
       >
         <div className='text-center'>
-          {!!streamError && (
+          {streamError != null && globalThis.__DEV__ && (
```
TalDerei: separately, possible to detect data clone errors by matching the error string and modifying the log message to a more user-friendly error that uniquely identifies this type of issue?

turbocrime:
> fell into an infinite loop of "querying local block height" after locking the wallet mid-sync, then reloading the page following the "connection closed" error. The connection then times out, and reloading the page doesn’t resolve the issue, leaving syncing stuck in a state where it can't self-correct or make progress. Using current prax main and minifront on localhost.

i'm not able to reproduce this.

locking the wallet will destroy the content-script context, as discussed in other threads.

reloading the page should restore connection. i wasn't able to observe a 'stuck' connection after reload.

how long did the 'stuck' connection persist after locking the extension and reloading?

clearing cache will also destroy content-script context, and genesis blocking may time out local requests or connection attempts.

how long did the 'stuck' connection persist after clearing cache and reloading?

if you can reproduce this, we should pair on investigation.

apps/minifront/src/components/syncing-dialog/index.tsx
turbocrime: specifically, data clone errors observed in the case of content script context loss are a failure of error reporting prax-side, which already has a fix landed in transport-dom.

future data clone errors would be indistinguishable, dapp-side, from these data clone errors.

adding special-case behavior for a data clone error here will be of short-lived utility, as clients update and become incapable of emitting this example. it would also obscure other possible unique root causes of data clone errors, which would be of high importance to discover.

in any case, data clone errors caused by content script context loss are transient, and are replaced within 5 seconds by a 'Connection closed' error, which is already special-cased.

TalDerei:
> if you can reproduce this, we should pair on investigation.

afaict, prax wasn't running locally which caused seems to have caused this.

TalDerei (approved):
looks good to me!

Issue #2064, opened 2025-02-19
'genesis sync' nomenclature


TalDerei:
the pending state of the minifront and prax currently displays 'genesis sync' when it hasn't received a status stream chunk yet, rather than only during actual genesis syncing. we should change the [naming convention](https://github.com/penumbra-zone/web/blob/main/packages/ui-deprecated/components/ui/block-sync-status/block-sync-status.tsx#L74) used here for clarity

----------

https://github.com/user-attachments/assets/2457184e-6f5c-445d-bf6f-85bb2495b301

TalDerei:
closed by #2081

PR #2081, opened 2025-02-25
minifront: detect genesis sync state more accurately

turbocrime:
## Description of Changes

Presently, the header progress bar will display a 'Genesis sync' status when it's waiting for status data. This change will cause the header progress bar to only display a 'Genesis sync' status when the synced height is zero/undefined, but a remote block height is known.

## Related Issue

https://github.com/penumbra-zone/web/issues/2064

### DIFF
--- a/packages/ui-deprecated/components/ui/block-sync-status/block-sync-status.tsx
+++ b/packages/ui-deprecated/components/ui/block-sync-status/block-sync-status.tsx
@@ -19,7 +19,7 @@ export const CondensedBlockSyncStatus = ({
     return <BlockSyncErrorState />;
   }
   if (!latestKnownBlockHeight || !fullSyncHeight) {
-    return <AwaitingSyncState genesisSyncing={!fullSyncHeight} />;
+    return <AwaitingSyncState genesisSyncing={!!latestKnownBlockHeight && !fullSyncHeight} />;
   }
 
   const isSyncing = latestKnownBlockHeight - fullSyncHeight > 10;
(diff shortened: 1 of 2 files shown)


packages/ui-deprecated/components/ui/block-sync-status/block-sync-status.tsx
```diff
@@ -19,7 +19,7 @@ export const CondensedBlockSyncStatus = ({
     return <BlockSyncErrorState />;
   }
   if (!latestKnownBlockHeight || !fullSyncHeight) {
-    return <AwaitingSyncState genesisSyncing={!fullSyncHeight} />;
+    return <AwaitingSyncState genesisSyncing={!!latestKnownBlockHeight && !fullSyncHeight} />;
```
TalDerei, line 22: why wouldn't this be just `!latestKnownBlockHeight`?

turbocrime: if `latestKnownBlockHeight` is undefined or zero, it probably means the query response is not populated.

technically it could mean that the global chain is at height zero, but i expect most users are connecting to chains with height.

so: if one of these values is falsy, we should only consider it genesis sync if the query is actually populated, and the full sync height is falsy.

without this change, users see the 'genesis sync' status while the query is pending.

TalDerei: still have a mismatch between frontend and extension when loading the genesis state

Screenshot 2025-02-26 at 3 31 02 PM

TalDerei: I think it's fine actually because I can't think of another way to do this, displaying the sync state is actually a good generalization.

TalDerei (approved):

PR #2083, opened 2025-02-26
feat(ui): #2051: `TransactionSummary`

### LINKED ISSUE #2051: Transaction Summary Component
There needs to be a skeleton component implementation that uses the V2 designs and fills in the list of action views.

At this stage, it's not important to fill in all of the action views - they can have "unsupported action" or similar text -- what's important is having a code scaffold that will allow filling in the action views one by one.

Once the V2 component is in place, it can be used in Veil, so that Veil can move off of the forked V1 transaction view designs. This will also unblock work on the PK labs block explorer.

VanishMax:
Closes #2051 and https://github.com/penumbra-zone/web/issues/1394

An independent PR but only https://github.com/penumbra-zone/web/pull/2000 will allow applying this component to the DEX or other products.

Preview of the component: https://penumbra-ui-preview--pr2083-feat-2051-tx-summar-rkethfkg.web.app/?path=/docs/ui-library-transactionsummary--docs

Needs a follow up work to support the whole range of transaction types but I'd like to switch to other features. At least this component works well on Send/Receive/Swap transaction types and partially on the rest of them.

image

image


packages/perspective/src/transaction/classify.ts
```diff
@@ -1,92 +1,196 @@
...
+  /** A type of the main action that defines a transaction */
+  type: TransactionClassification;
+  /** A main action that defines the transaction */
+  action?: ActionView;
+}
+
+/**
+ * Takes a transaction view, finds the most relevant action that defines its type, and returns it.
+ */
+export const classifyTransaction = (txv?: TransactionView): ClassificationReturn => {
```
VanishMax, line 18: a powerful function, it was returning only a classified type of a transaction. I added the return of the most relevant action type given transaction

packages/ui/src/AddressView/index.tsx
```diff
@@ -5,14 +5,27 @@ import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
...
+  return 'strong';
+};
+
 export interface AddressViewProps {
   addressView: AddressView | undefined;
   copyable?: boolean;
   hideIcon?: boolean;
   truncate?: boolean;
+  /** If true, takes the `altBech32m` field from the address and renders it as-is */
+  external?: boolean;
```
VanishMax, line 28: added support for AddressView component to render addresses from other blockchains like `osmo1awefasdf...`

packages/ui/src/TransactionSummary/use-classification.ts
```diff
@@ -0,0 +1,169 @@
...
+  liquidityTournamentVote: 'Liquidity Tournament Vote',
+};
+
+/**
+ * A hook that prepares data from TransactionInfo to be rendered in TransactionSummary.
+ */
+export const useClassification = (
+  info: TransactionInfo,
+  getMetadataByAssetId?: GetMetadataByAssetId,
+) => {
```
VanishMax, line 66: the main data modeling logic is here - the hook prepares a transaction for rendering

packages/perspective/src/action-view/relevant-assets.ts
```diff
@@ -0,0 +1,111 @@
...
+    return returnAssets([pair?.asset1, pair?.asset2]);
+  }
+
+  if (view.case === 'delegatorVote') {
+    const note =
+      view.value.delegatorVote.case === 'visible' ? view.value.delegatorVote.value.note : undefined;
+    return returnAssets([getMetadata.optional(note?.value)]);
+  }
+
+  // TODO: check this for validity
```
TalDerei, line 75: comment: this looks right to me – we've been using `getValidatorIdentityKeyFromValueView`, but that doesn't seem to work here.

packages/perspective/src/action-view/relevant-assets.ts
```diff
@@ -0,0 +1,111 @@
...
+ * - actionDutchAuctionSchedule
+ * - actionDutchAuctionWithdraw
+ * - actionLiquidityTournamentVote
+ *
+ * Some action views can compute asset information from its data:
+ * - delegate
+ * - undelegate
+ * - position close/withdraw (not implemented but can compute lpNft from positionId)
+ */
+export const findRelevantAssets = (action?: ActionView): RelevantAsset[] => {
```
TalDerei, line 35: ACK – returns assets from `AssetId` and `Metadata`

TalDerei (approved):
this is really great work! left a few minor comment, and now that transaction summary support has landed, we can proceed with more testing with actual data. 

additionally, we should ensure the asset type is not truncated in the compact variant in the storybook.

PR #2092, opened 2025-03-04
fix(ui): #2051: `TransactionSummary` improvements

VanishMax:
Improvements to `TransactionSummary`:
1. Takes summary effects and sums them together while also categorizing. If there were multiple lpNft outputs, it now simply shows '+3 LPNFT', for example. The same now works for auctions.
2. Truncates addresses – before it was taking much space
3. Corrects the negation display
4. Adds a Storybook example of a PositionOpen transaction rendering correctly 
5. Adds rendering to LPNFT and Auction metadata in `AssetIcon`
6. Adds stub icons for positions close/withdraw state

Storybook: https://penumbra-ui-preview--pr2092-fix-2051-summary-fi-4icx8mnc.web.app/?path=/docs/ui-library-transactionsummary--docs

image

This is how it looks in practice for my test mainnet account:

image


packages/ui/src/TransactionSummary/sum-balances.ts
cronokirby, line 1: Maybe instead of doing this work here, we should rely on the view server doing this kind of filtering instead. I think we should just have this portion of the code be responsible for rendering what was passed to it as it is.

TalDerei: +1 for removing the redundant `sumBalances` method — we should directly consume the already aggregated contents of `TransactionSummary_Effects` instead.

VanishMax: removed the summing logic, now filtering out the lpNFTs and auctionNFTs

TalDerei (approved):
logic looks correct to me! to reduce scope, we can address the supplemental processing for IBC assets separately.

PR #2095, opened 2025-03-05
minifront: reduce status stream timeout

TalDerei:
## Description of Changes

reverts https://github.com/penumbra-zone/web/pull/2080/commits/7fa28d100c8c6be9c92c24a643e33be0fa0d9955

I think the timeout may be too long — I can't consistently replicate the behavior I was seeing in https://github.com/penumbra-zone/web/pull/2080#discussion_r1974112717 which the user should be actively seeing while the status stream attempts to self-correct.

## Related Issue

### DIFF
--- a/.changeset/friendly-mugs-push.md
+++ b/.changeset/friendly-mugs-push.md
@@ -1,0 +1,5 @@
+---
+'minifront': minor
+---
+
+reduce status stream timeout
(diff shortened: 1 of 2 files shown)


.changeset/friendly-mugs-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'minifront': minor
```
turbocrime, line 2: nit

```suggestion
'minifront': patch
```

turbocrime (approved):

PR #2056, opened 2025-02-14
lint: shellcheck on minifront bundled asset script

conorsch:
## Description of Changes
I ran `shellcheck` on the script used to export a bundled minifront, suitable for inclusion in `pd`. These are small changes intended to make the script more resilient, e.g. failing fast and handling directories with spaces in the name.

Prepared while working on
https://github.com/penumbra-zone/penumbra/issues/5089.

## Related Issue

Refs https://github.com/penumbra-zone/penumbra/issues/5089.

## Checklist Before Requesting Review

- [x] I have ensured that any relevant minifront changes do not cause the existing extension to break.

No app-code changes, only a script to export artifacts is changed. Still, I did consume the results of this modified script, as part of testing for https://github.com/penumbra-zone/penumbra/issues/5089.

PR #2098, opened 2025-03-05
build: make zip file exports deterministic

conorsch:
## Description of Changes

As long as the zip files are exported from the same git commit, the resulting archives should be byte-for-byte identical. Having fully reproducible builds will require comprehensive dependency pinning, e.g. via nix flake or similar, to ensure that the wasm artifacts are built identically, but for now, simply clobbering the timestamps of the exported asset files provides decent reproducibility.

## Related Issue

No request in an issue, simply making opportunistic improvements, as with #2056. Was motivated to try this while updating the bundled assets in https://github.com/penumbra-zone/penumbra/pull/5119. 

^ N/A, as this script only makes changes to build scripts, not to app code.

### DIFF
--- a/scripts/build-for-deployment.sh
+++ b/scripts/build-for-deployment.sh
@@ -29,12 +29,16 @@ while getopts "o:r:" opt; do
 done
 
 cd "$REPO_ROOT" || exit 2
+pnpm install
 pnpm build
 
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
reproducibility is good 👍

PR #2105, opened 2025-03-08
session client timeout

turbocrime:
use a timeout instead of a recorded time

### DIFF
--- a/packages/transport-chrome/src/session-client.ts
+++ b/packages/transport-chrome/src/session-client.ts
@@ -42,10 +42,10 @@ export class CRSessionClient {
   private static signal: AbortSignal;
 
   /**
-   * Document-level recorded time of earliest pending connection attempt.
+   * Document-level timeout since earliest pending connection attempt.
    * Cleared when any service message arrives.
    */
-  private static connectingSince?: number;
+  private static timeout?: ReturnType<typeof setTimeout>;
 
   static {
     const staticAc = new AbortController();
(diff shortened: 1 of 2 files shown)


packages/transport-chrome/src/session-client.ts
```diff
@@ -134,7 +134,10 @@ export class CRSessionClient {
   }
 
   private connect = () => {
-    CRSessionClient.connectingSince ??= Date.now();
+    CRSessionClient.timeout ??= setTimeout(
```
TalDerei, line 137: general question; timeout triggers and disconnects all `CRSessionClients` in the document, but what's the affect on the session-side (ie. do sessions in background abort)? triggers onDisconnect() in `CRSession` or does the corresponding session remain active?

packages/transport-chrome/src/session-client.ts
```diff
@@ -242,8 +241,9 @@ export class CRSessionClient {
    * error.
    */
   private serviceListener = (tev: unknown) => {
-    // response clears the pending connection state
-    CRSessionClient.connectingSince = undefined;
+    // any response clears the document-level timeout progress
```
TalDerei, line 244: ACK – at least one client is active and making requests, the session is alive, so timeout shouldn't apply

TalDerei (approved):
thanks for the clarification here

packages/transport-chrome/src/session-client.ts
turbocrime: If the session is timing out, it's either failing validation and the session is cleaned up already, or something strange is happening in the background and whatever handling that should be happening is already broken.

packages/transport-chrome/src/session-client.ts
turbocrime: clarification:

the timeout is created by a request that must also start a connection.

if the service is responding, the timeout is satisfied and may be cleared.

PR #2107, opened 2025-03-09
Session client activity

turbocrime:


packages/transport-chrome/src/session-client.ts
```diff
@@ -104,26 +94,23 @@ export class CRSessionClient {
...
+        requestId: undefined,
+        error: errorToJson(ConnectError.from(this.signal.reason), undefined),
+      });
+      this.close();
     });
 
+    CRSessionClient.signal.addEventListener('abort', () => this.close(), { signal: this.signal });
+
     // connect to service
     this.servicePort = this.connect();
```
TalDerei, line 115: comment: can we add a comment to clarify that this establishes the initial session connection?

packages/transport-chrome/src/session-client.ts
```diff
@@ -163,20 +136,34 @@ export class CRSessionClient {
...
+      }
       if (isContextLossError(e)) {
         throw ConnectError.from(e, Code.Unavailable);
       }
       throw e;
     }
   };
 
+  /** Inform the page the connection is closed. */
+  private close = () => {
```
TalDerei, line 161: ACK – explicit method for session termination rather than triggering `abort()` based on timeout.

packages/transport-chrome/src/session-client.ts
```diff
@@ -163,20 +136,34 @@ export class CRSessionClient {
 
       port.onMessage.addListener(this.serviceListener);
       port.onDisconnect.addListener(() => {
-        // clean up for reconnect
-        this.servicePort = undefined;
+        if (this.active) {
```
TalDerei, line 139: ACK – differentiating between active vs. inactive sessions. 

wondering how to contextualize zombie sessions?

TalDerei (approved):
much simpler to reason about, LGTM

packages/transport-chrome/src/session-client.ts
turbocrime: the point of this is just to provide some kind of terminal state, so the session-client doesn't just keep trying forever, and can identify a condition to finally report a closed connection.

there are a few possible causes

1. session validation rejected
2. extension worker accepted connection, but failed to ever respond
3. client never made a request that returned a response (all requests were either cancelled or exceeded an explicit deadline)

PR #2108, opened 2025-03-10
assurance testing: dex-explorer migration

TalDerei:
## Description of Changes

migrates the [dex-explorer](https://github.com/penumbra-zone/dex-explorer) into the web monorepo using [git filter-repo](https://github.com/newren/git-filter-repo) tool and `--allow-unrelated-histories`. This preserves and interleaves commit histories chronologically. Some additional formatting for dex-explorer PR #'s in the [commit history](https://github.com/penumbra-zone/web/commits/migration-dex-retry/) is required.

_extra_ assurance should be done to ensure the migration was performed correctly. cc @conorsch 

## Related Issue

https://github.com/penumbra-zone/web/issues/2123


cronokirby:
Q: Do we want to also complete migration of some necessary support elements before merging this. For example, the CI will need to be moved, because Github doesn't support nested workflows (alas, I looked this up). We'll also want to include the package in the pnpm monorepo, and update dependencies to be workspace ones, and then also integrate this into to repo level build commands.

In terms of phasing:

1. Remove stuff that doesn't work because of the migration (nix config, Github ci, etc.).
2. Integrate the package into the monorepo fully.
3. Add CI again.

For review it might be easier to just separate 1, 2, and 3 out into subsequent PRs, making it easier to see their effects.

conorsch:
@cronokirby consider opening an issue to memorialize the pain points we're addressing by pulling the app into the monorepo. This will serve as a reference to future maintainers who try to make sense of this large PR. @TalDerei similarly, could you journal the specific steps you took with the git-filter mechanics? I'm mostly asking for posterity, but I also expect that those steps will need to be rerun before this lands, due to churn on main. It's not immediately clear to me whether we can naively `git rebase` this change set to keep it in sync with main.

cronokirby:
To memorialize some of the motivation:

# Tight Coupling

The dex explorer app has very tight coupling with the packages in this repo. We depend on them very closely, are in need of the latest versions of them. We often find ourselves making changes inside this repo for the purpose of supporting the dex explorer. The additional friction of having separate repositories makes changes like these take unnecessarily long to land. This coupling also leads to another anti-pattern:

# Papering over Shortcomings of the Web Repo

Often, the dex explorer encounters some short-coming or missing piece of functionality from the web repo. Rather than improving it for all users, we're instead encouraged to find a short workaround to continue progress on the explorer. This is not good. It creates a lot of unnecessary code in the explorer, and prevents other people from benefitting from improvements to the web repo.

# More Steady improvements to the Web SDK

By including the apps we write in the same repo, it's easy to identify improvements to the SDK, and include those along side feature-development. Ultimately, while we do want to have an excellent app, a more important longer term goal is having an excellent *foundation* on which to build better apps. It's much easier to do this when it's easy to improve the foundation, rather than it being a "I'll get to this someday after I work around it here" thing.

# Easier integration of GRPC

Less proven, but I have a hunch that a lot of our GRPC woes will be solved by being able to have a central source of truth in *this* repo about the interfaces for both:
- penumbra nodes,
- penumbra clients (ultra-light nodes),
- internal interfaces of the apps.

TalDerei:
holding off on merging until in-flight dex-explorer PRs land.

for posterity, reiterating that minifront and dex-explorer use different build systems – minifront uses **Vite** (vite build), while dex-explorer uses **Next.js** (next build). They handle module resolution differently, requiring weird webpack-specific aliasing to ensure the compiler correctly resolves workspace packages in node modules.

Consequently, the _initial_ merge should just treat the apps separately by aliasing the package.json scripts in the dex-explorer with a `:dex` moniker to ensure the builds succeeds. Afterwards, we can begin migrating to workspace dependencies and reintroducing CI.

TalDerei:
closing in favor of https://github.com/penumbra-zone/web/pull/2134

TalDerei:
reopening as testing ground for assurance purposes

PR #2114, opened 2025-03-13
refactor: #1622: bufbuild upgrade

### LINKED ISSUE #1622: consider `@bufbuild/protobuf@2.x`
https://github.com/bufbuild/protobuf-es/releases/tag/v2.0.0

provides schema and utility functions instead of classes.

this could improve wasm and idb interfaces.

pending support for `@connectrpc/connect` which we depend on.

VanishMax:
Closes #1622 

Upgrades bufbuild and connectrpc libraries to v2.

Steps of this PR:
1. Run `pnpx @connectrpc/connect-migrate@latest` – it updates all dependencies and paths. Didn't commit 200+ files updated by the script, it can always be redone and committed later
2. Updated protobuf generation schema to v2, moved everything to `buf.gen.yaml` file. 
3. Generated new protobufs using `pnpm proto` in the protobuf package
4. Updated protobuf package contents by changing imports from `*_connect.js` to `*_pb.js`. Fixed other issues in the protobuf package like adding Schema suffix to types, changed `ServiceType` to `DescService`
5. Refactored the `create.ts` file in transport-dom and its tests by following this code https://github.com/connectrpc/connect-es/blob/main/packages/connect/src/protocol-grpc/transport.ts

Further work:
1. Refactor the rest of the transport-dom and transport-chrome packages
2. Check other packages for inconsistencies
3. Do `pnpx @connectrpc/connect-migrate@latest` again and commit the changes

### DIFF
--- a/packages/protobuf/buf.gen.yaml
+++ b/packages/protobuf/buf.gen.yaml
@@ -1,12 +1,19 @@
-# buf.gen.yaml defines a local generation template.
-# For details, see https://docs.buf.build/configuration/v1/buf-gen-yaml
-version: v1
+# This file defines a Buf generation schema. Calling it with `pnpm gen`
+# results in downloading common protobuf definitions from the Penumbra protocol repo
+# and transforming them into a comfortable TypeScript format.
+
+# Learn more: https://buf.build/docs/configuration/v2/buf-gen-yaml
+version: v2
+clean: true
+inputs:
+  - module: buf.build/penumbra-zone/penumbra:f6682d4ef839bb06e2da4fae487eb9d565ea04d6
+  - module: buf.build/cosmos/cosmos-sdk:e7a85cef453e4b999ad9aff8714ae05f
+  - module: buf.build/cosmos/ibc:7ab44ae956a0488ea04e04511efa5f70
+  - module: buf.build/cosmos/ics23:55085f7c710a45f58fa09947208eb70b
+  - module: buf.build/noble-assets/forwarding:5a8609a6772d417584a9c60cd8b80881
 plugins:
-  # This will invoke protoc-gen-es and write output to src/gen
-  - plugin: es
+  - local: protoc-gen-es
     out: gen
-    opt: target=ts
-  # This will invoke protoc-gen-connect-es
-  - plugin: connect-es
-    out: gen
-    opt: target=ts
+    opt:
+      - import_extension=js
+      - target=js+dts
--- a/apps/minifront/package.json
+++ b/apps/minifront/package.json
@@ -15,8 +15,8 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@bufbuild/protobuf": "^1.10.0",
-    "@connectrpc/connect": "^1.4.0",
+    "@bufbuild/protobuf": "^2.2.2",
+    "@connectrpc/connect": "^2.0.0",
     "@cosmjs/proto-signing": "^0.32.4",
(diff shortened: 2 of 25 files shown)


packages/protobuf/buf.gen.yaml
```diff
@@ -1,12 +1,20 @@
...
+
+# Learn more: https://buf.build/docs/configuration/v2/buf-gen-yaml
+version: v2
+clean: true
+inputs:
+  - module: buf.build/penumbra-zone/penumbra:f6682d4ef839bb06e2da4fae487eb9d565ea04d6
+  - module: buf.build/cosmos/cosmos-sdk:e7a85cef453e4b999ad9aff8714ae05f
+  - module: buf.build/cosmos/ibc:7ab44ae956a0488ea04e04511efa5f70
+  - module: buf.build/cosmos/ics23:55085f7c710a45f58fa09947208eb70b
+  - module: buf.build/noble-assets/forwarding:5a8609a6772d417584a9c60cd8b80881
```
TalDerei, line 13: comment: should we be using 
[module: buf.build/protocolbuffers/wellknowntypes:fe2b451906844d42b2d078c4d35d87f6](https://buf.build/protocolbuffers/wellknowntypes/commits) here for `Any` and `AnyMessage` imports in transport-dom for instance?

apps/minifront/src/components/shared/input-token.tsx
```diff
@@ -35,7 +37,12 @@ export default function InputToken({
   loading,
 }: InputTokenProps) {
   const setInputToBalanceMax = () => {
-    const match = balances.find(b => b.balanceView?.equals(selection?.balanceView));
+    const match = balances.find(
+      b =>
+        b.balanceView &&
+        selection?.balanceView &&
+        equals(ValueViewSchema, b.balanceView, selection.balanceView),
```
TalDerei, line 44: comment: `equals`, `fromBinary`, `toBinary`, etc. API changes generally make the code harder for me to read compared to the previous version 🤷‍♂️

TalDerei (commented):
the updated bufbuild and connectrpc changes seem reasonable to me, except `transport-dom` which is understandably complex and incomplete. Since that package was recently refactored and has a sensitive diff, I'll defer that review to @turbocrime.

turbocrime:
if it's possible to split this out into a set of automated changes and a set of manual refactors, that would be helpful for review.

VanishMax:
@turbocrime it turned out to be 95% manual - most files generated by migration tool needed changes. Specifically, the tool didn't update .toJson, . toBinary, .equals methods to their respective standalone functions

packages/protobuf/buf.gen.yaml
VanishMax: well-known types are included by default in the `@bufbuilf/protobuf/wkt` since their transition to v2. I constantly use these types throught the PR, for example here: https://github.com/penumbra-zone/web/pull/2114/files#diff-96a26e478b65e8f9919aa9a0a98e316ce41841ebb1770da1bae156799166c39dR7

cronokirby:
Maybe as a pre-migration step we could make sure that communication boundaries that we don't control, like between prax and a website, or with skip, etc. use *serialized* protos, which are the *actual* stable communication layer, and not whatever generated types we happen to have. That way there shouldn't be any interop at all, because the actual messages don't change, just the code we use to work with them.

VanishMax:
@cronokirby thanks for the idea. Will investigate later

PR #2125, opened 2025-03-19
feat(ui): #2124: pagination

### LINKED ISSUE #2124: UI v2: Pagination
https://www.figma.com/design/QSiF2XSvHAWwzBJS5C3f4S/Penumbra-Web-UI?node-id=1030-51107&m=dev

VanishMax:
Closes #2124 

Add Pagination UI component with many props controlling its behavior. Implements this design: https://www.figma.com/design/QSiF2XSvHAWwzBJS5C3f4S/Penumbra-Web-UI?node-id=1030-51107&m=dev. However, goes further and makes it look fine on mobile screens.

image

### DIFF
--- a/packages/ui/src/Pagination/index.tsx
+++ b/packages/ui/src/Pagination/index.tsx
@@ -22,8 +22,8 @@ export interface PaginationProps {
   onChange: (value: number) => void;
   /** The number of items per page. If 0 or less, doesn't render the limit selector */
   limit: number;
-  /** The total number of pages */
-  pages: number;
+  /** The total number of items. If provided, will be used to calculate the total number of pages */
+  totalItems: number;
   /** Callback function to update the number of items per page */
   onLimitChange?: (limit: number) => void;
   /** The available options list for the limit */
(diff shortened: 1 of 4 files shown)


packages/ui/src/Pagination/index.tsx
```diff
@@ -0,0 +1,261 @@
...
+        {!hideLimitSelector && limit > 0 && (
+          <>
+            <label
+              role='button'
+              className='flex cursor-pointer items-center justify-end gap-1 whitespace-nowrap text-text-secondary'
+              htmlFor={selectId}
+              aria-haspopup='listbox'
+              tabIndex={0}
+              onClick={onLabelClick}
+              onKeyDown={onLabelEnter}
```
TalDerei, line 235: comment: do we have / want keyboard support for page navigation?

TalDerei (approved):
Implementation looks great 👍

would be good to apply this to https://github.com/penumbra-zone/web/issues/2132 at some point.

packages/ui/src/Pagination/index.tsx
VanishMax: i usually double check my components for a full keyboard support, so the DEX at this point is ~95% keyboard-covered. Not everything there works perfectly but it's definitely accessible

PR #2133, opened 2025-03-25
DEX migration setup

VanishMax:
Notice that this PR is not for main but for `migration-dex-retry` branch – into https://github.com/penumbra-zone/web/pull/2108

While mentioned PR moves the commit history, this PR makes the DEX actually build and run correctly in local environment.

What's done: 
- Dependencies syncpack-ed
- Removed unnecessary dependencies
- Updated all `package.json`s to NOT export unbuilt files. While it worked well with Minifront, it is super complicated with the DEX. Long story short: `exports` field points dev scripts to look for the raw codebase and delegates dev-build to NextJS's SWC compiler, while `publishConfig.exports` field is used only when publishing to NPM, so that `dist` directories are never applied in a monorepo. My changes make all packages to use built code from `dist` directories of all other packages.
- Made sure that HMR works in both DEX and Minifront
- ESLint was updated to v9 to match DEX's version. This bump changed several linting rules, so I fixed them in this PR

How to run the apps now:
- Call `pnpm dev:pack` from your terminal – this will watch for changes within the packages. Don't stop this while working.
- Call `pnpm dev` from DEX or Minifront – this will run dev script with HMR support

Further work: add required build step to CI correctly and efficiently


VanishMax:
Closed in favor of https://github.com/penumbra-zone/web/pull/2135

Issue #2121, opened 2025-03-13
Current Voting Results


garwalsh:

VanishMax:
Implemented in https://github.com/penumbra-zone/web/pull/2145

PR #2145, opened 2025-03-27
LQT Voting Results

JasonMHasperhoven:
## Description of Changes

Implements https://github.com/penumbra-zone/web/issues/2121

image

## Related Issue


apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+          Current Voting Results
+        </Text>
+        <Density compact>
+          <div className='grid grid-cols-5 h-auto overflow-auto'>
+            <div className='grid grid-cols-subgrid col-span-5'>
+              <TableCell heading>Asset</TableCell>
+              <TableCell heading>Gauge Value</TableCell>
+              <TableCell heading>Casted Votes</TableCell>
+              <TableCell heading>Estimated Incentive</TableCell>
+              <TableCell heading>Vote</TableCell>
```
VanishMax, line 111: todo: add sorting to the table with `useSortableTableHeaders`. Otherwise it'll be more work for the future

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+              Array.from({ length: 6 }).map((_, i) => (
+                <div key={i} className='grid grid-cols-subgrid col-span-5'>
+                  {Array.from({ length: 5 }).map((_, j) => (
+                    <TableCell loading key={j}>
+                      <div className='w-24 h-5'>
+                        <Skeleton />
+                      </div>
+                    </TableCell>
+                  ))}
+                </div>
```
VanishMax, line 124: suggestion (non-blocking): would be easier to add `loading={isLoading}` to every TableCell instead of creating new table. Especially having TableRow, it needs only one prop passing. Also a good thing in this approach is that columns will stay of fixed width if not changed

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  ];
+
+  return (
+    <Card>
+      <div className='flex flex-col p-3 gap-4'>
+        <Text xxl color='text.primary'>
+          Current Voting Results
+        </Text>
+        <Density compact>
+          <div className='grid grid-cols-5 h-auto overflow-auto'>
```
VanishMax, line 105: issue: Tailwind's `grid-cols-{number}` expands as `grid-template-columns: repeat(<number>, minmax(0, 1fr))`, which allows shrinking of columns to fit the viewport. This leads to a table without horizontal scroll, preferable in our case. I'd change it to `grid-cols-[auto_auto_auto_auto_auto]` or `grid-cols-[1fr_1fr_1fr_1fr_1fr]`. If this is implemented along with `whitespace: no-wrap` within the rows, it'd look better for mobile screens.

image

VanishMax (changes requested):

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
JasonMHasperhoven: How do you make sure to render a couple of rows when there's no data yet?

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
JasonMHasperhoven: Wasn't aware, thanks for the tip!

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
VanishMax: Similar to this: https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/previous-epochs.tsx#L43

VanishMax (approved):
With sorting in place, I don't see any blockers

PR #2134, opened 2025-03-25
migration: dex-explorer

TalDerei:
## Description of Changes

migrates the [dex-explorer](https://github.com/penumbra-zone/dex-explorer) into the web monorepo using [git filter-repo](https://github.com/newren/git-filter-repo) tool and `--allow-unrelated-histories`. This preserves and interleaves commit histories chronologically. Some additional formatting for dex-explorer PR #'s in the [commit history](https://github.com/penumbra-zone/web/commits/migration-dex-retry/) is required.

_extra_ assurance should be done to ensure the migration was performed correctly. cc @conorsch 

## Related Issue

https://github.com/penumbra-zone/web/issues/2123

## Checklist Before Requesting Review

- [x] I have ensured that any relevant minifront changes do not cause the existing extension to break.

PR #2147, opened 2025-03-28
ci: update deployment workflows for dex-explorer

conorsch:
## Description of Changes

Follow-up to https://github.com/penumbra-zone/web/pull/2134. Ports the CI workflows to the repo root, so they'll run on the right events. 

The substantive changes here are to the `Containerfile`. Given the migration of Veil to a monorepo setup, we now have a much larger build context to handle: we must build all relevant dependencies prior to building Veil, including using Rust tooling. That's handled now.  The resulting image is larger: ~1.2GB compared to ~300MB for the old dex-explorer repo, but I'll punt on further refinements until we have CD flowing again.

## Related Issue
https://github.com/penumbra-zone/web/pull/2134


TalDerei:
tagging https://github.com/penumbra-zone/web/pull/2148 as relevant

conorsch:
The container build story still needs some work: the module resolution for the node deps has changed, and I haven't been able to sort it yet. Will work on resolving locally, then ask CI to build it.

conorsch:
After a surprising amount of wrestling, I've got successful container build, posting to the GHCR, under new image name `veil`. The new build context is expansive: because we're using monorepo workspace deps for veil—the point of this migration—we need to ensure that we can build the entire workspace, including the wasm deps, when building veil.

Not done yet, though: the built artifact doesn't serve, due to a misalignment on the `server.js` paths. Once that's straightened out, I'll put up a trial instance hooked up to testnet data.

.github/workflows/container.yml
```diff
@@ -12,6 +12,9 @@ on:
   workflow_call:
   workflow_dispatch:
   repository_dispatch:
+
+  # Temporary: build on PRs, to make sure workflow succeeds
+  pull_request:
```
conorsch, line 17: The comment says "temporary" but I'd still like to merge it: I'll follow up with smaller tweaks in subsequent PRs, after observing how the migrated deploy workflows behave, once they land in main.

.github/workflows/deploy-dex-explorer.yml
```diff
@@ -51,9 +51,10 @@ jobs:
       - name: deploy veil
         run: >
           nix develop --command
-          kubectl -n testnet rollout restart deployment veil &&
+          kubectl -n testnet rollout restart deployment dex-explorer &&
           nix develop --command
-          kubectl -n testnet rollout status deployment veil
+          kubectl -n testnet rollout status deployment dex-explorer
+
```
conorsch, line 57: The `dex-explorer` name is correct for the current deployments; can always rename later on. I'll handle any string substitutions in the deploy workflows as necessary, in follow-ups.

packages/keys/download-keys
```diff
@@ -88,3 +88,6 @@ fi
 cp -v "$cacheDir/$keysVersion/"*_pk.bin "$outDir"
 
 check_keys "$outDir"
+
+echo "Finished 'download-keys' script from 'penumbra-zone/keys' package"
+exit 0
```
conorsch, line 93: Informational output only, no change to script behavior. I shelved rewriting this script to bail out on errors more readily, in order to focus on shipping the container rewrite. Might circle back to this script on a rainy day, though.

conorsch (commented):
Left some in-line comments to guide review.

conorsch:
I'm able to use the new container, built locally, on my workstation to view mainnet data:

veil-screenshot-1

Calling this suitable for merge. Post-merge, I'll need to follow up with some tweaks to the deployment logic, but this changeset is enough to get the container building and serving again.

cronokirby (approved):
nice!

TalDerei:
LGTM! wondering what [this commit](https://github.com/penumbra-zone/infrastructure/commit/d9417072b443963fa46d11b7f9a26e341c7e76ee) does though?

TalDerei (approved):

Issue #2150, opened 2025-03-17
LQT transaction plan construction


TalDerei:
flush out the core LQT logic needed for transaction planning that can be integrated following https://github.com/penumbra-zone/web/issues/2115, which initially will be using dummy data.

PR #2144, opened 2025-03-27
[LQT] Round card

JasonMHasperhoven:
## Description of Changes

Implements https://github.com/penumbra-zone/web/issues/2116 (with dummy data)

image

## Related Issue

Same as  https://github.com/penumbra-zone/dex-explorer/pull/406

## Checklist Before Requesting Review

- [x] I have ensured that any relevant minifront changes do not cause the existing extension to break.

PR #2191, opened 2025-03-31
LQT: voting dialogue

TalDerei:
## Description of Changes

Implements voting dialogue and transaction construction for LQT

------------------

https://github.com/user-attachments/assets/2e7c4e80-0530-4647-94ff-9a3ed2c8ed11

## Related Issue

references https://github.com/penumbra-zone/web/issues/2150 and https://github.com/penumbra-zone/web/issues/2122


apps/veil/src/pages/tournament/ui/landing-card/voting-footer.tsx
```diff
@@ -104,42 +118,50 @@ export const VotingFooter = observer(({ isBanned }: { isBanned: boolean }) => {
...
-        </Text>
-      </div>
-      <div className='flex gap-2'>
-        <Button actionType='accent' icon={ExternalLink}>
-          Delegate
-        </Button>
-        <Button actionType='default'>Details</Button>
-      </div>
-    </div>
+      <VoteDialogueSelector isOpen={isVoteDialogueOpen} onClose={closeVoteDialog} />
```
TalDerei, line 164: the `VoteDialogueSelector` is temporarily wired to the `Delegate` button for testing purposes. It will be hooked up to the `Vote` button once https://github.com/penumbra-zone/web/pull/2144 lands. cc @JasonMHasperhoven

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const showSearchResultsHeader =
+    (searchQuery && !hasResults) ||
+    (!searchQuery &&
+      filteredAssets.length > 0 &&
+      filteredAssets.every(a => a.percentage < threshold));
+
+  return (
+    <div className='relative flex items-center gap-2 text-text-primary'>
+      <Dialog isOpen={isOpen} onClose={handleClose}>
+        <Dialog.Content title={`Vote in epoch #${epochIndex}`}>
```
TalDerei, line 203: noticed that using `Dialog` causes the entire dialog box to be left-shifted here. That's also the case for the asset selector on the trading page. Manual attempts to force centering with a outer wrapper div didn't work – might be worth adding a prop like `centered={true}` to the dialogue UI component to handle this?

VanishMax: Somehow it doesn't happen for me. Not for this dialog, not for pair selector 🤷‍♂️

I'd like to dig dipper into this, just not sure how to reproduce

image

apps/veil/src/pages/tournament/api/use-voting-notes.ts
```diff
@@ -0,0 +1,37 @@
...
+  return await Array.fromAsync(
+    penumbra.service(ViewService).lqtVotingNotes({ accountFilter, epochIndex }),
+  );
+};
+
+/**
+ * Must be used within the `observer` mobX HOC
+ */
+export const useLQTNotes = (epochIndex: bigint, subaccount?: number) => {
+  const queryKey = ['lqt-notes', subaccount];
```
VanishMax, line 25: suggestion / question: probably should include epochIndex?

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
TalDerei: https://github.com/user-attachments/assets/5ddd691d-a57e-4f54-bf58-84a315940812

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
```diff
@@ -0,0 +1,296 @@
...
+          const isSelected = selectedAssetId === asset.id && !disableSelection;
+
+          return (
+            <button
+              key={asset.id}
+              onClick={() => !disableSelection && onSelect(asset.id)}
+              className={`w-full text-left rounded-xl px-4 py-2 ${
+                isSelected ? 'bg-[#1e2c2b]' : ''
+              } ${disableSelection ? 'opacity-50 cursor-not-allowed' : 'hover:bg-[#202425] transition'}`}
+              disabled={disableSelection}
```
VanishMax, line 58: suggestion: within Dialog, you can use `Dialog.RadioGroup` and `Dialog.RadioItem` components. They are not critically different from designs, but give much better UX in comparison with writing your own components. 

Take a look at the example here: https://preview.ui.penumbra.zone/?path=/docs/ui-library-dialog--docs#with%20radio%20items – the bottom of the page, example called "With Radio Items". Another example is wallet selector dialog – opens in Veil if you have 2+ wallet extensions installed.

For your usecase, you could substitute the `description` property with the progress block, and `title` – another div with useful info. Like this:

```tsx
<Dialog.Content title='Select wallet'>
          <Dialog.RadioGroup>
            <div className='flex flex-col gap-1 pt-1'>
              <Dialog.RadioItem
                value='1'
                title={
                  <div className='flex justify-between'>
                    <Text technical color='text.primary'>{metadata.symbol}</Text>
                    <Text technical color='text.secondary'>2.5%</Text>
                 </div>
                }
                description={<ProgressBar />}
                startAdornment={<AssetIcon metadata={PIZZA_METADATA} size='lg' />}
              />
          </Dialog.RadioGroup>
</Dialog.Content>
```

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+              <div className='flex items-center justify-center gap-2 mt-1'>
+                <div className='rounded-sm bg-[#cc5500] w-5 h-5 flex items-center justify-center relative'>
+                  <input
+                    type='checkbox'
+                    id='reveal-vote'
+                    checked={revealVote}
+                    onChange={() => setRevealVote(!revealVote)}
+                    className='opacity-0 absolute w-full h-full cursor-pointer'
+                  />
```
VanishMax, line 270: todo: https://github.com/penumbra-zone/web/issues/2192

for now, we can leave it as-is, maybe add a todo comment to replace it in the future

apps/veil/src/pages/tournament/ui/landing-card/voting-footer.tsx
```diff
@@ -72,8 +83,11 @@ export const VotingFooter = observer(({ isBanned }: { isBanned: boolean }) => {
           </Text>
         </div>
         <div className='flex gap-2'>
-          <Button actionType='accent' icon={WalletMinimal}>
-            Connect Prax Wallet
+          <Button actionType='accent'>
+            <div className='flex items-center gap-2 whitespace-nowrap'>
```
JasonMHasperhoven, line 87: I feel like `whitespace-nowrap` should be applied to all buttons, what do you think? (also cc @VanishMax )

VanishMax: Agree that it should be there by default, even no prop needed to disable

apps/veil/src/pages/tournament/api/use-voting-notes.ts
TalDerei: sure, it won't make a meaningful difference here since we're refetching every block anyway, but it doesn't hurt to include it. Fetching every block is preferable to fetching every epoch because it lets us display the user's recent delegation note balance, but that correspondingly means we'll need clear UI messaging around delegation notes that _aren't_ eligible to vote in the current epoch. 

There's three cases to consider: 
1. the user's entire delegation note balance is eligible for voting – everything proceeds as normal.
3. none of the user's delegation notes are eligible for voting because they were delegated in the current epoch – we need to show messaging to reflect that. 
4. only some delegation notes are eligible — in this case, we should clearly show the partial eligibility.

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
TalDerei: I actually originally considered using the `Dialog.RadioGroup` and `Dialog.RadioItem` components, but they introduce a grey overlay bar that diverges from the figma design.

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
TalDerei: added a comment for follow-up work

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
TalDerei: > It also doesn’t seem to be enforcing consistent widths for the dialogue box between our screenshots?

@VanishMax what causes these width differences in the voting dialogue? is there a canonical way to enforce uniformity there?

VanishMax: > is there a canonical way to enforce uniformity there?

not that I know of. The width is based on a grid system from Jesse, I'm not fully familiar with how it should work. We can definitely revise it

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
VanishMax: @TalDerei That's ok. If the design lacks an overlay, it must only mean the design is not ready. Each clickable element must have an overlay on focus for accessibility purposes

apps/veil/src/pages/tournament/ui/landing-card/voting-footer.tsx
VanishMax: well, i'd be hesitant to add `truncate` by default, but as a prop it might work well

packages/wasm/crate/src/planner.rs
```diff
@@ -683,21 +683,12 @@ pub async fn plan_transaction_inner<Db: Database>(
                 .note
                 .ok_or_else(|| anyhow!("missing note in liquidity tournament"))?
                 .try_into()?;
-            let staked_note_position: Position = staked_note.position.into();
+            let note_position: Position = staked_note.position.into();
```
erwanor, line 686: This seems fine because it should call the bitpacking conversion trait

erwanor: iirc there's a `From<u64>`

packages/wasm/crate/src/planner.rs
```diff
@@ -683,21 +683,19 @@ pub async fn plan_transaction_inner<Db: Database>(
...
-                    staked_note: note,
-                    staked_note_position,
-                    start_position,
-                    randomizer: Fr::rand(&mut OsRng),
-                    proof_blinding_r: Fq::rand(&mut OsRng),
-                    proof_blinding_s: Fq::rand(&mut OsRng),
-                },
-            ));
+            let note_position: Position = staked_note.position.into();
+            let start_position: Position = Position::from((epoch_index as u16, 0, 0));
```
TalDerei, line 687: comment: bug ([https://github.com/penumbra-zone/web/pull/2191/commits/feec9d50438c3c1e6b8792412679c3e9b69c9c1f#diff-ec51d55a0b0ab2ce263[…]c963213a0899cca79a0f0R687](https://github.com/penumbra-zone/web/pull/2191/commits/feec9d50438c3c1e6b8792412679c3e9b69c9c1f#diff-ec51d55a0b0ab2ce263357c3d43509ddfe1d88ace8bc963213a0899cca79a0f0R687)) where the conversion of the start_position was incorrect – it produced `Position(Tree { epoch: 0, block: 0, commitment: EPOCH_INDEX }) `instead of `Position(Tree { epoch: EPOCH_INDEX, block: 0, commitment: 0 })`, which has been fixed.

packages/wasm/crate/src/tx.rs
```diff
@@ -80,6 +80,10 @@ pub fn witness_inner(plan: TransactionPlan, stored_tree: StoredTree) -> WasmResu
             plan.delegator_vote_plans()
                 .map(|vote_plan| vote_plan.staked_note.commit()),
         )
+        .chain(
+            plan.lqt_vote_plans()
+                .map(|vote_plan| vote_plan.staked_note.commit()),
+        )
```
TalDerei, line 86: comment: bug fix – chain lqt commitments in witness data

apps/veil/src/pages/tournament/api/vote.ts
```diff
@@ -0,0 +1,81 @@
...
+  } catch (e) {
+    openToast({
+      type: 'error',
+      message: 'Error with withdraw action',
+      description: String(e),
+    });
+  }
+};
+
+export const checkIfAlreadyVoted = ({
```
TalDerei, line 62: detecting if all delegation notes have been used to vote in the current voting period

----------------------

Screenshot 2025-04-09 at 1 28 47 PM

packages/wasm/crate/src/keys.rs
```diff
@@ -29,6 +29,7 @@ pub fn load_proving_key(key: &[u8], key_type: &str) -> WasmResult<()> {
         "swap" => &SWAP_PROOF_PROVING_KEY,
         "swapClaim" => &SWAPCLAIM_PROOF_PROVING_KEY,
         "undelegateClaim" => &CONVERT_PROOF_PROVING_KEY,
+        "actionLiquidityTournamentVote" => &DELEGATOR_VOTE_PROOF_PROVING_KEY,
```
TalDerei, line 32: fixes action building

apps/veil/src/pages/tournament/ui/vote-dialogue.tsx
TalDerei: would like to investigate uniformity in https://github.com/penumbra-zone/web/issues/2233

VanishMax (approved):

PR #2197, opened 2025-04-02
feat(ui): #2195: LiquidityTournamentVote ActionView

### LINKED ISSUE #2195: ActionView component: `ActionLiquidityTournmentVote`
Design:

Image

VanishMax:
Closes #2195 

Goes along with https://github.com/prax-wallet/registry/pull/133

Live version: https://penumbra-ui-preview--pr2197-feat-2195-lqt-vote-oyu4mi88.web.app/?path=/docs/ui-library-actionview--docs

image


TalDerei (commented):
for LQT actions, the only difference between the opaque and visible cases is that the visible case provides more information about the note. 

If we compare the visible and opaque cases in the [storybook](https://penumbra-ui-preview--pr2197-feat-2195-lqt-vote-oyu4mi88.web.app/?path=/docs/ui-library-actionview--docs), the visible case shows the address that cast the vote, while the opaque case shows the rewards recipient address. This distinction blurs the definition of what each address represents, and we should clarify this more explicitly. We shouldn't be intermingling address definitions here. 

Instead, proposing we (1) add a dedicated **rewards recipient** field to both the visible and opaque cases, and (2) remove the address view from the header in the **opaque** case to align with the figma design.

.github/workflows/deploy-ui-preview-main.yml
```diff
@@ -7,18 +7,29 @@ on:
     branches:
       - main
 jobs:
+  turbo-compile:
```
TalDerei, line 10: comment: ACK – github preview deployments related to https://github.com/penumbra-zone/web/issues/2200

VanishMax:
@TalDerei thanks, that's a very important feedback! I guess I didn't understand it fully before – the designs were also based on my incorrect requirements. It should be better now

packages/ui/src/ActionView/actions/liquidity-tournament-vote.tsx
```diff
@@ -79,12 +86,13 @@ export const LiquidityTournamentVoteAction = ({
       },
     });
 
-    if (value.liquidityTournamentVote.case === 'visible') {
-      return value.liquidityTournamentVote.value.note?.address ?? defaultAddress;
+    const visible = spendAddress?.addressView.value?.address;
+    if (visible && receipient && visible.equals(receipient)) {
```
TalDerei, line 90: suggestion: remove `visible.equals(receipient)` – there's nothing enforcing this requirement at the protocol level. The address used for voting can be different from the recipient address.

packages/ui/src/ActionView/actions/liquidity-tournament-vote.tsx
```diff
@@ -1,10 +1,138 @@
...
+    <ActionWrapper
+      title='Liquidity Tournament Vote'
+      opaque={value.liquidityTournamentVote.case === 'opaque'}
+      infoRows={[
+        !!epoch && <ActionRow label='Epoch' info={`#${epoch.toString()}`} />,
+        !!voteValue && (
+          <ActionRow
+            label='Voting Power'
+            info={
+              <Density slim>
```
TalDerei, line 111: comment: sparse and compact density handling in the storybook don't match the figma for this action view.

TalDerei (approved):
left some suggested improvements, otherwise looks good!

packages/ui/src/ActionView/actions/liquidity-tournament-vote.tsx
VanishMax: i know, but if there's a match, which is a simple comparison operation, then we have a better rendering for a user

Issue #2198, opened 2025-04-02
bug: portfolio page failed loading


hitchhooker:
```sh
 Failed to select value from "[object Object]" with ""
    at t (
```

Application error: a client-side exception has occurred (see the browser console for more information).

![Image](https://github.com/user-attachments/assets/dc363365-e94a-4379-aa59-c91d43e255ec)

![Image](https://github.com/user-attachments/assets/fc91c6bf-c29d-4713-867f-035f596b32b1)

![Image](https://github.com/user-attachments/assets/6097a4ed-6c0c-434b-b14f-e82fd40e9f99)

reloading worked out:

![Image](https://github.com/user-attachments/assets/70936135-9736-4cac-9e4d-19e61671d2a0)

TalDerei:
@hitchhooker thanks for the bug report, we're internally tracking this and investigating the cause.

conorsch:
This is a great bug report, thank you for including the details. :pray:

PR #2199, opened 2025-04-02
filter unknown balances

TalDerei:
## Description of Changes

filters out unknown assets from balances view service implementation when querying the [unified assets](https://github.com/penumbra-zone/dex-explorer/blob/main/src/pages/portfolio/api/use-unified-assets.ts#L82-L83) on the portfolio page. 

## Related Issue

fixes https://github.com/penumbra-zone/web/issues/2198#issuecomment-2773216896

### DIFF
--- a/.changeset/ten-apes-kneel.md
+++ b/.changeset/ten-apes-kneel.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/services': patch
+---
+
+filter unknown assets from balances view service"
--- a/.changeset/ten-apes-kneel.md
+++ b/.changeset/ten-apes-kneel.md
@@ -1,5 +1,0 @@
----
-'@penumbra-zone/services': patch
----
-
-filter unknown assets from balances view service"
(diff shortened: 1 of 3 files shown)


cronokirby (changes requested):
https://github.com/penumbra-zone/dex-explorer/blob/main/src/pages/portfolio/api/use-unified-assets.ts#L122 we should instead filter this out in the component itself, by adding a check that the metadata is not unknown here, for example.

TalDerei:
> we should instead filter this out in the component itself

good point, modified the fix to be more localized; the issue stemmed from `getMetadataFromBalancesResponse` being called on unknown assets. We now apply two filtering passes: the first filters out balances with unknown assets, and the second excludes specific [asset types](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/portfolio/api/use-unified-assets.ts#L56) like LP NFTs, voting tokens, etc.

cronokirby (approved):

PR #2201, opened 2025-04-02
fix: unbonding penalty calculation

turbocrime:
## Description of Changes

Port of pcli fix https://github.com/penumbra-zone/penumbra/pull/5084 to the minifront planning UI

>This fixes an edge-case in pcli tx undelegate-claim that would compute an invalid penalty range for dated delegation tokens.

>Instead of using the current epoch as the end of the penalty range, we compute a more precise bound using the block delay chain parameter. This fixes processing old delegation tokens bound to a validator that transitioned in and out of penalty states.

## Related Issue

https://github.com/penumbra-zone/penumbra/pull/5084

### DIFF
--- a/.changeset/forty-laws-refuse.md
+++ b/.changeset/forty-laws-refuse.md
@@ -1,0 +1,5 @@
+---
+'minifront': minor
+---
+
+fix unbonding penalty range calculation
--- a/.changeset/slow-wings-glow.md
+++ b/.changeset/slow-wings-glow.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/types': minor
+---
+
+remove `getUnbondingStartHeight` export from assets regex module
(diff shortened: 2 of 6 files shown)


TalDerei (approved):
compared with rust version and they appear functionally equivalent 👍

turbocrime:
i don't have an example case to actually test functionality. cc @erwanor

erwanor:
I don't either, it's a somewhat niche edge case. I think if you run this locally and can delegate/undelegate/claim (i.e. no ostensible regression) then this is good to go and we should merge it

TalDerei:
> I don't either, it's a somewhat niche edge case. I think if you run this locally and can delegate/undelegate/claim (i.e. no ostensible regression) then this is good to go and we should merge it

successfully tested the workflows

erwanor:
Thanks a lot @turbocrime and @TalDerei, merging optimistically

PR #2204, opened 2025-04-03
use node setup action in workflows

turbocrime:
use node setup action in deploy-ui workflows, to match others. these workflows broke recently due to new engine requirements in veil.

should probably centralize tool-versions with something like mise.

https://github.com/penumbra-zone/web/actions/runs/14231070370/job/39914124285


VanishMax (approved):
question: should it also be applied to the `deploy-ui-preview-pr.yml` file?

Issue #2097, opened 2025-03-05
minifront: syncing status dialog will never reopen once closed


turbocrime:
if the user sees the syncing dialog early in a session, and then closes it, the dialog will never reopen because the 'closed dialog' state is permanent.

this may be confusing to a user that expects to see the status dialog appear when there are problems.

for example:

1. the dialog appears briefly at the start, as syncing begins, is pending, and then streams
2. the user clicks away before it disappears automatically. the closed dialog state is now permanent.
4. later, possibly minutes or hours later, an error causes syncing to fail.
5. the status header displays this state by turning red, but is less informative.
6. the dialog won't open again, for any reason.

it's not appropriate to simply reset the dialog closed state when an error is present, because the user may wish to inspect local state while offline, which would continually produce a syncing error and continually open the dialog.

## solution A

using a long constant-duration timeout to unset the dialog closed state might be a simple solution. it would achieve reopening without any logic or state inference.

negative: the user may be annoyed by a reappearing dialog showing the same state. a longer timeout would reduce this, but decrease responsiveness to a changing situation.

## solution B

maybe, some logic could conditionally do this without a timer.

1. if the user closes the dialog during a pending state
    - no action
2. if the stream begins (regardless of user action)
    - immediately clear closed state when an error appears
    - immediately clear closed state when sync reaches present
3. if the user closes the dialog during an error state
    - immediately clear closed state if streaming begins

Issue #2096, opened 2025-03-05
Syncing modal doesn't always spin after an interruption


TalDerei:
the syncing modal animation is configured to stop when 'syncing interrupted' and should have some kind of animation. 

---------------

https://github.com/user-attachments/assets/1ba6cfd8-94d8-4463-938f-e87e45746fd2

varunguleriaCodes:
Hi @TalDerei would like to work on this what kind of animation would you like to see when an error occurs?

PR #2205, opened 2025-04-03
Improve syncing modal

turbocrime:
address some issues with the syncing modal.

- [x] fixes https://github.com/penumbra-zone/web/issues/2097
- [x] fixes https://github.com/penumbra-zone/web/issues/2096

~~also, some arbitrary changes to the appearance.~~

keep in mind the orange error text is only visible in debug mode.

### DIFF
--- a/apps/minifront/src/components/syncing-dialog/index.tsx
+++ b/apps/minifront/src/components/syncing-dialog/index.tsx
@@ -52,7 +52,7 @@ export const SyncingDialog = () => {
   }, [isUnavailable, isSynced, streamError, didClose]);
 
   const dialogText = useMemo(() => {
-    const title = `Syncing ${streamError ? 'interrupted' : '. . .'}`;
+    const title = `Syncing ${streamError ? 'interrupted' : '...'}`;
 
     let error = '';
     if (globalThis.__DEV__ && streamError) {
@@ -86,9 +86,10 @@ export const SyncingDialog = () => {
     return { title, error, detail, instructions };
   }, [didClose, initialStatus.data, isSynced, isUnavailable, streamStatus.data, streamError]);
 
+  // TODO: 'zIndex' is deprecated — update to use zIndex on buttons instead
   return (
     <Dialog isOpen={isOpen} onClose={() => setDidClose(streamError ?? true)}>
-      <Dialog.Content title={dialogText.title}>
+      <Dialog.Content title={dialogText.title} zIndex={9999}>
         <div className='text-center'>
           {dialogText.error && (
             <Text technical color={theme => theme.caution.main} as='div'>
--- a/apps/minifront/src/components/syncing-dialog/index.tsx
+++ b/apps/minifront/src/components/syncing-dialog/index.tsx
@@ -10,22 +10,19 @@ import { Code, ConnectError } from '@connectrpc/connect';
 
 export const SyncingDialog = () => {
   const initialStatus = useInitialStatus();
-  const status = useStatus();
+  const streamStatus = useStatus();
   const { error: streamError } = useStore(statusStreamStateSelector);
 
   const [didClose, setDidClose] = useState(false);
 
-  const syncData = useMemo(
-    () => ({ ...initialStatus.data, ...status.data }),
-    [initialStatus.data, status.data],
+  const { fullSyncHeight = 0n, latestKnownBlockHeight = 0n } = useMemo(
+    () => ({ ...initialStatus.data, ...streamStatus.data }),
+    [initialStatus.data, streamStatus.data],
   );
 
   const isSynced = useMemo(
-    () =>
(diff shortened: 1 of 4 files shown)


apps/minifront/src/components/syncing-dialog/sync-animation.tsx
```diff
@@ -1,20 +1,60 @@
...
+    } else if (!running) {
+      return 'starting';
+    } else {
+      return 'streaming';
+    }
+  }, [error, running]);
+
+  return (
+    <div className={cn('absolute', 'left-0 top-0', 'size-full', '-z-10', 'overflow-hidden')}>
+      <div className={cn('pt-8 px-4', '[mask-image:linear-gradient(black,transparent_65%)]')}>
```
TalDerei, line 25: understand the design is subjective, but I'd like to experiment with some suggestions here; 

------------

1. different modes currently cause inconsistent or buggy renderings. you'll notice for example during a refresh, the syncing modal visually shifts and adjusts itself. Additionally, it cuts off UI elements like the send / receive buttons.  Instead we should increase the modal size to _accommodate_ all modes consistently and avoid readjustments. The syncing modal skeleton should be static in some sense across rerenders. 

https://github.com/user-attachments/assets/cda4dead-d149-4a98-ac72-294431473582


https://github.com/user-attachments/assets/21a8d285-4589-4bda-a799-2073cefe217e

-------------------------------------

2. different sync modes currently display varying modal sizes, leading to inconsistent layouts. Notice the size differences between streaming and retrying states. 

https://github.com/user-attachments/assets/d8c7f726-c13d-48af-8aaa-7ce26d483d67

-------------------------------------

3. let’s remove the estimated time remaining and block height display during sync interruptions. Currently, it shows up inconsistently. 

Screenshot 2025-04-04 at 4 42 28 AM

Screenshot 2025-04-04 at 4 43 35 AM

-------------------------

4. In the starting state, the continuous animation isn’t as smooth / fluid as it was before.

https://github.com/user-attachments/assets/8852970d-51bc-42e5-8af9-185585d3d437

https://github.com/user-attachments/assets/5d0dfdeb-7500-4600-8c94-1b9b98e3ffbd

-----------------

5. there may be some legibility issues with the current “text over image” layout. Can we experiment with increasing the modal size and adjusting the layout – moving "Syncing interrupted" higher up, and placing the supporting text below the illustration? That said, I really like the color blending in this version compared to the previous one. 

------------------

6. ETA sometimes wraps to the next line

apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -29,82 +30,86 @@ export const SyncingDialog = () => {
...
-        zIndex={9999}
-      >
-        <div className='text-center'>
-          {streamError != null && globalThis.__DEV__ && (
-            <Text technical color={theme => theme.caution.main} as='div'>
-              Error: {dialogErrorMessage}
-            </Text>
-          )}
-        </div>
+    <Dialog isOpen={isOpen} onClose={() => setDidClose(stream.error ?? true)}>
```
TalDerei, line 80: suggestion: user can't manually close the syncing modal; currently lacks explicit dismissible behavior

apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -29,82 +30,86 @@ export const SyncingDialog = () => {
...
+      <Dialog.Content title={dialogText.title} zIndex={9999}>
```
TalDerei, line 81: comment: `zIndex` deprecated

turbocrime:
cosmetic changes reverted. PR now focused on issues

- syncing modal animation always spins
- syncing modal will conditionally reopen

apps/minifront/src/components/syncing-dialog/index.tsx
turbocrime: this was present previously, and i was avoiding changing it. but it's removed now.

now the segmented picker overlays the modal.

apps/minifront/src/components/syncing-dialog/index.tsx
turbocrime: ah, you're right. the object is not comparable like i assumed. the logic is corrected and now it compares error message strings which should be more forgiving

VanishMax (approved):

TalDerei (commented):
some modal overlay issues with UI when refreshing or cutting connection 

---------------------

Screenshot 2025-04-08 at 7 25 31 AM
Screenshot 2025-04-08 at 7 27 01 AM
Screenshot 2025-04-08 at 7 27 31 AM

apps/minifront/src/components/syncing-dialog/index.tsx
TalDerei: added back the `zIndex`

Issue #2212, opened 2025-04-04
undisclosed analytics in skip app


hitchhooker:
come a cross with amplitude analytics tracking in skip codebase without clear user disclosure. while it doesn't directly save wallet addresses, it does track:
- wallet connections/disconnections
- address retrieval attempts 
- error messages

for privacy-preserving apps, users should know when they're being tracked. no opt-out option visible.

ref: https://github.com/skip-mev/skip-go/blob/e3b7f978039a7289fdf415632a03c14e8d7357a0/packages/widget/src/hooks/useCreateCosmosWallets.tsx#L303-L308

im not sure if app.penumbra.zone ui even builds with this version and might be nothing burger, but transparency about analytics kinda appreciated/expected in privacy tools.

PR #2215, opened 2025-04-06
stub skip-go dep

turbocrime:
Provides an empty package to override `@amplitude/analytics-browser` and stubs for `@skip-go/widget` imports from `@amplitude/analytics-browser` to completely omit the package from build output

references https://github.com/penumbra-zone/web/issues/2212

### DIFF
--- a/.changeset/old-frogs-flash.md
+++ b/.changeset/old-frogs-flash.md
@@ -1,0 +1,5 @@
+---
+'@repo/stubs': major
+---
+
+introduce stubs
--- a/apps/minifront-v2/package.json
+++ b/apps/minifront-v2/package.json
@@ -48,6 +48,7 @@
   "devDependencies": {
     "@chain-registry/types": "^0.50.93",
     "@eslint/compat": "^1.1.0",
+    "@repo/stubs": "workspace:*",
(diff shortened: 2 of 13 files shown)


TalDerei (commented):
organizationally, I'd prefer a more self-contained approach than creating a dedicated package and placing stub files in the minifront root

turbocrime:
replacing it requires some kind of stub, because the imported functions are called.

TalDerei:
> replacing it requires some kind of stub, because the imported functions are called.

wondering how other projects handle dependency overrides and package omissions?

TalDerei:
> wondering how other projects handle dependency overrides and package omissions?

wondering how we can also make progress on this; not a fan of the current approach organizationally (empty package + stub in minifront root)

turbocrime:
would like this to merge.

i think adding a repo package is not terrible. it enables reuse, and i expect more.

TalDerei (approved):

TalDerei:
dedicated package seems like the most practical approach here, and it also allows us to pull in other dependencies like cosmos.

turbocrime:
go ahead and merge when you're satisfied with behavior as discussed

PR #2218, opened 2025-04-07
simplify key script

turbocrime:
Simplify key download script.

for some reason, downloading repo archive tarballs such as https://github.com/penumbra-zone/penumbra/archive/refs/tags/v1.0.0.tar.gz is prevented by CI, but the former technique of downloading the glob URL does work. so, tarballs are still not used.


packages/keys/package.json
```diff
@@ -5,20 +5,16 @@
...
     "action-keys.json",
-    "download-keys",
     "keys/*_pk.bin"
   ],
   "exports": {
     ".": "./action-keys.json",
     "./*_pk.bin": "./keys/*_pk.bin"
-  },
-  "bin": {
-    "penumbra-download-keys": "./download-keys"
```
turbocrime, line 22: tool is no longer provided. keys are bundled in package and shouldn't change.

packages/keys/download-keys
turbocrime, line 1: download script is much simpler.

packages/keys/README.md
turbocrime, line 1: readme omits information on removed tool. other information is simplified.

turbo.json
turbocrime, line 1: task dependency changes

TalDerei (approved):
👍 remind me, why do we store two copies of the keys in the package again

_update_: reasoning explained in https://github.com/penumbra-zone/web/pull/434#discussion_r1472812909

PR #2220, opened 2025-04-08
fix(types, ui): better swap getters

VanishMax:
Fixes the issue with certain swap ActionViews that don't have metadata inside them, usually in opaque views.

image


TalDerei (approved):
fallback to unknownAssetId when metadata is unavailable looks correct to me

Issue #2164, opened 2025-01-17
Input overflow in the order form


TalDerei:
`buy` and `pay with` input fields in the order form can overflow, cutting off user input. We should implement input validation to limit the number of decimal places or total input length.

https://github.com/user-attachments/assets/712695e5-bc41-4852-8b52-d8c95edcee18

Issue #2163, opened 2025-02-03
'pay with' should reduce decimal precision


TalDerei:
https://github.com/user-attachments/assets/a981b89a-93d8-43e8-b3be-8f2cddb6eac2

VanishMax:
@TalDerei can you elaborate on this? Like, what should be the requirements and expected results for this task?

TalDerei:
> [@TalDerei](https://github.com/TalDerei) can you elaborate on this? Like, what should be the requirements and expected results for this task?

the `pay with` amount should be capped to 5 decimal places

PR #2223, opened 2025-04-08
Order form: Set max decimals of input to exponent

JasonMHasperhoven:
## Description of Changes

Implements https://github.com/penumbra-zone/web/issues/2163 and https://github.com/penumbra-zone/web/issues/2164

- Sets the max decimals of an input to the assets exponent

https://github.com/user-attachments/assets/f0fddf31-c9b2-4111-92ce-3cee6ef9d0b8

## Related Issue

### DIFF
--- a/apps/veil/src/pages/trade/ui/order-form/store/LimitOrderFormStore.ts
+++ b/apps/veil/src/pages/trade/ui/order-form/store/LimitOrderFormStore.ts
@@ -4,7 +4,7 @@ import { limitOrderPosition } from '@/shared/math/position';
 import { makeAutoObservable, reaction } from 'mobx';
 import { AssetInfo } from '@/pages/trade/model/AssetInfo';
 import { parseNumber } from '@/shared/utils/num';
-
+import { round } from '@penumbra-zone/types/round';
 export type Direction = 'buy' | 'sell';
 
 export enum SellLimitOrderOptions {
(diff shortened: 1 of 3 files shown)


apps/veil/src/pages/trade/ui/order-form/store/MarketOrderFormStore.ts
```diff
@@ -123,7 +126,9 @@ export class MarketOrderFormStore {
   };
 
   get quoteInput(): string {
-    return this._quoteAssetInput;
+    return this._quoteAssetInput.length
+      ? round({ value: Number(this._quoteAssetInput), decimals: this._quoteAsset?.exponent ?? 6 })
+      : '';
   }
```
cronokirby, line 132: I think the store should keep the field as is, and the component should be responsible for displaying it.

apps/veil/src/pages/trade/ui/order-form/store/LimitOrderFormStore.ts
```diff
@@ -73,23 +73,29 @@ export class LimitOrderFormStore {
...
 
   setQuoteInput = (x: string) => {
     this._input.inputB = x;
   };
 
   get priceInput(): string {
-    return this._priceInput;
+    return this._priceInput.length
+      ? round({ value: Number(this._priceInput), decimals: this._quoteAsset?.exponent ?? 6 })
+      : '';
```
cronokirby, line 98: Same here ; we want to move presentation to the components displaying this information.

cronokirby (commented):
I think we should factor the logic from the model into the view, and we should create a little reusable component which can contain this logic, to avoid repetition.

JasonMHasperhoven:
Closing. Replacement PR here: https://github.com/penumbra-zone/web/pull/2238

PR #2228, opened 2025-04-09
fix(perspective): lqt plan to action view

VanishMax:
Prax wasn't rendering LQT Action View in approval dialog. That's how it renders it after the changes:

image


TalDerei:
@VanishMax we should fix the `delUM` getting cut off in the approval dialogue

PR #2229, opened 2025-04-09
rm custody service

turbocrime:
custody service implementation ~~focuses on validation. actual authorization logic is provided via context, so service host can be more flexible in implementation.~~ is removed


TalDerei:
> custody service implementation focuses on validation. actual authorization logic is provided via context, so service host can be more flexible in implementation.

validation in service, logic in context is a good seperation of concerns here

turbocrime:
implemented more validation and extensive tests of validation. prepares custody service to support more varied implementations

packages/services/src/custody-service/util/validate-request.ts
```diff
@@ -0,0 +1,132 @@
...
+    case 'spend':
+    case 'undelegate':
+    case 'undelegateClaim':
+    case 'validatorDefinition':
+    case 'validatorVote':
+      // no specific assertions
+      if (!Object.values(action.value).some(v => v != null)) {
+        throw new TypeError('Missing action plan', { cause: action });
+      }
+      return true;
```
turbocrime, line 126: a lot of plan validation will effectively be handled by the transaction builder, which will fail to build transactions that don't make sense.

so, specific validation is only applied to a few action types. most here are simply asserted to have some kind of content.

packages/services/src/custody-service/util/validate-response.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const assertValidAuthorizationData = (
+  actionCounts: Record<NonNullable<ActionPlan['action']['case']>, number>,
+  data: PlainMessage<AuthorizationData> = toPlainMessage(new AuthorizationData({})),
+): PlainMessage<AuthorizationData> => {
+  for (const field of Object.keys(authValidations) as (keyof PlainMessage<AuthorizationData>)[]) {
+    authValidations[field](field, data, actionCounts);
+  }
+  return data;
+};
```
turbocrime, line 62: a lot of auth validation will effectively be handled by the final transaction build step and the remote rpc, which will reject transactions that don't satisfy proofs.

packages/services/src/custody-service/util/validate-response.ts
```diff
@@ -0,0 +1,62 @@
...
+    const count = counts[actionType];
+
+    if (count < auths.length) {
+      throw new Error(`Unexpected ${actionType} authorization`);
+    } else if (count > auths.length || auths.some(s => isZeroBytes(s.inner))) {
+      throw new ReferenceError(`Missing ${actionType} authorization`);
+    } else if (auths.some(s => s.inner.length !== 64)) {
+      throw new TypeError(`Invalid ${actionType} authorization`);
+    }
+  };
```
turbocrime, line 45: a lot of auth validation will effectively be handled by the final transaction build step and the remote rpc, which will reject transactions that don't satisfy proofs.

so this just checks basic dimensions and counts expected auths needed for the plan.

TalDerei (approved):
discussed offline; looks good to me and can land for the next prax release with the refactored custody.

TalDerei:
this is also good to land after the prax ledger changes merge

Issue #2236, opened 2025-04-10
fix broken delegations


TalDerei:
following https://github.com/penumbra-zone/web/pull/2201, delegations and undelegations were initially functioning properly but now appear to be broken. Additionally, claims don't seem to have been tested and are erroring erroring due to an unimplemented `ValidatorStatus` rpc query service.

we should spend time fixing this today, cc @turbocrime 

-------------

https://github.com/user-attachments/assets/7ac4bcd0-ed2c-474b-85ad-4de7c1db2c2e

PR #2240, opened 2025-04-10
validator status method implementation

turbocrime:
## Description of Changes

stake service validator status method implementation somehow dropped from https://github.com/penumbra-zone/web/pull/2201 🙃

## Related Issue

https://github.com/penumbra-zone/web/issues/2236

### DIFF
--- a/packages/services/src/stake-service/validator-status.ts
+++ b/packages/services/src/stake-service/validator-status.ts
@@ -1,28 +1,22 @@
 import { Impl } from './index.js';
 import { servicesCtx } from '../ctx/prax.js';
-import { Code, ConnectError } from '@connectrpc/connect';
 import {
   GetValidatorInfoRequest,
   ValidatorInfo,
 } from '@penumbra-zone/protobuf/penumbra/core/component/stake/v1/stake_pb';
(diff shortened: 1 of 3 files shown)


packages/services/src/stake-service/validator-status.ts
```diff
@@ -0,0 +1,29 @@
...
+  const remoteValidatorInfo = querier.stake
+    .validatorInfo(new GetValidatorInfoRequest({ identityKey }))
+    .then(({ validatorInfo }) => validatorInfo);
+
+  const { status } = new ValidatorInfo(
+    // try to get latest, but suppress request error
+    (await remoteValidatorInfo.catch()) ??
+      // if unavailable, use local info
+      (await indexedDb.getValidatorInfo(identityKey)),
+  );
```
turbocrime, line 26: on second thought, this could probably make the local check first, since the block-processor should update validator info every epoch.

TalDerei (approved):
undelegation claims are working now 👍

TalDerei:
@erwanor delegation / undelegation issues seems to predate https://github.com/penumbra-zone/web/pull/2201 because nothing there would be effecting

TalDerei:
merging as this fixes the problem. improper epoch handling tracked in https://github.com/penumbra-zone/web/pull/2242

PR #2242, opened 2025-04-11
Improve idb epoch queries

turbocrime:
## Description of Changes

Improve storing and selecting epochs.

edit by Tal:
- performs db schema migration for `EPOCHS` table for preserving safety and make reasoning easier,
- `getEpochByHeight` performs a range query (is there a better approach here?), `getBlockHeightByEpoch` performs a direct key lookup.

## Related Issues

- [x] https://github.com/penumbra-zone/web/issues/2262
- [x] https://github.com/penumbra-zone/web/issues/2263
- [x] https://github.com/penumbra-zone/web/issues/2264
- [x] https://github.com/penumbra-zone/web/issues/2265

expected to fix failures observed in https://github.com/penumbra-zone/web/issues/2341 and https://github.com/penumbra-zone/web/issues/2236 as a side-effect.


packages/storage/src/indexed-db/index.ts
```diff
@@ -755,65 +758,113 @@ export class IndexedDb implements IndexedDbInterface {
...
-    for await (const cursor of this.db.transaction('EPOCHS', 'readonly').store) {
-      const currentEpoch = Epoch.fromJson(cursor.value);
+    if (findHeight > latestHeight) {
+      throw new Error('No epoch for future height', { cause: { latestHeight, findHeight } });
+    }
 
-      if (currentEpoch.startHeight <= height) {
-        epoch = currentEpoch;
-      } else if (currentEpoch.startHeight > height) {
-        break;
```
turbocrime, line 812: previously, this would break on any epoch newer than the queried height. since there are no order guarantees, this meant that the returned epoch was possibly older than the correct epoch, or possibly undefined even if the correct epoch was present.

TalDerei:
looks like LQT construction is also failing due to selecting a future epoch in https://github.com/penumbra-zone/web/pull/2191

----------

image_720

packages/storage/src/indexed-db/index.ts
```diff
@@ -755,65 +758,93 @@ export class IndexedDb implements IndexedDbInterface {
...
+   *
+   * Epoch duration is specified in appParameters, but the actual duration of
+   * any given epoch may vary. Calculations using that parameter must be treated
+   * as estimates.
    */
-  async addEpoch(startHeight: bigint): Promise<void> {
-    const cursor = await this.db.transaction('EPOCHS', 'readonly').store.openCursor(null, 'prev');
-    const previousEpoch = cursor?.value ? Epoch.fromJson(cursor.value) : undefined;
-    const index = previousEpoch?.index !== undefined ? previousEpoch.index + 1n : 0n;
+  async addEpoch(epoch: PlainMessage<Epoch>): Promise<void> {
```
turbocrime, line 770: since `epochIndex` is available on the compact block when this method is called, using heuristics to estimate an epoch index is unnecessary.

packages/storage/src/indexed-db/index.ts
turbocrime: changing this behavior seems to break some transaction planning.

the reasonable behavior of `getEpochByHeight` when no epoch is identified in storage is unclear, but i think locations calling `getEpochByHeight` should not rely on the return of an incorrect epoch from this method.

packages/storage/src/indexed-db/index.ts
```diff
@@ -755,65 +758,93 @@ export class IndexedDb implements IndexedDbInterface {
...
-  async getEpochByHeight(
-    /**
-     * The block height to query by. Will return the epoch with the largest
-     * start height smaller than `height` -- that is, the epoch that contains
-     * this height.
-     */
-    height: bigint,
-  ): Promise<Epoch | undefined> {
-    let epoch: Epoch | undefined;
+  async getEpochByHeight(findHeight: bigint): Promise<Epoch> {
```
turbocrime, line 792: this implementation will
- if a height after the start of the present epoch is queried,
    - if the block is synced, return the present epoch
    - if the block is in the future but within the expected duration of the present epoch, return the present epoch
    - throw.
- if a height before the start of the present epoch is queried,
    - identify the nearest-starting epoch.
        - if an epoch index immediately after the nearest-starting epoch is located, return the nearest-starting epoch.
        - throw.

turbocrime: various other implementation options:

- if the block height is in the future
    - if the future block height is expected to be within the present epoch, is it acceptable to return the present epoch?
    - if the future block height is not within the present epoch, should the future epoch index be estimated?
    - fail?
    - return undefined?
    - return the genesis epoch?
- if the block height is not in the future,
    - but is not yet synced,
        - estimate/fail/undefined/genesis?
    - but no epoch is identified,
        - estimate/fail/undefined/genesis?

turbocrime (commented):
some open questions and other problems were revealed by this work. will discuss in sync meeting.

turbocrime:
updated with comments and some initial wip tests of the new implementation i had sitting around. (the old tests relied on the incorrect behavior)

packages/storage/src/indexed-db/index.ts
TalDerei: why we weren't doing this in the first place is beyond me...this is an important change.

TalDerei: why we weren't doing this in the first place is beyond me...this is a super important improvement.

packages/storage/src/indexed-db/index.ts
```diff
@@ -1069,16 +1044,15 @@ export class IndexedDb implements IndexedDbInterface {
       return;
     }
 
-    const epoch =
-      (await this.getEpochByHeight(blockHeight)) ?? new Epoch({ startHeight: 0n, index: 0n });
```
TalDerei, line 1073: comment: need to think some more about this. I don't see a good technical reason for this fallback.

packages/storage/src/indexed-db/index.ts
```diff
@@ -159,7 +159,11 @@ export class IndexedDb implements IndexedDbInterface {
         }).createIndex('nullifier', 'nullifier.inner');
         db.createObjectStore('GAS_PRICES', { keyPath: 'assetId.inner' });
         db.createObjectStore('POSITIONS', { keyPath: 'id.inner' });
-        db.createObjectStore('EPOCHS', { autoIncrement: true });
+
+        /** @todo migrate to { autoIncrement: false }, and use Number(epoch.index) as key */
```
TalDerei, line 163: comment: remove comment

packages/storage/src/indexed-db/index.ts
```diff
@@ -820,83 +824,54 @@ export class IndexedDb implements IndexedDbInterface {
...
-    }
-
-    const newEpoch = {
-      startHeight: startHeight.toString(),
-      index: index.toString(),
-    };
+    // todo: relax the strict invariant for the very first write: when table is
+    // empty, allow starting from any index (fresh wallet skip sync), and then
+    // enforce monotonicity.
+    // if (epochCount !== Number(epoch.index)) { ... }
```
TalDerei, line 842: comment: this restriction prevented seeding the database from any epoch height because it failed the conditional. we should think of another check here and maybe consider a `put`? @turbocrime

turbocrime: the conditional helps enforce the implicit assumption that all historical epochs are available, up to the present epoch.

to maintain that assumption, when initializing the database from a non-zero epoch, all historical epochs should be populated. otherwise, many other things are likely to fail.

packages/storage/src/indexed-db/index.ts
turbocrime: the most correct approach would be to throw if no epoch is found.

turbocrime: also, the epoch index relevant within `addNewNotes` should be available on the compact block. pass a new parameter here.

packages/types/src/indexed-db.ts
```diff
@@ -331,8 +331,11 @@ export interface PenumbraDb extends DBSchema {
     value: PositionRecord;
   };
   EPOCHS: {
-    key: number; // auto-increment
+    key: number; // key path is 'index'
     value: Jsonified<Epoch>;
+    indexes: {
+      startHeight: number;
```
turbocrime, line 337: startHeight on a jsonified epoch will be a string, not a number

TalDerei: we're explicitly [casting](https://github.com/penumbra-zone/web/pull/2242/files#diff-8298d111bc94531ec943d5f024a9ab57657ad28670caea51e3aa9eda88bf3954R844-R846) to a number, but the broader question is: should we really be storing the full `Jsonified<Epoch>`, or just a plain object with numeric types instead? the original concern came up when trying to store the genesis block epoch `{ index: 0n, startHeight: 0n }`, and it was complaining treating it as Jsonified<Epoch> in that case for some reason.

packages/storage/src/indexed-db/index.ts
TalDerei: > the conditional helps enforce the implicit assumption that all historical epochs are available, up to the present epoch.

I don't think we need to enforce the conditional that all historical epochs are present for fresh wallets using quick sync, because the whole point is that they can operate on the chain without needing that historical state. What else could realistically fail in that context? I agree we should guard against duplicates, but supporting historical epochs is likely extraneous unless we can identify a clear failure mode.

packages/storage/src/indexed-db/index.ts
TalDerei: yeah, good catch

TalDerei: `getEpochByHeight` is querying the full epoch which includes the associated starting height, so querying by epoch index alone will require an additional lookup to retrieve the start height. The epoch should always be available from storage, but I’ve added a condition to throw just in case. We can eventually go throw that file and optimize the method calls

packages/types/src/indexed-db.ts
TalDerei:
```ts
  EPOCHS: {
    key: number; // key path is 'index'
    value: { index: number; startHeight: number };
    indexes: {
      startHeight: number;
    };
  };
```

```ts
  async addEpoch(epoch: PlainMessage<Epoch>): Promise<void> {
    ...

    await tx.store.add({
      index: Number(epoch.index),
      startHeight: Number(epoch.startHeight),
    });
  }
  ```
  
  this is the modified construction.

packages/storage/src/indexed-db/index.ts
TalDerei: opted for `put` here with idempotent behavior in the presence of duplicate keys:

TalDerei:
@turbocrime ready for another round of rereview

TalDerei:
closing as unplanned

Issue #2065, opened 2025-02-19
roll over notes for delegation and LQT voting


TalDerei:
references follow-up work from https://github.com/penumbra-zone/web/pull/2041#issuecomment-2660364791 to reduce the number of voting actions during transaction construction and prevent nullifier likability.

PR #2255, opened 2025-04-12
LQT: prevent nullifier linkability 

TalDerei:
## Description of Changes

roll over delegation notes in the wasm planner for LQT actions

## Related Issue

references https://github.com/penumbra-zone/web/issues/2065

### DIFF
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -706,6 +706,11 @@ pub async fn plan_transaction_inner<Db: Database>(
 
         // By setting the change address, all of the excess balance we've created
         // from spending the notes will be directed back to our account.
+        //
+        // Note: currently, we default to using the same account address as the reward recipient
+        // in the transaction planner request. If we decide to allow users to set the reward recipient
+        // address dynamically, we’ll need to detect when that field is set and bypass the change address
+        // mechanism accordingly.
         change_address = fvk.ephemeral_address(OsRng, Default::default()).0;
     }
 
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -690,13 +690,23 @@ pub async fn plan_transaction_inner<Db: Database>(
                 &mut OsRng,
                 incentivized,
                 rewards_recipient,
-                note,
+                note.clone(),
                 note_position,
                 start_position,
             );
 
             actions_list.push(vote_plan);
+
+            // We also want to go ahead and do the consolidation thing,
+            // to reduce the number of votes we need in the next epoch.
+            // To do so, we need to spend all of these notes, and produce one output per
+            // delegator token.
+            actions_list.push(SpendPlan::new(&mut OsRng, note, note_position));
         }
+
+        // By setting the change address, all of the excess balance we've created
+        // from spending the notes will be directed back to our account.
+        change_address = fvk.ephemeral_address(OsRng, Default::default()).0;
     }
 
     // Phase 2: balance the transaction with information from the view service.
--- a/packages/wasm/crate/src/planner.rs
+++ b/packages/wasm/crate/src/planner.rs
@@ -684,6 +684,8 @@ pub async fn plan_transaction_inner<Db: Database>(
                 .ok_or_else(|| anyhow!("missing note in liquidity tournament"))?
                 .try_into()?;
             let note_position: Position = staked_note.position.into();
+
+            // Domain type conversion: `Position(Tree { epoch: EPOCH_INDEX, block: 0, commitment: 0 })`
             let start_position: Position = Position::from((epoch_index as u16, 0, 0));
 
             let vote_plan = ActionLiquidityTournamentVotePlan::new(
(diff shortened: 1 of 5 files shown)


packages/wasm/crate/src/planner.rs
```diff
@@ -690,13 +690,28 @@ pub async fn plan_transaction_inner<Db: Database>(
...
         }
+
+        // By setting the change address, all of the excess balance we've created
+        // from spending the notes will be directed back to our account.
+        //
+        // Note: currently, we default to using the same account address as the reward recipient
+        // in the transaction planner request. If we decide to allow users to set the reward recipient
+        // address dynamically, we’ll need to detect when that field is set and bypass the change address
+        // mechanism accordingly.
+        change_address = fvk.ephemeral_address(OsRng, Default::default()).0;
```
TalDerei, line 714: derive a random ephemeral address (IBC deposit address)

erwanor: this is the change address though, why does this control where the rewards are minted? why aren't rewards minted to the account that owns the notes we are voting with?

packages/wasm/crate/src/planner.rs
```diff
@@ -690,13 +690,28 @@ pub async fn plan_transaction_inner<Db: Database>(
...
                 rewards_recipient,
-                note,
+                note.clone(),
                 note_position,
                 start_position,
             );
 
             actions_list.push(vote_plan);
+
+            // We also want to go ahead and do the consolidation thing,
```
erwanor, line 700: nit: this comment isn't really clear, on one hand, we're trying to rollover notes AND we want to consolidate them for future voting, can we isolate how each of those things is achieved? "the consolidation thing" doesn't really add anything helpful for future readers.

packages/wasm/crate/src/planner.rs
cronokirby: Yeah, I don't think this has that effect ; the plan itself contains an address to direct rewards to. We should instead keep the change address as is here.

TalDerei: these changes are directly lifted from the pcli implementation https://github.com/penumbra-zone/penumbra/pull/5079/files#diff-2d0c5138b423c7a2b0122cedcf5bcd3052a78b521d9ef64d6cc89150b05b6c7aR103-R118

packages/wasm/crate/src/planner.rs
TalDerei: we can isolate them, but I was trying to avoid introducing an additional loop and repeating domain conversions

packages/wasm/crate/src/planner.rs
cronokirby: Right ; it's still true that setting the change address doesn't effect the minting of rewards, only where the notes being rolled over are sent to. As to why we should not set the change address to `fvk.ephemeral(Default)`:

1. The pcli implementation uses an ephemeral address unnecessarily. This is a mistake on my part, because I forgot that outputs are encrypted to the address you're outputting to, so there's no need to use an ephemeral address, and in fact you want to use a normal address to help with syncing (once we exploit dag sync + clues etc.)
2. Even worse, the pcli implementation doesn't respect accounts at all! The wasm planner currently does, and correctly sets the change address to be that of the account doing the planning. This diff would redirect notes used by voting in one account back to the 0 account, which we don't want.

erwanor: I think it will be helpful to pause and write down a short blurb about what we are trying to do here, just to reset our posture from "porting to pcli" to "amending the reference implementation of the penumbra wallet software". It will also be helpful for future readers of the code.

TalDerei: noted, will internalize the feedback and take a more principled and documentation-oriented approach to these kinds of changes moving forward. I'll flush this out today

TalDerei: "a randomized address to mint rewards to" was an accidental miscomment here

cronokirby:
For posterity: what's our thinking behind having this code be in the *planner*, versus having this being a strategy we call the planner with in the code that actually creates the vote plan for all the user's voting notes?

TalDerei:
> For posterity: what's our thinking behind having this code be in the _planner_, versus having this being a strategy we call the planner with in the code that actually creates the vote plan for all the user's voting notes?

good question; for some context, we have stringent safety invariants related to _spend-max_ behavior. 

certain [spend](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/planner.rs#L585) checks had to be enforced within the planner, since we can’t directly query the actions contained in the planner’s `ActionList` – it's a private field, and [exposing](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/action_list.rs#L23) it via a getter doesn't seem ideal. Importantly, [one such check](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/planner.rs#L601) ensures that a TPR trying to spend it's entire balance is _constructed_ with exactly _one_ spend request, to guard against a class of edge cases.

Anyway, this would be the limitation _only_ if we attempt to handle it by constructing spend plans directly at the request level, which wouldn't work because of the previously stated reasons. It doesn’t prevent what you’re suggesting because alternatively, we can append an output plan to achieve that either at the planner request level or during request construction. 

It currently happens at the planner level because I thought it provides better encapsulation – you only need to construct the voting plans, and the planner handles the rest. Unless there's a specific reason to separate concerns here, implementing it in the planner felt more idiomatic.

TalDerei:
would like to merge this if there aren't any objections. complicates more in-depth testing of the voting dialog wiring in https://github.com/penumbra-zone/web/pull/2191 due to action blowup during planning.

cronokirby (approved):
I think this looks good and well-scoped now. Given the current constraints around how the planner is used, having this logic live inside the planner is fine, especially given how narrow the places we might produce an LQT vote action are.

One thing we do step on a little, which we should be aware of towards the future, is how we do break composability of actions a little bit here, in that if you try and plan a transaction involving both voting and spending a note, things will break. Fine in this context, but a general thorn to avoid running into in the future.

packages/wasm/crate/src/planner.rs
```diff
@@ -684,18 +684,36 @@ pub async fn plan_transaction_inner<Db: Database>(
...
+            // - The spend plan spends the note, enabling us to *roll over* its value into new notes.
+            //
+            // The latter has two purposes:
+            // (1) Enables note *consolidation*: if a user has multiple staked notes of the same asset,
+            //     they will all be spent and reissued in a smaller number of outputs, reducing note fragmentation
+            //     and minimizing the number of inputs required for future votes.
+            // (2) Prevents nullifier linkability: the voting action doesn't require a spend, but it does reveal
+            //     information about the note, making repeated use across epochs linkable. To avoid this, you can
+            //     optionally spend the notes during voting, consolidating your voting power into a single, unlinkable
+            //     note—allowing future votes with fewer actions.
```
erwanor, line 714: Great!

erwanor (approved):

PR #2258, opened 2025-04-14
Atomic validator status update

turbocrime:
provide an idb method to atomically update all validator information at the epoch boundary

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -159,7 +159,10 @@ export class IndexedDb implements IndexedDbInterface {
         }).createIndex('nullifier', 'nullifier.inner');
         db.createObjectStore('GAS_PRICES', { keyPath: 'assetId.inner' });
         db.createObjectStore('POSITIONS', { keyPath: 'id.inner' });
+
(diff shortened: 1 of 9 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -865,19 +871,54 @@ export class IndexedDb implements IndexedDbInterface {
...
-   */
-  async upsertValidatorInfo(validatorInfo: ValidatorInfo): Promise<void> {
-    // bech32m conversion asserts length
-    const identityKeyAsBech32 = bech32mIdentityKey(getIdentityKeyFromValidatorInfo(validatorInfo));
+  async updateValidatorInfo(
+    validatorInfoResponses: AsyncIterable<ValidatorInfoResponse>,
+  ): Promise<void> {
+    const validatorInfos = (await Array.fromAsync(validatorInfoResponses))
+      .map(({ validatorInfo }) => validatorInfo)
+      .filter(x => x != null);
```
turbocrime, line 879: this accumulates the entire query, before running the idb transaction. unfortunately this is necessary because async operations cannot be performed during an idb transaction.

TalDerei:
closing in favor of simplification in https://github.com/penumbra-zone/web/pull/2360

Issue #2127, opened 2025-03-20
onboarding guide should integrate the skip widget API


TalDerei:
"shield your assets" card should link to skip deposits

-------------------------

<img width="785" alt="Image" src="https://github.com/user-attachments/assets/75cc0921-981c-41c4-9b2d-355654ad0fec" />

JasonMHasperhoven:
This is done with the update on https://github.com/orgs/penumbra-zone/projects/20/views/23?pane=issue&itemId=101112296&issue=penumbra-zone%7Cdex-explorer%7C387

JasonMHasperhoven:
Accidentally moved it, assigning it to @vacekj since he's working on it

PR #2259, opened 2025-04-14
feat: deposits and withdrawals

### LINKED ISSUE #2128: deposit / withdraw actions on the portfolio assets card
Note 30.3.2025:
This PR would normally soft-depend on the expanded row feature, but I am choosing a simpler version of just implementing deposit/withdraw buttons on the normal row, as per screenshot below:

Image

skip deposits are relatively lightweight, whereby integrating the skip widget API is only few lines of code. Lifting the withdrawal functionality from minifront will be a bit more involved but still straightforward if we use the existing code as a guide.

-----------

Image

vacekj:
Fixes #2128 #2127 

includes:
- expanded row for both shielded and public assets
- improved styling, more aligned to figma
- deposits using the skip widget, with asset and chain prefilling for both origin and destination (USDC on Osmosis -> USDC on Penumbra)

Deliberately doesn't include withdrawal functionality, that is addressed by [this issue and branch](https://github.com/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+author%3A%40me+archived%3Afalse&issue=penumbra-zone%7Cweb%7C2257)


apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
```diff
@@ -0,0 +1,228 @@
...
+  };
+
+  return (
+    <Button onClick={handleClick} disabled={asset.shieldedBalances.length === 0}>
+      Unshield
+    </Button>
+  );
+}
+
+const theme = {
```
JasonMHasperhoven, line 55: I think it's better to link these colors by importing and using the vars of our theme.

JasonMHasperhoven (approved):
I will pre-emptively approve this to move things along but next time please split up PRs if you can. Everything related to the next config shouldve been a separate PR imo.

Also cc-ing @conorsch as it seems like the CI running doesnt have enough memory for the testing job.

VanishMax:
Will quickly add some UI bugs:

image

1. the alignment within an expanded row is off

2. there's no indication that a row can be expanded – no hover background or a pointer cursor

3. somehow a dialog with a Skip widget needs 2-∞ clicks to close

4. when using Skip widget, I entered the "connect Prax and set destination address" step. My Prax extension was locked and needed a log in, so an error couldn't let me go next. The problem is that this error only popped into the dev console, – should open a toast

Having this, I managed to deposit some Osmo using Skip widget, the core flow worked for me 👍

apps/veil/package.json
```diff
@@ -97,5 +99,8 @@
     "typescript": "5.5.3",
     "typescript-eslint": "^8.17.0",
     "vitest": "^1.6.0"
+  },
+  "engines": {
```
VanishMax, line 103: fyi: this field in Veil was causing GitHub actions to fail, so we removed it recently and did this: https://github.com/penumbra-zone/web/pull/2204

apps/veil/src/pages/portfolio/ui/asset-row.tsx
```diff
@@ -0,0 +1,290 @@
...
+              </TableCell>
+              <TableCell variant={'lastCell'}>
+                <div className='flex items-center gap-3 justify-between w-full'>
+                  <div className={'py-3'}>
+                    <ValueViewComponent
+                      valueView={bal.valueView}
+                      trailingZeros={false}
+                      priority={'tertiary'}
+                      density={'slim'}
+                      context={'table'}
```
VanishMax, line 195: question/nitpick: i wonder how did you start using this style of attribute syntax? I mean, didn't see you using it before, but here's clearly many attributes are written like this. Not all though, so it's a mix

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { useRegistry } from '@/shared/api/registry.ts';
+import { getMetadata } from '@penumbra-zone/getters/value-view';
+
+/**
+ * Computes the IBC denom hash for Penumbra-1
+ * If the denom already starts with "ibc/", it's assumed to be already computed.
+ * If the denom starts with "transfer/", it's assumed to be a full trace path.
+ * Otherwise, constructs the trace path using channelId and denom.
+ */
+async function computeIbcDenom(denom: string, channelId: string): Promise<string> {
```
VanishMax, line 16: question: i understand that this function hashes the denom with SHA-256 and that this hash later goes to the Skip widget as `defaultRoute.destAssetDenom` prop. Is it the recommended and required way to do so?

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
```diff
@@ -0,0 +1,228 @@
...
+  // Check if denom is already an IBC hash
+  if (denom.startsWith('ibc/')) {
+    return denom;
+  }
+  // Check if denom already is a full trace path
+  const ibcTraceStr = denom.startsWith('transfer/') ? denom : `transfer/${channelId}/${denom}`;
```
VanishMax, line 22: thought: many assets have a `/factory/` substring in them like here: https://github.com/prax-wallet/registry/blob/main/registry/chains/penumbra-1.json#L129

question: should we support it too? Won't the things break if we don't consider this `factory` thing?

eslint.config.js
```diff
@@ -16,4 +16,10 @@ export default [
       'tailwindcss/no-custom-classname': ['error', { callees: ['cn', 'cva'] }],
     },
   },
+  {
+    name: 'custom:no-unused-vars',
+    rules: {
+      'no-unused-vars': 'off',
```
VanishMax, line 22: question: why?

todo: start a discussion with a team about removing an eslint rule or bring it back. If we remove it now, I don't think we'll remember to put it back afterwards

VanishMax (commented):
Don't want to recommend to merge or block the PR, so leaving the comments only. The main flow worked for me, and there's also a room for improvement in terms of UI and code style

TalDerei:
> Don't want to recommend to merge or block the PR, so leaving the comments only. The main flow worked for me, and there's also a room for improvement in terms of UI and code style

@vacekj in favor of addressing that feedback before proceeding with merge

eslint.config.js
JasonMHasperhoven: I agree that this rule shouldn't be in here.

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
vacekj: Done

apps/veil/package.json
vacekj: Great, it was causing painful issues with pnpm for me as well. This is a rebase artifact.

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
vacekj: Good catch - this is one of the sources of complexity for deposits and withdrawals. I will support factory as well, wondering if there are already utilities for this somewhere - either in our repos, or in 3rd party cosmos projects.

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
vacekj: It is required if we want to prefill the route in Skip, didn't find a way to do this better - if you or anyone else has improvements, i'm all for them.

apps/veil/src/pages/portfolio/ui/asset-row.tsx
vacekj: I write some of my code with AI, so maybe that's why. Is there a preferred style you would prefer I use?

.npmrc
```diff
@@ -5,5 +5,5 @@ enable-pre-post-scripts=true
 
 # node-linker=hoisted
 strict-peer-dependencies=false
-engine-strict=true
+# engine-strict=true
```
vacekj, line 8: I disabled this because, even after several reinstalls using different methods, pnpm fails to detect my node version, doesn't matter if i use `pnpm env use`, homebrew node, manual node install, it just fails. according to some issues in pnpm, this seems to be a long-standing bug.

```shell
❯ pnpm i
Scope: all 23 workspace projects
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your Node version is incompatible with "@babel/runtime@7.27.0".

Expected version: >=6.9.0
Got: false

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.
```

apps/veil/src/pages/portfolio/ui/asset-row.tsx
VanishMax: the preferred style is any style that's consistent throughout the codebase. I'd say curly braces are only needed when you want to pass a variable in them

eslint.config.js
VanishMax: how often do you see it with the rule enabled?

apps/veil/src/pages/portfolio/ui/shield-unshield.tsx
VanishMax: don't mind if it's required, could be nice to add a comment with a link to the docs recommending it

vacekj: Done

eslint.config.js
vacekj: It happened for some files in veil

VanishMax: are you bringing the rule back? Maybe disable eslint for some specific lines?

vacekj:
Diagnosing a memory leak in the veil build

vacekj:
Fixed the OOM by some tree-shaking, now found another bug with public balances displaying as 0 in the summed amount. Investigating.

Issue #2268, opened 2025-04-17
LQT: ViewService


VanishMax:
Implement a ViewService method called `lqt-rewards` that will return all LQT rewards for a user by default.

Additionally, it should accept these filters:
1. `accountFilter` (naming is similar to `lqt-notes`) – subaccount filter
2. `epochFilter`

This method is needed to let users know how much rewards they earned for tournament voting. Otherwise, this info is not accessible from anywhere

VanishMax:
Closing because ViewService's `tournamentVotes` method already exists and implemented

PR #2269, opened 2025-04-17
feat(veil): #2249: previous epochs data wiring

### LINKED ISSUE #2249: LQT: Wire up previous epochs with real data
https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/api/use-previous-epochs.ts needs to be linked up with pindexer to use actual data.

VanishMax:
Closes #2249 

How to test:
1. Make sure your Prax is up-to-date and points to Phobos 3 testnet
2. Change `.env` `PENUMBRA_CHAIN_ID` to `penumbra-testnet-phobos-3`

This PR implements previous epochs table with both Pindexer data and asynchronously-loading personal rewards coming from ViewService, if the user is connected.

Result in one screenshot:

image

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
+++ b/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
@@ -6,6 +6,9 @@ import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys
 
 /**
  * Retrieves voting rewards from the ViewService.
+ *
+ * TODO: update rewards calculation based on this comment:
+ * https://github.com/penumbra-zone/web/pull/2269#pullrequestreview-2780489876
  */
 export const usePersonalRewards = (epoch?: number) => {
   const { connected, subaccount } = connectionStore;
--- a/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
+++ b/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
@@ -1,0 +1,27 @@
+import { useQuery } from '@tanstack/react-query';
+import { connectionStore } from '@/shared/model/connection';
+import { penumbra } from '@/shared/const/penumbra';
+import { ViewService } from '@penumbra-zone/protobuf';
+import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+
+/**
+ * Retrieves voting rewards from the ViewService.
+ */
+export const usePersonalRewards = (epoch?: number) => {
+  const { connected, subaccount } = connectionStore;
+
+  return useQuery({
+    queryKey: ['personal-rewards', subaccount, epoch],
+    enabled: connected && !!epoch,
+    queryFn: async () => {
+      const res = await penumbra.service(ViewService).tournamentVotes({
+        epochIndex: epoch ? BigInt(epoch) : undefined,
+        accountFilter: new AddressIndex({
+          account: subaccount,
+        }),
+      });
+
+      return res.votes;
+    },
+  });
+};
--- a/apps/veil/src/pages/tournament/server/previous-epochs.ts
+++ b/apps/veil/src/pages/tournament/server/previous-epochs.ts
@@ -1,0 +1,144 @@
+import { sql } from 'kysely';
+import { NextRequest, NextResponse } from 'next/server';
+import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { addressFromBech32m, isAddress } from '@penumbra-zone/bech32m/penumbra';
+import { Address } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import { serialize, Serialized } from '@/shared/utils/serializer';
+import { FinishedEpochs, Votes } from '@/shared/database/schema';
+import { pindexerDb } from '@/shared/database/client';
+import { uint8ArrayToBase64 } from '@penumbra-zone/types/base64';
+import { uint8ArrayToHex } from '@penumbra-zone/types/hex';
+
+const SORT_KEYS = ['epoch', 'lpReward', 'votingReward', ''] as const;
+export type PreviousEpochsSortKey = (typeof SORT_KEYS)[number];
+
+const DIRECTIONS = ['asc', 'desc', ''] as const;
+export type PreviousEpochsSortDirection = (typeof DIRECTIONS)[number];
+
+export interface PreviousEpochsRequest {
+  limit: number;
+  page: number;
+  address?: string;
+  sortKey: PreviousEpochsSortKey;
+  sortDirection: PreviousEpochsSortDirection;
+}
+
+interface AggregatedVotes extends Votes {
+  reward: number;
+}
+
+export type PreviousEpochsApiResponse = FinishedEpochs[];
+
+const DEFAULT_LIMIT = 10;
+
+export const getQueryParams = (req: NextRequest): PreviousEpochsRequest => {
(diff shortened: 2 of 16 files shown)


apps/veil/src/pages/tournament/server/previous-epochs.ts
```diff
@@ -0,0 +1,147 @@
...
+    limit,
+    page,
+    sortKey,
+    sortDirection,
+  };
+};
+
+const previousEpochsQuery = async ({ limit, page, sortDirection }: PreviousEpochsRequest) => {
+  // 1. Take the 'gauge' table and sort it by 'epoch' and then by 'portion', map asset_id to base64
+  const sortedGauge = pindexerDb
```
VanishMax, line 67: note: using `pindexerDb` variable instead of `pindexer` class to move away from this class as it become too big and needs splitting into smaller files. Also the DB usage is now placed right where it needs to be used.

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
```diff
@@ -29,17 +30,19 @@ const TABLE_CLASSES = {
 export const PreviousEpochs = observer(() => {
   const [page, setPage] = useState(BASE_PAGE);
   const [limit, setLimit] = useState(BASE_LIMIT);
-  const { getTableHeader, sortBy } = useSortableTableHeaders<keyof Required<EpochVote>['sort']>();
+  const { getTableHeader, sortBy } = useSortableTableHeaders<'epoch'>();
 
   const { connected } = connectionStore;
   const tableKey = connected ? 'connected' : 'default';
 
+  // TODO: use ViewService to request delegator rewards for connected users
```
VanishMax, line 38: follow-up work: #2268

apps/veil/src/pages/tournament/server/previous-epochs.ts
cronokirby: Yeah, I think the pindexer class is superfluous, the individual api routes can be responsible for making their own queries.

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
VanishMax: implemented in the latest commits

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
```diff
@@ -1,45 +1,144 @@
...
+
+    const asset = getMetadata(reward.reward.assetId);
+
+    const value = asset
+      ? new ValueView({
+          valueView: {
+            case: 'knownAssetId',
+            value: {
+              amount: reward.reward.amount,
+              metadata: getMetadata(reward.reward.assetId),
```
VanishMax, line 58: @TalDerei I see that `tournamentVotes` ViewService's method takes `stakingTokenAssetId` for the reward field: https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/tournament-votes.ts#L32.
 
As a result, this table uses `getMetadata` to transform assetId to Metadata, and it turns out to be UM token. Question: is it correct? Shouldn't it be a delUM token?

TalDerei: > @TalDerei I see that `tournamentVotes` ViewService's method takes `stakingTokenAssetId` for the reward field: https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/tournament-votes.ts#L32.

technically that's correct — the reward values are released from the community pool as UM, but they're distributed in the staking token, `delUM`. This gets to the core question of whether we want to display rewards in UM or delUM.

Currently, the `LQT_HISTORICAL_VOTES` table doesn't store any information about which delegation token the reward is associated with. If we want to perform the estimated cumulative conversion from delUM to UM, as we currently do in minifront's assets list, we'll need that information. If we instead choose to display rewards directly in delUM, we can simply hardcode the delUM symbol in the frontend. 

I’m currently reworking `tournamentVotes`, so I’ll make a note of it there.

erwanor: i think we should store the validator metadata either way so we can do the conversion back and forth on a whim

VanishMax: hardcoded a delUM in this specific case since we're sure it's delUM and don't really care about which one

VanishMax: That's not from the database though. And as Gar mentioned, if rewards are sent in delUM, it should be displayed in delUM to avoid confusion

cronokirby: If we do that, we should have several records per epoch, with each reward received in its respective delegated um amount since users can receive rewards with a different delUM denom, and you can't just sum these.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -0,0 +1,27 @@
...
+    enabled: connected && !!epoch,
+    queryFn: async () => {
+      const res = await penumbra.service(ViewService).tournamentVotes({
+        epochIndex: epoch ? BigInt(epoch) : undefined,
+        accountFilter: new AddressIndex({
+          account: subaccount,
+        }),
+      });
+
+      return res.votes;
```
cronokirby, line 24: I think we want the level of abstraction here to directly return rewards, rather than returning votes (of which another hook exists, I think?) and then having callers have to get rewards

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
```diff
@@ -11,35 +12,120 @@ import { Density } from '@penumbra-zone/ui/Density';
...
+  isLoading: boolean;
+  className: string;
+  connected: boolean;
+}
+
+const PreviousEpochsRow = ({ row, isLoading, className, connected }: PreviousEpochsRowProps) => {
+  const { data: rewards, isLoading: rewardsLoading } = usePersonalRewards(row.epoch);
+
+  const votingRewards = useMemo(() => {
+    const reward = rewards?.[0];
```
cronokirby, line 44: Not sure if this will work once we have different rewards for different votes based on the delegation asset.

cronokirby (commented):
I think https://github.com/prax-wallet/prax/pull/328 will break the calculation of personal rewards here :thinking:

TalDerei:
> I think [prax-wallet/prax#328](https://github.com/prax-wallet/prax/pull/328) will break the calculation of personal rewards here 🤔

this definitely affects the calculation, and I've messaged @VanishMax about the downstream side effects from the block processor logic changes. I'll take over the PR to modify it accordingly

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
TalDerei: agreed, hook should return accumulated rewards

TalDerei (approved):
as pointed out, https://github.com/prax-wallet/prax/pull/328 and https://github.com/penumbra-zone/web/pull/2272 will break reward calculations – the latter due to the generalization of the `tournamentVotes` implementation. Previously, the view server request only accepted an epoch index and returned all votes for that specific epoch. It now uses a streaming rpc that supports both epoch index and block height inputs, returning an async iterator of votes up to the epoch associated with the provided block height.

suggest adding comments to flag the follow-up work here, merging this PR as-is, and then handling those changes directly in #2272. flagging necessary follow-up work:
* create a separate hook within usePersonalRewards to handle proper reward accumulation, 
* convert delUM into UM: determine the conversion rate at the given epoch to convert each reward amount into UM

VanishMax:
Added several todo comments mentioning this discussion. Agree that we can revise it later after Prax updates

PR #2270, opened 2025-04-17
feat(veil): #2247: delegator leaderboard wiring

### LINKED ISSUE #2247: LQT: Delegator Leaderboard with real data
https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/api/use-delegator-leaderboard.ts should be augmented to use real data instead of dummy data, connecting to pindexer.

VanishMax:
Closes #2247 

Requests data to delegators leaderboard and, if user is connected to Prax, checks asynchorously, if some of the addresses in the table belongs to them.

image


apps/veil/src/pages/tournament/api/use-index-by-address.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Address } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { uint8ArrayToBase64 } from '@penumbra-zone/types/base64';
+import { ViewService } from '@penumbra-zone/protobuf';
+import { connectionStore } from '@/shared/model/connection';
+import { penumbra } from '@/shared/const/penumbra';
+
+/**
+ * Takes and address and checks if it belongs to user
+ */
+export const useIndexByAddress = (address?: Address) => {
```
cronokirby, line 11: This implementation is ok for now, but we should avoid checking each address in the leaderboard, and instead use the fact that we know the address we've voted with, fetch that, and check if it's among that set. We can then revalidate that query every epoch or something.

cronokirby (approved):
Looks good, we should create an issue to revisit the method we use to display addresses though.

apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
```diff
@@ -1,50 +1,155 @@
...
+                  exponent: 6,
+                },
+              ],
+              display: RANDOM_DELUM_DENOM,
+              symbol: 'delUM',
+            },
+          },
+        },
+      });
+    }, [loading, row.total_rewards]);
```
TalDerei, line 86: comment: out of the scope here, but let's add a comment to remove the hardcoded delUM value view. we'll replace it with a separate helper for converting delUM to UM, similar to how we display delegation token equivalents on the staking page using the staking token as the numeraire.

TalDerei (approved):

TalDerei:
nice work 👍

Issue #2251, opened 2025-04-11
LQT: wire up voting rewards with real data


cronokirby:
https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/api/use-voting-rewards.ts should be wired up with real data from pindexer.

PR #2272, opened 2025-04-20
feat(LQT): wire up voting rewards

TalDerei:
## Description of Changes

rework tournament rewards logic

- Implemented `usePersonalRewards` hook: fetches the user's votes (and rewards) from the view service for every epoch they participated in. Implemented `usePersonalRewardsByEpoch` hook:  fetches the user’s vote for a specific epoch (ie. to be used by the social card, cc @JasonMHasperhoven ). Both hooks query the pindexer `delegate_history` table and then filter rewards locally by the user’s address. We use pindexer instead of the local reward aggregation logic because it relies on _historical_ validator exchange rates, which aren’t available via RPCs (it only exposes _current_ rates). To access historical rates, we have to instead query pindexer.
- Refactored `tournamentVotes` view server method: support the `blockHeight` field, enabling iteration through and returning all votes, with internal calls to the `iterateLQTVotes` IDB method for streaming votes. 
- Extends reward logic to delegator leaderboard and previous epochs tables.
- FSD: moved server calls into shared folder, moved transaction building into entities folder.

pairs with https://github.com/prax-wallet/prax/pull/328

-------------------------

Screenshot 2025-04-24 at 11 11 21 PM

## Related Issue

references https://github.com/penumbra-zone/web/issues/2251


apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -1,30 +1,69 @@
...
+  );
+
+  const groupedRewards = await aggregateRewardsByEpoch(accountFilter, { votes: responses });
+
+  return groupedRewards;
+};
+
+// Retrieves voting rewards from the view service.
+export const usePersonalRewards = (subaccount?: number) => {
+  // TODO: pass the epoch index from the `useCurrentEpoch` hook into the query key
```
TalDerei, line 53: comment: TODO – ingest `useCurrentEpoch` hook

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+      if (!matched) {
+        console.warn(`No matching delegation for assetId: ${assetIdHex}`);
+        continue;
+      }
+
+      // Convert the vote amount using the validator’s exchange rate (with 8 decimal places)
+      const amount = vote.amount;
+      const rate = matched.validatorInfo.rateData?.validatorExchangeRate;
+      const rateVid = pnum(rate, { exponent: 8 }).toNumber();
```
TalDerei, line 105: comment: we should double check this conversion is correct. I avoided replicating the estimation machinery found from our [balances view server method](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/balances.ts#L35), which internally calls `BalancesAggregator` to supply delegation token equivalents on the staking page using the staking token as the numeraire. 

It sparsely references https://github.com/penumbra-zone/web/pull/816, which actually performs this conversion slightly differently by calling [this.indexedDb.getPricesForAsset](https://github.com/penumbra-zone/web/blob/main/packages/services/src/view-service/balances.ts#L218-L223).

apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
```diff
@@ -23,13 +23,18 @@ import type {
...
+import { useStakingTokenMetadata } from '@/shared/api/registry';
 
+// TODO: extend generic delUM --> UM conversion to here.
 const LeaderboardRow = observer(
   ({ row, loading }: { row: DelegatorLeaderboardData; loading: boolean }) => {
     const { connected } = connectionStore;
     const { data: subaccountIndex, isLoading: indexLoading } = useIndexByAddress(row.address);
 
+    // TODO: Convert delUM to UM – we are currently using the staking token,
+    // but the conversion from delUM to UM has not been applied here yet.
```
TalDerei, line 35: comment: `usePersonalRewards` hasn't been applied to the delegator leaderboard yet.

packages/types/src/indexed-db.ts
```diff
@@ -191,17 +191,39 @@ export interface IndexedDbInterface {
...
+      epoch: string;
+      TransactionId: TransactionId;
+      AssetMetadata: Metadata;
+      VoteValue: Value;
+      RewardValue: Amount;
+      id: string | undefined;
+      subaccount?: number;
+    },
+    void
+  >;
```
TalDerei, line 226: comment: we should double check this doesn't require bumping IDB, which we should avoid.

apps/veil/src/pages/tournament/ui/my-rewards.tsx
```diff
@@ -20,6 +20,7 @@ import { VotingRewards } from './voting-rewards';
 export const MyRewards = observer(() => {
   const { connected } = connectionStore;
 
+  // TODO: accumulate total rewards and remove `useTotalRewards` API.
```
TalDerei, line 23: comment: TODO: total rewards is still hardcoded and should use `usePersonalRewards`

TalDerei:
with respect to the number of TODOs flagged here, perhaps this PR can serve as the scaffolding, with the extended reward wiring to other tables handled in a separate PR?

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
```diff
@@ -0,0 +1,143 @@
...
+  addressIndex?: AddressIndex,
+  data?: { votes: TournamentVotesResponse[] },
+): Promise<
+  {
+    epochIndex: bigint;
+    total: Amount;
+  }[]
+> {
+  const grouped = new Map<bigint, { amount: Amount; assetId: AssetId }[]>();
+  const seen = new Map<bigint, Set<string>>(); // Track unique reward keys per epoch
```
VanishMax, line 31: thought: after using `useCurrentEpoch`, we already made sure that `tournamentVotes` method filters by epoch correctly, and the result of this method are votes per one single epoch and subaccount. 

suggestion: remove grouping by epoch

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  // referencing https://github.com/penumbra-zone/web/pull/816, we multiply each delegation
+  // note by its corresponding validator’s exchange rate to estimate the conversion to UM.
+  //
+  // TODO: plit this into a separate utility that can be externally used by other tables,
+  // such as delegation and previous epochs.
+  for await (const response of penumbra.service(ViewService).delegationsByAddressIndex({
+    addressIndex,
+    filter: DelegationsByAddressIndexRequest_Filter.ALL,
+  })) {
```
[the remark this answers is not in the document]
VanishMax: alternatively, `delegationsByAddressIndex` can be called within one `Promise.all()` with `tournamentVotes`

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
```diff
@@ -0,0 +1,143 @@
...
+      if (!rate) {
+        console.warn(`No exchange rate for assetId: ${assetIdHex}`);
+        continue;
+      }
+
+      // Accumulate the converted amount for the epoch
+      const converted = multiplyAmountByNumber(amount, rateVid);
+      totalAmount = new Amount({
+        ...addAmounts(totalAmount, converted),
+      });
```
VanishMax, line 118: nitpick

```suggestion
      totalAmount = addAmounts(totalAmount, converted)
```

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
VanishMax: suggestion: get exponent from delUM token

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    for (const vote of votes) {
+      const assetIdHex = Buffer.from(vote.assetId.inner).toString('hex');
+
+      // Find the delegation that matches the vote's asset ID
+      const matched = delegationPairs.find(pair => {
+        const delAssetIdHex = Buffer.from(
+          (pair.delegation.valueView.value as ValueView_KnownAssetId).metadata?.penumbraAssetId
+            ?.inner ?? [],
+        ).toString('hex');
```
VanishMax, line 94: suggestion (code style): use `uint8ArrayToHex` for consistency

packages/services/src/view-service/tournament-votes.ts
```diff
@@ -7,17 +7,18 @@ import { servicesCtx } from '../ctx/prax.js';
...
+  //
+  // Yields a group of votes for a specific epoch: [vote1, vote2, ...]
+  if (req.epochIndex) {
+    const tournamentVote = new TournamentVotesResponse();
 
-  // Retrieve the vote cast in the liquidity tournament for the current epoch.
-  const tournamentVote = new TournamentVotesResponse();
-  if (epoch?.index) {
-    const votes = await indexedDb.getLQTHistoricalVotes(epoch.index);
+    const votes = await indexedDb.getLQTHistoricalVotes(req.epochIndex, req.accountFilter?.account);
```
VanishMax, line 21: note: added filtering by account here

VanishMax (commented):
Couldn't test the whole PR as Prax doesn't store reward values for me: https://github.com/prax-wallet/prax/pull/328#discussion_r2060041176

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
TalDerei: comment: the change you made in https://github.com/penumbra-zone/web/pull/2272/commits/b41f79850ae4b937c61c473f0c3b96d2c6ee92c0#diff-75d13929a00f691eba21a9c78996504aaf0dec332670d7ceea1b04400113278aL43, switching from `.tournamentVotes({ accountFilter, blockHeight: latestKnownBlockHeight })` to `.tournamentVotes({ accountFilter, epochIndex: BigInt(epoch) })` carries an important behavioral effect in view server. The `tournamentVotes` method actually supports both `blockHeight` and `epochIndex` as arguments:

- when called with a block height, it uses `iterateLQTVotes` IDB method under the hood to return all votes up to the corresponding epoch, grouping them by epoch before yielding,
- when called with an epoch index, it returns only the grouped votes for that specific epoch.

consequently, `usePersonalRewards` should be split into two distinct hooks:
- one for components like tables that require _all_ past rewards, supplying the `blockHeight`,
- another for the social card, which only needs rewards for the current epoch, supplying the `epochIndex`.

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
TalDerei: changed to `uint8ArrayToHex`

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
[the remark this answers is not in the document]
TalDerei: yeah, this will necessarily need to be split into a dedicated helper if it's going to be used by other components that need to perform the reward conversion to UM.

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
TalDerei: > thought: after using `useCurrentEpoch`, we already made sure that `tournamentVotes` method filters by epoch correctly, and the result of this method are votes per one single epoch and subaccount.

I think we want to avoid situations where we have to individually call `tournamentVotes` multiple times (once for each epoch), and in turn multiple storage calls.

apps/veil/src/pages/tournament/ui/my-rewards.tsx
TalDerei: @VanishMax do you have a suggested approach here?

apps/veil/src/pages/tournament/ui/voting-rewards.tsx
```diff
@@ -1,78 +1,74 @@
...
+
+                return (
+                  <div key={index} className='grid grid-cols-subgrid col-span-4'>
+                    <TableCell cell loading={isLoading}>
+                      Epoch #{epochIndex}
+                    </TableCell>
+
+                    <TableCell cell loading={isLoading}>
+                      -
+                    </TableCell>
```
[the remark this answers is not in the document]
cronokirby: If we go that route, why not just use the delegator history on pindexer to tell us what our rewards are?

apps/veil/src/pages/tournament/ui/my-rewards.tsx
cronokirby: Use the caching that react query can provide?

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
VanishMax: I see. We know an epoch in any case, to me it would sound better to have an epochFilter accepting an array of epoch indexes. Without it, the tables would need to fetch all epochs, possibly large data structure.

apps/veil/src/pages/tournament/ui/my-rewards.tsx
VanishMax: If the hook that tables use fetches all rewards, it can also calculate the total amount. I'd put it there

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
TalDerei: outdated; aggregation logic was removed in favor of calling pindexer directly for rewards

apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
TalDerei: applied hook to delegator leaderboard

apps/veil/src/pages/tournament/ui/my-rewards.tsx
TalDerei: takes advantage of `useQuery` cache

apps/veil/src/pages/tournament/ui/aggregate-rewards.tsx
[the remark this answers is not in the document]
TalDerei: outdated; aggregation logic was removed in favor of calling pindexer directly for rewards

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -1,30 +1,105 @@
...
+      penumbra.service(ViewService).tournamentVotes({
+        accountFilter,
+        blockHeight: epochOrHeight.value,
+      }),
+    );
+  }
+
+  const { address } = await penumbra
+    .service(ViewService)
+    .addressByIndex({ addressIndex: { account: accountFilter?.account } });
```
VanishMax, line 41: suggestion: move this async request into a promise.all along with the previous .tournamentVotes call

apps/veil/src/shared/api/server/tournament/summary.ts
VanishMax, line 1: question: why are endpoint files moves to `shared` from `pages/tournament/server`? I don't see these files reused, so by the code structure it would be better placed next to the code using it

VanishMax:
Good work! Understanding the implementation, it makes sense to leave it as it. Just want to finish the product discussion and improve the UX of UM conversion rendering

TalDerei:
> Good work! Understanding the implementation, it makes sense to leave it as it. Just want to finish the product discussion and improve the UX of UM conversion rendering

I think we can follow-up in separate PR to avoid ballooning scope creep

VanishMax (approved):
Alright, it's good to have a decision of rendering estimated UM. Ready for a merge after lint fixes and maybe threads from previous review

Issue #2192, opened 2025-03-31
UI v2: Checkbox


VanishMax:
https://www.figma.com/design/QSiF2XSvHAWwzBJS5C3f4S/Penumbra-Web-UI?node-id=314-41839&t=pL1dM3HxO7dCkPcf-0

PR #2274, opened 2025-04-23
feat(veil): #2271: LQT landing card & voting dialog data wiring

### LINKED ISSUE #2271: LQT: wire up voting dialog with real data
Load voting dialog with real votable assets and make sure all voting states are covered in UI

VanishMax:
Closes #2271 

In this PR:
- New texts from Figma applied to the landing card
- Landing card now uses real data and updates
- Tracking of new epochs with an informative toast opening when new epoch starts
- Voting dialog now uses real data from the registry
- Voting dialog now distinguishes between the search screen and "select asset" screen

Follow up work: wait for Prax to land the new ViewService method with rewards and fixes to notes, update the `VotingInfo` to correctly display if a user voted in this epoch, and other states of this component. Maybe optimize the loading times.

https://github.com/user-attachments/assets/6fcdb688-8b6d-4b00-a608-3c5fefde6a8f


apps/veil/src/pages/tournament/ui/landing-card/index.tsx
```diff
@@ -1,64 +1,59 @@
...
+    limit: 1,
+    page: 1,
+  });
+
+  const { epoch, isLoading: epochLoading } = useCurrentEpoch(newEpoch => {
+    openToast({
+      type: 'info',
+      message: `New epoch has started: ${newEpoch}. Vote for your favorite asset!`,
+    });
+  });
```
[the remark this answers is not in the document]
JasonMHasperhoven: Nice

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+      setSelectedAsset(asset);
+    };
+
+    return (
+      <Dialog isOpen={isOpen} onClose={handleClose}>
+        <Dialog.Content
+          title={`Vote in epoch #${epoch}`}
+          headerChildren={
+            <>
+              {/* Focus catcher. If this button wouldn't exist, the focus would go to the first input, which is undesirable */}
```
JasonMHasperhoven, line 99: Did you try to wrap this in a div with tabIndex="-1"?

apps/veil/src/pages/tournament/ui/vote-dialog/vote-dialog-asset.tsx
```diff
@@ -0,0 +1,48 @@
...
+export interface VoteDialogAssetProps {
+  asset: MappedGauge;
+  onSelect: VoidFunction;
+}
+
+export const VoteDialogAsset = ({ asset, onSelect }: VoteDialogAssetProps) => {
+  const isSecondary = asset.portion < VOTING_THRESHOLD;
+
+  return (
+    <Dialog.RadioItem
```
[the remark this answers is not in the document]
JasonMHasperhoven: Improvements are always welcome

JasonMHasperhoven (approved):
Looks good to me :+1:

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -162,7 +164,7 @@ export const VotingInfo = observer(({ epoch }: { epoch: number }) => {
               <Coins className='w-full h-full' />
             </div>
             <Text variant='small' color='text.secondary'>
-              You’ve delegated UM and are now eligible to vote in this epoch.
+              You&#39;ve delegated UM and can now vote with your delUM in this epoch.
             </Text>
             {firstNoteValueView && <ValueViewComponent valueView={firstNoteValueView} />}
```
TalDerei, line 169: comment: can the `ValueView` component display only the delUM symbol?

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+      setSearchQuery('');
+      setIsSearchOpen(false);
+      setSelectedDenom(asset.asset.base);
+      setSelectedAsset(asset);
+    };
+
+    return (
+      <Dialog isOpen={isOpen} onClose={handleClose}>
+        <Dialog.Content
+          title={`Vote in epoch #${epoch}`}
```
TalDerei, line 96: suggestion: I'd push back on the premise that the default screen the user sees in the voting dialogue should be the blank search screen rather than showing the "select assets" view. This isn't intuitive – it requires the user to click into the search field just to see which assets are available to vote on. It also doesn’t match the Figma, where assets are displayed by default even if the threshold isn't reached or no asset is selected. Point being, something should be shown by default – I think a blank screen creates unnecessary confusion from a UX perspective.

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const { subaccount } = connectionStore;
+    const valueAddress = subaccounts?.find(
+      account => getAddressIndex(account).account === subaccount,
+    );
+
+    // Temporarily hardcode the same account address as the reward recipient.
+    const rewardsRecipient = valueAddress?.addressView.value?.address;
+
+    // Fetch user's spendable voting notes for this epoch
+    const { notes } = useLQTNotes(subaccount);
```
TalDerei, line 49: suggestion: since we're already puling the latest epoch from `useCurrentEpoch` which calls into pindexer, we can remove the epoch return field from the `useLQTNotes` hook. Let's also move the call to `useLQTNotes` below `const { epoch } = useCurrentEpoch()` and pass the epoch index as a parameter into `useLQTNotes` for reuse. The way we currently query the epoch index from the SCT service is not efficient.

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const [searchQuery, setSearchQuery] = useState('');
+
+    const { data: subaccounts } = useSubaccounts();
+
+    const { subaccount } = connectionStore;
+    const valueAddress = subaccounts?.find(
+      account => getAddressIndex(account).account === subaccount,
+    );
+
+    // Temporarily hardcode the same account address as the reward recipient.
```
TalDerei, line 45: suggestion: remove this comment – we don’t currently have a product requirement that allows users to select the reward recipient.

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -12,13 +12,15 @@ import { connectionStore } from '@/shared/model/connection';
...
+import { useCurrentEpoch } from '../api/use-current-epoch';
+import { VoteDialogueSelector } from './vote-dialog';
 import { checkIfAlreadyVoted } from '../api/vote';
 
 // TODO: use it for both landing and round page, apply all component states from Figma
-export const VotingInfo = observer(({ epoch }: { epoch: number }) => {
+export const VotingInfo = observer(() => {
   const { connected, subaccount } = connectionStore;
 
+  const { epoch = 0 } = useCurrentEpoch();
```
TalDerei, line 23: comment: why do we destructure the epoch like this?

TalDerei (approved):
left some feedback, great work 👍

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -162,7 +164,7 @@ export const VotingInfo = observer(({ epoch }: { epoch: number }) => {
               <Coins className='w-full h-full' />
             </div>
             <Text variant='small' color='text.secondary'>
-              You’ve delegated UM and are now eligible to vote in this epoch.
+              You&#39;ve delegated UM and can now vote with your delUM in this epoch.
```
TalDerei, line 167: comment: think we've had a regression somewhere where the landing and round cards are not enforcing the warning message: `You have already voted in this epoch. Come back next epoch to vote again.` after a vote has been submitted?

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+                    : 'Select asset to vote'}
+                </Button>
+
+                <Checkbox
+                  title='Reveal my vote to the leaderboard.'
+                  description='Voting each epoch grows your streak.'
+                  checked={revealVote}
+                  onChange={value => setRevealVote(value as boolean)}
+                />
+              </div>
```
TalDerei, line 144: comment: ACK – checkbox uses updated UI component (https://github.com/penumbra-zone/web/issues/2192)

apps/veil/src/pages/tournament/api/use-current-epoch.ts
```diff
@@ -0,0 +1,26 @@
...
+  const prevEpoch = useRef(epoch);
+
+  useEffect(() => {
+    if (epoch && !prevEpoch.current) {
+      prevEpoch.current = epoch;
+    } else if (epoch && prevEpoch.current !== epoch) {
+      prevEpoch.current = epoch;
+      onChange?.(epoch);
+    }
+  }, [epoch, onChange]);
```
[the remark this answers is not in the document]
VanishMax: Agree with the suggestion, created an issue: https://github.com/penumbra-zone/web/issues/2282

For context, the idea of this toast was to let users online and on page know why the data on the page has changed, because there aren't much of animation demonstrating the state change.

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: done! Displaying only 'delUM' without the validator id. Also added the notes aggregation

image

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
VanishMax: nice idea, implemented!

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: fixed

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: VotingInfo requires even more work, created an issue to follow up with this component: https://github.com/penumbra-zone/web/issues/2283

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
VanishMax: Waited for several days for Artur to implement empty screens in Figma but got no response, as he is OOO. Thanks to your feedback I found a simple way to make an empty screen – just showing "search results" with all assets now.

Next product question: if there's only 1 vote for let's say USDC, voting dialog will open with only one option to vote for – USDC. Users might ignore the "search asset" input and won't notice other options, thinking they can only vote for USDC. I'd rethink the UX in the future

image

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
VanishMax: Tabindex=-1 has a negative side-effect of the input not being focusable afterwards. This catcher might look weird in the code, but it does what it needs to do

VanishMax:
Thinking I addressed all PR feedback, thanks for it @JasonMHasperhoven @TalDerei. Created two follow-up issues #2283 and #2282.

PR #2278, opened 2025-04-24
feat(veil): #2277: round results data wiring

### LINKED ISSUE #2277: LQT: implement current results in the round page with real data
Fill this section with real data from pindexer + ViewService

Image

VanishMax:
Closes #2277

This PR focuses on the LQT round page:
1. In round card, reuses the `IncentivePool` component to render epoch summary data
2. In current results section, implements data request for a specific epoch's gauge data
3. Allows voting from current results section

Todo: implement "estimated incentive" column for this table. @cronokirby can we do it in the `gauge` view in Pindexer?

image


apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -52,82 +27,44 @@ const TABLE_CLASSES = {
...
+
+export const CurrentVotingResults = observer(({ epoch }: CurrentVotingResultsProps) => {
   const [page, setPage] = useState(1);
-  const [limit, setLimit] = useState(10);
-  const { getTableHeader, sortBy } = useSortableTableHeaders();
+  const [limit, setLimit] = useState(BASE_LIMIT);
+  const { getTableHeader, sortBy } = useSortableTableHeaders<EpochResultsSortKey>();
+
+  const { connected } = connectionStore;
+  const { data, isLoading } = useEpochResults({
```
TalDerei, line 40: requested change: `useEpochResults` isn't returning any voting data for me unless I navigate back to the landing card and then return. It doesn’t seem to re-render or update the vote data while remaining on the same page.

are we missing `useCurrentEpoch()` call before here? or possibly related to the static `!!params.epoch` param in the useQuery? there seems to be some stale state in `useCurrentEpoch`, where the hook is not updating the epoch. 

Screenshot 2025-04-24 at 5 50 40 PM

-------------

as a side-effect, if you remain on the page, the displayed epoch doesn't update, and the vote data remains stale even after multiple epochs have passed.

Screenshot 2025-04-24 at 5 50 52 PM

Screenshot 2025-04-24 at 5 58 49 PM

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -138,19 +75,30 @@ export const CurrentVotingResults = observer(() => {
         <Density compact>
           <div className={cn('grid h-auto overflow-auto', TABLE_CLASSES.table[tableKey])}>
             <div className={cn('grid grid-cols-subgrid', TABLE_CLASSES.row[tableKey])}>
-              {getTableHeader('symbol', 'Asset')}
-              {getTableHeader('gaugeValue', 'Percentage of Votes')}
+              <TableCell heading>Asset</TableCell>
+              {getTableHeader('portion', 'Percentage of Votes')}
               {getTableHeader('votes', 'Votes Cast')}
```
TalDerei, line 80: comment: should we rename this to "Casted Votes" based on Artur's suggestion? Also, it'd be useful to add a note that this field currently reflects an incorrect count of casted votes returned by pindexer.

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
```diff
@@ -148,18 +154,17 @@ export const VoteDialogueSelector = observer(
           <div className='flex flex-col pt-2'>
             {!isSearchOpen && (
               <VotingAssetSelector
-                value={selectedDenom}
                 selectedAsset={selectedAsset}
                 loading={isLoading}
-                gauge={assets ?? []}
+                gauge={assets?.data ?? []}
```
TalDerei, line 159: suggestion: If `isLoading === true` but `assets?.data` is still undefined, the component renders with []. The check should incorporate checking against `isLoading` flag: `gauge={!isLoading && assets?.data ? assets.data : undefined}`.

we should update this pattern in other places where we're relying on `isLoading` from `useQuery`.

apps/veil/src/pages/tournament/server/epoch-results.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      return {
+        asset,
+        epoch: item.epoch,
+        votes: item.votes,
+        portion: item.portion,
+        missing_votes: Number(item.missing_votes),
+      };
+    })
+    .filter<MappedGauge>(item => !!item);
```
TalDerei, line 132: comment: instead of map + filter, we can use the `flatMap` pattern to reduce verbosity.

TalDerei (changes requested):

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
VanishMax: That's a nice find! Took me several hours to figure out. Turns out, two queries had the same query name, but `useRefetchOnNewBlock` was picking only one of the queries, leaving the other one without updates. Fixed it and stopped voting dialog refetches if the dialog is closed.

As for the side effect, the idea is that this page should not jump into the latest block, but show historical data. Will add more "finished epoch" states later in "VotingInfo component update" issue.

apps/veil/src/pages/tournament/ui/round/page.tsx
```diff
@@ -24,7 +24,7 @@ export const TournamentRoundPage = () => {
     <section className='flex flex-col gap-6 p-4 max-w-[1168px] mx-auto'>
       <PenumbraWaves />
       <RoundCard epoch={epoch} />
-      <CurrentVotingResults />
+      <CurrentVotingResults epoch={epoch} />
       <LeaderboardTable startBlock={startBlock} endBlock={endBlock} />
```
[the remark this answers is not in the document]
VanishMax: Implemented this real data, but for now works only for the current unfinished epoch. Do we have a method to get block height by epoch?

apps/veil/src/pages/tournament/server/epoch-results.ts
VanishMax: Got excited about this idea at first, but it doesn't seem to work that way – flattening doesn't filter undefined values.

image

apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx
VanishMax: There's a check inside the component, `{!loading && !gaugeWithValue.length && <VotingDialogNoResults />}`, preventing this from happening

TalDerei (approved):
this is looking in a better state now, great work 👍

apps/veil/src/pages/tournament/api/use-round-block-info.ts
```diff
@@ -0,0 +1,75 @@
...
+        sctService.epochByHeight({ height: currentEpoch.epoch.startHeight + epochDifference + 1n }),
+        sctService.timestampByHeight({ height: currentEpoch.epoch.startHeight }),
+        sctService.timestampByHeight({ height: prevEpoch.epoch.startHeight }),
+      ]);
+
+      let nextEpochBlock = nextEpoch.epoch?.startHeight;
+      nextEpochBlock =
+        nextEpochBlock && nextEpochBlock > currentEpoch.epoch.startHeight
+          ? nextEpochBlock - 1n
+          : undefined;
```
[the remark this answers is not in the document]
VanishMax: Ideally, pindexer data would fix all issues possible with this approach. Added the comment to not forget

PR #2289, opened 2025-04-29
feat(veil): #2283: VotingInfo

### LINKED ISSUE #2283: LQT: wire VotingInfo with real data
VotingInfo component is used in the tournament landing card and tournament round page card – https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/voting-info.tsx

It has many states that need to be implemented, check with Figma for different states.

Reference: https://github.com/penumbra-zone/web/pull/2274#discussion_r2059406643

VanishMax:
Closes #2283

Implements all design states of the VotingInfo component for both tournament landing and round pages. Here are some new states in screenshots:

Landing page, voted:

image

Round page, epoch ended, voted

image

Round page, epoch ended, not voted

image


apps/veil/src/pages/tournament/api/use-voting-notes.ts
```diff
@@ -17,7 +17,7 @@ const fetchQuery = async (subaccount = 0, epoch: number): Promise<LqtVotingNotes
 /**
  * Must be used within the `observer` mobX HOC
  */
-export const useLQTNotes = (subaccount = 0, epoch?: number) => {
+export const useLQTNotes = (subaccount = 0, epoch?: number, disabled?: boolean) => {
```
TalDerei, line 20: comment: should this be `enabled: connectionStore.connected && !!epoch && !disabled`?

apps/veil/src/pages/tournament/api/use-tournament-votes.ts
```diff
@@ -0,0 +1,35 @@
+import { useQuery } from '@tanstack/react-query';
+import { ViewService } from '@penumbra-zone/protobuf/penumbra/view/v1/view_connect';
+import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { connectionStore } from '@/shared/model/connection';
+import { penumbra } from '@/shared/const/penumbra';
+import { useRefetchOnNewBlock } from '@/shared/api/compact-block';
+
+export const useTournamentVotes = (epoch?: number, disabled?: boolean) => {
+  const { connected, subaccount } = connectionStore;
+  const enabled = connected && !!epoch;
```
TalDerei, line 10: comment: should this be `enabled = connected && !!epoch && !disabled`?

apps/veil/src/pages/tournament/api/use-delegations.ts
```diff
@@ -0,0 +1,32 @@
...
+import { DelegationsByAddressIndexRequest_Filter } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+import { connectionStore } from '@/shared/model/connection';
+import { penumbra } from '@/shared/const/penumbra';
+import { useRefetchOnNewBlock } from '@/shared/api/compact-block';
+
+export const useAccountDelegations = (disabled?: boolean) => {
+  const { connected, subaccount } = connectionStore;
+
+  const query = useQuery<ValueView[]>({
+    enabled: connected,
```
TalDerei, line 14: comment: should this be `enabled: connected && !disabled`? does passing disabled here actually short-circuit the grpc query?

apps/veil/src/pages/tournament/api/use-tournament-votes.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+
+      const res = await Array.fromAsync(
+        penumbra.service(ViewService).tournamentVotes({
+          accountFilter: new AddressIndex({ account: subaccount }),
+          epochIndex: BigInt(epoch),
+        }),
+      );
+
+      return res[0]?.votes ?? [];
```
TalDerei, line 28: comment: Is there a more efficient way to consume the async stream? calling `tournamentVotes` returns an async stream, and wondering if we can short-circuit the stream by reading just the first yielded item, since as you correctly pointed out every vote will have the same incentivized asset field. 

see comment below, we may not even need to do this because we're already storing the incentivized asset metadata.

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -64,6 +73,89 @@ export const VotingInfo = observer(() => {
     });
   }, [getMetadata, notes]);
 
+  const votedFor = useMemo(() => {
+    const vote = votes?.[0];
+    const metadata = getMetadata(vote?.incentivizedAsset);
```
TalDerei, line 78: comment: IDB already stores the incentivized asset metadata for each vote – it was an oversight not to store just the asset ID, since the caller can query the metadata when needed: https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L524-L525

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -206,6 +292,34 @@ export const VotingInfo = observer(() => {
     );
   }
 
+  if (isDelegated) {
```
TalDerei, line 295: question: are we confident this state is actually still reachable? I wasn’t able to reproduce it after voting in the current epoch and then delegating afterward.

apps/veil/src/pages/tournament/api/use-delegations.ts
```diff
@@ -0,0 +1,32 @@
...
+    queryKey: ['delegations-by-address-index', subaccount],
+    staleTime: Infinity,
+    queryFn: async () => {
+      const res = await Array.fromAsync(
+        penumbra.service(ViewService).delegationsByAddressIndex({
+          addressIndex: new AddressIndex({ account: subaccount }),
+          filter: DelegationsByAddressIndexRequest_Filter.ALL_ACTIVE_WITH_NONZERO_BALANCES,
+        }),
+      );
```
TalDerei, line 23: comment: not convinced this is the optimal approach for checking  if delegation has already occurred. This call is relatively heavy. Since we're already calling `useLQTNotes`, wouldn't it be lighter to simply filter those notes (similar to `checkIfAlreadyVoted`) and verify whether all have already been used for voting?

TalDerei (commented):
flagged a few things that stood out to me, curious to hear your thoughts

rebase once https://github.com/penumbra-zone/web/pull/2291 lands to pass CI

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: uhm, I didn't expect this case to be possible, so covered only "if not voted but delegated" – your case seem to require more calculations. Not sure if `use-delegations` can cover that

apps/veil/src/pages/tournament/api/use-delegations.ts
VanishMax: i don't think so – `useLQTNotes` doesn't return **_not eligible_** notes for voting: https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L717-L722

VanishMax: i agree that this way is not efficient, but can't find another option

apps/veil/src/pages/tournament/api/use-voting-notes.ts
VanishMax: So the idea of distinction between `!enabled` and `disabled` here is that the first one is used in the query itself – it blocks only the initial call. `disabled`, on the other hand, is used in `useRefetchOnNewBlock`, which bypasses the `enabled` property of a query. In my tests, it is preferable to allow the initial query, but disable constant refetches.

VanishMax: In addition to that idea, I'd say that `disabled` argument was added to limit the amount of endpoint calls when we're sure there's no more fresh data on the server. This is crucial for all finished epochs that only render historical data.

apps/veil/src/pages/tournament/api/use-delegations.ts
VanishMax: Replied in a thread above

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: Created a PR in the protocol to support it: https://github.com/penumbra-zone/penumbra/pull/5185

apps/veil/src/pages/tournament/api/use-voting-notes.ts
TalDerei, line 20: suggestion: can we change this to `useLQTNotes = (subaccount: number)`

apps/veil/src/pages/tournament/ui/voting-info.tsx
TalDerei: one way we could test this is just transfer all delegation notes from main account to subaccount, and then try delegating again in the main account after epoch has started

TalDerei (approved):
good to merge after consuming https://github.com/penumbra-zone/penumbra/pull/5185 👍

apps/veil/src/pages/tournament/ui/voting-info.tsx
VanishMax: agree on the testing method, but it would also break the UX. Imagine you voted, then staked, and instead of "you voted for XXX", you see "you can't vote in this epoch". Don't really want to cover this case without design changes

apps/veil/src/pages/tournament/api/use-voting-notes.ts
TalDerei: yeah we should sanitize the code and explicitly expect the subaccount to be passed in as a field

apps/veil/src/pages/tournament/api/use-tournament-votes.ts
VanishMax: done

Issue #2290, opened 2025-04-29
ci: wasm-pack missing from turbo jobs


conorsch:
**Describe the bug**
Recently the Turbo CI jobs have started failing, complaining about `wasm-pack` missing. Here's an example failure job: https://github.com/penumbra-zone/web/actions/runs/14735714321/job/41362408719?pr=2289 from #2289. 

Not sure why it broke, but it broke, so let's fix it. As usual I'd prefer to spike on making the tool versions hardcoded somewhere, like in a nix flake dev shell. Will timebox that approach to prioritize speedy resolution.

conorsch:
>  As usual I'd prefer to spike on making the tool versions hardcoded somewhere, like in a nix flake dev shell.

Same spirit as https://github.com/penumbra-zone/web/commit/417e1b230a37928f401659549f8b5fc6368f2a67 although that commit suggests [mise](https://github.com/jdx/mise) and I'd prefer [nix](https://nixos.org/download/) to match the tooling in other repos. Please holler if you have a strong preference for mise, @turbocrime ; would love to hear more if so.

conorsch:
The changes in https://github.com/penumbra-zone/web/pull/2291 indeed got CI passing, but will need further refinements for new deps like https://github.com/penumbra-zone/web/pull/2256/files, specifically the failure in https://github.com/penumbra-zone/web/actions/runs/14766684259/job/41459518960?pr=2256.

conorsch:
We're good, will handle further refinements as necessary.

PR #2291, opened 2025-04-29
ci: use nix devshell for workflows

conorsch:
## Description of Changes

We want a single declaration of tooling dependencies required to work on
the apps in the web monorepo. Extending the nix flake so that it's used
in CI, as well, ensuring that CI failures are easily replicable.

There's a single omission, for the sake of velocity: I haven't ported
the playwright tests to use the nix env. That'll take another pass, but
punting on it for now, so that we have green CI again on the web
monorepo.

## Related Issue

Refs #2290.

No changes to app code, this is just CI-wrangling.

### DIFF
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -1,0 +1,28 @@
+name: 'Setup Nix Environment'
+description: 'Installs Nix and configures the cache'
+inputs:
+  primary-key:
+    description: Primary key for the cache
+    required: false
+    default: nix-${{ runner.os }}-${{ hashFiles('**/*.nix') }}
+  restore-prefix:
+    description: Restore prefix for the cache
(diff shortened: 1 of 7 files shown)


conorsch:
Calling this ready for review. Once it lands, other PRs can be rebased on top of it, to get green CI again. As noted in the PR submission text, I haven't ported the playwright deps yet; furthermore, the "release" workflow is still using the old github action helpers, but I'll update those when we're at a lower velocity.

TalDerei:
@conorsch general question, any idea why https://github.com/penumbra-zone/web/pull/2289 fails CI, but https://github.com/penumbra-zone/web/pull/2297 passes?

TalDerei (approved):
👍

VanishMax (approved):

Issue #2295, opened 2025-04-29
Bug: useAssets hook can result in stale data


cronokirby:
Right now, the `useAssets` hook returns either the assets surfaced from the view service, or those of the registry:
https://github.com/penumbra-zone/web/blob/4a51a461033e2e91f57a631dafd8992ef4f47e83/apps/veil/src/shared/api/assets.ts#L34

The problem is that assets from the registry only flow into the view service when the wallet is initialized:
https://github.com/penumbra-zone/web/blob/4a51a461033e2e91f57a631dafd8992ef4f47e83/packages/storage/src/indexed-db/index.ts#L189

This means that if the registry is updated, the view service may not have metadata for those assets.

We should instead *merge* the results of the two queries, not relying on just the view service.

(Ideally, we should not rely on the view service at all, right now I don't see a use case within Veil where we need to display assets that don't feature in the registry)

cronokirby:
One point I'm not completely sure of is if the view service will register metadata for notes that get sent to it later, as long as they appear in the registry at the point where it's syncing. This mitigates the issue somewhat. However, there can still be assets that the user wants to see in Veil that they may have never received, so the issue persists!

TalDerei:
> One point I'm not completely sure of is if the view service will register metadata for notes that get sent to it later

It's a nuanced observation. [IDB initialization](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L120-L124) occurs as part of starting the wallet services during service worker init, and [saveRegistryAssets](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L189) is then used to persist the latest `REGISTRY_VERSION` to storage. 

The registry version in the database is only updated in two cases: (i) a manual resync, or (ii) a service worker restart. 
- 1. clearing the cache resets local storage params, stops the block processor, tears down the database instance, and reloads the background service worker, which kicks of wallet services again. 
- 2. restarting the service worker reinitializes wallet services, which reads the database and [compares the stored registry version](https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L336-L346) against remote, and updates local version if it's out of date. The extension docs state that chrome terminates the service worker after [30 seconds of inactivity](https://developer.chrome.com/docs/extensions/develop/concepts/service-workers/lifecycle), but in reality the service worker stays alive as long as it's receiving events. 

[...]

Issue #2287, opened 2025-04-28
reward metadata renders too slowly


TalDerei:
‘Rewards Earned’ momentarily renders unknown-asset metadata before updating to the correct staking metadata. Apply the fix outlined in this comment to prevent the flicker: https://github.com/penumbra-zone/web/pull/2278#discussion_r2059432022

--------------------

https://github.com/user-attachments/assets/7087cad5-c34e-4a8b-b083-e9dfc5970f65

PR #2298, opened 2025-04-30
bug(veil): fetch registry object unconditionally

TalDerei:
## Description of Changes

the `useStakingTokenMetadata` hook was originally calling into [getAssetMetadataById](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/shared/api/metadata.ts#L20-L24), which attempts to fetch the metadata from the view service (in-memory registry cache). However, if the wallet wasn't connected and the remote registry hadn't been fetched yet, this would return undefined and display unknown assets.

this updates the hook to explicitly call `registryQueryFn`, which pays a one-time cost to fetch the full remote registry when the wallet isn't connected. Then it calls the registry helper `getMetadata` to retrieve the staking token metadata. React query then caches subsequent hook invocations. 

additionally, adds more stringent `isLoading` checks to the reward cell within `LeaderboardRow` to make sure rewards are only rendered when metadata is fully loaded.

----------------

https://github.com/user-attachments/assets/7b3fad3a-eaf6-4141-a620-2c25a5a246bd

## Related Issue

https://github.com/penumbra-zone/web/issues/2287 (_related_ to motivation in https://github.com/penumbra-zone/web/issues/2295, but doesn't close it)

### DIFF
--- a/apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
+++ b/apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
@@ -24,6 +24,7 @@ import { useDelegatorLeaderboard, BASE_PAGE, BASE_LIMIT } from '../api/use-deleg
 import { useSortableTableHeaders } from './sortable-table-header';
 import { useIndexByAddress } from '../api/use-index-by-address';
 import { useStakingTokenMetadata } from '@/shared/api/registry';
+import { Skeleton } from '@penumbra-zone/ui/Skeleton';
 
 const LeaderboardRow = observer(
   ({ row, loading }: { row: DelegatorLeaderboardData; loading: boolean }) => {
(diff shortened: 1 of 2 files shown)


apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
```diff
@@ -110,7 +110,10 @@ const LeaderboardRow = observer(
           {row.streak}
         </TableCell>
         <TableCell cell loading={loading}>
-          {row.total_rewards && <ValueViewComponent valueView={totalRewards} priority='tertiary' />}
+          {/* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- required to ensure staking token is loaded before rendering */}
+          {row.total_rewards && stakingToken && !stakingLoading && (
+            <ValueViewComponent valueView={totalRewards} priority='tertiary' />
+          )}
```
VanishMax, line 116: suggestion: display a Skeleton in this table cell if staking token is loading instead of blocking the whole table rendering

VanishMax (approved):
One minor UX suggestion

apps/veil/src/pages/tournament/ui/delegator-leaderboard.tsx
TalDerei: good call, changed to use skeleton

PR #2303, opened 2025-05-01
feat(quick-sync): TCT frontier support 

TalDerei:
## Description of Changes

quick sync using the state commitment tree snapshots to initialize the view server for freshly generated wallets.

currently failing on mainnet with `ConnectError: [unimplemented]` when attempting to call the SCT frontier service, but works with testnet rpc. cc @conorsch 

--------------------

https://github.com/user-attachments/assets/b4f333e2-f80a-49d6-bd2b-af258e2fdf23

## Related Issue

references https://github.com/penumbra-zone/penumbra/issues/1128 and https://github.com/penumbra-zone/penumbra/issues/1131

extension of https://github.com/penumbra-zone/penumbra/pull/5179 and pairs with https://github.com/prax-wallet/prax/pull/336


packages/storage/src/indexed-db/index.ts
```diff
@@ -1040,7 +1048,7 @@ export class IndexedDb implements IndexedDbInterface {
       txs.add({ table: 'TREE_HASHES', value: h });
     }
 
-    // TODO: What about updates.delete_ranges?
+    // TODO: What about updates.delete_ranges (https://github.com/penumbra-zone/web/issues/818)?
```
TalDerei, line 1051: comment: doesn't functionally impact syncing, but adds extraneous tree hashes in storage that should eventually be removed. Just linked to the relevant issue here.

packages/wasm/crate/src/view_server.rs
```diff
@@ -85,9 +85,41 @@ impl ViewServer {
...
+            storage: init_idb_storage(constants).await?,
+            last_position: None,
+            last_forgotten: None,
+        };
+        Ok(view_server)
+    }
+
+    /// Create new instances of `ViewServer` from SCT frontier snapshot.
+    #[wasm_bindgen]
+    pub async fn new_snapshot(
```
TalDerei, line 107: comment: separated view server initializing from existing indexedDB instance versus ingesting the tree construction from a snapshot.

packages/wasm/crate/src/view_server.rs
```diff
@@ -85,9 +85,41 @@ impl ViewServer {
...
+        compact_frontier: &[u8],
+    ) -> WasmResult<ViewServer> {
+        utils::set_panic_hook();
+
+        let fvk: FullViewingKey = FullViewingKey::decode(full_viewing_key)?;
         let constants = serde_wasm_bindgen::from_value(idb_constants)?;
+
+        let tree: Tree = bincode::deserialize(&compact_frontier)
+            .map_err(|e| JsValue::from_str(&format!("Failed to deserialize frontier: {}", e)))
+            .expect("frontier snapshot");
```
TalDerei, line 119: comment: bincode serialization

packages/wasm/crate/src/view_server.rs
```diff
@@ -85,9 +85,41 @@ impl ViewServer {
...
+
+        // TODO: optionally verify merkle inclusion proof against provided anchor before initializing view server.
```
TalDerei, line 121: comment: once the rpc supports `{ with_proof = true }`, we can check the merkle inclusion proof before initializing the view server state

packages/services/src/sct-service/index.ts
```diff
@@ -5,7 +5,7 @@ import { timestampByHeight } from './timestamp-by-height.js';
 
 export type Impl = ServiceImpl<typeof SctService>;
 
-export const sctImpl: Omit<Impl, 'anchorByHeight'> = {
+export const sctImpl: Omit<Impl, 'anchorByHeight' | 'sctFrontier'> = {
```
erwanor, line 8:
```typescript
/**
 * Construct a type with the properties of T except for those in type K.
 */
type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
```

I don't understand this interface, why are we excluding these two methods?

I don't completely understand this, we're omitting the impl for those methods but we need them

TalDerei: > I don't understand this interface, why are we excluding these two methods?

these are excluded in the services package. To understand why, compare the following:
- https://github.com/penumbra-zone/web/blob/main/packages/services/src/sct-service/timestamp-by-height.ts
- https://github.com/prax-wallet/prax/blob/main/packages/query/src/queriers/sct.ts

they perform the same function. Since prax is the only consumer of that method (none of that other web packages that service definition), you can use the querier in prax's query package directly:

```ts
const querier = new RootQuerier({ grpcEndpoint });
const compact_frontier = await querier.sct.sctFrontier( new SctFrontierRequest({ withProof: false })
```

rather than constructing it from the services package:

```ts
 const { ... } = await penumbra
    .service(SctService)
    .sctFrontier({withProof: false);
```

we can support it, but it'll just remain unused.

packages/wasm/crate/src/view_server.rs
erwanor: We don't need this because we don't have a root of trust for the root anyway

erwanor (approved):
Optimistically, the 405 is suspicious but if it was tested, we can move this to beta testing.

TalDerei:
> Optimistically, the 405 is suspicious but if it was tested, we can move this to beta testing.

yeah, that's probably a @conorsch question. this wasn't presenting as an issue on testnet

packages/wasm/crate/src/view_server.rs
```diff
@@ -86,9 +86,41 @@ impl ViewServer {
...
+        utils::set_panic_hook();
+
+        let fvk: FullViewingKey = FullViewingKey::decode(full_viewing_key)?;
         let constants = serde_wasm_bindgen::from_value(idb_constants)?;
+
+        let tree: Tree = bincode::deserialize(compact_frontier)
+            .map_err(|e| JsValue::from_str(&format!("Failed to deserialize frontier: {}", e)))
+            .expect("frontier snapshot");
+
+        // TODO: optionally verify merkle inclusion proof against provided anchor before initializing view server.
```
TalDerei, line 122: comment: @erwanor still is still an open todo

erwanor: We should fix the RPC in the full node, but we don't need to have this TODO here. This is fine to leave as is.

TalDerei: that comments misleading, check doesn't happen here obviously yeah

PR #2319, opened 2025-05-07
experiment(storage): idb improvements

turbocrime:
remove `IbdUpdater` in favor of events and direct transaction management.

intention is to improve scan/flush epoch transition handling.

draft work from april i'm not currently focused on, pushed up for visibility.


TalDerei:
closing as unplanned

Issue #2317, opened 2025-05-05
LP Leaderboard on details page


TalDerei:
currently still stubbed, should be wired up to pindexer to use real data. cc @JasonMHasperhoven 

----------------

<img width="1014" alt="Image" src="https://github.com/user-attachments/assets/de5531dd-06d1-44e6-9b47-820635068f0f" />

PR #2315, opened 2025-05-05
LP Rewards

JasonMHasperhoven:
## Description of Changes

Closes https://github.com/orgs/penumbra-zone/projects/20/views/23?pane=issue&itemId=106126979&issue=penumbra-zone%7Cweb%7C2248

Tested:
image

## Related Issue

Link to the issue this PR addresses.

## Checklist Before Requesting Review

- [ ] I have ensured that any relevant minifront changes do not cause the existing extension to break.

PR #2320, opened 2025-05-07
LP Leaderboard - Data Integration

JasonMHasperhoven:
## Description of Changes

Wires up the LP Leaderboard with pindexer's data.

Depends on https://github.com/penumbra-zone/web/pull/2315
Closes https://github.com/penumbra-zone/web/issues/2317

image
image

## Related Issue


apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
cronokirby, line 1: I think this file duplicates the functionality in use-lp-rewards?

apps/veil/src/entities/leaderboard/api/use-lp-leaderboard.ts
```diff
@@ -0,0 +1,73 @@
...
+    penumbra.service(DexService).liquidityPositionsById({ positionId: positionIds }),
+  );
+  const positions = positionsRes.map(r => r.data).filter(Boolean) as Position[];
+
+  return data.map((lp, index) => ({
+    ...lp,
+    positionIdString: bech32mPositionId(lp.positionId),
+    position: positions[index] as unknown as Position,
+  }));
+}
```
cronokirby, line 37: we should avoid duplicating this

cronokirby (changes requested):
Functionally, the lp leaderboard should work without a wallet connected, but right now loads for a long time before eventually returning:

image

My hunch is that the cause of this is that the leaderboard table is too tightly coupled with the assumption that we need to fetch both streams of data, when we actually can only be guaranteed to fetch just the generic one.

apps/veil/src/entities/leaderboard/api/use-lp-leaderboard.ts
cronokirby: We can get around to this in a subsequent refactor.

apps/veil/src/entities/leaderboard/api/use-lp-leaderboard.ts
```diff
@@ -0,0 +1,73 @@
...
+        page,
+        limit,
+        sortKey,
+        sortDirection,
+      } as LpLeaderboardRequest).then(async resp => ({
+        ...resp,
+        data: await enrichLpLeaderboards(resp.data),
+      }));
+    },
+    enabled: typeof epoch === 'number' && isActive,
```
VanishMax, line 69: suggestion: refetch the leaderboard on a new block if the epoch is not over. Disable refetch if epoch has ended

apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
```diff
@@ -0,0 +1,109 @@
...
+        page,
+        limit,
+        sortKey,
+        sortDirection,
+      } as LpLeaderboardRequest).then(async resp => ({
+        ...resp,
+        data: await enrichLpLeaderboards(resp.data),
+      }));
+    },
+    enabled: typeof epoch === 'number' && positionIds !== undefined && isActive,
```
VanishMax, line 105: suggestion: refetch on new block for the current epoch

apps/veil/src/entities/leaderboard/ui/table.tsx
```diff
@@ -1,76 +1,84 @@
...
+  MyLPs: 'My LPs',
+} as const;
+
+type Tab = (typeof Tabs)[keyof typeof Tabs];
+
+export const LeaderboardTable = observer(() => {
+  const { connected, subaccount } = connectionStore;
+  const totalRef = useRef<number>(0);
+  const params = useParams<{ epoch: string }>();
+  const epoch = Number(params?.epoch);
```
VanishMax, line 38: suggestion (consistency): accept epoch as a prop to align with another components of the round page

apps/veil/src/entities/leaderboard/ui/table.tsx
```diff
@@ -1,76 +1,84 @@
...
-        },
-      })),
-      `sortValues.${sortBy.key}`,
-      sortBy.direction,
-    );
-  }, [positions, sortBy]);
+  const [positions, total, error, isLoading] = isMyTab
+    ? [myLeaderboard?.data, myLeaderboard?.total, myLeaderboardError, myLeaderboardLoading]
+    : [leaderboard?.data, leaderboard?.total, leaderboardError, leaderboardLoading];
+  totalRef.current = total ?? totalRef.current;
```
VanishMax, line 81: question: why is `totalRef` a ref and not a state since it changes so often?

apps/veil/src/entities/leaderboard/ui/table.tsx
```diff
@@ -1,76 +1,84 @@
...
-export const LeaderboardTable = () => {
-  const totalCountRef = useRef<number>(0);
+const Tabs = {
+  AllLPs: 'All LPs',
+  MyLPs: 'My LPs',
+} as const;
+
+type Tab = (typeof Tabs)[keyof typeof Tabs];
+
+export const LeaderboardTable = observer(() => {
```
VanishMax, line 34: proposal (non-blocking): split this into two components: an actual table and a card with two tabs and a table, – for now it looks like the component does extra work with data mutation (e.g. totalRef) just to not reuse JSX

apps/veil/src/entities/leaderboard/ui/table.tsx
```diff
@@ -1,76 +1,84 @@
...
-      `sortValues.${sortBy.key}`,
-      sortBy.direction,
-    );
-  }, [positions, sortBy]);
+  const [positions, total, error, isLoading] = isMyTab
+    ? [myLeaderboard?.data, myLeaderboard?.total, myLeaderboardError, myLeaderboardLoading]
+    : [leaderboard?.data, leaderboard?.total, leaderboardError, leaderboardLoading];
+  totalRef.current = total ?? totalRef.current;
 
   if (error) {
```
VanishMax, line 83: suggestion: wrap error content into Card too. For now the error hangs within the page and it's not easy to understand what it relates to

VanishMax (commented):
Overall, the tables work, loving the result! Added mostly code-related comments, none of them are strictly blocking

apps/veil/src/entities/leaderboard/api/post-lp-leaderboard.ts
```diff
@@ -0,0 +1,84 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { serialize, Serialized } from '@/shared/utils/serializer';
+import { pindexerDb } from '@/shared/database/client';
+import { PositionId } from '@penumbra-zone/protobuf/penumbra/core/component/dex/v1/dex_pb';
+import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import { positionIdFromBech32 } from '@penumbra-zone/bech32m/plpid';
+import { LpLeaderboardRequest, LpLeaderboardApiResponse } from './utils';
+
+async function queryLqtLps({
+  positionIds = [],
```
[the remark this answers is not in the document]
JasonMHasperhoven: This api is used for both tabs of the LP leaderboard:
- All LPs tab (without position ids), and for
- My LPs (with provided position ids)

apps/veil/src/entities/leaderboard/ui/table.tsx
JasonMHasperhoven: When you click on another page, the total becomes undefined for a brief moment.

JasonMHasperhoven:
> My hunch is that the cause of this is that the leaderboard table is too tightly coupled with the assumption that we need to fetch both streams of data, when we actually can only be guaranteed to fetch just the generic one.

You're right. I did look at joining positions from the `dex_ex_position_state` (because we have to show the state), but I have no idea how to get `PositionState_PositionStateEnum` from the data in this table. Perhaps you can point me in the right direction.

apps/veil/src/entities/leaderboard/api/use-lp-leaderboard.ts
JasonMHasperhoven: When `use-lp-leaderboard.ts`, no longer depends on the DexService for position states, we can remove this and there will be no duplicated code.

apps/veil/src/entities/leaderboard/api/post-lp-leaderboard.ts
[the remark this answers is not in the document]
JasonMHasperhoven: I suppose we could split this if we're going to use a table join for the All LPs tab

cronokirby:
I think we don't need any server modifications, the frontend code will just need to be changed to not require the wallet to be connected, I think.

apps/veil/src/entities/leaderboard/ui/table.tsx
JasonMHasperhoven: TODO in next PR with asset selector.

apps/veil/src/entities/leaderboard/ui/table.tsx
JasonMHasperhoven: TODO, upgrade logic to account for switching tabs

JasonMHasperhoven:
Please approve this one so that it can be merged. Follow up PR is here that moves the merging of Position to the server side: https://github.com/penumbra-zone/web/pull/2362.

cronokirby:
I tested and looked at the two PRs jointly, in their current state ; good stuff!

cronokirby (approved):

Issue #2322, opened 2025-05-07
Delegator voting rewards UI bugs


vacekj:
**Describe the bug**
Loading state is missing for the delegator voting rewards tab.
When no rewards are available, the text is black.

<img width="1279" alt="Image" src="https://github.com/user-attachments/assets/c36adb97-ceb0-408d-b301-f55294de71c0" />

TalDerei:
@vacekj can you see if https://github.com/penumbra-zone/web/pull/2332 resolves this for you as a side-effect?

vacekj:
Fixed on latest main.

Issue #2292, opened 2025-04-29
rewards card: add casted vote field


TalDerei:
"my rewards" card is missing the "casted vote" column per https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx#L22-L24

-------------

<img width="490" alt="Image" src="https://github.com/user-attachments/assets/62bf0d83-294a-4353-9f74-d9c8c636b3c6" />

PR #2332, opened 2025-05-08
feat(veil): add "casted vote" column to delegator rewards card

TalDerei:
## Description of Changes

- added "casted vote" column to the delegator rewards card, with support for server-side pagination and sorting of delegator voting rewards,
- updated delegator history route to the conventional approach of returning structured, serialized types instead of raw `Buffer` for certain fields,
- modified the summary route to optionally accept a range of epochs, enabling targeted queries against specific epochs,
- cleanup extraneous `useCurrentEpoch` calls

-------------------

https://github.com/user-attachments/assets/1e8dfd01-6752-42ef-b558-ad4976ba4997

## Related Issue

references https://github.com/penumbra-zone/web/issues/2292 (and possibly https://github.com/penumbra-zone/web/issues/2322?)


apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
```diff
@@ -2,79 +2,165 @@ import { observer } from 'mobx-react-lite';
...
+    epochLoading,
+    page,
+    limit,
+    sortBy.key,
+    sortBy.direction,
+  );
+  const { data: stakingToken, isLoading: stakingLoading } = useStakingTokenMetadata();
+
+  // Extract epochs for summary lookup
+  const epochs = useMemo(() => rewardsData?.map(r => r.epoch).filter(Boolean) ?? [], [rewardsData]);
```
TalDerei, line 51: comment: extracts an array of epochs to be used in `useTournamentSummary` for requesting a summary of each epoch the user has voted in.

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
```diff
@@ -2,79 +2,165 @@ import { observer } from 'mobx-react-lite';
...
-  // user’s voting power (from `delegator_history`) by the epoch’s total voting
-  // power (from `delegator_summary`).
+  const loadingArr = new Array(5).fill({
+    address: {},
+    asset_id: {},
+    metadata: {},
+    epoch: {},
+    power: {},
+    reward: {},
+  }) as LqtDelegatorHistoryData[];
```
TalDerei, line 74: comment: loading pattern extracted from other pages

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
```diff
@@ -2,79 +2,165 @@ import { observer } from 'mobx-react-lite';
...
 
+  const displayData = isLoading ? loadingArr : (rewardsData ?? []);
+
+  // TODO: populate casted votes with metadata.
```
TalDerei, line 78: comment: currently, "casted votes" is missing metadata until registry support is resolved

apps/veil/src/pages/tournament/server/summary.ts
```diff
@@ -1,53 +1,94 @@
...
 import { LqtSummary } from '@/shared/database/schema';
 
+const DIRECTIONS = ['asc', 'desc'] as const;
+export type TournamentSummaryDirection = (typeof DIRECTIONS)[number];
+
 export interface TournamentSummaryRequest {
   limit: number;
   page: number;
-  epoch?: number;
+  epochs?: number[];
```
TalDerei, line 12: comment: previously, the summary only returned data for the latest epoch or a range of recent epochs. It now optionally supports an epochs field, allowing the caller to request a summary for a specific epoch or range of epochs.

apps/veil/src/pages/tournament/server/summary.ts
```diff
@@ -1,53 +1,94 @@
...
   return {
-    epoch,
     limit,
     page,
+    sortDirection,
   };
 };
 
 export interface TournamentSummaryApiResponse {
+  total: number;
```
TalDerei, line 39: comment: required field for pagination

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
```diff
@@ -2,79 +2,166 @@ import { observer } from 'mobx-react-lite';
...
+            return (
+              <div key={rowKey} className='grid grid-cols-subgrid col-span-4'>
+                <TableCell cell loading={isLoading}>
+                  {!isLoading && `Epoch #${rewardData.epoch.toString()}`}
+                </TableCell>
+
+                <TableCell cell loading={isLoading || !hasVoteWeightData}>
+                  {!isLoading &&
+                    hasVoteWeightData &&
+                    ((rewardData.power / matchingSummary!.total_voting_power) * 100).toFixed(3) +
```
TalDerei, line 129: comment: Is this the correct interpretation of casted votes — “% of voting power (for any asset) that was directed toward that asset”? @cronokirby

apps/veil/src/pages/tournament/server/summary.ts
```diff
@@ -1,53 +1,93 @@
...
+  let query = pindexerDb.selectFrom('lqt.summary').select(expr => [expr.fn.countAll().as('total')]);
+
+  if (epochs && epochs.length > 0) {
+    query = query.where('epoch', 'in', epochs);
+  }
+
+  return query.executeTakeFirst();
+};
+
+export async function POST(
```
TalDerei, line 72: comment: changed to `POST` method since we now need to pass epoch information to the server when requesting summaries

TalDerei:
temporarily converting back to draft to fix something

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -39,33 +39,33 @@ export const useVotingInfo = (defaultEpoch?: number) => {
...
   const votingNote = useMemo(() => {
-    const values = (notes ?? []).map(note => note.noteRecord?.note?.value).filter(item => !!item);
+    if (!notes?.length) {
+      return undefined;
+    }
 
-    const amount = values.reduce(
-      (accum, current) => {
-        if (current.amount) {
-          return addAmounts(accum, current.amount);
```
TalDerei, line 47: comment: `current` may have been undefined previously – the CI linter didn’t complain, but the local linter did.

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
```diff
@@ -83,12 +83,6 @@ export const VotingRewards = observer(() => {
             <TableCell heading> </TableCell>
           </div>
 
-          {!isLoading && (!rewardsData || rewardsData.length === 0) && (
-            <div className='col-span-4 text-sm text-muted-foreground py-4'>
-              No voting rewards found for this account.
-            </div>
-          )}
-
```
TalDerei: comment: not sure why we had this. The figma designs show that empty wallets without voting rewards should display zero balance and disable the dropdown menu.

-------------

Screenshot 2025-05-08 at 6 09 32 PM

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -91,7 +81,11 @@ const fetchRewards = async (
     return delegatorHistory;
   }
 
-  return undefined;
+  return {
+    data: [],
+    totalItems: 0,
+    totalRewards: 0,
+  };
```
TalDerei, line 88: comment: rather than returning undefined, always return empty object if no votes are found. Easier to render and reason about this in the UI.

TalDerei:
now this is in a more complete state and opening back up for testing!

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
cronokirby: It should actually be the total amount of voting power you directed toward that asset.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
cronokirby: This seems the correct thing to do to me.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -69,55 +71,101 @@ const fetchRewards = async (
...
-    },
+      ),
   });
+
+  return {
+    query,
+    data: query.data?.data ?? [],
+    total: query.data?.totalItems ?? 0,
+    totalRewards: query.data?.totalRewards ?? 0,
+  };
```
cronokirby, line 128: I think instead we should return something like:
```
{...query, data: {data: ..., total: ..., totalRewards: ...}}
```
so that it aligns with what useQuery will return by default.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -69,55 +71,101 @@ const fetchRewards = async (
...
     enabled: connectionStore.connected && !!epoch,
-    queryFn: async () => {
-      return await fetchRewards(
+    queryFn: async () =>
+      fetchRewards(
         {
           type: 'epoch',
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- based on `enabled`, epoch is always defined
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- epoch parameter is always defined
           value: BigInt(epoch!),
```
cronokirby, line 155: What utility does this function provide over the previous one?

cronokirby: I guess I see it behaves differently, but I think we should just merge the two, and use the presence of the "epoch" param to differentiate.

apps/veil/src/pages/tournament/api/use-tournament-summary.ts
```diff
@@ -1,19 +1,38 @@
...
   disabled?: boolean,
 ) => {
+  const epochsKey = useMemo(
+    () => (params?.epochs ? JSON.stringify(params.epochs) : undefined),
+    [params?.epochs],
+  );
+
   const query = useQuery({
-    queryKey: ['tournament-summary'],
+    queryKey: ['tournament-summary', params?.limit, params?.page, epochsKey],
```
cronokirby, line 20:
```suggestion
    queryKey: ['tournament-summary', epochsKey, ...params?.limit ?? [], ...params?.page ?? []],
```

query keys can't be undefined.

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -109,24 +107,43 @@ function processEpochResults(
     totalReward += match.reward;
   }
 
-  return { results, totalReward };
+  // We need to perform pagination after fetching all data from pindexer,
+  // since we require the total count. This is the correct approach for
+  // implementing server-side pagination while preserving accurate totals and sorting.
```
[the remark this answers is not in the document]
cronokirby: We should just do a separate query for the total reward. We don't want to pull all the results from the disk, that defeats the performance aspects of pagination.

apps/veil/src/pages/tournament/server/summary.ts
```diff
@@ -1,53 +1,94 @@
...
   req: NextRequest,
 ): Promise<NextResponse<Serialized<TournamentSummaryApiResponse | { error: string }>>> {
-  const params = getQueryParams(req);
-  const response = await tournamentSummaryQuery(params);
+  const urlParams = getQueryParams(req);
+  const bodyData = (await req.json()) as Partial<TournamentSummaryRequest>;
```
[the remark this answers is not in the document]
cronokirby: Why is this the proper way? I think using GET makes sense with the query params in the URL is what I'd expect.

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
```diff
@@ -82,15 +85,15 @@ const PreviousEpochsRow = observer(
...
+            {rewards.find(r => BigInt(r.epoch) === (row.epoch && BigInt(row.epoch))) && (
               <ValueViewComponent
                 valueView={
                   new ValueView({
                     valueView: {
                       case: 'knownAssetId',
                       value: {
                         amount: pnum(
-                          rewards.data.find(r => BigInt(r.epoch) === BigInt(row.epoch))?.reward,
+                          rewards.find(r => BigInt(r.epoch) === BigInt(row.epoch))?.reward,
```
cronokirby, line 96: I don't like us doing a scan here, we should instead create a Map or something before.

cronokirby (changes requested):
I think having the asset you voted for is important, and we should handle the undefined assets by either
- ignoring the icon and symbol _for that row only_,
- using an unknown icon and symbol.

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
TalDerei: > It should actually be the total amount of voting power you directed toward that asset.

doesn't the interpretation in the implementation match the figma design?

---------------

Screenshot 2025-05-09 at 11 02 45 AM

cronokirby: You're absolutely right, I was misremembering this when I started the review, and forgot to update this comment!

apps/veil/src/pages/tournament/server/delegator-history.ts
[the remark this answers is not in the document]
TalDerei: we totally can do that! forgot we can just do a separate pindexer query for just that field

apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx
TalDerei: > I recall our UI components having the ability to display unknown assets ; I think we should do that.

for visibility, when displaying unknown metadata, we want to construct a `ValueView` with `unknownAssetId` (instead of `knownAssetId`), and leave the metadata field empty.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
TalDerei: yeah, returning undefined opened a can of worms that I didn't want to deal with, so this just returns an empty `TournamentDelegatorHistoryResponse` object.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
TalDerei: > so that it aligns with what useQuery will return by default.

this is a purely stylistic choice – the first approach keeps the query object intact in the sense that you don't have to do more complicated nested destructing on the caller-side. You'll see what i'm referring to if you modify it

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
TalDerei: > I guess I see it behaves differently, but I think we should just merge the two, and use the presence of the "epoch" param to differentiate.

was following DRY here – `usePersonalRewards` fetches all votes up to specified block height, while `usePersonalRewardsForEpoch` fetches votes for the specified epoch.

apps/veil/src/pages/tournament/api/use-tournament-summary.ts
TalDerei: react query uses the full array as the cache key, but it's certainly fine for them to be undefined if the caller doesn't supply it. we don't need to stub this with empty values

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
TalDerei: could certainly change to a map, and it would probably yield more efficient lookups

cronokirby (approved):

PR #2336, opened 2025-05-08
Improve registry / asset fetching

### LINKED ISSUE #2295: Bug: useAssets hook can result in stale data

cronokirby:
## Description of Changes

This relies solely on our registry for asset fetching. This closes #2295 This also greatly simplifies the client side code, and improves performance, by fetching the registry once on the server side (atm, for any page), and including it in the bundle, avoiding the need for an additional round trip on the client, and removing code which needlessly checks whether or not the registry is loaded, by guaranteeing that it always is loaded.


JasonMHasperhoven (approved):
LGTM, well done 👍

erwanor:
We can schedule a refresh every six hours, but caching is a must

erwanor (approved):

cronokirby:
> > by fetching the registry once on the server side (atm, for any page)
> 
> So this means that for future registry updates, a service bounce is required for the new registry to take effect, right? That's not a problem—given how quickly code is changing in Veil, service bounces happen several times a day—I just want to make sure I understand the data flow.

The server should do a fetch every time you do a request, so this shouldn't be needed afaict

PR #2337, opened 2025-05-08
storage: store incentivized asset id instead of metadata

TalDerei:
## Description of Changes

references https://github.com/penumbra-zone/web/blob/main/packages/storage/src/indexed-db/index.ts#L524-L534. Recent registry changes exacerbated this issue by exposing breakage whereby the asset Id used at the time of voting no longer exists in the current registry, preventing metadata (and by extension the vote) from being saved.

pairs with https://github.com/prax-wallet/prax/pull/339

## Related Issue

N/A

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -433,9 +433,9 @@ export class IndexedDb implements IndexedDbInterface {
     subaccount?: number,
   ): Promise<
     {
+      incentivizedAsset: AssetId;
       epoch: string;
       TransactionId: TransactionId;
-      AssetMetadata: Metadata;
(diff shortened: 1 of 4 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -507,9 +509,9 @@ export class IndexedDb implements IndexedDbInterface {
    * Saves historical liquidity tournament votes and rewards for a given epoch.
    */
   async saveLQTHistoricalVote(
+    incentivizedAsset: AssetId,
```
cronokirby, line 512: Nit, let's preserve the argument order?

cronokirby (approved):
LGTM

Issue #2254, opened 2025-04-12
reduce intrusiveness of syncing modal


TalDerei:
currently, syncing modal behavior feels overly aggressive – stalling on densely populated blocks even with fast internet connections. Should we make it more conservative? Should timeouts be increased?

additionally, can we detect signal sources originating from wallet locking or clearing cache events and possibly avoid surfacing the syncing modal in those cases. Instead, maybe we should opt towards auto-reloading (https://github.com/penumbra-zone/web/issues/1885) similar to how we handle disconnections.

cc @turbocrime 

------------------

https://github.com/user-attachments/assets/f9838abc-4fc3-43ad-ba4e-048788b5af16

erwanor:
The syncing modal is kind of disruptive. It sometimes shows up on caught up wallets, presumably to sync a few blocks, and then disappears. I'd rather it never appeared in those cases. Is this what this ticket tracks? The recording seems unrelated, with a Keplr popup.

We don't need to do anything about it now, but it's good to track for a Praxv20 or 21.

TalDerei:
we should increase the timeout here, seeing the syncing modal surface often on slower networks.

----------------

<img width="777" height="640" alt="Image" src="https://github.com/user-attachments/assets/f1411773-7303-45eb-bb3d-107973524c03" />

PR #2347, opened 2025-05-12
minifront: sync dialog text

erwanor:
## Description of Changes
Screenshot 2025-05-11 at 10 41 49 PM

We need to hammer that this is a local first experience. And the instruction should be worded in terms of a positive.

>Copy changes

### DIFF
--- a/apps/minifront/src/components/syncing-dialog/index.tsx
+++ b/apps/minifront/src/components/syncing-dialog/index.tsx
@@ -81,9 +81,7 @@ export const SyncingDialog = () => {
 
     const text = 'All of your private data is local.';
 
-    const instructions = isUnavailable
-      ? 'Please reload the page.'
-      : 'You can leave this tab open to speed things up.';
+    const instructions = isUnavailable ? 'Please reload the page.' : 'You can leave this tab open';
 
     return { title, error, subheading, text, instructions };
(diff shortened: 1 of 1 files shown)


apps/minifront/src/components/syncing-dialog/index.tsx
```diff
@@ -62,28 +62,30 @@ export const SyncingDialog = () => {
...
+        subheading = 'Unknown state.';
       }
     }
 
+    const text = 'All of your private data is local.';
+
     const instructions = isUnavailable
       ? 'Please reload the page.'
-      : 'You can click away, but your data may not be current.';
+      : 'You can leave this tab open to speed things up.';
```
TalDerei, line 86: this _isn't_ actually true in practice — the main extension handler, the service worker, operates in a constrained single-threaded runtime, and keeping the extension open reduces context switching for the service worker, allowing it to stay focused on syncing. In reality, leaving the extension open actually speeds things up. That's kind of a moot point though since we’re removing initial sync for fresh wallets, and existing wallets should always be caught up with periodic background syncing.

pertinently, we can address https://github.com/penumbra-zone/web/issues/2254 by reducing the timeout parameter that causes the modal to reappear

erwanor: Got it, I thought it would keep the worker hot but I suppose there's no keep-alive link at all

TalDerei (approved):

PR #2349, opened 2025-05-12
veil: bug: distinguish no lp rewards and loading

### LINKED ISSUE #2348: LP rewards bug
If the `useLpRewards` hook doesn't find owned positions in the view server , it fails to properly handle the empty positions case, returning `undefined` and falling into a rendering loop. In that case, we should return an empty `LpRewardsApiResponse` object (as shown in the second screenshot) and handle it appropriately in the UI. cc @cronokirby @JasonMHasperhoven 

Image

----------

Image

cronokirby:
The bug here is that we were treating the case where we knew the user had no positions and the case where we were waiting to learn if the user had any positions identically, causing an infinite load.

## Description of Changes

This handles things correctly now, returning an empty rewards response when you have no positions.

Closes #2348.

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-lp-rewards.ts
+++ b/apps/veil/src/pages/tournament/api/use-lp-rewards.ts
@@ -84,6 +84,9 @@ export const useLpRewards = (
     // but it's fine to let people reload the page to see that.
     staleTime: Infinity,
     queryFn: async () => {
+      if (!positionIds?.length) {
+        return { data: [], total: 0, totalRewards: 0 };
+      }
       return apiPostFetch<LpRewardsApiResponse>('/api/tournament/lp-rewards', {
         positionIds,
         page,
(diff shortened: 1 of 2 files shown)


TalDerei:
threw this error

Screenshot 2025-05-11 at 9 38 45 PM

erwanor (approved):

TalDerei (changes requested):
threw this error

Screenshot 2025-05-11 at 9 38 45 PM

cronokirby:
On a fresh wallet it successfully loads for me, returning 0. I'm not quite sure how to reproduce the issue.

cronokirby:
https://github.com/penumbra-zone/web/blob/e3bbefbf60e74fd5c29eb44391a2d73d9ea4b4d1/apps/veil/src/pages/tournament/ui/lp-rewards.tsx#L127-L151

I think this part is the culprit, but seems to be somewhat race conditiony to trigger. The issue is that you cast something to `LpReward`, incorrectly, because it's missing a bunch of fields, and then the type checker no longer helps you from runtime errors.

This is a bad pattern and we should not be doing it.

TalDerei (approved):
👍

PR #2355, opened 2025-05-13
fix(veil): #2340: implement better gauge rendering

### LINKED ISSUE #2340: Gauge voting results are not showing that certain assets are below the incetivisation threshold
**Describe the bug**
When I look at the voting results and there are multiple assets that have received votes, even votes with 0% are not denoted as below the incentivisation threshhold

**To Reproduce**

1. Ensure that more than one asset has received a vote in a given epoch. Because USDC is being automatically voted on, the easiest thing to do is choose another asset for your own vote. In the attached screenshot, I have voted for OSMO

**Expected behavior**
OSMO should be denoted as below the threshold by having it's % shown in grey text while USDCs should be in white text. Please see Figma designs

**Screenshots**

Image

VanishMax:
Closes #2340 

Now, current results on the landing tournament page will always contain 5 assets. Also, improved and reused voted assets rendering between the mentioned component and the voting dialog. Assets below the threshold are now grayscaled. If the percentage is less than 1 but not 0, it renders as "<1%".

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-epoch-results.ts
+++ b/apps/veil/src/pages/tournament/api/use-epoch-results.ts
@@ -1,13 +1,26 @@
+import { useMemo } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { apiFetch } from '@/shared/utils/api-fetch';
+import { useRegistryAssets } from '@/shared/api/registry';
 import { useRefetchOnNewBlock } from '@/shared/api/compact-block';
 import { EpochResultsRequest, EpochResultsApiResponse } from '../server/epoch-results';
+import type { MappedGauge } from '@/pages/tournament/server/previous-epochs';
+import { Metadata } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+import { assetPatterns } from '@penumbra-zone/types/assets';
 
+/**
+ * Requests voting results of a given epoch. Returns percentages of each
+ * asset within the epoch and an array of `assetGauges` – results merged
+ * with registry asset metadata that usually gives an array of asset with 0 votes.
+ */
 export const useEpochResults = (
   name: string,
   params: Partial<EpochResultsRequest>,
   disabled?: boolean,
+  search = '',
 ) => {
+  const { data: assets } = useRegistryAssets();
+
   const query = useQuery({
     queryKey: [name, params.epoch, params.limit, params.page, params.sortKey, params.sortDirection],
     enabled: !!params.epoch && !disabled,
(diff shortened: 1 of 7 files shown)


apps/veil/src/pages/tournament/api/use-epoch-results.ts
```diff
@@ -18,5 +31,47 @@ export const useEpochResults = (
...
+      }
+
+      return {
+        asset,
+        epoch: 0,
+        votes: 0,
+        portion: 0,
+        missing_votes: 0,
+      };
+    });
```
cronokirby, line 70: It's not correct to do this. The amount of missing votes is not 0. We should just not display assets that have received no votes.

apps/veil/src/pages/tournament/api/use-epoch-results.ts
```diff
@@ -18,5 +31,47 @@ export const useEpochResults = (
...
+  // filters assets from the registry that are IBC-assets and match the search query
+  const filteredAssets = useMemo<Metadata[]>(() => {
+    return assets.filter(asset => {
+      return (
+        assetPatterns.ibc.matches(asset.base) &&
+        (asset.symbol.toLowerCase().includes(search.toLowerCase()) ||
+          asset.description.toLowerCase().includes(search.toLowerCase()))
+      );
+    });
+  }, [assets, search]);
```
cronokirby, line 53: If we don't display assets having received no votes, then this code can be removed as well.

apps/veil/src/pages/tournament/ui/vote-dialog/vote-dialog-asset.tsx
```diff
@@ -1,48 +1,70 @@
...
-const VOTING_THRESHOLD = 0.05;
+export const VOTING_THRESHOLD = 0.05;
 
 export interface VoteDialogAssetProps {
   asset: MappedGauge;
   onSelect: VoidFunction;
 }
 
-export const VoteDialogAsset = ({ asset, onSelect }: VoteDialogAssetProps) => {
+export const getVoteAssetComponents = (asset: MappedGauge) => {
```
cronokirby, line 15: I think it's probably more idiomatic to split this into two components instead

apps/veil/src/pages/tournament/api/use-epoch-results.ts
VanishMax: What do you mean? The field `missing_votes` is not used anywhere in the frontends, needed here only for typing reasons

cronokirby (commented):
I suggested a couple small refactors, otherwise this looks fine.

apps/veil/src/pages/tournament/api/use-epoch-results.ts
cronokirby: The issue with creating a meaningless value to satisfy the type checker, under the guise that the field isn't actually used, is that then someone subsequently ends up using the field, and now you have a bug.

In this case, probably fine to display 0 somewhere, but this was the source of an actual runtime error recently.

If we actually want to include a value here, the correct type is something like `MappedGauge | Asset`, actually representing the possibilities.

I think in this case, what we want is just to not consider assets which don't have gauge information available to them at all.

apps/veil/src/pages/tournament/api/use-epoch-results.ts
VanishMax: we need this to display asset with no votes and improve overall UX – users need to see options to vote for

protocol might filted for ibc assets, but the registry doesn't

apps/veil/src/pages/tournament/ui/vote-dialog/vote-dialog-asset.tsx
VanishMax: good point, refactored

apps/veil/src/pages/tournament/api/use-epoch-results.ts
VanishMax: I still don't see an issue here, but removed the `missing_votes` field from types and from the usage. 

> the correct type is something like MappedGauge | Asset

this would only bring confusion, and we'd still need mapping in some underlying component

> what we want is just to not consider assets which don't have gauge information available to them at all

we need it for UI/UX

apps/veil/src/pages/tournament/api/use-epoch-results.ts
cronokirby: Yeah, that's a good point about needing to see the options with no votes. In practice the only asset this will filter out is UM, but it's fine to have this logic.

apps/veil/src/pages/tournament/api/use-epoch-results.ts
cronokirby: I think the actual situation here is that you have all the assets, some of which may have the information including the votes, their portion, etc. So the type you want is something like:
```ts
interface Example {
  asset: Asset;
  voteInfo?: MappedGauge | undefined;
}
```
but, eh, this isn't blocking for the PR.

cronokirby (approved):

PR #2358, opened 2025-05-13
fix: filter out unknown assets in asset bars

### LINKED ISSUE #2329: Uncaught error on Portfolio page, can't build getter
**Describe the bug**
In both dev and prod I cannot access the `/portfolio` route when my wallet is connected. 

> `Error: Failed to select value from "[object Object]" with ""`

which I encountered on this route: http://localhost:3000/portfolio

**To Reproduce**`

1. Run dev env 
2. Navigate to http://localhost:3000/portfolio
3. Make sure your wallet is connected; I didn't see the error until I took that step

**Expected behavior**
I expect the page to load so that I can interact with the website.

**Screenshots**

Image

**Additional context**
Encountered this error while checking the Portfolio page after #2308 was closed. The good news is, the bug in #2308 is no longer occurring, but now there's a new one.

vacekj:
fixes #2329

### DIFF
--- a/apps/veil/src/pages/portfolio/ui/asset-bars.tsx
+++ b/apps/veil/src/pages/portfolio/ui/asset-bars.tsx
@@ -4,6 +4,7 @@ import { BalancesResponse } from '@penumbra-zone/protobuf/penumbra/view/v1/view_
 import {
   getMetadataFromBalancesResponse,
   getBalanceView,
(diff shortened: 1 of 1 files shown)


conorsch:
Tested in the local dev env, and with this change, the `/portfolio` route now renders correctly:

veil-portfolio-displays

Nice work, @vacekj!

PR #2360, opened 2025-05-14
atomic validator status updates

TalDerei:
## Description of Changes

directly lifts only the `updateValidatorInfo` method from https://github.com/penumbra-zone/web/pull/2258. I haven't tested this functionally yet and am not comfortable with merging, but we should proceed to the testing stage. 

question: can this proceed and land after a release that includes the epoch handling work? I think that will be the plan for now, cc @turbocrime 

## Related Issue

complementary to https://github.com/penumbra-zone/web/pull/2242

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -33,7 +33,10 @@ import { IbdUpdater, IbdUpdates } from './updater.js';
 
 import { IdbCursorSource } from './stream.js';
 
-import { ValidatorInfo } from '@penumbra-zone/protobuf/penumbra/core/component/stake/v1/stake_pb';
+import {
+  ValidatorInfo,
+  ValidatorInfoResponse,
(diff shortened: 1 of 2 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -874,6 +880,56 @@ export class IndexedDb implements IndexedDbInterface {
     return result ? Epoch.fromJson(result) : undefined;
   }
 
+  async updateValidatorInfo(
+    validatorInfoResponses: AsyncIterable<ValidatorInfoResponse>,
+  ): Promise<void> {
+    const validatorInfos = (await Array.fromAsync(validatorInfoResponses))
+      .map(({ validatorInfo }) => validatorInfo)
+      .filter(x => x != null);
```
TalDerei, line 888: references https://github.com/penumbra-zone/web/pull/2258#discussion_r2042627867

TalDerei:
closing as unplanned

Issue #2363, opened 2025-05-14
UI v2: Portfolio Page incl. PorftolioBalance & AssetCard


DDX1:
This issue tracks the work to update the v2 portfolio page and its components in Minifront to align with the latest Figma designs. TransactionCard is not yet included here. PortfolioBalance and USDC values in AssetCard should be hidden as long as associated data can't be fetched easily. 

Link to the Figma Design:
[Minifront v2 Portfolio Page](https://www.figma.com/design/pTXDKhfOkXXwpUL8GQRIO4/Minifront-V2?node-id=12-22307&m=dev)

Issue #1937, opened 2024-12-07
UIv2 Library Designs — Tracking List


smmhrdmn:
**Is your feature request related to a problem? Please describe.**
New ui designs need to be implemented so minifront can be updated and others can utilize the library.

**Describe the solution you'd like**
Call the current state of the UIv2 Library done, and begin implementing changes. This tracking list can be used to mark off components that have been **_fully implemented_** into the ui library. As components are implemented or issues are created from here, please make note of them here. 

**Describe alternatives you've considered**
n/a

**Additional context**

- [UIv2 Library designs](https://www.figma.com/design/WKgXSVVK3L0EI0sr9Hwjoa/penumbra-UI-library?m=auto&node-id=1600-80188&t=10ultq3wW4cPHe3R-1)
- Further details on what changes have been made to each component are outlined in Figma. Some of these changes are minor and some are major. If you have any questions about changes, or run into any issues, please reach out.

### Primitives _(ready for dev)_

- [ ] Color _(alternative black added)_
- [ ] Icon _(a few miscellaneous icons added that are not Lucide icons)_

### Semantics _(ready for dev)_

- [ ] Color _(Several color additions, including gradients)_
- [ ] Typography _(additional text styles added based on DEX designs)_

### Components _(ready for dev)_

- [ ] Account selector
- [ ] Action Views — _(new component)_
- [ ] Address view
- [ ] Alert — _(new component)_
- [x] Asset Icon — _(no changes)_
- [ ] Asset Group — _(new component)_
- [ ] Asset Selector
- [ ] Asset Value Input 
- [ ] Auction Routing
- [ ] Block sync bar
- [ ] Block sync popover
- [ ] Button
- [ ] Button group
- [ ] Card base — _(previously Card)_
- [ ] Dialog
- [ ] Input field — _(major changes)_
- [ ] Icon
- [ ] Icon Adornment
- [x] Identicon — _(no changes)_
- [ ] List item — _(major changes)_
- [ ] Menu item
- [ ] Pill
- [ ] Popover
- [ ] Segmented control
- [ ] Slider
- [ ] Swap asset selector — _(new component)_
- [ ] Swap Claim — _(new component)_
- [ ] Swap Value Input 
- [ ] Swap routing
- [ ] Navigation bar
- [ ] Tab
- [ ] Table cell — _(previously Table)_
- [ ] Toast — _(new component)_
- [x] Toggle — _(no changes)_
- [ ] Tooltip
[...]

PR #2364, opened 2025-05-14
Feat/v2 portfolio balance and assets

DDX1:
## Description of Changes

feat(ui): Added v2 Portfolio Page layout including PortfolioBalance and AssetCard components. Also had to setup a dual-theme build process to avoid tailwind conflicts in Minifront app (legacy vs. v2 theme).

## Related Issue

[Link to the issue this PR addresses.
](https://github.com/penumbra-zone/web/issues/2363)


packages/ui/src/TextInput/index.tsx
```diff
@@ -45,77 +45,79 @@ export interface TextInputProps
...
-      {...rest}
-      value={value}
-      onChange={e => onChange?.(e.target.value)}
-      placeholder={placeholder}
-      disabled={useDisabled(disabled)}
-      type={type}
-      max={max}
-      min={min}
-      ref={ref}
+export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
```
VanishMax, line 48: issue: we moved to React 19 recently with Veil and removed `forwardRef` in favor of native ref props: https://imrankhani.medium.com/react-19-deprecated-forwardref-a-guide-to-passing-ref-as-a-standard-prop-7c0f13e6a229

i'm afraid we need to migrate all Minifront components to new APIs and remove forwardRef, upgrade Minifront to React 19. Do you think it's possible within this PR or should be done separately?

packages/ui/src/AssetCard/AssetListItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+        {!isNaN(Number(asset.value)) && (
+          <Text color="text.secondary" xs>
+            {asset.value}
+          </Text>
+        )}
+      </div>
+      {/* Buttons group - absolutely positioned */}
+      <div className="absolute right-3 top-4 flex items-center gap-2 opacity-0 group-hover:opacity-100 transition-opacity duration-200">
+        <Button 
+          iconOnly
```
VanishMax, line 65: issue (ui): the icons look off in the Storybook:

image

packages/ui/package.json
```diff
@@ -8,7 +8,9 @@
     "node": ">=22"
   },
   "scripts": {
-    "build": "vite build",
+    "build": "vite build && pnpm run build:css && pnpm run build:css:prefixed",
+    "build:css": "pnpm exec tailwindcss -c ./tailwind.config.ts -i ./src/styles/main.css -o ./dist/style.css --minify",
+    "build:css:prefixed": "pnpm exec tailwindcss -c ./tailwind.config.prefixed.ts -i ./src/styles/main.css -o ./dist/style-prefixed.css --minify",
```
VanishMax, line 13: praise: amazing work with splitting v1 from v2 👏

packages/ui/src/AssetCard/index.tsx
VanishMax, line 1: issue (code structure, blocking): i'd suggest moving Minifront-related component from UI package to Minifront. Let's say that UI library should have only component listed in "Penumbra Web UI" Figma file. Generally, these components are defined in the tracking issue #1937. If, on the other hand, the components are defined only in "Minifront V2" Figma file, let's keep them in Minifront.

This means that AssetCard, InfoButton, InfoCardTitle, and PortfolioBalance need to be moved.

VanishMax (commented):
Some files need to be moved to Minifront from UI library, the rest is 10/10 code 🔥

packages/ui/src/TextInput/index.tsx
DDX1: I see! Thanks for pointing out. I'll take care of the upgrade and cleanup. Should work in this PR.

packages/ui/src/AssetCard/AssetListItem.tsx
DDX1: Good catch! Fix on the way.

VanishMax (approved):

PR #2365, opened 2025-05-14
fix(veil): #2350: personal rewards tables

### LINKED ISSUE #2350: rendering rewards lagging
a [conditional](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/total-delegator-rewards.tsx#L95) was removed during the refactor of https://github.com/penumbra-zone/web/pull/2332 that prevented rewards briefly displaying as zero before rendering the correct amount.

additionally, "no voting rewards found for this account" was [accidentally removed](https://github.com/penumbra-zone/web/pull/2332/files#diff-7ca50a7b5908d778409d54be927510ac77a5f259b8d55feef970a0d2ced23b4aL72-L75) and should be restored. We can also add a similar message for zero reward positions. cc @cronokirby 

additionally, the the loading states for paginating between states seems to have been affected post refactor.

-----------------------

https://github.com/user-attachments/assets/be06ff4a-8af7-4e5c-a1af-dd32a6f3786d

https://github.com/user-attachments/assets/dbbc9285-ff30-4e99-8899-147ebdc913f5

VanishMax:
Closes #2350 

In this PR:
1. Fix: Returned correct loading and empty states to "LPs rewards" and "Voting rewards" tables within the tournament landing page, as the issue suggests
2. Feat: added sorting by rewards, not only by epoch to "Voting rewards" table. Previously, only sorting by epoch was supported
3. Refactor: removed all JS-based rewards and power calculation from the server, delegated this logic to SQL. Works faster now, easier to maintain and read. Tested that the result is the same as it was before.
4. Fixed a bug of pagination limit not being correctly applied in the voting rewards table

image

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
+++ b/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
@@ -15,6 +15,7 @@ import {
   AddressByIndexResponse,
   TournamentVotesResponse,
 } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
 
 export const BASE_LIMIT = 10;
 export const BASE_PAGE = 1;
@@ -69,8 +70,7 @@ const fetchRewards = async (
       '/api/tournament/delegator-history',
       {
         epochs: Array.from(epochs),
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- address is defined
-        address: { inner: Array.from(address!.inner) },
+        address: address ? bech32mAddress(address) : '',
         page,
         limit,
         sortKey,
--- a/apps/veil/src/pages/tournament/server/delegator-history.ts
+++ b/apps/veil/src/pages/tournament/server/delegator-history.ts
@@ -1,14 +1,13 @@
 import { NextRequest, NextResponse } from 'next/server';
+import { Address } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
+import { isAddress, addressFromBech32m } from '@penumbra-zone/bech32m/penumbra';
+import { AssetId } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
 import { Serialized, serialize } from '@/shared/utils/serializer';
 import { pindexerDb } from '@/shared/database/client';
 import { LqtDelegatorHistory } from '@/shared/database/schema';
-import { Address } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
-import { bech32mAddress } from '@penumbra-zone/bech32m/penumbra';
-import { AssetId, Metadata } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
 import { BASE_LIMIT, BASE_PAGE } from '../api/use-personal-rewards';
-import { ChainRegistryClient } from '@penumbra-labs/registry';
 
-export const SORT_KEYS = ['epoch', 'power', 'reward', ''] as const;
+export const SORT_KEYS = ['epoch', 'reward'] as const;
 export type DelegatorHistorySortKey = (typeof SORT_KEYS)[number];
 
 export const DIRECTIONS = ['asc', 'desc'] as const;
@@ -16,7 +15,7 @@ export type DelegatorHistorySortDirection = (typeof DIRECTIONS)[number];
 
 export interface DelegatorHistoryRequest {
   epochs: string[];
-  address: Address;
+  address: string;
   limit?: number;
   page?: number;
   sortKey?: DelegatorHistorySortKey;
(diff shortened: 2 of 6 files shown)


apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -32,162 +31,118 @@ export interface TournamentDelegatorHistoryResponse {
...
- * delegator address.
- *
- * For each epoch we:
- *    * filter rows belonging to the delegator,
- *    * tally their total voting power,
- *    * capture that epoch’s reward and asset id.
- *
- * Finally, we aggregate the rewards earned by the address across all epochs.
- */
-function processEpochResults(
```
VanishMax, line 59: removed this function completely in favor of vanilla SQL calculations

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -32,162 +31,118 @@ export interface TournamentDelegatorHistoryResponse {
...
-  const registryClient = new ChainRegistryClient();
-  const registry = await registryClient.remote.get(chainId);
-
-  // Map only the paginated results with metadata
-  const mapped = await Promise.all(
-    paginatedResults.map(item => {
-      // TODO: remove hardcoded staking asset metadata when registry is fixed.
-      const stakingAssetId = registryClient.bundled.globals().stakingAssetId;
-      const asset_id = new AssetId({ inner: Uint8Array.from(item.asset_id) });
-      const metadata = registry.tryGetMetadata(stakingAssetId);
```
VanishMax, line 169: removed metadata searching from the backend as frontend duplicates this work anyway

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
```diff
@@ -84,26 +84,27 @@ const PreviousEpochsRow = observer(
             </Tooltip>
           )}
         </TableCell>
-        {connected && (rewardsLoading || reward !== undefined) && (
-          <TableCell cell loading={isLoading || rewardsLoading}>
```
VanishMax, line 88: found a rendering bug that was hiding a cell in a certain condition

apps/veil/src/pages/tournament/server/delegator-history.ts
cronokirby: love this!

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -69,8 +70,7 @@ const fetchRewards = async (
       '/api/tournament/delegator-history',
       {
         epochs: Array.from(epochs),
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- address is defined
-        address: { inner: Array.from(address!.inner) },
+        address: address ? bech32mAddress(address) : '',
```
cronokirby, line 73: I think we should throw here, or explicitly mark this function as being able to return undefined, because otherwise we'll just get an error when the server sees an empty string and doesn't like that.

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -32,162 +31,118 @@ export interface TournamentDelegatorHistoryResponse {
...
-      asset_id: Buffer.from(match.assetId),
-      reward: match.reward,
-    });
+  const sortDirectionParam = body.sortDirection;
+  const sortDirection =
+    sortDirectionParam && DIRECTIONS.includes(sortDirectionParam) ? sortDirectionParam : 'desc';
 
-    totalReward += match.reward;
+  const epochs = Array.isArray(body.epochs)
+    ? body.epochs.filter(epoch => typeof epoch === 'string')
```
cronokirby, line 52: Generally not a fan of doing `typeof` comparisons instead of explicitly comparing to undefined, because it creates unnecessary code churn if the type changes.

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -32,162 +31,118 @@ export interface TournamentDelegatorHistoryResponse {
...
+    : [];
+
+  if (!isAddress(body.address)) {
+    return 'Invalid address';
   }
```
cronokirby, line 57: Instead of doing a check like this, we should instead transform the type so that it's an actual Address type. This isn't blocking though, but a point to think of when we refactor the domain type library and what not.

apps/veil/src/pages/tournament/ui/lp-rewards.tsx
```diff
@@ -1,86 +1,77 @@
...
-            null
-          </TableCell>
-          <TableCell cell loading>
-            null
-          </TableCell>
-        </div>
-      ))}
-    </>
-  );
-}
```
cronokirby, line 49: This was explicitly factored out, because the pattern of merging in `isLoading` deep down into a component's display makes it much harder to figure out what the component is actually display as it goes through distinct states in its lifecycle.

The modification here should just be to add the case for when the rewards are known to be 0.

apps/veil/src/pages/tournament/ui/previous-epochs.tsx
cronokirby: This illustrates my gripe against nesting the loading condition inside of the component, versus making a top level decision.

cronokirby (changes requested):
In `lp-rewards.tsx`, I think we should keep a separate LoadingRows construction. Maybe the way it's done in `total-delegator-rewards`, for example.

The other comments are minor.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
VanishMax: the endpoint throws 400 in this rare case, so it's handled

apps/veil/src/pages/tournament/ui/lp-rewards.tsx
VanishMax: How blocking this change is? Didn't hear from the web team that pattern was bad.

I suggest sticking to the [Conventional Comments](https://conventionalcomments.org/) when doing code reviews – it generally helps preventing subjective change requests and promotes discussion rather then submission to another opinion. Sticking to conventional comments, what label would you put on this comment? How blocking would you make it?

VanishMax: In retrospect, I see the concern – been trying to refactor the refactored, erasing your code while doing so. Will change it tomorrow

cronokirby: I would consider this a blocking suggestion. I think we've already run into a couple of bugs related to unclear handling of what is and isn't loading, so I think we should apply a soft "ratchet" of at least not regressing in terms of our handling of this.

This doesn't directly impact the current functionality, but it makes any kind of bug finding wrt the lifecycle much more tedious, in our recent experience.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
cronokirby: Yeah, I guess it's fine to defer to the endpoint

apps/veil/src/pages/tournament/ui/lp-rewards.tsx
JasonMHasperhoven: That's nice :+1:

cronokirby (approved):

apps/veil/src/pages/tournament/ui/lp-rewards.tsx
```diff
@@ -151,20 +127,25 @@ export const LpRewards = observer(() => {
...
-              <LpRewardRow key={index} lpReward={lpReward} umMetadata={umMetadata} />
-            ))
+          {!loading && !total && (
+            <div className='col-span-5 text-sm text-muted-foreground py-4'>
+              No LP rewards found for this account.
+            </div>
           )}
+
+          {loading &&
+            new Array(BASE_LIMIT).fill({}).map((_, index) => <LoadingRow cells={5} key={index} />)}
```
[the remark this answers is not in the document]
VanishMax: I'd leave it for a parent component to deal with, at least for now

PR #2367, opened 2025-05-14
fix: only allow voting in current results when able

### LINKED ISSUE #2357: LQT round page: hide 'vote' column in current results table
Hide 'vote' column in the "current results table" if:
1. User voted already
2. The epoch is over

Image

vacekj:
fixes #2357

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-tournament-votes.ts
+++ b/apps/veil/src/pages/tournament/api/use-tournament-votes.ts
@@ -1,4 +1,4 @@
-import { useQuery } from '@tanstack/react-query';
+import { useQuery, UseQueryResult } from '@tanstack/react-query';
 import { ViewService } from '@penumbra-zone/protobuf/penumbra/view/v1/view_connect';
 import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import { connectionStore } from '@/shared/model/connection';
@@ -6,11 +6,14 @@ import { penumbra } from '@/shared/const/penumbra';
 import { useRefetchOnNewBlock } from '@/shared/api/compact-block';
 import { TournamentVotesResponse_Vote } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
 
-export const useTournamentVotes = (epoch?: number, disabled?: boolean) => {
+export const useTournamentVotes = (
+  epoch?: number,
+  disabled?: boolean,
+): UseQueryResult<TournamentVotesResponse_Vote[]> => {
   const { connected, subaccount } = connectionStore;
   const enabled = connected && !!epoch;
 
-  const query = useQuery<TournamentVotesResponse_Vote[]>({
+  const query = useQuery({
     enabled,
     queryKey: ['tournament-votes', subaccount, epoch],
     staleTime: Infinity,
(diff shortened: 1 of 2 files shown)


apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -48,8 +49,13 @@ export const CurrentVotingResults = observer(({ epoch }: CurrentVotingResultsPro
     sortDirection: sortBy.direction,
   });
 
-  // TODO: `canVote` should be true when connected and has delUM for this epoch
-  const canVote = connected;
+  const { isVoted, isDelegated, isEnded } = useVotingInfo();
```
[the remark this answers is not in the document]
vacekj: afaik useVotingInfo uses the current epoch by default, which is what we want here.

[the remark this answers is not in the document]
VanishMax: But why? We're talking about the round page for a specific epoch. Then we check if a user can vote for this specific, not current, epoch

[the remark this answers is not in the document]
vacekj: You're right - my mistake. Will fix.

apps/veil/src/pages/tournament/api/use-tournament-votes.ts
```diff
@@ -4,12 +4,13 @@ import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys
...
 import { penumbra } from '@/shared/const/penumbra';
 import { useRefetchOnNewBlock } from '@/shared/api/compact-block';
+import { TournamentVotesResponse_Vote } from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
 
 export const useTournamentVotes = (epoch?: number, disabled?: boolean) => {
   const { connected, subaccount } = connectionStore;
   const enabled = connected && !!epoch;
 
-  const query = useQuery({
+  const query = useQuery<TournamentVotesResponse_Vote[]>({
```
TalDerei, line 13: comment: proto type is infered here, doesn't need to be strictly typed

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
```diff
@@ -48,8 +49,13 @@ export const CurrentVotingResults = observer(({ epoch }: CurrentVotingResultsPro
...
+
+  /* A user can vote if they:
+   * - have enough delUM
+   * - epoch hasn't ended yet
+   * - user hasn't voted already for this epoch */
+  const canVote = connected && !isEnded && isDelegated && !isVoted;
```
TalDerei, line 58: comment: vote button seems to disappear and then pop back up after epoch has ended.

----------

Screenshot 2025-05-15 at 2 27 23 PM

TalDerei (changes requested):

apps/veil/src/pages/tournament/api/use-tournament-votes.ts
cronokirby: I would also rather see the top level function annotated with `UseQueryResult<...>` so that you can see this without inspecting the function body.

TalDerei: > I would also rather see the top level function annotated with `UseQueryResult<...>` so that you can see this without inspecting the function body.

impartial; this isn't blocking for me

apps/veil/src/pages/tournament/ui/round/current-voting-results/index.tsx
vacekj: Fixed by passing in the correct epoch.

TalDerei (approved):

TalDerei:
👍

PR #2368, opened 2025-05-15
Wire up specific delegator page

### LINKED ISSUE #2343: LQT: delegator reward views don't filter by delegator
**Describe the bug**
When I access the LQT Tournament page and try to inspect the historical rewards for a specific delegator, I instead see historical rewards for all delegators, including from epochs that the delegator I chose did not vote in.

**To Reproduce**

1. Navigate to LQT `/tournament` page on testnet Veil
2. Scroll down to the "Delegator Leaderboard"
3. Choose a delegator that has a very smallnumber of votes, e.g. ~1
4. Click that delegator row to bring up a specific view; [example URL](https://dex-explorer.testnet.plinfra.net/tournament/delegator/penumbra10y652m6339l5quq87nl6zu23hju3gttplncpuhu90hdwwpx29a4dj35s7vlppdruhnj4l0c7ffsfjrpe6c8gujpx64yx7re0822rnnxe686uel5wjfsjkdvy6wkgx5pkhyxmwx)

**Expected behavior**
I expected to see only a single result because the delegator chosen had only voted once. Instead, I see rewards listing from `Epoch #1`.

**Screenshots**

Image

**Additional context**
N/A

cronokirby:
## Description of Changes

Closes #2343.

The logic here ended up needing to be substantially different than the existing code path, which was tightly coupled to fetching rewards for a specific delegator.

I also took the freedom to do some light refactoring, and to try out "use server" as a way to avoid the tedium of writing out an API route. I think it works pretty nicely!


cronokirby:
image

apps/veil/src/pages/tournament/ui/delegator-table/delegator-rewards.tsx
```diff
@@ -21,39 +51,134 @@ export const DelegatorRewards = () => {
             Total Rewards Earned
           </Text>
           <Text small color='text.secondary'>
-            Cumulative from all epochs, voting and LPs rewards
+            Cumulative from all epochs, and voting rewards
```
VanishMax, line 54: question: shouldn't it be "Cumulative voting rewards (in UM) from all epochs" like in the design?

apps/veil/src/pages/tournament/ui/delegator-table/delegator-rewards.tsx
```diff
@@ -21,39 +51,134 @@ export const DelegatorRewards = () => {
...
+
+          const metadata = getMetadata(vote.asset);
+          const valueView = toValueView(
+            metadata
+              ? { metadata, amount: new Amount({}) }
+              : { assetId: vote.asset, amount: new Amount({}) },
+          );
+
+          return (
+            <div key={rowKey} className='grid grid-cols-subgrid col-span-4'>
```
VanishMax, line 141: todo: the whole row should be a link to the `/tournament/${epoch}` page with a highlighted background on hover

apps/veil/src/pages/tournament/ui/delegator-table/delegator-rewards.tsx
```diff
@@ -21,39 +51,134 @@ export const DelegatorRewards = () => {
...
+  const [page, setPage] = useState(BASE_PAGE);
+  const [limit, setLimit] = useState(BASE_LIMIT);
+  const { getTableHeader, sortBy } = useSortableTableHeaders<SortKey>('epoch', 'desc');
 
-        {tab === 'voting' && <VotingRewards />}
+  const onLimitChange = (l: number) => {
+    setLimit(l);
+    setPage(BASE_PAGE);
+  };
+  const { data, isPending, error } = useSpecificDelegatorRewards({
```
VanishMax, line 74: praise: wow, calling Pindexer without writing an endpoint manually just works!

thoughts (non-blocking, only trying to udnerstand): This API here looked like magic to me, thanks for this new perspective. Previously, I was trying SSR multiple times, and each time it turned out to be a real pain. The magic of this API is that it's actually not the SSR, it's [a server action](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) with only job is to save some lines of code of creating and calling the endpoint.

Under the hood it creates a POST endpoint for the same page route as we have in NextJs. And calling the function sends a request to this endpoint from client-side environment, what a great idea!

apps/veil/src/pages/tournament/server/specific-delegator-rewards.ts
```diff
@@ -0,0 +1,98 @@
...
+      // TODO: upstream this into pindexer.
+      eb =>
+        sql<number>`power / ${eb.fn.sum('power').over(ob => ob.partitionBy('epoch'))}`.as('share'),
+    ])
+    .where('address', '=', Buffer.from(address.inner))
+    .orderBy(sortKey ?? 'epoch', sortDirection ?? 'desc')
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Kysely limitation
+    .$if(limit !== undefined, qb => qb.limit(limit!))
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Kyseley limitation
+    .$if(page !== undefined && limit !== undefined, qb => qb.offset(page! * limit!))
```
VanishMax, line 54: question: why not giving default values to limit & page?

VanishMax (changes requested):
Requesting changes only for some rendering issues labeled with `todo` comment. The rest is an inspiring work!

apps/veil/src/pages/tournament/ui/delegator-table/delegator-rewards.tsx
cronokirby: Yeah, absolutely!

apps/veil/src/pages/tournament/server/specific-delegator-rewards.ts
cronokirby: For the page, you could make it 0. For the limit, there's no obvious value. Infinity doesn't work well, because then you'd need to do 0 * Infinity, which is NaN

conorsch:
Drove this around a bit in the dev env, and it resolves the problem I reported in https://github.com/penumbra-zone/web/issues/2343. One discrepancy I noticed while testing is that there's at least one example of a historical delegator that allegedly participated twice, yet clicking through shows only a single event:

[lqt-2-delegators-only-1-result.webm](https://github.com/user-attachments/assets/3951cb6a-a47b-47fc-8b72-645bb889c853)

But the issue as described is definitely resolved.

cronokirby:
> Drove this around a bit in the dev env, and it resolves the problem I reported in #2343. One discrepancy I noticed while testing is that there's at least one example of a historical delegator that allegedly participated twice, yet clicking through shows only a single event:
> lqt-2-delegators-only-1-result.webm
> 
> But the issue as described is definitely resolved.

In this video, I see two different epochs they've received a reward in?

cronokirby:
> > In this video, I see two different epochs they've received a reward in? This matches what I'd expect with "rounds participated: 2"
> 
> Oops, that video was truncated, and only showed the working entry of 2. Here's just the quirky one:
> lqt-2-delegators-only-1-result-fixed.webm

image

This seems to actually be expected behavior, for that particular delegator.
They voted twice, but they only received rewards once.

PR #2372, opened 2025-05-15
storage: store remote epoch

TalDerei:
## Description of Changes

Fresh wallets leverage instant sync by consuming the latest frontier, and therefore need to persist the latest remote epoch fetched from the full node. This provides an auxiliary helper storage method to do that. 

## Related Issue

pairs with https://github.com/prax-wallet/prax/pull/346

### DIFF
--- a/packages/types/src/indexed-db.ts
+++ b/packages/types/src/indexed-db.ts
@@ -111,8 +111,7 @@ export interface IndexedDbInterface {
     newState: PositionState,
     subaccount?: AddressIndex,
   ): Promise<void>;
-  addRemoteEpoch(startHeight: bigint, epochIndex: bigint): Promise<void>;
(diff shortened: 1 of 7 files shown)


packages/types/src/indexed-db.ts
```diff
@@ -111,6 +111,7 @@ export interface IndexedDbInterface {
     newState: PositionState,
     subaccount?: AddressIndex,
   ): Promise<void>;
+  addRemoteEpoch(startHeight: bigint, epochIndex: bigint): Promise<void>;
   addEpoch(startHeight: bigint): Promise<void>;
```
turbocrime, line 115: comment: i think we could easily just update the `addEpoch` method instead of maintaining both.

turbocrime (approved):

PR #2373, opened 2025-05-16
services: make tests use type-correct idb mock

turbocrime:
Replaces the arbitrarily-declared `IndexedDbMock` in the services package with a preconstructed mock object that enforces the actual interface type.

This still isn't ideal, because it's still not using a real database underneath. But, this change creates no functional diff.

This is important because now the mock no longer has to be manually maintained in parallel to the storage interface - it updates automatically when the type updates, requiring the tests to also update.

### DIFF
--- a/packages/services/src/sct-service/epoch-by-height.test.ts
+++ b/packages/services/src/sct-service/epoch-by-height.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { epochByHeight } from './epoch-by-height.js';
-import { IndexedDbMock, MockServices } from '../test-utils.js';
+import { mockIndexedDb, MockServices } from '../test-utils.js';
 import { createContextValues, createHandlerContext, HandlerContext } from '@connectrpc/connect';
 import { SctService } from '@penumbra-zone/protobuf';
 import { servicesCtx } from '../ctx/prax.js';
@@ -12,15 +12,12 @@ import type { ServicesInterface } from '@penumbra-zone/types/services';
 
 describe('EpochByHeight request handler', () => {
   let mockServices: MockServices;
-  let mockIndexedDb: IndexedDbMock;
+
   let mockCtx: HandlerContext;
 
   beforeEach(() => {
     vi.resetAllMocks();
 
-    mockIndexedDb = {
-      getEpochByHeight: vi.fn(),
-    };
     mockServices = {
       getWalletServices: vi.fn(() =>
         Promise.resolve({ indexedDb: mockIndexedDb }),
@@ -41,7 +38,7 @@ describe('EpochByHeight request handler', () => {
   it('returns an `EpochByHeightResponse` with the result of the database query', async () => {
     const expected = new Epoch({ startHeight: 0n, index: 0n });
 
-    mockIndexedDb.getEpochByHeight?.mockResolvedValue(expected);
+    mockIndexedDb.getEpochByHeight.mockResolvedValue(expected);
     const req = new EpochByHeightRequest({ height: 0n });
 
     const result = await epochByHeight(req, mockCtx);
--- a/packages/services/src/stake-service/validator-info.test.ts
+++ b/packages/services/src/stake-service/validator-info.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { validatorInfo } from './validator-info.js';
-import { IndexedDbMock, MockServices } from '../test-utils.js';
+import { mockIndexedDb, MockServices } from '../test-utils.js';
 import { createContextValues, createHandlerContext, HandlerContext } from '@connectrpc/connect';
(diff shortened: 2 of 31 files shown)


packages/services/src/view-service/tournament-votes.test.ts
```diff
@@ -74,11 +67,12 @@ const epoch = new Epoch({
...
 const mockVote = {
+  incentivizedAsset: new AssetId({}),
   id: 'test-uuid-or-any-string',
-  epoch: 100n,
-  TransactionId: {
+  epoch: '100',
+  TransactionId: new TransactionId({
     inner: new Uint8Array([1, 2, 3, 4]),
-  },
+  }),
```
turbocrime, line 75: this change demonstrated that the `mockVote` here contained the wrong type. cc @TalDerei

packages/services/src/sct-service/epoch-by-height.test.ts
turbocrime, line 1: large diff, but mostly simple/automated removals as seen in this file

packages/services/src/stake-service/validator-info.test.ts
```diff
@@ -32,21 +32,13 @@ describe('ValidatorInfo request handler', () => {
...
+      ]);
     });
-    mockIterateValidatorInfos.next.mockResolvedValueOnce({
-      value: mockValidatorInfoResponse2.validatorInfo,
-    });
-    mockIterateValidatorInfos.next.mockResolvedValueOnce({ done: true });
 
-    mockIndexedDb = {
-      iterateValidatorInfos: () => mockIterateValidatorInfos,
-    };
```
turbocrime, line 49: the strange partial-iterator recursive mocks are now just much simpler generator implementations

packages/services/src/test-utils.ts
turbocrime, line 1: the mock is manually defined here, and its type is no longer arbitrary. every method is present, and will throw `expect.unreachable()` unless a test explicitly enables it by providing a mock implementation.

the use of 'any' within the stub mocks is okay because it doesn't disturb the interface type. test-provided implementations are still asked to meet the interface.

packages/services/src/view-service/asset-metadata-by-id.test.ts
```diff
@@ -6,25 +6,21 @@ import {
...
   let mockShieldedPool: ShieldedPoolMock;
   let request: AssetMetadataByIdRequest;
 
   beforeEach(() => {
     vi.resetAllMocks();
 
-    mockIndexedDb = {
-      getAssetsMetadata: vi.fn(),
-      saveAssetsMetadata: vi.fn(),
-    };
```
turbocrime, line 27: tests no longer have to repeatedly define partial mocks

packages/services/src/view-service/asset-metadata-by-id.test.ts
```diff
@@ -55,15 +51,15 @@ describe('AssetMetadataById request handler', () => {
   });
 
   test('should successfully respond with metadata when idb record is present', async () => {
-    mockIndexedDb.getAssetsMetadata?.mockResolvedValue(metadataFromIdb);
+    mockIndexedDb.getAssetsMetadata.mockResolvedValue(metadataFromIdb);
```
turbocrime, line 54: optional access could silently fail, and mislead a developer into thinking they were configuring a method that was not actually mocked. so now, access is guaranteed and it's not possible to 'optionally' configure a mock.

packages/services/src/view-service/fees.test.ts
```diff
@@ -64,9 +52,8 @@ describe('extractAltFee', () => {
       }),
     ];
 
-    mockIndexedDb.saveGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.getAltGasPrices?.mockResolvedValueOnce(gasPrices);
-    mockIndexedDb.hasTokenBalance?.mockResolvedValue(true);
+    mockIndexedDb.getAltGasPrices.mockResolvedValueOnce(gasPrices);
+    mockIndexedDb.hasTokenBalance.mockResolvedValue(true);
```
turbocrime, line 56: it doesn't actually make sense to mock a return value for `saveGasPrices`, and that was correctly rejected by the new mock type

packages/services/src/view-service/latest-swaps.test.ts
```diff
@@ -68,38 +57,35 @@ describe('LatestSwaps request handler', () => {
...
-      }
-      mockIterateSwaps.next.mockResolvedValueOnce({
-        done: true,
-      });
-    };
+    /*
+      mockIndexedDb.iterateSwaps.mockImplementationOnce(async function* () { yield* await Promise.resolve([
+
+      ]); });
+      */
```
turbocrime, line 64:
```suggestion
```

oops

TalDerei (approved):
just missing a changeset, otherwise looks good to me 👍

TalDerei:
@turbocrime any blockers on this? looks good to land to me, this is a good improvement. What would be required to use the same db underneath?

PR #2374, opened 2025-05-16
suggestions for store remote epoch

turbocrime:
suggestions for https://github.com/penumbra-zone/web/pull/2372

instead of defining a new method, updates the definition of the existing method, and improves retrieval without updating schema.

rebased on top of the services mock PR to ensure consistency and correct use within services. PR branch targets the mock branch to show the functional diff.

### DIFF
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -820,6 +820,21 @@ export class IndexedDb implements IndexedDbInterface {
   }
 
   /**
+   * Adds a new remote epoch from full node with the given start height.
+   */
+  async addRemoteEpoch(startHeight: bigint, epochIndex: bigint): Promise<void> {
+    const remoteEpoch = {
+      startHeight: startHeight.toString(),
+      index: epochIndex.toString(),
+    };
+
+    await this.u.update({
+      table: 'EPOCHS',
+      value: remoteEpoch,
+    });
+  }
+
+  /**
    * Adds a new epoch with the given start height. Automatically sets the epoch
    * index by finding the previous epoch index, and adding 1n.
    */
--- a/packages/storage/src/indexed-db/index.ts
+++ b/packages/storage/src/indexed-db/index.ts
@@ -59,7 +59,6 @@ import type {
   ScanBlockResult,
   StateCommitmentTree,
 } from '@penumbra-zone/types/state-commitment-tree';
-import { sctPosition } from '@penumbra-zone/wasm/tree';
 import {
   AuctionId,
   DutchAuctionDescription,
@@ -225,7 +224,7 @@ export class IndexedDb implements IndexedDbInterface {
 
     this.addSctUpdates(txs, updates.sctUpdates);
     this.addNewNotes(txs, updates.newNotes);
-    await this.addNewSwaps(txs, updates.newSwaps, updates.height);
+    this.addNewSwaps(txs, updates.newSwaps);
     txs.add({ table: 'FULL_SYNC_HEIGHT', value: updates.height, key: 'height' });
 
     await this.u.updateAll(txs);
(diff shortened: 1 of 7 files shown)


packages/storage/src/indexed-db/index.ts
```diff
@@ -820,83 +819,80 @@ export class IndexedDb implements IndexedDbInterface {
...
-      index: index.toString(),
-    };
-
-    await this.u.update({
+  async addEpoch(epoch: Epoch): Promise<void> {
+    await this.db.put('EPOCHS', {
       table: 'EPOCHS',
-      value: newEpoch,
+      value: epoch.toJson() as Jsonified<Epoch>,
+      key: Number(epoch.index),
```
TalDerei, line 828: comment: intuitively raised a red flag at first glance – we’re specifying a key manually on a table that doesn’t have an explicit `keyPath` defined in the object store (instead relying on `autoIncrement`). Additionally, reference the concern for `Jsonified<Epoch>` on genesis (https://github.com/penumbra-zone/web/pull/2242/files#r2089880545). Anyways, confirmed my suspicion after inspecting the db: the epoch indices are incorrect. 

--------------

Screenshot 2025-05-15 at 6 52 23 PM

----------------

something like this also leads to the same index duplication:

```ts
  async addEpoch(startHeight: bigint, epochIndex: bigint): Promise<void> {
    const tx = this.db.transaction('EPOCHS', 'readwrite');

    await tx.store.put({
      startHeight: startHeight.toString(),
      index: epochIndex.toString(),
    });
  }
``` 

----------

Screenshot 2025-05-15 at 6 56 36 PM

TalDerei: however, this construction seems to be working properly:

```ts
  async addEpoch(startHeight: bigint, epochIndex: bigint): Promise<void> {
    const tx = this.db.transaction('EPOCHS', 'readwrite');

    await tx.store.put(
      {
        startHeight: startHeight.toString(),
        index: epochIndex.toString(),
      },
      Number(epochIndex)
    );    
  }
```

--------------

Screenshot 2025-05-15 at 7 00 25 PM

TalDerei: but it's not a jsonified epoch like you wanted for consistency

TalDerei: the way you're doing it doesn't match the definition of `this.db.put()` which expects (`table: string, value: any, key?: IDBValidKey)`, so I think instead of

```ts
await this.db.put('EPOCHS', {
  table: 'EPOCHS',
  value: epoch.toJson() as Jsonified<Epoch>,
  key: Number(epoch.index),
});
```

it should instead be 

```ts
await this.db.put('EPOCHS', epoch.toJson() as Jsonified<Epoch>, Number(epoch.index));
```

the former seems to be doing some weird object wrapping?

packages/storage/src/indexed-db/index.ts
```diff
@@ -820,83 +819,80 @@ export class IndexedDb implements IndexedDbInterface {
...
-    let epoch: Epoch | undefined;
-
+  async getEpochByIndex(epochIndex: bigint): Promise<Epoch | undefined> {
     // Iterate over epochs and return the one with the matching epoch index.
-    for await (const cursor of this.db.transaction('EPOCHS', 'readonly').store) {
-      const currentEpoch = Epoch.fromJson(cursor.value);
-      if (currentEpoch.index === epoch_index) {
-        epoch = currentEpoch;
-        break;
+    for await (const cursor of this.db.transaction('EPOCHS').store.iterate(null, 'prev')) {
```
TalDerei, line 889: comment: prev is more efficient iterator

packages/storage/src/indexed-db/index.ts
```diff
@@ -1053,35 +1049,20 @@ export class IndexedDb implements IndexedDbInterface {
     // TODO: What about updates.delete_ranges (https://github.com/penumbra-zone/web/issues/818)?
   }
 
-  private addNewNotes(txs: IbdUpdates, notes: SpendableNoteRecord[]): void {
-    for (const n of notes) {
-      assertCommitment(n.noteCommitment);
-      txs.add({ table: 'SPENDABLE_NOTES', value: n.toJson() as Jsonified<SpendableNoteRecord> });
+  private addNewNotes(txs: IbdUpdates, newNotes: SpendableNoteRecord[]): void {
```
TalDerei, line 1052: comment: we'll fold `addNewSwaps` changes into https://github.com/penumbra-zone/web/pull/2242 (eg. directly passing in epoch index and removing genesis epoch fallback)

packages/storage/src/indexed-db/index.ts
```diff
@@ -820,83 +819,80 @@ export class IndexedDb implements IndexedDbInterface {
...
-  async getEpochByHeight(
-    /**
-     * The block height to query by. Will return the epoch with the largest
-     * start height smaller than `height` -- that is, the epoch that contains
-     * this height.
-     */
-    height: bigint,
-  ): Promise<Epoch | undefined> {
-    let epoch: Epoch | undefined;
+  async getEpochByHeight(height: bigint): Promise<Epoch | undefined> {
```
TalDerei, line 839: comment: logic is more complicated for me to reason about because of prev iterator, defering for now

TalDerei:
@turbocrime I implemented a modified version of some of the suggestions in https://github.com/penumbra-zone/web/pull/2372. We can fold in follow-up optimizations as part of https://github.com/penumbra-zone/web/pull/2242 once we spend more time reasoning through those changes.

packages/storage/src/indexed-db/index.ts
turbocrime: this approach is the minimal diff with no modified schema. the format in the table is unchanged.

the table isn't configured with a keypath, just an autoincrementing index, and an autoincrementing index is always an integer.

this is definitely the most questionable part of the suggestions, i haven't completely exercised it, so maybe just go with the separate method.

packages/storage/src/indexed-db/index.ts
turbocrime: well, in some cases, and only because it's likely to be queried with a recent epoch, which may be later in the table. and maybe not if the cache has been reset and the other method implementations are used (in which case autoincrementing index may leave some epochs above present index height)

packages/storage/src/indexed-db/index.ts
turbocrime, line 828:
```suggestion
```

keep auto-incrementing and this eliminates some of the concerns

still inferior to a migration

Issue #2273, opened 2025-04-22
Route book shows incorrect data (on testnet)


cronokirby:
**Describe the bug**
There's a clear mismatch between the results of simulation on minfront, which matches pcli:

![Image](https://github.com/user-attachments/assets/97497000-07dc-422c-baa5-05ef30e32058)

and between what the route book shows on Veil:

![Image](https://github.com/user-attachments/assets/2cd23316-cc26-4425-967c-c557586a1450)

**To Reproduce**
1. Open a position near the spread, with a fee, and with a large amount of liquidity, in both assets, on the GM : UM pair
2. Observe that it doesn't appear in the route book, but does appear on minifront.

**Expected behavior**
We should see the position appear. Also, I don't see why we should ever observe `NaN`. That seems like some other bug to me, like a division by 0 or something odd like that creeping through.

JasonMHasperhoven:
I've looked into this:

<img width="759" alt="Image" src="https://github.com/user-attachments/assets/85bf5367-85d1-472e-99c9-96a4e996d890" />

<img width="333" alt="Image" src="https://github.com/user-attachments/assets/0c53c80f-c63e-4d8e-ac01-232d2ca688a4" />

It only appears to be different, but when you calculate the prices you'll see that it matches:

First entry minifront to first entry (from midprice) in veil
1/0.999026 = 1.0009749496009 = 1.000975

Fixed the NaN caused by commas in the value:
- https://github.com/penumbra-zone/web/pull/2383

PR #2383, opened 2025-05-16
Fix Route Book

JasonMHasperhoven:
## Description of Changes

Closes https://github.com/penumbra-zone/web/issues/2273

Fixes wrong amounts caused by a comma in the string (which cant be parsed properly).

## Related Issue

### DIFF
--- a/apps/veil/src/shared/api/server/book/helpers.ts
+++ b/apps/veil/src/shared/api/server/book/helpers.ts
@@ -118,12 +118,15 @@ export const processSimulation = ({
     : sortedTraces.slice(-limit).reverse();
 
   let cumulativeTotal = new BigNumber(0);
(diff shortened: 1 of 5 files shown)


apps/veil/src/shared/api/server/book/helpers.ts
```diff
@@ -65,7 +65,9 @@ export const buildTrace = (
 
   return {
     price: removeTrailingZeros(price),
-    amount: pnum(baseValueView).toFormattedString(),
+    amount: pnum(baseValueView).toFormattedString({
+      commas: false,
```
JasonMHasperhoven, line 69: This solves the problem of showing wrong amounts & NaN in the route book.

TalDerei (approved):
👍

PR #2385, opened 2025-05-16
Handle vote revelation correctly

### LINKED ISSUE #2266: Use LQT-specific randomizer as the stable voting address
If a user requests a stable voting address, we should use a specific fixed randomizer, and not just "0".

We should also make sure that when viewing this address (at least in the context of the LQT, but probably simplest to just do in the AddressView component) that we handle this case differently than any other. The name isn't important for now, but something like "Lqt Voting Address" "Tournament Address" etc. vs "IBC Deposit Address", which is what I think we would get now.

### LINKED ISSUE #2333: lqt: votes are always revealed
**Describe the bug**
https://github.com/penumbra-zone/web/blob/7c8e30ecaad601a5714f2c8b4349bd56601e47bf/apps/veil/src/pages/tournament/ui/vote-dialog/index.tsx#L33 Looking at the code, `revealVote` is never actually used to construct an ephemeral vs a default address, so you always reveal your vote!

**To Reproduce**
You can observe the effects of this, as you don't see several entries even if you uncheck revealing your vote.

**Expected behavior**
Choosing to not reveal your vote should mean you use an ephemeral address.

cronokirby:
## Description of Changes

- Uses a specific LQT randomizer if vote is revealed.
- Adjusts UI components to render it correctly.
- Uses an ephemeral randomizer if vote is not revealed.
- Fixes a bug in the view service where getting the address by index would ignore the randomizer.

## Related Issue

Closes #2266.
Closes #2333.

### DIFF
--- a/.changeset/beige-cloths-try.md
+++ b/.changeset/beige-cloths-try.md
@@ -1,0 +1,11 @@
+---
+'@penumbra-zone/wasm': major
+'@penumbra-zone/services': patch
+---
+
+Make the view service request for mapping indices to addresses take the randomizer into account.
+
+The WASM package has a breaking change in that the `get_address_by_index` function now
+_requires_ you to pass a randomizer.
+This can be an empty slice, indicating a randomizer consistent of all 0s.
+Clients should upgrade by adding a `&[]` parameter, which will preserve their current behavior.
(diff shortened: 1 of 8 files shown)


TalDerei (approved):
👍

TalDerei:
@cronokirby can we add a changeset here?

PR #2388, opened 2025-05-17
fix: don't redirect to dashboard after connecting a wallet

### LINKED ISSUE #2375: delegate button should redirect to the staking page
currently, the delegate button in veil redirects to the minifront homepage instead of taking users directly to the staking page.

vacekj:
now it just stays on the current page

useful for redirecting to minifront staking from veil, without a wallet connected

fixes #2375

### DIFF
--- a/apps/minifront/src/components/extension-not-connected.tsx
+++ b/apps/minifront/src/components/extension-not-connected.tsx
@@ -63,7 +63,7 @@ export const ExtensionNotConnected = () => {
   const connect = async (provider: string) => {
     try {
       await penumbra.connect(provider);
-      navigate('/');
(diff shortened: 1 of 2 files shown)


TalDerei:
@vacekj there doesn't seem to be a good way to to this. currently, the redirection will _already_ work if wallet is connected to the frontend provider, otherwise it redirects to the homepage.

TalDerei:
> @vacekj there doesn't seem to be a good way to to this. currently, the redirection will _already_ work if wallet is connected to the frontend provider, otherwise it redirects to the homepage.

I was testing on `app.penumbra.zone` rather than localhost, this actually does seem to work!

PR #2390, opened 2025-05-17
feat: implement caching for registry data in epoch results

### LINKED ISSUE #2310: throttle network connectivity
we should be more explicit about performance under slower network conditions like public wifi and evaluate how veil behaves under network throttling. In that context, the motivation is that fetching the remote registry (this is just one example for instance) can occasionally timeout inside some of our server-side API hooks, eg. [here](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/server/epoch-results.ts#L104-L110). 

this will surface a 400 error in the client logs, and display the following warning in the server. 

-------------------

Image

Image

Image

vacekj:
following https://github.com/prax-wallet/registry/pull/146, pass in 'force-cache' option to the registry client to get server-side caching.

### DIFF
--- a/apps/minifront/package.json
+++ b/apps/minifront/package.json
@@ -22,7 +22,7 @@
     "@cosmos-kit/core": "^2.15.0",
     "@cosmos-kit/react": "^2.20.1",
     "@interchain-ui/react": "^1.26.1",
-    "@penumbra-labs/registry": "^12.7.0",
+    "@penumbra-labs/registry": "^12.8.0",
     "@penumbra-zone/bech32m": "workspace:*",
     "@penumbra-zone/client": "workspace:*",
     "@penumbra-zone/crypto-web": "workspace:*",
--- a/apps/veil/package.json
+++ b/apps/veil/package.json
@@ -34,7 +34,7 @@
     "@grpc/proto-loader": "^0.7.13",
     "@interchain-ui/react": "^1.26.1",
     "@keplr-wallet/types": "^0.12.199",
-    "@penumbra-labs/registry": "^12.7.0",
+    "@penumbra-labs/registry": "^12.8.0",
     "@penumbra-zone/bech32m": "workspace:*",
     "@penumbra-zone/client": "workspace:*",
     "@penumbra-zone/crypto-web": "workspace:*",
--- a/apps/veil/src/pages/tournament/server/epoch-results.ts
+++ b/apps/veil/src/pages/tournament/server/epoch-results.ts
@@ -1,6 +1,6 @@
 import { sql } from 'kysely';
 import { NextRequest, NextResponse } from 'next/server';
-import { ChainRegistryClient } from '@penumbra-labs/registry';
+import { ChainRegistryClient, Registry } from '@penumbra-labs/registry';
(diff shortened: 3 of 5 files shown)


TalDerei:
@vacekj https://github.com/penumbra-zone/web/issues/2310 should've modified accordingly after discussions that surfaced out of https://github.com/penumbra-zone/web/pull/2336. That intent of that issue was more along the lines of: "observe how the app behaves under slow network conditions and flag any resulting issues."

currently, `fetchJsonRegistryWithGlobals` runs on every SSR page request, always fetching the latest registry bundle per request. That's what we want to avoid holding a stale registry. Client nav on the _same_ page caches the registry, but it does refetch across page reloads. That said, the bundle is relatively compact (~200 kb uncompressed), and we can likely rely on some kind of http request caching there? cc @cronokirby

cronokirby:
> @vacekj #2310 should've been modified accordingly after discussions that surfaced out of #2336. That intent of that issue was more broadly along the lines of: "observe how the app behaves under slow network conditions and flag any resulting issues."
> 
> currently, `fetchJsonRegistryWithGlobals` runs on every SSR page request, always fetching the _latest_ registry bundle per request. That's what we want to avoid holding a stale registry. Client nav on the _same_ page caches the registry, but it does refetch across page reloads. That said, the bundle is relatively compact (~200 kb uncompressed), and we can likely rely on some kind of http request caching there? Implementing some kind of caching layer would be more involved to my understanding. Defering to @cronokirby

Yeah, my thinking here is that we want the registry client to implementing caching, so that any consumer of the registry library benefits.
Ideally, this would be done leaning on HTTP caching, that way the choices we make there are controlled by the actual deployment we choose for the registry resource.
Then consuming code can "just" fetch a remote registry always, and we can use the existing code in `https://github.com/penumbra-zone/web/tree/apps/veil/src/shared/api-fetch-registry.ts` to avoid re-writing this, and to have a shared client.

cronokirby:
c.f. https://github.com/penumbra-zone/web/issues/2310#issuecomment-2889215482 as well ; doing caching on the server in this case should be about reducing latency, which doesn't help #2310.

TalDerei:
> c.f. [#2310 (comment)](https://github.com/penumbra-zone/web/issues/2310#issuecomment-2889215482) as well ; doing caching on the server in this case should be about reducing latency, which doesn't help #2310.

let's closed this as unplanned for now?

cronokirby:
> > c.f. [#2310 (comment)](https://github.com/penumbra-zone/web/issues/2310#issuecomment-2889215482) as well ; doing caching on the server in this case should be about reducing latency, which doesn't help #2310.
> 
> let's closed this as unplanned for now?

Yeah, I think that ticket is not quite specified enough, so closing is fine.
We should keep performance in mind, and spend time doing a comprehensive audit of things, to figure out what's worth squashing and in what order.

cronokirby:
https://github.com/prax-wallet/registry/pull/146 merging this would allow us to just modify:
https://github.com/penumbra-zone/web/blob/b78a73f7c59f7c9f81d092f38034128bc5ec6818/apps/veil/src/shared/api/fetch-registry.ts#L12 to pass in the options, and then get caching.
Then, we could make sure all the server endpoints use the shared client.

cronokirby (approved):

PR #2393, opened 2025-05-19
fix(veil): #2387: "ends in" clock fix

### LINKED ISSUE #2387: "ends in" clock on social card should include hours and days
needs to support formatting the longer epoch times (~2 days) on mainnet. Currently, this helper method is only designed for the shorter epoch times on testnet.

https://github.com/penumbra-zone/web/blob/main/apps/veil/src/shared/utils/format-time.ts

VanishMax:
Closes #2387 

Adds support for hours and days in the `formatTimeRemaining` function


TalDerei (approved):
looks good!

Issue #2386, opened 2025-05-17
rerendering voting button on landing card during sync


TalDerei:
behavior observed only when wallet is installed, but _disconnected_, during sync. Wondering if there was a regression somewhere that might have caused this rendering issue to resurface? Or is this something that's already tracked but just hasn't been addressed yet? Sanity check whether this behavior is real or hallucinated. 

--------------

https://github.com/user-attachments/assets/463112d5-74f3-4530-be0c-968cb4162322

PR #2400, opened 2025-05-20
Fix Perpetual Voting Buttons Loading state

JasonMHasperhoven:
## Description of Changes

Closes: https://github.com/penumbra-zone/web/issues/2386

Solution is to only render the loading state when the requests haven't been called yet using the `isFetched` prop.

## Related Issue

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-current-epoch.ts
+++ b/apps/veil/src/pages/tournament/api/use-current-epoch.ts
@@ -5,6 +5,7 @@ export const useCurrentEpoch = (onChange?: (newEpoch: number) => void) => {
   const {
     data: summary,
     isLoading,
+    isFetched,
     status,
   } = useTournamentSummary({
     limit: 1,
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
It doesn't seem to function properly across page reloads because the `isFetched` flag resets to false? Optimistically approving since it's still an improvement over the current behavior if you stay on the page, but we should be cautious about layering in more loading hook states which can spiral complexity. cc @cronokirby 

------------------

https://github.com/user-attachments/assets/fb6006b9-1f2b-4db5-9a3d-9fbb144c6e27

JasonMHasperhoven:
> It doesn't seem to function properly across page reloads because the `isFetched` flag resets to false? Optimistically approving since it's still an improvement over the current behavior if you stay on the page, but we should be cautious about layering in additional loading hook states which can spiral complexity. cc @cronokirby
> 
>  Screen.Recording.2025-05-20.at.12.28.35.PM.mov

@TalDerei That's fine, I believe the average user has no expectation to not see loading states on page reloads.

Issue #2399, opened 2025-05-19
My LPs tab reloads


TalDerei:
noticed some odd behavior when navigating to the "My LPs" section on the round page.

---------------------

https://github.com/user-attachments/assets/78259554-1b25-4242-a910-933887b09e85
https://github.com/user-attachments/assets/e37afd2c-1407-4300-a8a6-84ec93bdd64b

PR #2401, opened 2025-05-21
Fix My LP tab loading states

JasonMHasperhoven:
## Description of Changes

Closes https://github.com/penumbra-zone/web/issues/2399

Caused by updating blockHeights which was part of the query keys. The fix is to use a different approach that keeps the keys the same while refreshing.

## Related Issue

### DIFF
--- a/apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
+++ b/apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
@@ -4,7 +4,6 @@ import { apiPostFetch } from '@/shared/utils/api-fetch';
 import { penumbra } from '@/shared/const/penumbra';
 import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
 import { ViewService } from '@penumbra-zone/protobuf/penumbra/view/v1/view_connect';
-import { statusStore } from '@/shared/model/status';
 import { useCurrentEpoch } from '@/pages/tournament/api/use-current-epoch';
 import {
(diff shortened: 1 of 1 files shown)


apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
```diff
@@ -85,11 +82,8 @@ export const useMyLpLeaderboard = ({
         assetId,
       } as LpLeaderboardRequest);
     },
-    enabled:
-      typeof epoch === 'number' &&
-      positionIds !== undefined &&
-      isActive &&
-      (epoch === currentEpoch ? !!latestKnownBlockHeight : true),
+    enabled: typeof epoch === 'number' && positionIds !== undefined && isActive,
+    refetchInterval: epoch === currentEpoch ? 10000 : false,
```
TalDerei, line 86: comment 10,000ms?

TalDerei (approved):
Overall looks good! As part of this scope, could we also add a smooth transition when switching between "All LPs" and "My LPs"? It currently feels a bit janky.

------------

https://github.com/user-attachments/assets/4e5bd6c3-2322-45d2-84ba-56ebefaeebc4

apps/veil/src/entities/leaderboard/api/use-my-lp-leaderboard.ts
JasonMHasperhoven: Yes

TalDerei:
> not lazy loading it

yeah, we should be lazily loading the data, let's prerender to make this transition smoother.

PR #2408, opened 2025-05-22
chore: bump wasm deps to v2.0.0-alpha.11

conorsch:
## Description of Changes
Bumps the wasm dependencies to the latest LQT-related tag, which includes a rebase on `v1.5.2` changes. No anticipated breakage for web code; merely updating to keep things well in sync, in anticipation of final release.

Refs:

  * https://github.com/penumbra-zone/penumbra/issues/5203
  * https://github.com/penumbra-zone/penumbra/issues/5010

## Related Issue

See above.

I have done no such thing. Also, that reminds me: we're due for a minifront update in the protocol repo. Will do that again ahead of release.

### DIFF
--- a/packages/wasm/crate/Cargo.lock
+++ b/packages/wasm/crate/Cargo.lock
@@ -826,8 +826,8 @@ dependencies = [
 
 [[package]]
 name = "decaf377-fmd"
-version = "2.0.0-alpha.9"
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
👍 can we add a changeset?

PR #2409, opened 2025-05-22
hotfix: veil

VanishMax:
1. Correct account index comparison in Veil order form
2. CSS build correction for Veil – as a result breaks Minifront pages related to v2 only, so no users will be affected. Will revisit this later


TalDerei (approved):
👍

PR #2417, opened 2025-05-23
fix(veil): #2414: filter delegator rewards in the client side

### LINKED ISSUE #2414: client-side reward filtering
after https://github.com/penumbra-zone/web/pull/2385, we use a ephemeral randomizer when the vote _isn’t revealed_. Because of this, the current method of [requesting the address by index](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/api/use-personal-rewards.ts#L46-L48) and then [filtering by that single address](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/server/delegator-history.ts#L78) directly in pindexer won't work anymore. Consequently, voting in newer epochs without revealing the vote means the rewards won't be returned from pindexer and therefore won't appear in the UI under "My Total Rewards".

Instead, the backend should return all relevant rewards for each epoch we voted in and perform _client-side_ filtering to identify the reward that applies to us. 

For each reward, we need to perform a similar address query to determine if the reward applies to us:

```ts
    const { addressIndex } = await viewClient.indexByAddress({
      address: new Address({ altBech32m: bech32mAddress }),
    });

    if (!addressIndex) {
      return { isValidAddress: true, belongsToWallet: false };
    }
```

We should revert https://github.com/penumbra-zone/web/pull/2365 and migrate the previous filtering logic to the client-side accordingly.

VanishMax:
Closes #2414

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
+++ b/apps/veil/src/pages/tournament/api/use-personal-rewards.ts
@@ -1,4 +1,4 @@
-// File: src/pages/tournament/api/use-personal-rewards.ts
+import { useMemo } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { connectionStore } from '@/shared/model/connection';
 import { AddressIndex } from '@penumbra-zone/protobuf/penumbra/core/keys/v1/keys_pb';
@@ -25,8 +25,6 @@ export interface PersonalRewardsData {
 
 const fetchRewards = async (
   epochOrHeight: { type: 'epoch'; value: bigint } | { type: 'blockHeight'; value: bigint },
-  page: number = BASE_PAGE,
-  limit: number = BASE_LIMIT,
   sortKey: DelegatorHistorySortKey = 'epoch',
   sortDirection: DelegatorHistorySortDirection = 'desc',
   subaccount?: number,
@@ -66,8 +64,6 @@ const fetchRewards = async (
       '/api/tournament/delegator-history',
       {
         epochs: Array.from(epochs),
-        page,
-        limit,
         sortKey,
         sortDirection,
       },
@@ -109,7 +105,7 @@ export const usePersonalRewards = (
   const blockHeight = statusStore.latestKnownBlockHeight;
 
   const query = useQuery({
-    queryKey: ['total-voting-rewards', subaccount, page, limit, sortKey, sortDirection],
+    queryKey: ['total-voting-rewards', subaccount, sortKey, sortDirection],
     enabled:
       connectionStore.connected &&
       !!epoch &&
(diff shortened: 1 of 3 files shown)


apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -50,103 +50,81 @@ export const getBodyParams = async (
...
 const tournamentDelegatorHistoryQuery = async ({
   epochs,
   sortKey,
   sortDirection,
-  address,
-  limit,
-  page,
+  // todo: apply limit and page to the aggregated query below - probably needs a lateral join or a partition by
+  // limit,
+  // page,
```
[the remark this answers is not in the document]
cronokirby: The current query seems vastly over-complicated to me. All that should be needed on the server side is:
list of epochs -> list of (epoch, address, reward)
then, on the client side, you filter this down to (epoch, reward), figuring out which reward is yours in each address.
Then you link this with your personal voting information.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
```diff
@@ -57,28 +63,36 @@ const fetchRewards = async (
...
+    // filter the history by address – find user's subaccount index
+    const historyByAddress = await Promise.any<TournamentDelegatorHistoryResponse | undefined>(
+      delegatorHistory.map(async item => {
+        const index = await getIndexByAddress(item.address);
+        if (!index || index.account !== subaccount) {
+          throw new Error('Address index does not match');
+        }
+        return item;
+      }),
+    );
```
cronokirby, line 89: I think throwing here is not what we want. We should instead just filter, since we may have multiple addresses. Maybe using flatMap would be simplest, to do the merge instead.

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -15,17 +15,17 @@ export type DelegatorHistorySortDirection = (typeof DIRECTIONS)[number];
...
   limit?: number;
   page?: number;
   sortKey?: DelegatorHistorySortKey;
   sortDirection?: DelegatorHistorySortDirection;
 }
 
 export interface TournamentDelegatorHistoryResponse {
+  address: Address;
+  total_items: number;
+  total_rewards: number;
```
cronokirby, line 27: We should not let the database dictate our domain model. We should just map the result of the database query onto the convention here.

apps/veil/src/pages/tournament/server/delegator-history.ts
```diff
@@ -50,103 +50,81 @@ export const getBodyParams = async (
...
+      jsonArrayFrom(
+        eb
+          .selectFrom(filteredQuery.as('delegators'))
+          .selectAll()
+          .whereRef('delegators.address', '=', 'grouped.address')
+          .orderBy(sortKey, sortDirection)
+          .limit(limit)
+          .offset(limit * (page - 1)),
+      ).as('data'),
     ]);
```
cronokirby, line 98: This doesn't work because the pagination needs to happen across all addresses controlled by a given user. This assumes that each user controls one address, which is not correct.

We have to do this on the client side, in which case the db query should be pared down to just return address, power, reward.

cronokirby (changes requested):
I think the current logic is incorrect, returning the wrong result, in two ways. But, the rest looks good, so shouldn't be too hard to correct.

apps/veil/src/pages/tournament/api/use-personal-rewards.ts
TalDerei: we certainly shouldn't throwing here, and `Promise.any()` will only return the first matching address. Suggesting something like:

```ts
  const pairs = await Promise.all(
      delegatorHistory.map(item =>
        getIndexByAddress(item.address)    
          .then(index => ({ item, index })),  
      ),
    );

 const historyByAddress: TournamentDelegatorHistoryResponse[] = pairs
    .filter(({ index }) => index?.account === subaccount)
    .map(({ item }) => item);        
 ```
 
 and then the API hook should be returning `TournamentDelegatorHistoryResponse[]` objects rather than `LqtDelegatorHistoryData[]` for a single `TournamentDelegatorHistoryResponse`.

apps/veil/src/pages/tournament/server/delegator-history.ts
TalDerei: beat me to it, I wasn't sure why we were paginating here either.

TalDerei:
@VanishMax opened https://github.com/penumbra-zone/web/pull/2430 which addresses all the issues;  the only remaining concern is that the rewards still load slowly on first fetch.

TalDerei (approved):

TalDerei:
looks like we're still merged blocked by @cronokirby

cronokirby (approved):

Issue #2410, opened 2025-05-22
ci: it's too slow


conorsch:
The CI runs in the web monorepo take too long. Specifically, the "Turbo CI" job averages about 15-20 minutes per PR. This should not be the case! Given that we use [turborepo](https://turborepo.com/), we should be able to leverage its [smart caching](https://turborepo.com/docs/crafting-your-repository/caching) to keep builds fast. Note that I do _not_ propose we add support for turbo's [remote caching](https://turborepo.com/docs/core-concepts/remote-caching), which is a SaaS that we've no need of, given our heavy use of [BuildJet caching](https://buildjet.com/for-github-actions/docs/guides/migrating-to-buildjet-cache). 

Looking at historical CI runtimes, a lot of time is spent on unnecessary serialization of jobs:

![Image](https://github.com/user-attachments/assets/90f41897-dfcc-4049-a4f1-cfc268583fcb)

In [that run](https://github.com/penumbra-zone/web/actions/runs/15198685057), there's 5m spent on "compiling," followed by another 5m spent "building," and finally another 5m spent on running tests. What, praytell, is the difference between building and compiling? Do we really need those as two separate steps? Nay, I say, we do not. 

## Suggested changes

As long as we're careful to declare inputs and outputs in the `turbo.json`, which largely we already are:

https://github.com/penumbra-zone/web/blob/713a9ad524ef835ce82cff022ecdf796b564324b/turbo.json#L21-L24

then we can trust turbo to build things in the correct order as defined in its dependency graph. In order for that to work, and be fast, we must:

1. Cache and restore the entirety of the `.turbo/` directory
2. Provide all necessary build tooling, such as `pnpm`, `playwright`, browsers, and the like

And that's about it. UX-wise, I find having separate jobs within a workflow to be somewhat helpful: did linting fail, or did testing fail? But as I understand it, we can run _all_ Turbo actions within a single job, to maximize use of the turbo cache. What I feel most strongly about is that overall wall time for a CI on a given PR should be as low as possible; we can finetune how that information is displayed.

## Appendix 1

<details>
<summary>
Here's some historical data
</summary>

```
❯ gh run list --branch main --workflow "Turbo CI" --status success --json databaseId,startedAt,updatedAt -L 20 | /tmp/jawn.sh
Run  Runtime  Started
1    17m30s   2025-05-15 13:59:13
2    16m55s   2025-05-15 18:06:05
3    16m38s   2025-05-16 03:26:20
[...]

PR #2419, opened 2025-05-23
dev: support nix env for playwright tests

### LINKED ISSUE #2411: dev/ci: playwright tests don't work on linux
**Is your feature request related to a problem? Please describe.**
The existing [playwright](https://playwright.dev/) tests don't run on my computer, which is running Fedora Workstation 42. This is because Playwright [only supports Ubuntu and Debian for Linux options](https://playwright.dev/docs/intro#system-requirements). That's a reasonable choice for their project; I'd probably do the same in their shoes.

**Describe the solution you'd like**
I demand to be able to execute the web integration tests on my machine. Ideally, I'd be using the same setup locally as is used in CI, to aid in debugging. 

**Describe alternatives you've considered**
I could just watch tests run in CI and assume that's good enough. I could distro-hop, or use [distrobox](https://distrobox.it/#what-it-does) to wrap the test suite, or run [playwright in a container](https://playwright.dev/docs/docker). For many teams, the containerized approach would be sufficient, but running containers on dev workstations hasn't gotten a lot of traction at PL, which is the team I support, so I'm bearish on that as a solution. 

**Additional context**
For any developer using macOS, they should just be able to do the sane thing, which is to use playwright [exactly as described in its docs](https://playwright.dev/docs/intro#updating-playwright). Any workarounds to support esoteric Linux distros should be careful _not_ to disturb that first-class workflow. 

This is kind of a prereq for #2410: I don't want to change how tests are executed in CI unless I can confidently run them locally myself. 

Any solution should make sure to incorporate strong caching of installed browsers, as with #2366. Already I feel like a nix-based solution would suit me just fine, especially if it can be made to work in CI, too.

conorsch:
## Description of Changes

It's somewhat cursed what I'm doing here, but the goal was:

* get playwright tests running under linux, so i can run them locally
* use the nix env to provide tooling
* ergo: run the playwright tests within the devshell

Surprisingly difficult! Playwright demands binaries in a very specific location, and the binaries it downloads are dynamically linked, to paths that will be filtered out of the nix devshell's LD_LIBRARY_PATH, and therefore the browsers won't run. Even if we override the devshell's library path, we'd still be referencing incompatible shared objects.

We can instead install suitable browsers within nix and cache them, although instructing Playwright to use them is actually quite difficult. So we cheat, and create symlinks at the paths Playwright watches, pointing to the nix-downloaded browsers.

Cursed? You bet. Does it work? It does! Although it'll break whenever the `playwright` node deps are bumped. Made top-level var in the wrapper script so that updating the playwright browser version string is straightforward.

## Related Issue

Closes #2411. Refs #2410. 

## Testing and reviewing

In this changeset I've _avoided_ updating the behavior of CI, as assurance that the changes I've made don't break the existing playwright tests. In future work (#2410) I'll revisit CI and pull this logic in. 

If you're reviewing this PR on a mac, verify that nothing in it affects you. If you're reviewing this PR and you use a non-[Playwright-supported Linux distro](https://playwright.dev/docs/intro#system-requirements), and you feel comfortable using a nix devshell, go ahead and blast `just test` and see how it treats you. 

I've not done so, but was careful not to alter actual app code when preparing the diff.

### DIFF
--- a/apps/veil/src/pages/tournament/ui/social-card-dialog.tsx
+++ b/apps/veil/src/pages/tournament/ui/social-card-dialog.tsx
@@ -124,7 +124,6 @@ Join now 👇`;
                 Copy Image
               </Button>
               <Button actionType='accent' onClick={() => shareToX(text, url)}>
-                {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- Xcom */}
                 <Icon IconComponent={Xcom} size='sm' />
                 Share
               </Button>
--- a/apps/veil/tsconfig.json
+++ b/apps/veil/tsconfig.json
@@ -23,12 +23,14 @@
   },
   "extends": ["@tsconfig/strictest/tsconfig.json", "@tsconfig/vite-react/tsconfig.json"],
   "include": [
+    "app/**/*.ts",
+    "app/**/*.tsx",
+    "global.d.ts",
     "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
-    ".next/types/**/*.ts",
+    "src/**/*.ts",
+    "src/**/*.tsx",
     "styles/styles.d.ts",
-    "global.d.ts"
+    ".next/types/**/*.ts"
   ],
   "exclude": ["node_modules"]
 }
--- a/scripts/benchmark-tests
+++ b/scripts/benchmark-tests
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# Developer script to benchmark running the test suite and gather some timing info,
+# to aid in optimizing CI runtimes. Running multiple times illustrates "warm cache"
+# behavior.
+set -euo pipefail
+
+
+# The lint task passes completely, or has at least once. Let's focus on tests for now.
+hyperfine --setup "pnpm install && pnpm build" --warmup 1 --runs 5 \
+  "just lint-rust" \
+  "just lint-turbo" \
+  "just test-rust" \
+  "just test-turbo"
+  "pnpm run all-check"
--- a/scripts/benchmark-tests
+++ b/scripts/benchmark-tests
@@ -6,9 +6,9 @@ set -euo pipefail
 
 
 # The lint task passes completely, or has at least once. Let's focus on tests for now.
-hyperfine --setup "pnpm install && pnpm build" --warmup 1 --runs 5 \
+hyperfine --setup "just clean && pnpm install && pnpm build" --warmup 1 --runs 5 \
   "just lint-rust" \
   "just lint-turbo" \
   "just test-rust" \
-  "just test-turbo"
+  "just test-turbo" \
   "pnpm run all-check"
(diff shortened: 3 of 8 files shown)


apps/veil/src/pages/tournament/ui/social-card-dialog.tsx
```diff
@@ -124,7 +124,6 @@ Join now 👇`;
                 Copy Image
               </Button>
               <Button actionType='accent' onClick={() => shareToX(text, url)}>
-                {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- Xcom */}
```
conorsch, line 127: At some point in my tooling experimentation, one of the linters informed me this lint was unnecessary. We're excluding the veil code from lint actions in some places https://github.com/penumbra-zone/web/blob/403196c0c897a142de6d462892e653937a18e934/package.json#L23 but I'm not changing that behavior in this PR. As always, expect CI to pass.

apps/veil/tsconfig.json
```diff
@@ -23,12 +23,14 @@
   },
   "extends": ["@tsconfig/strictest/tsconfig.json", "@tsconfig/vite-react/tsconfig.json"],
   "include": [
+    "app/**/*.ts",
+    "app/**/*.tsx",
+    "global.d.ts",
     "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
```
conorsch, line 28: Including _all_ typescript files via `**/*.ts` was causing git-ignored paths to be considered, yuck. Fixed.

scripts/benchmark-tests
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Developer script to benchmark running the test suite and gather some timing info,
+# to aid in optimizing CI runtimes. Running multiple times illustrates "warm cache"
+# behavior.
```
conorsch, line 4: Will use this script to gather timing info for e.g. #2410.

scripts/playwright-setup
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
+# Build script to handle configuring Playwright for running integration tests,
+# both as a "pretest" hook in `package.json` and via manual invocation in CI.
+#
```
conorsch, line 4: This script is the bulk of the diff.

scripts/playwright-setup
```diff
@@ -0,0 +1,150 @@
...
+_PLAYWRIGHT_CHROMIUM_HEADLESS_VERSION="1169"
+
+
+# If we're not in a nix shell, then just do the simple thing:
+# defer to playwright docs on default setup strategy.
+if [[ -z "${IN_NIX_SHELL:-}" ]] ; then
+  >&2 echo "Installing playwright..."
+  pnpm exec playwright install
+  exit 0
+fi
```
conorsch, line 31: Careful attention was paid to ensure that _only_ nix devshells get to grapple with the ratking of if statements below.

conorsch (commented):
Left some in-line comments to aid review.

cronokirby:
image

Installation seems to work without a hitch, then I can run the benchmarking with little fanfare.

Great stuff!

cronokirby (approved):

conorsch:
Thanks for taking it for a spin! Glad it worked well. In #2421 I've pulled the fancy nix logic into CI.

PR #2424, opened 2025-05-25
middleware routing: redirect to portfolio page as default route

TalDerei:
## Description of Changes

- Redirects to the portfolio page as the default route after onboarding,
- Took creative liberty to reorder header tabs to place portfolio first (seems more natural?),
- As a side-effect, the "Manage Portfolio" button in Prax will now redirect to the portfolio page by default, unless the user selects a different minifront frontend, in which case it will redirect to the assets page as before.

Notably, this _doesn't_ require any changes to prax onboarding since we already set the default frontend to veil in extension LS. 

## Related Issue

N/A


conorsch:
Why was this closed? @TalDerei Can you add a summary comment to explain?

Issue #2420, opened 2025-05-23
bug: manual flow for shielding assets not working


conorsch:
**Describe the bug**
Today I tried to shield IBC funds on the testnet, following the [steps in the guide](https://guide.penumbra.zone/dev/testnet#obtaining-funds-on-the-testnet), and was unable to, due to the "manual" deposit actions being greyed out. 

**To Reproduce**

1. Connect wallet to testnet minifront
2. Follow guide steps to deposit testnet noble: https://guide.penumbra.zone/dev/testnet#obtaining-funds-on-the-testnet
3. Observe gray/inactive chains

**Expected behavior**
I expected to be able to deposit testnet funds. I use this flow a few times a week, generally, and it's been behaving well. It's been a few days since I last had it working, though.

**Screenshots**

![Image](https://github.com/user-attachments/assets/151a5660-9979-4e67-9531-f00047d6bc2d)

**Additional context**
We should check whether this bug affect mainnet manual deposits, as well. I haven't done so yet.

conorsch:
> We should check whether this bug affect mainnet manual deposits

Yes, I'm seeing the same greyed-out behavior on a mainnet RPC:

![Image](https://github.com/user-attachments/assets/0bc0fde3-e6f7-4e61-8ec7-225b4bec7341)

Issue #2382, opened 2025-05-16
overflow IBC asset id for manual deposits


TalDerei:
<img width="453" alt="Image" src="https://github.com/user-attachments/assets/3596513c-84ca-41cf-8c3d-eefbebfd0ae5" />

hitchhooker:
regardless of being connected to dex/veil already with kepler, it will prompt your keplr wallet when shielding assets.
when clicking to connect keplr 
```err
There is no modular chain info for quasar-1
```
![Image](https://github.com/user-attachments/assets/269df707-57a4-4f35-866f-f23f5d6b6946)

withdraw bug in minifront:

![Image](https://github.com/user-attachments/assets/56e22426-5613-47a8-9b40-b7a92b3e9236)

TalDerei:
@hitchhooker thanks for flagging, we're internally tracking and investigating the source of the breakage!

PR #2426, opened 2025-05-25
IBC: fix manual flow for shielding assets

TalDerei:
## Description of Changes

- fixes side-effect of some breakage from https://github.com/penumbra-zone/web/issues/2420 where assets couldn't be selected when manually shielding,
- truncates overflow IBC asset id for manual deposits 

## Related Issue

https://github.com/penumbra-zone/web/issues/2420 and https://github.com/penumbra-zone/web/issues/2382

### DIFF
--- a/apps/minifront/src/components/ibc/deposit-manual/assets-table.tsx
+++ b/apps/minifront/src/components/ibc/deposit-manual/assets-table.tsx
@@ -45,7 +45,7 @@ export const AssetsTable = () => {
       <Table>
         <TableHeader>
           <TableRow>
-            <TableHead className='w-[100px]'>Denom</TableHead>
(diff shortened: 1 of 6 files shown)


conorsch:
Lint failures look legitimate to my eye. Testing should include confirming a working deposit when built from a local dev env, correct?

VanishMax (approved):
Works 👍

VanishMax:
Managed to fix the sync bar by removing artifacts from #2364, which introduced the regression.

image

image

VanishMax:
Tested all UI elements of Minifront locally – everything seems like is back to normal. Tested manual deposits and withdrawals too with Leap and Noble, it works.

VanishMax:
Merging to unblock Minifront IBC deposits

Issue #2436, opened 2025-05-28
lqt: social card not rendering


conorsch:
**Describe the bug**
While testing LQT workflows on the testnet Veil instance, I noticed that the "social card" (#2241) is not rendering properly: 

![Image](https://github.com/user-attachments/assets/ff29cb9c-218a-40df-a59f-0bde6525bb9c)

**To Reproduce**

1. Navigate to `/tournament` on LQT testnet Veil instance
2. Submit vote for epoch
3. Observe social card display

**Expected behavior**
I expect the image to contain readable text, rather than a blank gray slate.

**Screenshots**
Above.

**Additional context**
I'm using Chromium on Fedora Workstation 42.

conorsch:
Console shows errors that reference `localhost:3000`: ![Image](https://github.com/user-attachments/assets/694dc071-bc10-40a8-8e36-8f9876b083da)

There should not be _any_ references to localhost when running in prod-built environment like https://dex-explorer.testnet.plinfra.net. Perhaps this is another instance of client/server side vars getting mixed up? Cf. https://github.com/penumbra-zone/web/issues/2413#issuecomment-2917179654

conorsch:
Perhaps merge of https://github.com/penumbra-zone/web/pull/2429 will resolve? Will test again once that's deployed.

conorsch:
I'll take a look at this. As mentioned in the linked-to discussions, this is a recurring point of friction in the context of Penumbra web, so I'd like to generate a durable solution.

conorsch:
I've deployed a custom build directly to testnet infra to resolve; will follow up with changes to the repo logic so that CD is flowing again.

TalDerei:
closing as resolved

PR #2447, opened 2025-05-29
build: support build-time BASE_URL

conorsch:
## Description of Changes

Here we introduce a container build-arg that permits overriding the `NEXT_PUBLIC_BASE_URL` env var for the nextjs prod build. This is not a comprehensive solution: there's a slew of `process.env` calls scattered around the Veil codebase. Cleaning up the interface for obtaining runtime config in client-side can wait, however; right now, I'm shoving things around to keep the deploys flowing in their current forms, so that hotfixes to Veil land immediately.

## Related Issue

Refs #2436.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -16,7 +16,7 @@ on:
   # Temporary: build on PRs, to make sure workflow succeeds
   pull_request:
 jobs:
-  veil:
+  veil-testnet:
     runs-on: buildjet-16vcpu-ubuntu-2204
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/container.yml
```diff
@@ -52,5 +52,52 @@ jobs:
...
+          context: .
+          platforms: linux/amd64
+          file: ci/containerfiles/Containerfile-veil
+          push: true
+          tags: |
+            ${{ steps.meta.outputs.tags }}
+            ghcr.io/penumbra-zone/veil:mainnet
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            BASE_URL=https://dex-explorer.mainnet.plinfra.net
```
TalDerei, line 103: comment: should this just be `https://dex.penumbra.zone/`?

TalDerei (approved):
left question about mainnet BASE_URL, otherwise looks good

.github/workflows/container.yml
TalDerei: what's the status of this?

Issue #2450, opened 2025-05-30
Minifront - Fix Background Color, Wallet Connection Issues and Tabs


DDX1:
1. Fix app background color. Currently v2 background is applied, which is slightly lighter than the full black of v1. See https://github.com/penumbra-zone/web/pull/2445#issuecomment-2922091047

2. While testing the connect page, I noticed that although the wallet connected, the connect page didn't move away. When clicking the "Connect" button again Prax wallet didn't open anymore (because it was already connected). After refreshing a few times I saw the actual app pages. Bad UX. 

3. While testing I also noticed that the Tabs components are shown over the sync overlay bg (z-index of Tabs too big).

<img width="1118" alt="Image" src="https://github.com/user-attachments/assets/230a8efd-4998-49bc-92e2-0a95830ffaab" />

PR #2445, opened 2025-05-29
Feat/v2 Transactions UI & Transfer Page

DDX1:
## Description of Changes

This PR establishes a comprehensive minifront-v2 application with complete Portfolio page, new Transfer Assets page,  UI component library improvements and additions, and a legacy cleanup after moving v2-development into a standalone app. This represents a major step forward in the minifront-v2 application while maintaining core architectural decisions as described here: https://github.com/penumbra-zone/web/blob/main/docs/adrs/002-framework.md

### Minifront v2 setup
- **MobX state management**: Comprehensive reactive state management implementation with differente Stores

### Portfolio Page Enhancements
- **Transaction view improvements**: Enhanced TransactionSummary with better icon handling and address display
- **Delegation token display**: Improved rendering and metadata handling for delegated tokens
- **AssetIcon refinements**: Better support for delegation tokens with proper badge display and metadata integration
- **Type safety improvements**: Comprehensive protobuf getter integration with proper error handling

### Transfer Assets Page (New)
- **Complete Transfer page**: Tabbed interface with Send and Receive forms
- **SendForm**: Full transaction form with recipient input, asset selection, fee tier selection with ValueViewComponent integration, and memo handling
- **ReceiveForm**: AccountSelector with IBC deposit toggle and address display functionality
- **Fee estimation**: Dynamic fee calculation with loading states, fallback dummy data, and proper error handling

### Connect Service incl. New Pages
- **New v2 design**: Wallet connect/install/reload pages based on new v2 Figma designs

### UI Component Library Enhancements
- **AccountSelector component**: Integrates TextInput with navigation buttons and AddressView display
- **Button component**: Added `rounded` prop for circular button styling in AccountSelector
- **ValueViewComponent**: Added `textColor` prop for custom color control, overriding default and signed colors (useful for balance error handling on Transfer page)
- **AssetValueInput**: Enhanced with ValueViewComponent integration for balance display and comprehensive error handling

### Header & Navigation Improvements
- **Mobile navigation refactor**: Fixed overlay, added smooth animations and aligned design with Figma
- **Navigation state management**: Better handling of active states and transitions

### Legacy Cleanup
- **Removed conflicting packages**: Cleaned up @penumbra-zone/ui package conflicts between old and new minifront
- **Fixed broken imports**: Resolved AppParametersStore and other import issues
- **Lint and type safety**: Resolved all ESLint errors and improved TypeScript integration

### Technical Improvements
- Proper TypeScript integration with protobuf getters throughout the application
[...]

PR #2451, opened 2025-05-30
fix(minifront): bg-color, connection issues and SegmentPicker (z-index)

DDX1:
See request from @TalDerei in https://github.com/penumbra-zone/web/pull/2445#issuecomment-2922091047 and more details in this issue: https://github.com/penumbra-zone/web/issues/2450


apps/minifront/src/abort-loader.ts
```diff
@@ -47,7 +47,8 @@ const throwIfProviderNotInstalled = () => {
  */
 export const abortLoader = async (): Promise<null> => {
   throwIfProviderNotInstalled();
-  await retry(() => Boolean(penumbra.connected));
+  // Increased timeout from 500ms to 3000ms to allow reconnection to complete
+  await retry(() => Boolean(penumbra.connected), 3000);
```
TalDerei, line 51: question: @DDX1 could increasing the timeout here be contributing to https://github.com/penumbra-zone/web/issues/2460?

DDX1: Yes, unfortunately that is the case, my bad. I didn't notice it during testing. Will reset it to the previous timeout (500) right away. 

The other problem and the actual reason for the change, as described in https://github.com/penumbra-zone/web/pull/2451 (no response when clicking the connect button), remains randomly for me. Tested this again on prod and dev and noticed it sometimes works and sometimes doesn't. Can't tell why, but won't further investigate at this point and just restore previous timeout, since it might be an edge case just on my end. 

https://github.com/user-attachments/assets/0a9260c1-ffd7-40f5-a93e-3a7a7acb1169

TalDerei: thanks for flagging, we're actually investigating in https://github.com/prax-wallet/prax/issues/344, so the reversion here is appropriate.

PR #2465, opened 2025-06-02
Simple liquidity form

JasonMHasperhoven:
## Description of Changes

Adds the simple liquidity form, that forms the base for follow up PRs where we add the depth chart (using d3-scale)

Current behaviors:
- Price range handles cannot cross each other
- Handle texts x positions get adjusted so that they don't overlap
- Inputting an amount in the base input field will set the quote and vice versa
  - Subsequent changes in the price range will update the opposite field
  - When the last touched input is quote and the price range exceeds the market price, it resets the input and shows a warning toast (also true for the opposite case).

Demo here:
https://github.com/user-attachments/assets/cb32b69e-f608-46f4-9539-420c7b476e5f

## Related Issue

- [x] Verified that the position amounts are correct and sum equally to the desired inputs


VanishMax (commented):
Impressive work with the slider component! Left the comments mostly about figma-related todos

VanishMax (approved):
🚀

PR #2473, opened 2025-06-09
fix(veil): #2468: persist subaccount

### LINKED ISSUE #2468: veil: persist selected subaccount index
From user testing:

> Opening up the Tournament keeps defaulting back to Main Account when using the tournament. If a user refreshes the page or re-opens it, for some reason it defaults back to the main account. This is bad because if a user wanted to manage liquidity positions in a different subaccount, they can’t view them in the same account they might be delegating/voting from

We should add previously selected subaccount index to LocalStorage and fill MobX store with this data on initial Veil opening.

VanishMax:
Closes #2468 

Saves the selected subaccount index in LocalStorage and retrieves it on connect and reconnect to Prax. Stops persisting when user chooses to disconnect their wallet.


TalDerei (approved):
this seems good to me – deferring persisting the key _until_ the user selects a subaccount explicitly in the account selector.

PR #2480, opened 2025-06-11
fix(ui): proper font loading for prod builds

### LINKED ISSUE #2479: regression: flaky: Poppins font loading
Image

Something got broken, and the Poppins font isn't being loaded. If you don't already have the font installed (I don't think it's a systems font in general) the page renders incorrectly.

c.f. https://dex-explorer.mainnet.plinfra.net/explore

This seems to be related to CI though, and may be a combination of underspecified builds, so that clearing some kind of cache makes it pop up.

conorsch:
## Description of Changes
We noticed a regression in fallback fonts when using the prod builds. The same issue wasn't present in the dev env (i.e. `pnpm dev`), because the dev env loads assets directly from disk. For prod builds, we must explicitly prepare outputs, such as static assets, so here we make two changes:

  1. ensure that the UI package explicitly imports the fonts css, so it's included in the generated tailwind outputs
  2. ensure that the font file WOFFs are copied to the dist/ output directory of the ui package.

After rerunning a prod build locally with these changes, the font loading problem was resolved on my machine.

## Related Issue

Closes #2479.

I did not test anything related to minifront, only veil. 

## Testing and review

The way I tested this locally was to build and run a local container on main and this branch, and observe the font fix. The process I used was:

```
cd apps/veil
just run-container
```

which calls out to local `podman` on my machine. If someone is so inclined to do that locally, please do. Maybe I should wire up a `just standalone` target that doesn't use a container, but still wires up the prod build logic for easy local testing. I think it's fine to provide visual review, merging if nothing appears obviously wrong in the diff, and then verify the behavior against a live URL post-merge-and-deploy.

### DIFF
--- a/packages/ui/src/styles/main.css
+++ b/packages/ui/src/styles/main.css
@@ -1,3 +1,8 @@
+/* Explicitly load fonts so that generated CSS contains the necessary font-face declarations.
+ * The explicit import must precede the tailwind declarations.
+ * */
+@import '../theme/fonts.css';
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
--- a/packages/ui/src/styles/main.css
+++ b/packages/ui/src/styles/main.css
@@ -1,8 +1,0 @@
-/* Explicitly load fonts so that generated CSS contains the necessary font-face declarations.
- * The explicit import must precede the tailwind declarations.
- * */
-@import '../theme/fonts.css';
-
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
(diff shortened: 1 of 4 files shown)


packages/ui/src/styles/main.css
```diff
@@ -1,3 +0,0 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
```
conorsch, line 3: So the solution was to _remove_ `styles/main.css` from the UI package? I did not try that! Thanks for cleaning up my mess, @VanishMax. :innocent:

Issue #2403, opened 2025-05-21
Add Liquidity Shape to Form


JasonMHasperhoven:
This task is to upgrade the provide liquidity order form with the liquidity shape functionality.

Context here:
https://honey-thunder-537.notion.site/1-Click-LP-PRD-1ebf89a3170180b5b28cd9027b6cf6b1?pvs=4

The implementation is planned to be phased:
1. Simple version without Liquidity Shape
2. Upgrade with Liquidity Shape function
3. Upgrade with Depth Chart

And for this task, we just need to do step 2.

For this we need to:
1) add the options to the form, and
2) create a function that can make range liquidity positions with a shape.

The function is expected to create custom curves/shapes based on its received parameters. It should be well tested and abstracted, so that we can achieve a pyramid shape and an inverted pyramid shape and everything in between.

Design reference:
<img width="394" alt="Image" src="https://github.com/user-attachments/assets/aea61cb1-4bf5-4845-bd48-2dd21fb2999a" />

erwanor:
I think phasing is a good idea. LGTM. 

We should test this with a large number of positions e.g, 20 strategies (assuming each strat defaults to 10 LPs).

>The function is expected to create custom curves/shapes based on its received parameters. It should be well tested and abstracted, so that we can achieve a pyramid shape and an inverted pyramid shape and everything in between.

We should do this last, we can reuse the linear recipe that is currently deployed, lifting it out into a routine. This will let us move fast to get to Phase-2 as soon as possible, and polish the user flow so that it's perfect. We can pair on backfilling the strategies later, the best ROI is in getting the interface done.

PR #2485, opened 2025-06-13
Liquidity shape

JasonMHasperhoven:
## Description of Changes

Closes https://github.com/penumbra-zone/web/issues/2403

Implements the liquidity shape to the order form:

image

Pyramid shape (stable kind)
image

## Related Issue


TalDerei (commented):
LP distributions seem to work as expected 👍

apps/veil/src/shared/math/position.ts
```diff
@@ -152,6 +152,18 @@ interface RangeLiquidityPlan {
...
+export enum LiquidityDistributionShape {
+  /** Equal distribution across all positions */
+  FLAT = 'FLAT',
+  /** Higher liquidity near market price, decreasing towards range edges */
+  PYRAMID = 'PYRAMID',
+  /** Lower liquidity near market price, increasing towards range edges */
+  INVERTED_PYRAMID = 'INVERTED_PYRAMID',
+}
+
 interface SimpleLiquidityPlan {
```
TalDerei, line 167: question: isn't `rangeLiquidityPositions` redundant since `SimpleLiquidityPlan` includes a flat option (uniform distribution)? can we distill these into a single helper?

TalDerei: sounds good, thanks for the clarification.

PR #2487, opened 2025-06-14
fix (services): don't randomly select index in status stream test

turbocrime:
status stream test was flaky due to selection of a random value.

if the selected value happened to be the last item in the mock stream, the test expectations were violated. now, a constant value is selected.


TalDerei (approved):
noticed this failure anecdotally as well, rerunning failed CI usually fixed it

PR #2507, opened 2025-06-18
feat: improve cosmos rpc handling

### LINKED ISSUE #2506: Portfolio: speed up and improve cosmos rpc fetching
Currently, we have a lot of errors even on mainnet deployments, when trying to fetch assets from cosmos chains. This is due to many endpoints being out of date, unreachable, or unreliable.

vacekj:
fixes #2506 

- Introduced a mechanism to track failed RPC endpoints with a cooldown period.
- Added functions to check endpoint health and mark endpoints as failed.
- Updated balance fetching logic to prioritize healthy endpoints and implement retry logic with exponential backoff.
- Improved fallback mechanism to use the default RPC endpoint if all reliable endpoints fail.
- Add script to collect known-good RPCs, and add the results of this list to our default RPCs hard-coded list.

I am observing much faster cosmos asset loading on mainnet now.

### DIFF
--- a/apps/veil/src/scripts/find-working-ibc-rpcs.ts
+++ b/apps/veil/src/scripts/find-working-ibc-rpcs.ts
@@ -64,7 +64,7 @@ async function isEndpointAlive(endpoint: string): Promise<boolean> {
 
 async function main(): Promise<void> {
   const { chainId } = parseArgs();
-  console.debug(`Fetching Penumbra registry for chain-id: ${chainId}`);
+  console.info(`Fetching Penumbra registry for chain-id: ${chainId}`);
   let registry: unknown;
 
   try {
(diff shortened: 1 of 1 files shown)


apps/veil/src/scripts/find-working-ibc-rpcs.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  for (const conn of ibcConnections) {
+    const cid = (conn as { chainId?: string }).chainId;
+    if (cid) {
+      targetChainIds.add(cid.toLowerCase());
+    }
+  }
+
+  console.debug('Target chain IDs:', Array.from(targetChainIds));
+  console.debug(`Will test RPCs for ${targetChainIds.size} chain(s)`);
```
VanishMax, line 107: chore: cleanup debug logs

vacekj: Fixed and noted - will walk-thru.

Issue #2499, opened 2025-06-18
liquidity shape: planner request support


TalDerei:
The existing transaction planner requests for constructing position bundles need to be amended to incorporate the encrypted metadata proto field.

PR #2508, opened 2025-06-20
frontend: position metadata support

TalDerei:
## Description of Changes

**Frontend:** extends https://github.com/penumbra-zone/web/pull/2485 (cc @JasonMHasperhoven) by introducing a mapping between positions and their corresponding liquidity shape, which is then used to set the appropriate strategy tag in the planner (see UIP-9: https://github.com/penumbra-zone/penumbra/pull/5212). 

Suggested strategy tags to represent the different liquidity shapes:

```ts
export enum LiquidityDistributionStrategy {
  SKIP = 1,
  ARBITRARY = 2,
  FLAT = 3,
  PYRAMID = 4,
  INVERTED_PYRAMID = 5,
}
```

## Related Issue

references https://github.com/penumbra-zone/web/issues/2499 

### DIFF
--- a/apps/veil/src/pages/trade/ui/order-form/liquidity-shape.tsx
+++ b/apps/veil/src/pages/trade/ui/order-form/liquidity-shape.tsx
@@ -57,6 +57,8 @@ export const LiquidityShape = ({
   onClick: () => void;
   selected: boolean;
 }) => {
+  const [isHovered, setIsHovered] = useState(false);
+
   if (!isSupportedShape(shape)) {
     return null;
   }
@@ -69,7 +71,6 @@ export const LiquidityShape = ({
     disabled,
     tooltipMessage,
   } = shapeMapping[shape];
-  const [isHovered, setIsHovered] = useState(false);
 
   if (disabled) {
     return null;


apps/veil/src/shared/math/position.ts
```diff
@@ -164,6 +177,17 @@ export enum LiquidityDistributionShape {
...
+/**
+ * Defines associative numeric encoding of `LiquidityDistributionShape` representing the strategy tag
+ */
+export enum LiquidityDistributionStrategy {
+  SKIP = 1,
+  ARBITRARY = 2,
+  FLAT = 3,       
+  PYRAMID = 4,
+  INVERTED_PYRAMID = 5,
+}
```
TalDerei, line 189: comment:  Limit orders currently default to the `FLAT` shape and its corresponding strategy tag — but should we use a reserved tag for limit orders instead?

apps/veil/src/shared/math/position.ts
```diff
@@ -164,6 +177,17 @@ export enum LiquidityDistributionShape {
   INVERTED_PYRAMID = 'INVERTED_PYRAMID',
 }
 
+/**
+ * Defines associative numeric encoding of `LiquidityDistributionShape` representing the strategy tag
+ */
+export enum LiquidityDistributionStrategy {
```
erwanor, line 183: I think we should use the nomenclature from the UIP: https://uips.penumbra.zone/uip-9.html#protobuf-definition

here, there's one too many right?

TalDerei: https://github.com/penumbra-zone/web/pull/2508/commits/63949532fce5e7e1dac4f17a118d38f0c313a85c

apps/veil/src/shared/math/position.ts
erwanor: discussed out of band, we can use the arbitrary tag as convenience index for them

apps/veil/src/shared/math/position.ts
erwanor: SKIP = 1 -> any order or strategy that we don't want to index
ARBITRARY = 2 -> auto-closing orders
LINEAR = 3 -> "Volatile" strategy
STABLE = 4 -> "Concentrated" or "Stable rate" strategy

TalDerei: yeah, auto-closing orders map to the `ARBITRARY` strategy tag here

PR #2510, opened 2025-06-21
Refactor explore page

cronokirby:
## Description of Changes

This refactors the mechanism behind the explore page to:
- fetch data on the server side to avoid an extra round-trip,
- use server functions instead of API routes,
- deferring rendering decisions to the frontend.

We could probably do another round of pindexer simplification, I'm not satisfied with how complex the queries are here.


cronokirby:
https://github.com/user-attachments/assets/5dc397b0-c000-46b5-a6f6-9329644ec913


https://github.com/user-attachments/assets/b6c17dc7-6fe4-4acb-9d9d-de26b60982cd

here's a comparison of the behaviors ;

apps/veil/src/pages/explore/ui/pair-card.tsx
```diff
@@ -3,170 +3,136 @@ import { Star, CandlestickChart } from 'lucide-react';
...
+  if (!endMetadata) {
+    throw new Error(`unknown asset: ${summary.end.toJsonString()}`);
+  }
+  const liquidityMetadata = getMetadata(summary.liquidity.assetId);
+  const volumeMetadata = getMetadata(summary.volume.assetId);
 
   return (
     <Link
-      href={loading ? `/trade` : `/trade/${summary.baseAsset.symbol}/${summary.quoteAsset.symbol}`}
+      href={`/trade/${startMetadata.symbol}/${endMetadata.symbol}`}
```
VanishMax, line 62: suggestion (non-blocking): use [`getTradePairPath` function](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/shared/const/pages.ts#L34) to get the trade link

VanishMax (approved):
That looks really nice!

apps/veil/src/pages/explore/server/stats.ts
```diff
@@ -0,0 +1,78 @@
...
+  largestTradingPairByVolume: { pair: DirectedTradingPair; volume: Amount };
+  /** The pair with the largest move in price. */
+  topPriceMover: { pair: DirectedTradingPair; changePercent: number };
+}
+
+export async function fetchStats(): Promise<Serialized<Stats>> {
+  // Kick off the fetching of the indexing asset.
+  const indexingAssetP = indexingAsset();
+  const raw = await pindexerDb
+    .selectFrom('dex_ex_aggregate_summary')
```
TalDerei, line 41: comment: what's the distinction between `dex_ex_aggregate_summary` and `stats` pindexer routes?

apps/veil/src/shared/api/server/fetchDaySummaries.ts
TalDerei, line 1: empty file

apps/veil/src/pages/explore/ui/page.tsx
```diff
@@ -1,13 +1,25 @@
...
+  const summariesP = fetchDaySummaries();
+  const registryP = fetchRegistry(getClientSideEnv().PENUMBRA_CHAIN_ID);
+  const [stats, summaries, registry] = await Promise.all([statsP, summariesP, registryP]);
   return (
     <section className='flex flex-col gap-6 p-4 max-w-[1062px] mx-auto'>
       <PenumbraWaves />
-      <ExploreStats />
-      <ExplorePairs />
+      <ExploreStats stats={stats} registry={registry} />
+      <ExplorePairs summaries={summaries} />
```
TalDerei, line 22: comment: very clean abstraction here

apps/veil/src/pages/explore/ui/pairs.tsx
```diff
@@ -1,26 +1,42 @@
...
-  const { data, isLoading, isRefetching, isFetchingNextPage, fetchNextPage } = useSummaries(search);
-
-  const { observerEl } = useObserver(isLoading || isRefetching || isFetchingNextPage, () => {
-    void fetchNextPage();
-  });
+export const ExplorePairs = ({ summaries }: ExplorePairsProps) => {
+  const getMetadata = useGetMetadata();
+  const augmentedSummaries = useMemo(() => {
+    const deserialized = deserialize<SummaryWithPrices[]>(summaries);
+    const out: [SummaryWithPrices, string, string][] = deserialized.map(x => [
```
TalDerei, line 22: comment: maybe summary deserialization should be moved to the top-level component for consistency with stats?

apps/veil/src/pages/trade/api/use-summary.ts
```diff
@@ -1,9 +1,16 @@
...
+import { useAssets } from '@/shared/api/assets';
+import { deserialize, serialize } from '@/shared/utils/serializer';
+import { Metadata } from '@penumbra-zone/protobuf/penumbra/core/asset/v1/asset_pb';
+
+interface SummaryWithMetadata extends Summary {
+  start: Metadata;
+  end: Metadata;
+}
 
 export const useSummary = (
```
TalDerei, line 15: comment: In line with the other changes that use server functions and remove the API method entirely, is there a reason this case wasn’t updated as well?

apps/veil/src/shared/api/server/summary.ts
```diff
@@ -0,0 +1,203 @@
...
+      'recent_prices',
+      'recent_dates',
+      'd.price',
+      'd.price_then',
+      'd.high',
+      'd.low',
+    ]);
+}
+
+export async function fetchSummary(
```
TalDerei, line 140: ACK – consolidating `all.ts` and single.ts into a generic summary server function that employs `dex_ex_pairs_summary`.

TalDerei (approved):
really great consolidation 👍

apps/veil/src/pages/explore/ui/pairs.tsx
cronokirby: Important: it can't! This is because this component is a "client component" and needs to be rendered on the client side (potentially, NextJS, as an optimization, does pre-render it, but that's why we can use hooks inside of this component) and so the parameters need to be serializable!

TalDerei: ah missed this nuance, that makes sense!

PR #2511, opened 2025-06-23
feat: #2004: tailwind v4

VanishMax:
Closes #2004 

**NOTE: the CI fails because this PR waits for the Eslint fix PR (#2512) to be merged first!**

For simplicity of review, I'm leaving here the repeatable steps of how to easily come up with the same result as I did in this PR. Most of the files here are auto-generated, so in the write up below I'll explain step by step how Tailwind was upgraded to v4.

Side note: Tailwind introduced CSS-based configuration and improved many classes, requiring breaking changes. They also released native Vite plugin for better class parsing. 

Improvement: now it is possible to use `@penumbra-zone/ui` package without installed Tailwind! Users would only need to import `'@penumbra-zone/ui/style.css'` into their app 🚀

Additionally, Tailwind v4 is applied ONLY to UI, Veil, and Minifront v2. The old Minifront and ui-deprecated packages are allowed to use Tailwind v3 and stay the same.

## Reproduction steps

1. Run Tailwind upgrade CLI – most of the changed files are caused by this command:

```bash
pnpm dlx @tailwindcss/upgrade
pnpm exec syncpack fix-mismatches
```

2. Manually remove all changes to files in:
- ui-deprecated directory
- minifront directory
- node-status directory
- tailwind-config directory

3. [Update `.syncpackrc` file](https://github.com/penumbra-zone/web/pull/2511/files#diff-aac9fa94496fd53bd351b6dbf9fa82cfd55e6abcff3b0f27169590744b2c5747R38-R48) to allow Tailwind v3 for older applications like Minifront and node-status

```bash
pnpm exec syncpack fix-mismatches
pnpm i
```

At this stage older applications work just as they did before. Newer applications are broken.

4. In UI package:
- `pnpm add @tailwindcss/vite`
- update `vite.config.ts` to use native plugin instead of postcss
- remove `postcss.config.js`
- remove `tailwind.config.ts`
- [create `theme.css`](https://github.com/penumbra-zone/web/pull/2511/files#diff-f1a1a598e00c69c0143a3b123e05c9bb0b56ccd0728ad549870fb383cae7e618R3) with new css variables derived from the old tailwind js-styled theme (it was actually generated by the first command in step 1)
- remove `storybook/tailwind.css`
- update storybook imports to point to new theme.css

At this point, Storybook works the same way as it did before, all components are styled correctly.

[...]

Issue #2004, opened 2025-01-28
Upgrade Tailwind to v4


VanishMax:
Tailwind just released stable v4 with new great improvements:
1. New engine that performs 'up to 5x better'
2. CSS-first config and theme variables
3. First-party Vite plugin to opt out of postcss
4. And much more

Full changelog: https://tailwindcss.com/blog/tailwindcss-v4

TalDerei:
in that vain, we should upgrade to [React 19](https://react.dev/blog/2024/12/05/react-19) which has some nice auto-memoization properties

PR #2512, opened 2025-06-23
feat: #2004: tailwind eslint

VanishMax:
This PR moves from using [eslint-plugin-tailwindcss](https://github.com/francoismassart/eslint-plugin-tailwindcss) to [eslint-plugin-better-tailwindcss](https://github.com/schoero/eslint-plugin-better-tailwindcss) because the older plugin [doesn't support Tailwind v4](https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/325).

This PR is an addition to #2511. It is split to not overflow the mentioned PR with too many auto-generated changes.

Here, eslint-plugin-better-tailwindcss is applied to UI package, Veil and Minifront v2 apps, because these are the only constantly-updating packages that require Tailwind v4.


TalDerei (approved):
nice work 🎉

Issue #2501, opened 2025-06-18
Storage (IndexedDB): encrypted position metadata plumbing


TalDerei:
Prax storag already stores the relevant position data (X<>Y pair, position state, subaccount index) in the `POSITIONS` IndexedDB Table.

```ts
### Schema

  POSITIONS: {
    key: string; // base64 PositionRecord['id']['inner'];
    value: PositionRecord;
  };

// need to store PositionId and Position in the same table
export interface PositionRecord {
  id: Jsonified<PositionId>; // PositionId (must be JsonValue because ['id']['inner'] is a key )
  position: Jsonified<Position>; // Position
  subaccount?: Jsonified<AddressIndex>; // Position AddressIndex
```

```ts
db.createObjectStore('POSITIONS', { keyPath: 'id.inner' });
```

We'll need to (1) extend the schema defintion to optionally store the decrypted position metadata alongside the position entries, and (2) index the positions by strategy. This enables queries like "what are the positions associated with strategy XYZ" without doing an entire DB scan.

```ts
// need to store PositionId and Position in the same table
export interface PositionRecord {
  id: Jsonified<PositionId>; // PositionId (must be JsonValue because ['id']['inner'] is a key )
  position: Jsonified<Position>; // Position
  subaccount?: Jsonified<AddressIndex>; // Position AddressIndex
  positionMetdata?: Jsonified<PositionMetadata>; // 'Decrypted' Position Metadata
}
```

```ts
const positionsStore = db.createObjectStore('POSITIONS', { keyPath: 'id.inner' });

// fast look-ups by strategy
positionsStore.createIndex('strategyTag', 'strategyTag');                // 1-dimensional

// composite lookup by strategy and bundle identifier (probably extranous)
positionsStore.createIndex(
  'strategyTag_bundleId',
  ['strategyTag', 'bundleId'],
);    
```

Additionally, we'll need to extend the storage getters and setters to retrieve this information in block processor and view service respectively.

```ts
// Setter: save position to storage (CHANGE: append 'positionMetadata')
  async addPosition(
    positionId: PositionId,
    position: Position,
    positionMetadata: PositionMetadata,
    subaccount?: AddressIndex,
  ): Promise<void> {
    assertPositionId(positionId);
    const positionRecord = {
      id: positionId.toJson() as Jsonified<PositionId>,
      position: position.toJson() as Jsonified<Position>,
      subaccount: subaccount && (subaccount.toJson() as Jsonified<AddressIndex>),
[...]

PR #2517, opened 2025-06-24
storage and view service: position metadata support

TalDerei:
## Description of Changes

- adds `lpPositionBundle` and `lpStrategyCatalog` view service definitions, using additive filtering rather than strict conjunctive filtering for various request fields, 
- Implements `getPositionsByStrategyStream`, which iterates over relevant position metadata,
- adds a `strategy` index on `POSITIONS` table to support efficient querying.

## Related Issue

references https://github.com/penumbra-zone/web/issues/2501 and https://github.com/penumbra-zone/web/issues/2497 and https://github.com/penumbra-zone/web/issues/2498 


packages/services/src/view-service/lp-position-strategy.ts
```diff
@@ -0,0 +1,35 @@
...
+  LpPositionBundleResponse_Entry,
+} from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+
+export const lpPositionBundle: Impl['lpPositionBundle'] = async function* (req, ctx) {
+  const services = await ctx.values.get(servicesCtx)();
+  const { indexedDb } = await services.getWalletServices();
+
+  if (!req.subaccount) {
+    throw new Error('Missing required subaccount field');
+  }
```
VanishMax, line 14: question: shouldn't it just return all position bundles for all subaccounts?

packages/services/src/view-service/lp-strategy-catalog.ts
```diff
@@ -0,0 +1,33 @@
...
+  LpStrategyCatalogResponse_StrategyEntry,
+} from '@penumbra-zone/protobuf/penumbra/view/v1/view_pb';
+
+export const lpStrategyCatalog: Impl['lpStrategyCatalog'] = async function* (req, ctx) {
+  const services = await ctx.values.get(servicesCtx)();
+  const { indexedDb } = await services.getWalletServices();
+
+  if (!req.subaccount) {
+    throw new Error('Missing required subaccount field');
+  }
```
VanishMax, line 14: suggestion: allow `subaccount` to be an optional filter. If we are to create an API from Prax services, other devs (and maybe us) would benefit from all positions returned without this filter

VanishMax (commented):
Want to discuss the `subaccount` filter. If we decide to leave it as is, I'll willingly approve

packages/services/src/view-service/lp-strategy-catalog.ts
TalDerei: yeah, that seems like a reasonable suggestion, will modify.

VanishMax (approved):
All looks good after addressing the feedback 👍

Issue #2521, opened 2025-06-26
Issue with LQT incentive pool denominations


garwalsh:
**Describe the bug**
The incentive pool is incorrectly represented in the LQT interface, presumably a front end issue.

**To Reproduce**
Navigate to https://dex.penumbra.zone/tournament
Look at the current epoch

![Image](https://github.com/user-attachments/assets/8670625a-5574-4cbd-8807-3f75575bc62e)

**Expected behavior**
The pool should be 100k UM for this epoch

PR #2523, opened 2025-06-26
fix staking token exponent handling in Incentive Pool

TalDerei:
## Description of Changes

divides by `10 ** UM exponent` to normalize values instead of using base-unit (micro-UM)

## Related Issue

https://github.com/penumbra-zone/web/issues/2521


cronokirby:
CI seems to be failing on an unrelated issue, namely that the lint is running over directories that should be excluded. We should fix that in a separate pr

cronokirby (approved):
This should work.

This is, on the other hand, not what we should be doing. We should be using a common component to display a ValueView so that this confusion can't happen. I think the existing one in the UI library should be usable for that, maybe.

We can handle that subsequently.

PR #2533, opened 2025-06-27
feat(veil): #2527: chart pagination

### LINKED ISSUE #2515: Modify candlestick chart to be bound by open/close instead of high/low
We have seen some weird behaviour in the candle stick chart as described here:
https://discord.com/channels/824484045370818580/1346940085446381731/1386489762004734012

Specific types of limit orders (which are admittedly edge cases) which are filled by protocol arbs, can cause the chart to become unusable.

To mitigate against this we will bind the chart by open/close prices instead of high/low

### LINKED ISSUE #2527: Candlestick chart doesn't render for 1m, 15m, 1h candles
**Describe the bug**
The 1m, 15m, 1h candles do not render correctly

**To Reproduce**

- Navigate to https://dex.penumbra.zone/trade/UM/USDC (or any other pair with volume)
- Click 1m, 15m, or  1h above the candlestick chart

**Expected behavior**
A functiona, legible, chart that is scaled correctly and is generally useful

**Screenshots**

Image

**Additional context**
Reported in Discord, also noticed by @VanishMax

VanishMax:
Closes #2515 

Closes #2527 

Added paginated candle request. To make it work properly, we need 2 types of requests:
1. An infinite query that requests pages of 100 candles. When user drags or zooms out the chart, it loads the next page and appends the data to the beginning of the candles array. Tradingview charts unfortunately don't have a method to add partial data to the chart, so I had to play around with ascending data ordering by candle time.
2. An updatable query for the latest candles, I limited it to the 5 latest. These are refetched on each block.

Also, there's no way to bound the chart vertically by open/close instead of low/high. I did a trick by limiting the `high` value to be at most 3x of the `open` candle value. It doesn't go to the moon now, but we can consider lowering this value even more. Ideally, I think, this needs to be moved to pindexer as large candles start trust questions about the chart and generally about Veil

https://github.com/user-attachments/assets/86b1092d-0626-490d-bef4-345eb401f5a9


apps/veil/src/pages/trade/ui/chart/chart.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { ChartLoadingState } from './loading-chart';
+import { useChartConfig } from './use-chart-config';
+
+export const Chart = observer(() => {
+  const [duration, setDuration] = useState<DurationWindow>('1d');
+
+  // we need two queries to avoid overfetching. if we leave only the infinite query, it will
+  // be requested PAGE times on each block, causing many unnecessary requests.
+  const { data: latestCandles } = useLatestCandles(duration);
+  const { data: historyCandles, isLoading, error, fetchNextPage } = useInfiniteCandles(duration);
```
VanishMax, line 18: note: 2 separate queries for the chart

erwanor (approved):

PR #2542, opened 2025-06-30
fix(veil): #2541: share button

### LINKED ISSUE #2541: fix the share button in the social card
The proper denom handling fixed in https://github.com/penumbra-zone/web/pull/2536 also needs to be extended to the share button in the social card — see https://x.com/XiF_YuW/status/1938817356027031790.

VanishMax:
Closes #2541 

Fixed UM exponent display in the share button, and also fixed social card not showing if the address is not reveled because it couldn't calculate the streak. I assigned the default streak to "1" if address is not revealed.


erwanor (approved):
Optimistically

TalDerei (approved):

PR #2546, opened 2025-06-30
fix(veil): #2544: allow zeros in order input

### LINKED ISSUE #2544: Veil forms won't accept zeros after the decimal place
**Describe the bug**
None of the text fields in Market, Limit, or Provide Liquidity will accept 0 after the decimal place making it impossible to enter prices or amounts such as 0.305, 9.054 etc etc

When I type a zero, it disappears and the decimal place gets deleted

**To Reproduce**

1. Go to Market, Limit, or Provide Liquidity at https://dex.penumbra.zone/trade/
2. Attempt to enter a price or amount with a 0 after the decimal place

VanishMax:
Closes #2544 

Trailing zeros were adding an overhead for value rendering in order inputs. Now, it uses `round` function only if the input is not focuses. Previously, `round` was used to prevent entering more decimal numbers than the asset exponent, now `onKeyDown` controls it.

### DIFF
--- a/apps/veil/src/pages/trade/ui/order-form/order-form-limit.tsx
+++ b/apps/veil/src/pages/trade/ui/order-form/order-form-limit.tsx
@@ -1,7 +1,6 @@
 import { observer } from 'mobx-react-lite';
 import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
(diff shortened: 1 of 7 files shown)


TalDerei (commented):
this doesn’t seem to be working on the provide liquidity page, and we should also clear the inputs there since they’re currently defaulting to 0.

PR #2553, opened 2025-07-02
fix: allow withdrawing withdrawn positions with reserves

### LINKED ISSUE #2552: Users unable to withdraw rewards from a previously withdrawn position
**Describe the bug**
When LQT rewards are granted to an LP, the reserves of the position are credited with UM.
Users are reporting that they cannot withdraw these rewards if they have withdrawn the positions reserves before the reward was granted. The Withdraw button is greyed out.

**To Reproduce**

1. Open an LP which should receive rewards based on the incentivised assets for the current epoch and the current prices of those assets.
2. Wait for the executions on the LP
3. Withdraw the reserves before the end of the epoch
4. Once the epoch ends, inspect the position to verify that rewards were granted
5. Attempt to withdraw those rewards

**Expected behavior**
Users should be able to withdraw rewards from LP reserves regardless of whether or not they have previously withdrawn that position

cronokirby:
## Description of Changes

The logic was simply wrong, making the assumption that only closed positions can be withdrawn.

Instead, this adds a little helper to check if a position is "fullyWithdrawn", i.e. has no reserves, and is withdrawn.

This is the kind of thing where having an actual domain library would be helpful.

## Related Issue

Closes #2552.

### DIFF
--- a/apps/veil/src/entities/position/ui/action-button.tsx
+++ b/apps/veil/src/entities/position/ui/action-button.tsx
@@ -9,6 +9,7 @@ import {
 import { withdrawPositions } from '../api/withdraw-positions';
 import { closePositions } from '../api/close-positions';
 import { Dash } from './dash';
+import { fullyWithdrawn } from '@/shared/utils/position';
 
 export const ActionButton = observer(({ id, position }: { id: PositionId; position: Position }) => {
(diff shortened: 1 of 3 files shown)


erwanor (approved):
🙏 thanks

Issue #2550, opened 2025-07-02
Users still intermittently experience being unable to vote for incetivised assets


garwalsh:
This behaviour was first reported here:
https://github.com/penumbra-zone/web/issues/2522

Users were unable to vote and were seeing the message that said they could not do so because they had delegated after the beginning of the current epoch, which was not true. 

The cause of this happening consistently was identified and fixed here:
https://github.com/penumbra-zone/web/pull/2524

The behaviour persists, although inconsistently. 
We have received reports of users experiencing the behaviour intermittently, sometimes "<sorting> itself out".

![Image](https://github.com/user-attachments/assets/29b87ac4-2855-42e2-beef-1cdfa1c75606)

![Image](https://github.com/user-attachments/assets/9f2c017e-0595-46b9-b4fe-4b32a3ebebe2)

![Image](https://github.com/user-attachments/assets/4be87169-4362-4ead-b37a-143dd24f4206)

@TalDerei may have identified the cause, a specific route that is broken, although we can see that this error does not always cause the Vote Now button to be disabled

![Image](https://github.com/user-attachments/assets/94728423-a7bb-495d-8a90-054963bbb528)

PR #2557, opened 2025-07-03
Refactor voting info fetching

cronokirby:
## Description of Changes

This refactors the states for voting info (whether or not you can't vote, if you can't why, etc.) into a unified source of truth, which should help eliminate the weird errors we've been seeing, like #2550. The previous logic was starting to suffer from "big ball of booleans" syndrome, where it wasn't exactly clear what the state of things was.

I also fixed some unrelated lints, which would've made CI unhappy.


cronokirby:
I tested the main flows of delegating and voting, so I don't think this regresses behavior, at least.

apps/veil/src/pages/tournament/ui/voting-info.tsx
```diff
@@ -1,173 +1,43 @@
...
+    </div>
+  );
+};
 
-    return PagePath.TradePair.replace(
-      ':primary',
-      metadata.symbol === 'USDC' ? 'UM' : metadata.symbol,
-    ).replace(':numeraire', 'USDC');
-  }, [votedFor]);
+export const VotingInfo = observer(({ epoch }: VotingInfoProps) => {
```
VanishMax, line 36: issue (ui/ux): buttons are jumping and flickering, irritating the eye. The video shows a case when wallet is not connected and one button is missing on loading state, leaving empty space.

Additional case, not covered in video: when wallet is connected, there's even more jumping: firstly UI shows loading, then "connect wallet", then loading again, and only then the correct state.

https://github.com/user-attachments/assets/8267cabb-0b89-48ae-8363-5a961d8e4db4

VanishMax (commented):
Didn't check the logic in-depth, only UI

apps/veil/src/pages/tournament/ui/voting-info.tsx
TalDerei: > c.f. also the current behavior of the page

relevant context https://github.com/penumbra-zone/web/pull/2400#pullrequestreview-2855375777

cronokirby: I added some logic to figure out if the state of the connection store is still loading, to avoid displaying a flash of a connect button

apps/veil/src/pages/tournament/api/use-voting-info.ts
```diff
@@ -0,0 +1,165 @@
...
+    return { case: 'can-vote', ability: { notes: votingNotesR, epoch, account }, power };
+  }
+  if (hasDelegatedR) {
+    return { case: 'delegated-after-start' };
+  }
+
+  return { case: 'needs-to-delegate' };
+}
+
+export function useVotingInfo(epoch?: number): VotingInfo {
```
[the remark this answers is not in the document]
TalDerei: afaict, this was an important comment that resulted in this regression https://github.com/penumbra-zone/web/issues/2589. I should have pushed harder on this, fixing now.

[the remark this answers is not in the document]
TalDerei: disregard, this is a feature, not a bug. the regression comes from someplace else, will provide relevant context in the associative PR.

Issue #2538, opened 2025-06-28
fix time remaining rendering in round page


TalDerei:
rendering inconsistencies cause the displayed time remaining to sometimes change across page reloads.

------------------

https://github.com/user-attachments/assets/5c001c00-16a0-4694-b8eb-b8b5d08e64d0

https://github.com/user-attachments/assets/48d67d76-d5f2-4d0d-90d6-4d45724c82fa

<img width="1837" alt="Image" src="https://github.com/user-attachments/assets/b5fd94e3-f1c5-455b-90c4-9ba270959dcc" />

TalDerei:
can we fold https://github.com/penumbra-zone/web/issues/2547 in scope? cc @vacekj

vacekj:
Yes, but I'm out for the rest of this week, so I'll unassigned myself so someone else can pick it up

TalDerei:
have we determined this to be a pindexer bug?

VanishMax:
@TalDerei Yes, kinda-random numbers come from pindexer

PR #2559, opened 2025-07-03
fix(veil): #2547: round card end time and blocks

### LINKED ISSUE #2547: veil lqt page lists historical epochs with wrong block ranges
**Describe the bug**
When viewing a historical epoch for the LQT within Veil, the start and end blocks displayed are incorrect. They appear to map to the current block, not to the historical block.

**To Reproduce**

1. View historical epoch such as https://dex.penumbra.zone/tournament/208
2. View a more recent epoch such as https://dex.penumbra.zone/tournament/210 (current at time of writing)
3. Compare the start/end blocks; observe that they match, but they shouldn't

**Expected behavior**
I expect the info pages for two different epochs to display different start/end heights, because it's impossible for two epochs to share those data, given that epochs are sequential on Penumbra.

**Screenshots**

Image

**Additional context**
Looks like all epoch stats in the card are incorrectly referencing the "current" epoch data; note that the delegator vote breakdown is identical in the screenshot above.

VanishMax:
Closes #2547 

References #2538 

Uses `epochs: [epoch]` filter in the summary query for the round card instead of requesting current epoch summary.

Also, adds a nice little tooltip over "Ends in XX time" with the same formatting as in GitHub:

image

### DIFF
--- a/apps/veil/src/pages/tournament/ui/round/round-card.tsx
+++ b/apps/veil/src/pages/tournament/ui/round/round-card.tsx
@@ -1,9 +1,11 @@
 import Link from 'next/link';
 import { ArrowLeft } from 'lucide-react';
 import { observer } from 'mobx-react-lite';
+import { addSeconds, format } from 'date-fns';
 import { Skeleton } from '@penumbra-zone/ui/Skeleton';
 import { Icon } from '@penumbra-zone/ui/Icon';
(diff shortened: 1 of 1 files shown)


conorsch:
Looks great. Love the tooltip. :ok_hand:

PR #2563, opened 2025-07-05
ci: fly ephemeral deploys

erwanor:
## Description of Changes

This adds a minimal container image, and a fly.io deployment step for every PR. It is triggered by commenting /deploy and garbage collected every twelve hours. That's on purpose!


erwanor:
I guess it won't work until I merge.

Issue #2551, opened 2025-07-02
veil bugs in providing liquidity in new ui tab


hitchho:
reversed amounts in provide liquidity
![Image](https://github.com/user-attachments/assets/d8286387-8bc7-492c-9ad2-ad522dad8907)

also providing one sided liquidity not possible on new ui
![Image](https://github.com/user-attachments/assets/99ba7f72-c180-4e4a-8587-b7fafd35e86f)

would be also great addition to prax to have _humanreadable_ tab to understand what the actual
tx impact is like what positions i set if i sign this. now im just pretty much trusting the ui to serve me,
which it doesnt

PR #2580, opened 2025-07-07
Fix: Opposite input calculation in LP form

JasonMHasperhoven:
## Description of Changes

Addresses first issue reported in https://github.com/penumbra-zone/web/issues/2551

## Related Issue

### DIFF
--- a/apps/veil/src/pages/trade/ui/order-form/store/SimpleLPFormStore.ts
+++ b/apps/veil/src/pages/trade/ui/order-form/store/SimpleLPFormStore.ts
@@ -205,7 +205,7 @@ export class SimpleLPFormStore {
     return simpleLiquidityPositions({
       baseAsset: this._baseAsset,
       quoteAsset: this._quoteAsset,
-      baseLiquidity: Number(this.baseInput) / this.marketPrice,
+      baseLiquidity: Number(this.baseInput),
       quoteLiquidity: Number(this.quoteInput),
       upperPrice: this.upperPrice,
       lowerPrice: this.lowerPrice,


conorsch:
CI is now passing. In looking at the behavior of the new LP behavior locally, the numbers I see generated inside the input widgets are now much saner, thanks to the changes in this PR. However, I'm not convinced I can trust those numbers: when I generated a txplan and reviewed the described positions in the Prax pop-up, the numbers are significantly different from the values in the input widgets.

Looking at the simple LP code, I suspect a change such as the following is necessary:

```diff
--- a/apps/veil/src/pages/trade/ui/order-form/store/SimpleLPFormStore.ts                                                                                      
+++ b/apps/veil/src/pages/trade/ui/order-form/store/SimpleLPFormStore.ts                                                                                      
@@ -205,7 +205,7 @@ export class SimpleLPFormStore {                                                                                                          
     return simpleLiquidityPositions({                                                                                                                        
       baseAsset: this._baseAsset,                                                                                                                            
       quoteAsset: this._quoteAsset,                                                                                                                          
-      baseLiquidity: Number(this.baseInput) / this.marketPrice,                                                                                              
+      baseLiquidity: Number(this.baseInput),                                                                                                                 
       quoteLiquidity: Number(this.quoteInput),                                                                                                               
       upperPrice: this.upperPrice,                                                                                                                           
       lowerPrice: this.lowerPrice,  
```

otherwise it looks like we're munging the values to be something other than what's displayed in the app UI. wdyt @JasonMHasperhoven ?

TalDerei (approved):
optimistically

conorsch (approved):
I tested on the ephemeral deployment _after_ @JasonMHasperhoven's latest change, and am pleased with the honesty of the UI widgets when compared to the generated LPs that hit the chain. Specifically, I expect my wallet balances to be deducted by the amounts shown in the widgets, and that's now true.

PR #2593, opened 2025-07-08
feat: add client side env to ambient context

### LINKED ISSUE #2575: social card: fix url rendering
Continuation of https://github.com/penumbra-zone/web/pull/2542, which addressed the exponent handling — the URL still needs to be formatted correctly.

--------------------

Image

cronokirby:
This should fix the social card issues.

I used BASE_URL instead of NEXT_PUBLIC_BASE_URL to avoid the magic env var, in favor of an explicit one.

## Description of Changes

Please provide a clear and concise description of the changes made in this PR.
For UI changes, include images or videos to show the visual impact.

## Related Issue

Closes #2575


conorsch (changes requested):
The fly deploys aren't behaving well, so there's still some logic that needs to be ironed out here for prod builds. Requesting changes to block merge until I sort that out.

conorsch:
> @conorsch what's the status of this?

The last update above remains accurate: the logic in this PR is incompatible with the new Fly deployments, and I don't want to break the new Fly deployments, so further changes are required to this PR before it's ready to go. I'll work on it, but higher priority is https://github.com/penumbra-zone/web/issues/2617.

> I thought the immediate hotfix being proposed was to hardcode https://dex.penumbra.zone in the social card?

Good call, that's here: https://github.com/penumbra-zone/web/pull/2619 and buys us some time.

conorsch (approved):
Looks fantastic, thanks for taking the time. The fly deployment problems were strictly due to the healthcheck endpoint being `/api/env`, which is now removed.

Issue #2154, opened 2025-03-17
add version hash to the dex-explorer footer


TalDerei:
for [dex.penumbra.zone](http://dex.penumbra.zone/), it's not immediately clear if it's running the latest frontend version since we don't display a version hash, and we should consider appending that information to the footer. When there's dex breakage, this would provide more immediate visibility into whether the issue is related to the provider running an outdated frontend. 

-----------------

<img width="327" alt="Image" src="https://github.com/user-attachments/assets/80c29dda-3cab-4216-b3b3-484a6637722c" />

PR #2594, opened 2025-07-08
feat: add git version to veil footer

conorsch:
## Description of Changes

Adds a git hash in the footer for Veil, so that it's obvious which commit is running on a given instance. There are multiple frontends managed by multiple orgs, all on their own deploy schedule, so this change is intended to make it clear whether a given fix has in fact been deployed to a given environment.

Closes https://github.com/penumbra-zone/web/issues/2154

## Images

It looks like this:

veil-footer-2

## Testing and review

Functionality-wise, this looks "good enough" to me, but I'm mostly concerned about the shape of the implementation. What you see is mostly what the LLM thought was a good idea. In particular, please scrutinize the overriding the node env and the resulting process.env calls; should we instead use an api env route as described in https://github.com/penumbra-zone/web/pull/2593?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,6 +10,9 @@
 !.npmrc
 !.syncpackrc
 
+# permit git dir, for version info
+!.git/
+
 # allow ca
 !ci/cert/ca-certificate.crt
 
--- a/apps/veil/app/app.tsx
+++ b/apps/veil/app/app.tsx
@@ -6,6 +6,7 @@ import { QueryClientProvider } from '@tanstack/react-query';
 import { ToastProvider } from '@penumbra-zone/ui/Toast';
 import { TooltipProvider } from '@penumbra-zone/ui/Tooltip';
 import { Header, SyncBar } from '@/widgets/header';
+import { Footer } from '@/widgets/footer';
 import { queryClient } from '@/shared/const/queryClient';
 import { connectionStore } from '@/shared/model/connection';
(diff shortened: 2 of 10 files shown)


conorsch:
> I'd favor putting it into a window var or as a meta tag, so that it's not visible to the public.

Why, may I ask? There's already precedent for the public version footer in minifront:

minifront-footer

We could instead have a dedicated route to display the info, like `/build`, but that just feels like more clicks to me to get to the same answer. Given that the code is already open source, advertising the version in the footer and linking back to the source feels like a standard practice to me. 

Specifically, I'm trying to provide accountability and transparency on what code is running where, given that there are multiple deployments of Veil, with multiple networks in play. Without a clear signal of what code is running, it can be difficult to ascertain whether we can confidently inform users a given bug has been patched on a specific interface.

conorsch:
Will rebase on top of merged #2593 and revisit review comments.

conorsch:
I'm merging this is as a near-term best effort, to increase visibility on deployments. It's very much a shortest-path solution: it pulls the `.git/` dir into the container context, and runs the git commit lookup logic via the js build tooling. That's not ideal, but it works for now, and we can always circle back and improve it more later. cc @vacekj because his suggestions were great, and I'm stubbornly punting on them. 

I've also not massaged the formatting, but we should totally mute the colors as @TalDerei suggested above.

Issue #2589, opened 2025-07-08
can't navigate to details page 


TalDerei:
https://github.com/user-attachments/assets/4cbb3784-a434-4249-89f2-9ab42c84f7bb

PR #2595, opened 2025-07-09
fix details button 

TalDerei:
## Description of Changes

fixes regression from https://github.com/penumbra-zone/web/pull/2557

the `VotingInfo` component has two call sites: the [landing card](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/landing-card/index.tsx#L86) and [round card](https://github.com/penumbra-zone/web/blob/main/apps/veil/src/pages/tournament/ui/round/round-card.tsx#L128). Notably, the landing card renders `<VotingInfo />`, while the latter passes an epoch prop via `<VotingInfo epoch={epoch} />`. This distinction allows `VotingInfo` to infer which page it's on and conditionally render the details button in the UI only when used on the landing page.

the regression here is that if you inspect the previous code,

Screenshot 2025-07-08 at 10 27 45 PM

notice the details button used the epoch field from  `useVotingInfo` to construct its link. In the updated code, however, it relies directly on the epoch prop passed to `VotingInfo`, which again may be _undefined_ if the caller is the landing page (which is fine). 

Screenshot 2025-07-08 at 10 39 21 PM

the fix here is amending the `VotingInfo` union type so that every variant includes an `epoch` field, and then inside the `VotingInfo` component, using the epoch value returned from `useVotingInfo` instead to match the previous behavior. 

Screenshot 2025-07-08 at 10 41 16 PM

-------------------

https://github.com/user-attachments/assets/682e017f-1b16-46b5-895b-b11dc686c4ef

## Related Issue

https://github.com/penumbra-zone/web/issues/2589

### DIFF
--- a/apps/veil/src/pages/tournament/api/use-voting-info.ts
+++ b/apps/veil/src/pages/tournament/api/use-voting-info.ts
@@ -27,14 +27,14 @@ export interface VotingAbility {
 }
 
 export type VotingInfo =
-  | { case: 'loading'; epoch: number | undefined }
-  | { case: 'not-connected'; epoch: number | undefined }
+  | { case: 'loading'; epoch: number }
+  | { case: 'not-connected'; epoch: number }
   | { case: 'already-voted'; epoch: number; votedFor: ValueView }
   | { case: 'needs-to-delegate'; epoch: number }
   | { case: 'can-vote'; epoch: number; ability: VotingAbility; power: ValueView }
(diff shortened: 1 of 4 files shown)


TalDerei:
> What if instead we just always passed in an epoch?

we could, but then we'd need some other way to distinguish between the landing / round card. maybe passing an identifier, 

```ts
<VotingInfo epoch={epoch} 'round-page' />
```

and then inside `votingInfo` component,

```ts
 isRoundPage = identifier == 'round-page'
```

PR #2601, opened 2025-07-10
fix(veil): #2591: display unknown assets

### LINKED ISSUE #2591: current voting results don't add up to 100%
are we possibly not displaying all the assets?

-----------------------

Image

VanishMax:
Closes #2591 

Displays Unknown assets.

1. In previous epochs table

image

2. In current results tables

image

But I left these assets ignored in the landing card results section and the voting dialog

image


cronokirby:
Can we display the actual asset id in some way?

TalDerei:
closing as not planned; @garwalsh suggested out of band that a tooltip will be sufficient for sanitizing the previous epoch table instead. The issue will be opened once it clears design review.

Issue #2608, opened 2025-07-11
Add Epoch time remaining to LQT landing page


garwalsh:
Right now we only show the time remaining in the current epoch on the Round page.
We should also show it on the LQT landing page
Designs in Figma

<img width="1170" height="806" alt="Image" src="https://github.com/user-attachments/assets/619840ca-83db-4a1c-8b8d-ac44e7b7d79b" />

PR #2612, opened 2025-07-12
landing card: add epoch time remaining

TalDerei:
## Description of Changes

- add an hourglass icon to both the landing and round cards (details page), 
- show the epoch’s remaining time on the landing card with a tooltip,
- apply the same gradient used on the epoch number to the “Current Epoch” text.

## Related Issue

https://github.com/penumbra-zone/web/issues/2608


VanishMax (approved):
The "Ends in X" probably deserves its own component instead of copying the code, but I don't strictly mind it in this case.

Issue #2617, opened 2025-07-14
bug: NaN on explore page


conorsch:
**Describe the bug**
Multiple DEX frontends are throwing errors in the browser.

**To Reproduce**

1. run veil dev env
2. navigate to explore page at http://localhost:3000/explore
3. observe error

**Expected behavior**
Page loads fine, assets render.

**Screenshots**
Here's the exact error text:

```
 ○ Compiling /explore ...
 ✓ Compiled /explore in 5s (21095 modules)
 ⨯ SyntaxError: Cannot convert NaN to a BigInt
    at BigInt (<anonymous>)
    at Object.toAmount (../../packages/types/dist/pnum.js:103:40)
    at fetchStats (src/pages/explore/server/stats.ts:59:48)
    at async eval (src/pages/explore/ui/page.tsx:12:16)
    at async ExplorePage (src/pages/explore/ui/page.tsx:17:39)
  101 |         },
  102 |         toAmount() {
> 103 |             return new Amount(splitLoHi(BigInt(value.toFixed(0))));
      |                                        ^
  104 |         },
  105 |         toValueView(metadata) {
  106 |             if (metadata) { {
  digest: '3529525386'
```

**Additional context**

The error seems specific to the `/explore` route to me; I can load these URLs fine:

* http://localhost:3000/portfolio
* http://localhost:3000/tournament
* http://localhost:3000/trade/UM/USDC

PR #2618, opened 2025-07-14
Fix pnum case when input is NaN

JasonMHasperhoven:
## Description of Changes

Addresses https://github.com/penumbra-zone/web/issues/2617

pnum currently throws when the input is not a number (NaN).

Logs that shows that one of the input is NaN:
image

Fix is confirmed & testing by pointing to mainnet.

## Related Issue

### DIFF
--- a/.changeset/eighty-ears-start.md
+++ b/.changeset/eighty-ears-start.md
@@ -1,0 +1,5 @@
+---
+'@penumbra-zone/types': patch
+---
+
+Fix pnum case of NaN input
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
optimistically 👍

conorsch:
Side by side comparison of the behavior between prod and this PR branch: 

dex-hotfix-3

Ideally we'd still render as many of the price widgets as possible, but this is a hotfix, and we're still figuring out the source of the regression, so let's roll with it.

PR #2619, opened 2025-07-14
fix(veil): hardcode veil fqdn in social card

conorsch:
## Description of Changes
The social card dialog logic was using `process.env` to try to look up the site's base URL. That can't work. While we want to standardize on a pattern for app config that works in all cases, e.g. [0], we also want an immediate fix for near-term to resolve bugs [1], which is why I'm submitting this stopgap.

[0] https://github.com/penumbra-zone/web/pull/2593
[1] https://github.com/penumbra-zone/web/issues/2575


TalDerei (approved):
👍

PR #2622, opened 2025-07-15
feat: validate env vars upon startup

### LINKED ISSUE #2621: Validate env vars on startup.
Common bug: code depends on an environment variable to be present, without ensuring it is indeed present during startup. 

Fix: Check that necessary env vars are present and valid during startup. Abort startup otherwise, failing health checks and preventing a rollout.

vacekj:
fixes #2621 

This PR adds a validation check that runs when the app starts and checks that the required environment variables are set and correct (URLs are URLs etc.)

### DIFF
--- a/apps/veil/next.config.mjs
+++ b/apps/veil/next.config.mjs
@@ -1,4 +1,3 @@
-import './src/env.ts';
 import bundleAnalyzer from '@next/bundle-analyzer';
 const withBundleAnalyzer = bundleAnalyzer({
   enabled: process.env.ANALYZE === 'true',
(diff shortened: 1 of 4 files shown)


cronokirby:
The env is fetched at build time, a regression

vacekj:
@cronokirby after my latest commit, it should only run at runtime. instrumentation.js runs upon server start, not build, to the best of my knowledge.

cronokirby:
It seems like a bit of a hack to rely on the fact that NextJS has a conventional file for `instrumentation.{ts,js}`, to then import a file for the purpose of side effects, hoping that it's evaluation hasn't been cached, to then have values for a function that gets called dynamically, instead of just computing the values inside of the function.

TalDerei:
if I'm deriving context correctly, are we closing this as unplanned?

TalDerei:
closing as unplanned

PR #2624, opened 2025-07-16
ci: tooling updates for fly deployments

conorsch:
## Description of Changes

Some small tweaks to the fly logic:

* ci: port flyctl cleanup script out of yaml
* ci: port flyctl deploy script out of yaml
* ci: add flyctl to nix devshell
* ci: add BASE_URL to runtime env
* ci: fix fly cleanup script

These were prepared to ahead of config refactors like #2593, to aid in review over there.

## Related Issue

Follow up to fly logic introduced in https://github.com/penumbra-zone/web/pull/2563


conorsch:
This change is CI only, so I'm self-merging to keep going.

PR #2629, opened 2025-07-17
fix(veil): #2628: explain epochs with votes not adding up to 100%

### LINKED ISSUE #2628: Add context for misleading voting results before fix for unknown assets
There was a bug that meant that we were mistakenly dismissing votes for certain assets by attributing them to the wrong denom. That means that for epochs 207 - 214, the gauge vote results do not add up to 100%

This has now been fixed but we should put a visual cue somewhere so that someone looking back at historical data will understand why the data looks off.

Further context and links to designs in Tooltip for incorrect voting results - Mini PRD

VanishMax:
Closes #2628 

image

image


TalDerei (approved):
This looks good to me, optimistically approving barring any tooltip wording tweaks from from @garwalsh

Issue #2632, opened 2025-07-18
Destructuring import of `BigNumber` from `bignumber.js` causes type errors for consumers


turbocrime:
**Describe the bug**

Destructuring import of `BigNumber` from `bignumber.js` causes type errors for consumers, most evident in consumers of `@penumbra-zone/types/amount`.

**To Reproduce**

1. create a typescript package in another workspace
2. `pnpm add @penumbra-zone/types`
3. import a utility returning `BigNumber` such as `divideAmounts` from `@penumbra-zone/types/amount` and use it
4. observe the return type is named `BigNumber` but the type is unresolved

**Expected behavior**

Types of exports should be valid

**Additional context**

you may examine the dist output of `@penumbra-zone/types`, and see that `amount.d.ts` contains use of a `BigNumber` type with no corresponding import.

consumer prax experiences this issue

https://github.com/prax-wallet/prax/blob/401f682765fc5e6fca7de581c692725ea4386df3/packages/query/src/helpers/price-indexer.ts#L33-L38

PR #2633, opened 2025-07-18
Import `BigNumber` correctly

turbocrime:
## Description of Changes

Use default import for `bignumber.js` instead of destructuring import.

## Related Issue

https://github.com/penumbra-zone/web/issues/2632


TalDerei (approved):
can we remove the linting directives https://github.com/prax-wallet/prax/blob/401f682765fc5e6fca7de581c692725ea4386df3/packages/query/src/helpers/price-indexer.ts#L33-L38?

turbocrime:
consuming this change would enable removal of those lint directives

PR #2636, opened 2025-07-23
veil: only display version footer on `/inspect`

erwanor:
## Description of Changes

I am not a fan of having a git commit linking to a repository at all, but let's keep it for now. This PR improves the styling to make it more discreet and moves it to the nerds page.

## Related Issue

NA


TalDerei (approved):
LGTM

PR #2637, opened 2025-07-23
veil: use 10k units in trade simulation

erwanor:
## Description of Changes

This PR replaces the route book simulation query from trying to simulate a trade with infinite input to doing 10k units of base input. It's a bandaid we can fix by improving the simulate trade RPC, and using simple caching rather than forcing RTTs all the time - blocks are five seconds apart, but none of the backend takes advantage of this clear and clean invalidation rule.


TalDerei (commented):
@erwanor the temporary measure still appears to be loading the route book slowly, averaging ~25s for the UM/USDC pair on a fast internet connection. I haven’t been able to reproduce this slowdown on a local instance hitting mainnet.

PR #2638, opened 2025-07-24
fix(veil): #2598: eliminate layout shifts in portfolio page

### LINKED ISSUE #2598: Unreliable connection status handling
After calling [setup()](https://github.com/penumbra-zone/web/blob/12e187ced3d7aea9aa56e30be3d9d01155d0bbe4/apps/veil/src/shared/model/connection/index.ts#L15), connected is false by default. If there's no connection, it stays that way; if there is, a callback updates it.

This creates a UX issue: there's no reliable way to know when connection status has been resolved. Setting loading = false too early causes a flash of “please connect,” while waiting on the callback can hang forever if not connected. We likely need a separate “checked” flag to resolve this cleanly.

-----------

https://github.com/user-attachments/assets/4477f7e7-a0eb-432e-89cd-da2f7147635a

VanishMax:
Closes #2598 

After looking at the #2598 issue, I found that the UX issue won't be fixed by adding new state in the connection store, rather the portfolio page needed a major rework to eliminate content layout shift and jumps.

In this PR, I used `connectedLoading` variable introduced by @cronokirby to render better connection states in the header and in the portfolio pages. Here's the final result in a video – compare it with [how it was before](https://github.com/user-attachments/assets/4477f7e7-a0eb-432e-89cd-da2f7147635a).

I also fixed css issues here and there while working on portfolio page.

https://github.com/user-attachments/assets/8b85176d-fc0c-4ff5-97be-3bc58106c95f

### DIFF
--- a/apps/veil/src/pages/portfolio/index.tsx
+++ b/apps/veil/src/pages/portfolio/index.tsx
@@ -6,15 +6,17 @@ import { XCircle } from 'lucide-react';
 import { Button } from '@penumbra-zone/ui/Button';
 import { Text } from '@penumbra-zone/ui/Text';
 import { Density } from '@penumbra-zone/ui/Density';
-import { AssetsTable } from './ui/assets-table';
+import { AssetsTable, AssetsTableLayout } from './ui/assets-table';
 import { WalletConnect } from './ui/wallet-connect';
 import { useRegistry } from '@/shared/api/registry.tsx';
 import { IbcChainProvider } from '@/features/cosmos/chain-provider.tsx';
 import { PortfolioPositionTabs } from './ui/position-tabs';
 import { AssetBars } from './ui/asset-bars';
-import { useUnifiedAssets } from '@/pages/portfolio/api/use-unified-assets';
+import { useUnifiedAssets } from './api/use-unified-assets';
 import { PenumbraWaves } from '@/pages/explore/ui/waves.tsx';
 import { ShieldingTicker } from '@/widgets/shielding-ticker';
+import { useAutoAnimate } from '@formkit/auto-animate/react';
+import { PortfolioCard } from './ui/portfolio-card';
 
 interface PortfolioPageProps {
   isMobile: boolean;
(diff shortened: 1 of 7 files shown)


apps/veil/src/pages/portfolio/index.tsx
```diff
@@ -71,21 +73,33 @@ function MobilePortfolioPage() {
...
-        <WalletConnect />
+      {!isConnectionLoading && (
+        <div ref={parent} className='container mx-auto flex max-w-[1136px] flex-col gap-4 py-8'>
+          <WalletConnect />
 
-        {/* Asset Allocation Bars */}
-        {(isPenumbraConnected || isCosmosConnected) && <AssetBars />}
+          {/* Asset Allocation Bars */}
+          {(isPenumbraConnected || isCosmosConnected) && (
+            <PortfolioCard title={'Allocation'}>
```
TalDerei, line 91: comment: why was the `PortfolioCard` moved up the call stack here?

apps/veil/src/pages/portfolio/index.tsx
```diff
@@ -71,21 +73,33 @@ function MobilePortfolioPage() {
 }
 
 const DesktopPortfolioPage = observer(() => {
-  const { isPenumbraConnected, isCosmosConnected } = useUnifiedAssets();
+  const [parent] = useAutoAnimate();
```
TalDerei, line 76: comment: ACK – `useAutoAnimate` placed at the top of the calling component in few places.

TalDerei (approved):
left some comments

apps/veil/src/pages/portfolio/index.tsx
```diff
@@ -71,21 +73,33 @@ function MobilePortfolioPage() {
...
   return (
     <>
+      <PenumbraWaves />
+
       <ShieldingTicker />
-      <div className='container mx-auto flex max-w-[1136px] flex-col gap-4 py-8'>
-        <PenumbraWaves />
 
-        <WalletConnect />
+      {!isConnectionLoading && (
```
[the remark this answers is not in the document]
VanishMax: In my tests the wait is always less than a blink of an eye, that's why it was flashing a lot. I believe this is a worthy sacrifice

PR #2639, opened 2025-07-24
Increase syncing modal timeout

TalDerei:
## Description of Changes

The syncing modal appears too aggressively, especially on dense blocks. This change increases the timeout from 15s to 60s as a conservative adjustment.

## Related Issue

https://github.com/penumbra-zone/web/issues/2254

### DIFF
--- a/apps/minifront/src/fetchers/status.ts
+++ b/apps/minifront/src/fetchers/status.ts
@@ -20,7 +20,7 @@ export async function* getStatusStream(
 ): AsyncGenerator<PlainMessage<StatusStreamResponse>> {
   for await (const item of penumbra
     .service(ViewService)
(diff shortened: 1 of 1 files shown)


apps/minifront/src/fetchers/status.ts
```diff
@@ -20,7 +20,7 @@ export async function* getStatusStream(
 ): AsyncGenerator<PlainMessage<StatusStreamResponse>> {
   for await (const item of penumbra
     .service(ViewService)
-    .statusStream({}, { timeoutMs: 15_000, ...opt })) {
+    .statusStream({}, { timeoutMs: 60_000, ...opt })) {
```
VanishMax, line 23: nit: there's a comment on line 14 that still mentions 15 seconds

PR #2644, opened 2025-07-28
draft: split wasm package

turbocrime:
old branch pushed up for visibility


TalDerei:
closing as unplanned; I’ll provide more context offline. It’s fine to keep the wasm as a single consumable package since it won’t be changing, and based on empirical testing, it doesn’t noticeably impact performance.

